PR #111, opened 2015-06-22
Print the real exception when DBH->connect fails

eradman:
Using eval { } as an error handler prevents the code from accessing
$DBI::errstr. Instead instruct DBI->connect to try without throwing
errors.

It would be better to exit immediately following a system error (module
not found, etc.), but this change reduces the time waiting for the
server to come up to 20 seconds.

### DIFF
--- a/test/lib/BackRestTest/BackupTest.pm
+++ b/test/lib/BackRestTest/BackupTest.pm
@@ -54,7 +54,7 @@ sub BackRestTestBackup_PgConnect
     BackRestTestBackup_PgDisconnect();
 
     # Default
-    $iWaitSeconds = defined($iWaitSeconds) ? $iWaitSeconds : 30;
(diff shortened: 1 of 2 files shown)


dwsteele:
Yeah that's much better error-handling, thanks!  Merged in dcad0f7.
The delay is to wait for clusters to start up, especially when they are in recovery.  I set 30 seconds pretty much arbitrarily so I'm happy with 20 unless it starts causing problems in regression.

dwsteele:
BSD path added in 760baab.  I'm a little surprised by the lack of multi-version support, but each to his own I suppose.

eradman:
I didn't mean to imply that you cannot install multiple version on the BSDs, only that there is usually a default version. The same is true for most Linux distributions (e.g. /usr/bin/psql for Cent6)

dwsteele:
I see.  That makes more sense.

PR #143, opened 2015-09-21
Adding ToC to USERGUIDE.md

Dwaligon:
I added Table of Contents with links to the headers in USERGUIDE to make navigation easier. No worries if this is not something you want in your guide. :)

Check out the rendered version here: https://github.com/Dwaligon/backrest/blob/master/USERGUIDE.md

### DIFF
--- a/USERGUIDE.md
+++ b/USERGUIDE.md
@@ -1,31 +1,125 @@
 # pgBackRest - User Guide
 
-## Installation
+## Table of Contents
+1. [Installation](#installation)
+  * [Ubuntu 12.04/14.04 Setup](#ubuntu-12-14-setup)
+  * [CentOS 6 Setup](#centos-6-setup)
+  * [CentOS 7 Setup](#centos-7-setup)
+  * [Software Installation](#software-installation)
+  * [Regression Test Setup](#regression-test-setup)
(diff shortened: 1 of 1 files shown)


dwsteele:
I like this a lot, actually - definitely makes the file more readable and also makes me realize how desperately a TOC is needed.

What you may not have noticed is that the USERGUIDE.md is automatically generated from the doc/userguide.xml file (the command-line help also comes from here) and from Config.pm.  That means these changes would be wiped out the next time I build the docs for a release.  See doc/doc.pl.

This week I will start working on the HTML documentation which will also be generated from the xml doc and will include a real user guide (currently it's really only a reference).  I'm hoping to have that done mid-October.  My plan at that time is to go with a simple README and CHANGELOG and divert all users to the website for documentation.

It might be interesting for you to get links into the code (see doc/lib/DocRender.pm) but I can't guarantee I would keep that once the HTML version is in place - or that it would make it into a release.

Dwaligon:
Ah, didn't realize it was auto generated.  I'll take a look through your code and maybe I can help contribute.  Great project :+1:

dwsteele:
Thanks, glad you like!

Issue #141, opened 2015-09-15
Throw error when archiving is not configured


dwsteele:
Throw an error if the archive_command is not configured and a backup is started.  Disable the check if archive-check=n.

Perhaps offer a valid archive_command as a hint?

dwsteele:
Fixed by @Dwaligon in c7502f3.

PR #161, opened 2016-01-11
Throw error when archiving is not configured #141

Dwaligon:
Added the check and error when archiving is invalid.


dwsteele:
The patch exposed a problem with PG8.4 in BackRestTestBackup_ClusterStart() so I added a fix for that.  Other than that, looks good!  Committed c7502f3.

PR #169, opened 2016-02-10
update manifest copyright year

crunchyjohn:
Found a copyright year that needs to be updated.


dwsteele:
Hi John,

I think this patch would output "Copyright © 2016-2016".  What were you intending?

crunchyjohn:
Ah, my bad.  We can close this out-- I thought a copyright year was missed.

PR #179, opened 2016-03-09
A clarifiction of the delta restore option

scrummyin:
Since we were talking at the meetup I re-read the docs and the delta docs do not mention removing files. So I added that bit.


dwsteele:
Yes, thanks.  That's a very good clarification to make.  I changed the wording a little and committed 87dd0a1.

BTW - please don't forget to submit the error message you were seeing for that group issue.

PR #181, opened 2016-03-22
semicolon supertiny fix

crunchyjohn:
Though I realize the semicolon is optional, I'm adding it be consistent with the style in the document.


dwsteele:
Oh, the joys of Perl.

PR #193, opened 2016-05-19
Fix perl warnings in doc/ tree

ChristophBerg:
Somewhere between perl 5.14 and 5.20, constructs like this:
  perl -e '$a = {}; keys $a'
started to throw warnings:
  keys on reference is experimental at -e line 1.

Fix by adding a bunch of %{} and @{} casts

Also fix a "is used only once" warning on $XML::Checker::FAIL.

### DIFF
--- a/doc/lib/BackRestDoc/Common/Doc.pm
+++ b/doc/lib/BackRestDoc/Common/Doc.pm
@@ -86,7 +86,8 @@ sub new
 
             eval
             {
-                local $XML::Checker::FAIL = sub
+                local $XML::Checker::FAIL;
+                $XML::Checker::FAIL = sub
                 {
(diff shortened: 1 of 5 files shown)


ChristophBerg:
Fwiw, these warnings are actually fatal:

$ ./doc.pl 
keys on reference is experimental at lib/BackRestDoc/Common/Doc.pm line 157.
Compilation failed in require at ./doc.pl line 23.
BEGIN failed--compilation aborted at ./doc.pl line 23.

dwsteele:
Hi Christoph,

Yeah, unlike the main code base the doc code has only been run on Perl 5.18 since up to now it has not had an audience outside the project developers.  I appreciate the fixes.

Thanks,
-David

PR #195, opened 2016-05-19
Added type from source node and check for it to determine if special handling is needed.

cmwshang:


cmwshang:
Latex is no longer failing (was a problem with my VM). Tested all outputs and all work fine.

PR #197, opened 2016-05-20
Closes #196: Fix perl warnings/errors in String.pm

disco-stu:
Somewhere in perl 5.2x.x the behavior of sprintf changed. sprintf throws a
warning if more arguments are given then used within the format string.

The timestmapFormat function takes any format string.

Fix by special casing the only given format string "%4d" within
timestampFormat.

PR #202, opened 2016-05-26
Fix perl warnings in DocCustomRelease.pm.

disco-stu:
Commit c8d68bc introduced some perl warnings like those fixed in #197
(be1385c2f).


dwsteele:
Alright, I'll bump my doc environment up to 16.04 so I actually catch these errors.

dwsteele:
Merged in b19143b6eaf0a04f921c791a7e62e.

dwsteele:
BTW - I have also upgraded by doc environment to Ubuntu 16.04 and verified the errors before applying the patch.  Hopefully that will prevent this sort of problem in the future.

disco-stu:
Great, thanks. IMHO: You do a great job.

dwsteele:
I appreciate that.  I'm just thankful you are finding these problems before other users do!

PR #203, opened 2016-05-28
Use variables for FILES section in man output.

disco-stu:
Use variables like OPTION_DEFAULT_REPO_PATH instead of hard coded paths.
Also add OPTION_DEFAULT_LOCK_PATH to man page FILES section.


dwsteele:
Thanks, merged at f727d9a8b5ec2ee6348c133.  I need put a few finishing touches on the branch before it goes to master.

PR #241, opened 2016-08-16
Added Archive retention to the user guide and clarified retention-arc…

cmwshang:
…hive option in command reference.

### DIFF
--- a/lib/pgBackRest/Config/ConfigHelpData.pm
+++ b/lib/pgBackRest/Config/ConfigHelpData.pm
@@ -524,7 +524,8 @@ my $oConfigHelpData =
             summary =>
                 "Number of backups worth of WAL to retain.",
             description =>
-                "Number of backups worth of archive log to keep."
+                "If this value is not set, it will default to the value of retention-full regardless of the " .
+                    "retention-archive-type setting."
         },
 
         # RETENTION-ARCHIVE-TYPE Option Help
--- a/lib/pgBackRest/Expire.pm
+++ b/lib/pgBackRest/Expire.pm
@@ -131,12 +131,6 @@ sub process
     my $strArchiveRetentionType = optionGet(OPTION_RETENTION_ARCHIVE_TYPE, false);
     my $iArchiveRetention = optionGet(OPTION_RETENTION_ARCHIVE, false);
 
-    # If no archive retention type is set (default is BACKUP_TYPE_FULL) then error
-    if (!defined($strArchiveRetentionType))
-    {
-        confess &log(ASSERT, 'archive retention type default is not set');
-    }
-
     # Load or build backup.info
     my $oBackupInfo = new pgBackRest::BackupInfo($oFile->pathGet(PATH_BACKUP_CLUSTER));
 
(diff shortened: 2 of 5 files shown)


lib/pgBackRest/Config/ConfigHelpData.pm
```diff
@@ -524,7 +524,8 @@ my $oConfigHelpData =
             summary =>
                 "Number of backups worth of WAL to retain.",
             description =>
-                "Number of backups worth of archive log to keep."
+                "If this value is not set, it will default to the value of retention-full regardless of the " .
+                    "retention-archive-type setting."
         },
```
dwsteele: Wow, I sure hope not.  Sounds like a bug if so.

lib/pgBackRest/Expire.pm
```diff
@@ -131,6 +131,12 @@ sub process
     my $strArchiveRetentionType = optionGet(OPTION_RETENTION_ARCHIVE_TYPE, false);
     my $iArchiveRetention = optionGet(OPTION_RETENTION_ARCHIVE, false);
 
+    # If no archive retention type is set (default is BACKUP_TYPE_FULL) then error
+    if (!defined($strArchiveRetentionType))
+    {
+        confess &log(ASSERT, 'archive retention type default is not set');
+    }
+
```
dwsteele: This shouldn't be required since it has a default.

lib/pgBackRest/Expire.pm
```diff
@@ -224,24 +230,33 @@ sub process
...
+                         &OPTION_RETENTION_ARCHIVE, ERROR_OPTION_INVALID);
         }
     }
 
-    # If no archive retention type is set then exit
-    if (!defined($strArchiveRetentionType))
+&log (INFO, "retention-full: ".(defined($iFullRetention) ? $iFullRetention : 'UNDEF').
+            " retention-diff: ".(defined($iDifferentialRetention) ? $iDifferentialRetention : 'UNDEF').
+            " archive-type: ".(defined($strArchiveRetentionType) ? uc($strArchiveRetentionType) : 'UNDEF').
+            " retention-archive: ".(defined($iFullRetention) ? $iFullRetention : 'UNDEF'));
```
dwsteele: You should convert this to a logDebugMisc call if it stays in.  It takes care of all the pesky undef testing.

lib/pgBackRest/Expire.pm
```diff
@@ -224,24 +230,33 @@ sub process
         }
     }
```
dwsteele: We talked about moving this section to configLoad().  I think that's a good idea to allow the values will show up in the expire start list.  That also means you could take out the debug output line 251.

dwsteele:
Be sure to add tests to ConfigTest.pm to test the new behavior.

lib/pgBackRest/Expire.pm
cmwshang: Oh, crap, you weren’t supposed to see any of that. I forgot that my previous pull request would not be on that previous commit but rather the current state of the branch. My bad. Sorry, I was just testing everything out to verify the logic and then was going to decide what to put in ConfigLoad and what needed to stay in Expire.pm. I should have created a new branch…didn’t mean to waste your time...

Cynthia Shang
www.crunchydata.com

PR #253, opened 2016-09-06
Retention-archive-type being ignored.

cmwshang:
Fixed an issue where retention-archive-type was being ignored if retention-archive was not set, resulting in archive logs not being expired with expiring backups.


dwsteele:
Committed in d2add293c7959d8b191.

PR #254, opened 2016-09-07
suppress a warning on vagrant VM initialization

crunchyjohn:
A minor thing to fix, but I figure every little bit helps.


dwsteele:
Merged in f253a4d85ce47.

PR #266, opened 2016-09-27
Fixed the check command so an error message is not displayed if the backup directory does not exist.

cmwshang:


dwsteele:
Commit with minor revisions, 76673a8e626.

PR #292, opened 2016-12-13
Suppress compiler warnings while building the c library.

disco-stu:
The former assignment discards the const qualifier of *szPageBuffer.
Declaring *szPage as const suppresses this warning.


dwsteele:
Merged in 4e1b74688df11f31e2d5caf76bf7c1d53623ae45.

PR #296, opened 2016-12-19
Added code, tests and documentation for running the check command on …

cmwshang:
…a standby. Also removed the creation of the repository on the standby from the user-guide.


dwsteele:
Committed at 56144c9.

PR #300, opened 2016-12-27
Stanza-create modified to only request force be used when the archive…

cmwshang:
… or backup directory does not contain an info file and there are other files or sub-directories.


dwsteele:
Committed at be337c951aee0121dc3d1a251bdb5218d72769a0.

PR #332, opened 2017-03-15
Added stanza upgrade command.

cmwshang:


test/expect/expire-expire-002.log
```diff
@@ -0,0 +1,1230 @@
...
+P00   INFO: expire full backup set: [BACKUP-FULL-1], [BACKUP-INCR-1]
+P00   INFO: remove expired backup [BACKUP-INCR-1]
+P00   INFO: remove expired backup [BACKUP-FULL-1]
+P00 DETAIL: archive retention on backup [BACKUP-FULL-2], archiveId = 9.2-1, start = 00000001000000000000000C
+P00 DETAIL: remove archive: start = 000000010000000000000000, stop = 00000001000000000000000B
+P00 DETAIL: archive retention on backup [BACKUP-FULL-3], archiveId = 9.3-2, start = 000000010000000000000000, stop = 000000010000000000000002
+P00 DETAIL: archive retention on backup [BACKUP-INCR-2], archiveId = 9.3-2, start = 000000010000000000000006, stop = 0000000100000000000000FB
+P00 DETAIL: archive retention on backup [BACKUP-FULL-4], archiveId = 9.3-2, start = 0000000100000000000000FF
+P00 DETAIL: remove archive: start = 000000010000000000000003, stop = 000000010000000000000005
+P00 DETAIL: remove archive: start = 0000000100000000000000FC, stop = 0000000100000000000000FE
```
dwsteele: This doesn't look right to me.  If full retention is 3, then all WAL after the oldest unexpired full backup should be retained.  The expiration going on here looks like what would happen if retention-archive were set to 1, but I can't see any indication that is true.

cmwshang:
Hmmm, yeah, that does not look right. I'm tracing through now.

Cynthia Shang
www.crunchydata.com

dwsteele:
Committed at beead04.

PR #334, opened 2017-03-21
Added code and unit tests for pgbackrest.conf file validation.

cmwshang:


lib/pgBackRest/Config/Config.pm
```diff
@@ -2691,8 +2691,103 @@ sub optionValidate
...
+            my $strOptionAltName = optionAltName($strOption);
+            if (defined($strOptionAltName))
+            {
+                $strOption = $strOptionAltName;
+            }
+
+            # Is the option a valid pgbackrest option?
+            if (!(exists($oOptionRule{$strOption}) || defined($strOptionAltName)))
+            {
+                &log(WARN, "pgbackrest.conf file contains invalid option '${strOption}'");
```
dwsteele: Should use `optionGet(OPTION_CONFIG)` instead of static string.

lib/pgBackRest/Config/Config.pm
```diff
@@ -2691,8 +2691,103 @@ sub optionValidate
...
+                $bFileValid = false;
+            }
+            else
+            {
+                # Is the option valid for the command section in which it is located?
+                if (defined($strCommand) && $strCommand ne '')
+                {
+                    if (!defined($oOptionRule{$strOption}{&OPTION_RULE_COMMAND}{$strCommand}))
+                    {
+                        &log(WARN, "pgbackrest.conf valid option '${strOption}' is not valid for command '$strCommand'");
```
dwsteele: Should use `optionGet(OPTION_CONFIG)` instead of static string.

lib/pgBackRest/Config/Config.pm
```diff
@@ -2691,8 +2691,103 @@ sub optionValidate
...
+                        $bFileValid = false;
+                    }
+                }
+
+                # Is the valid option a stanza-only option and not located in a global section?
+                if ($oOptionRule{$strOption}{&OPTION_RULE_SECTION} eq CONFIG_SECTION_STANZA &&
+                    $strSection eq CONFIG_SECTION_GLOBAL)
+                {
+                    &log(WARN, "pgbackrest.conf valid option '${strOption}' is a stanza section option and is not valid under " .
```
dwsteele: Should use `optionGet(OPTION_CONFIG)` instead of static string.

lib/pgBackRest/Config/Config.pm
```diff
@@ -2691,8 +2691,103 @@ sub optionValidate
...
+                        "section ${strSection}\n" .
+                        "HINT: global options can be under global or stanza sections but not visa-versa");
```
dwsteele: I think `specified in` is better than `under`.  Thoughts?

lib/pgBackRest/Config/Config.pm
```diff
@@ -2691,8 +2691,103 @@ sub optionValidate
...
+####################################################################################################################################
+sub configFileValidate
+{
+    my $oConfig = shift;
+
+    my $bFileValid = true;
+
+    foreach my $strSectionKey (keys(%$oConfig))
+    {
+        my ($strSection, $strCommand) = ($strSectionKey =~ m/(\w+):*(\w*-*\w*)/);
```
dwsteele: Is this correct?  A stanza name could have `-` in it.  Maybe anything but `:`?

cmwshang: https://msdn.microsoft.com/en-us/windows/hardware/drivers/install/general-syntax-rules-for-inf-files#section-names indicates the format of section names and allows for string tokens (e.g. %sectionname%) and such. Should we be validating the section name against the Windows INI file formatting conventions?

lib/pgBackRest/Config/Config.pm
cmwshang: Will do - to this and the others like it.

lib/pgBackRest/Config/Config.pm
cmwshang: Sounds good.

dwsteele:
Committed at af7e4f4.

PR #336, opened 2017-03-23
Added ability to test warning messages.

cmwshang:


test/lib/pgBackRestTest/Common/RunTest.pm
```diff
@@ -341,6 +344,59 @@ sub testResult
             $bDone = true;
         }
     } while (!$bDone);
+
+    # If we get here then test any warning message
+    if (defined($strWarnMessage))
+    {
+        testWarning($fnSub, $strWarnMessage, $bTestExactWarnMessage);
+    }
```
dwsteele: I don't think it is safe to make two calls to `$fnSub` during a test.  If the function is stateful it may not behave the same twice in a row.  Also, this make performance worse.

The logic in test warning should be embedded in this function (or abstracted to multiple functions) so that fnSub is only called once.  Of course, only if `defined($strWarnMessage)`.

lib/pgBackRest/Common/Log.pm
```diff
@@ -763,4 +763,54 @@ sub testCheck
...
+####################################################################################################################################
+sub logFileCacheTestExact
+{
+    my $strTest = shift;
+
+    return (($strLogFileCache eq $strTest) ? true : false);
+}
+
+push @EXPORT, qw(logFileCacheTestExact);
+
```
dwsteele: Both of these functions should be moved the run test and in Log.pm there should just be a function called `logFileCacheClear` to return the log cache.  I'd like to keep test code out of production to the maximum extent possible.

doc/xml/release.xml
```diff
@@ -171,6 +171,10 @@
                     <release-item>
                         <p>Updated vagrant to new version and image.</p>
                     </release-item>
+
+                    <release-item>
+                        <p>Added ability to test warning messages.</p>
+                    </release-item>
```
dwsteele: Need attribution here.

test/lib/pgBackRestTest/Common/RunTest.pm
```diff
@@ -341,6 +344,59 @@ sub testResult
...
+    logLevelSet(WARN, undef, undef, false);
+
+    # Clear the cache just in case
+    logFileCacheClear();
+
+    # Run the function
+    $fnSub->();
+
+    # Restore the log levels
+    logLevelSet($strLogLevelFile, $strLogLevelConsole, $strLogLevelStdErr, $bLogTimestamp);
```
dwsteele: This needs to be in an exception block so log levels are restored even on exception.

test/lib/pgBackRestTest/Common/RunTest.pm
```diff
@@ -303,12 +303,14 @@ sub testResult
     {
         eval
         {
+# CSHANG so maybe here we save the current log level settings, then turn the console off and the file on to have things sent to the file cache buffer
+# So we'll get a result from the function -
```
dwsteele: Yes, that's the idea.

dwsteele:
Committed at b63ede5.

PR #339, opened 2017-04-06
Modifications to stanza-upgrade per code review

cmwshang:


dwsteele:
Committed at b620e6b.

I moved the BackupInfo unit test into the BackupTest directory.  Other than that, committed as-is.

PR #340, opened 2017-04-10
Removed exceptions for perlcritic.

cmwshang:
Separate commits for each of the following.

Removed exception for perlcritic policy for last element in array.			6d12f16
Require all variables to be used.			8ab5b35
Require simple assignments in declarations.			42abbcd
Require conditional use statements to use dynamic load logic.  	bea8ad3


dwsteele:
Committed at 21a000d.

Issue #349, opened 2017-05-05
exception 124


tvernick:
I got a backup to run on 5/4 successfully and now I'm getting a different error.

-------------------PROCESS START-------------------
2017-05-05 02:45:01.893 P00   INFO: backup command begin 1.18: --archive-timeout=3000 --backup-standby --buffer-size=8388608 --compress-level=3 --db-host=10.100.131.136 --db-path=/dbdata/9.6/data/ --db-port=5433 --db-socket-path=/tmp --db-timeout=30000 --db-user=postgres --log-level-file=info --process-max=8 --protocol-timeout=35000 --repo-path=/data/backups/postgres/pgbackrest_repo --resume --retention-diff=12 --retention-full=7 --stanza=data --stop-auto --type=full
2017-05-05 02:45:03.606 P00   INFO: execute non-exclusive pg_start_backup() with label "pgBackRest backup started at 2017-05-05 02:45:01": backup begins after the next regular checkpoint completes
2017-05-05 02:46:30.524 P00   INFO: backup start archive = 000000020000DF0B00000009, lsn = DF0B/9E32220
2017-05-05 02:46:30.734 P00  ERROR: [124]: process terminated due to an unhandled exception:
                                        Can't call method "info" on an undefined value at /usr/share/perl5/pgBackRest/Backup.pm line 656.
                                         at /usr/bin/pgbackrest line 15.
                                                main::__ANON__('Can\'t call method "info" on an undefined value at /usr/share...') called at /usr/share/perl5/pgBackRest/Backup.pm line 656
                                                pgBackRest::Backup::process('pgBackRest::Backup=HASH(0x139ecb8)') called at /usr/bin/pgbackrest line 259
                                                eval {...} called at /usr/bin/pgbackrest line 289

2017-05-05 02:46:30.735 P00   INFO: backup command end: aborted with exception [124]

tvernick:
Seems to work when I comment out 
#backup-standby=y

dwsteele:
You'll definitely want to configure a standby before enabling that option.  Even so, there should be a better error message.

tvernick:
Hi David,
We actually already had the standby running for a while. We upgraded to 1.18 (from 1.11) and it seemed to break.

dwsteele:
Hmm, I don't see any configuration for a standby in the log from the first entry in this ticket.  Can you send the `pgbackrest.conf` from your backup server?

tvernick:
[...]

PR #355, opened 2017-05-12
Resetting backup-standy to false during backup process if standby unreachable (bug #349)

cmwshang:
Resetting option backup-standby to false if the standby is unreachable.
Removed unnecessary backup from full real test that was creating a backup for the purpose of creating the backup.info file which is no longer the case give stanza create.


dwsteele:
Committed at 7d0c50fcd6dd6d7800b53ae2a1abf35f923ec7da.

Issue #347, opened 2017-05-03
Error with exception 066


tvernick:
Hello, we upgraded from 1.11 to 1.18 (using required steps from version 1.13 first). We are getting an error towards the end of backup.

131/16402/1033065205 (0B, 86%)
2017-05-03 10:00:32.410 P00  ERROR: [066]: raised on local-1 host: remote process terminated on 10.100.131.136 host: WARN: /etc/pgbackrest.conf file contains invalid option 'thread-timeout'
                                        unable to read line after 35000 second(s)
2017-05-03 10:00:32.412 P00  ERROR: [066]: raised on local-1 host: remote process terminated on 10.100.131.136 host: WARN: /etc/pgbackrest.conf file contains invalid option 'thread-timeout'
                                        unable to read line after 35000 second(s)
2017-05-03 10:00:32.462 P00  ERROR: [066]: raised on local-1 host: remote process terminated on 10.100.131.136 host: WARN: /etc/pgbackrest.conf file contains invalid option 'thread-timeout'
                                        unable to read line after 35000 second(s)
2017-05-03 10:00:32.462 P00   INFO: backup command end: aborted with exception [066]

dwsteele:
The first thing to do is remove `thread-timeout` from your config files.  It hasn't been used for a very long time.

You shouldn't be getting an error, only a warning.  However, the warning is getting thrown into the protocol layer and messing things up.  We'll get that fixed, but removing the invalid option should get you going again.

gintoddic:
Thanks for the quick response. We did see the warning and I ended up commenting the option out. I am running the backup again now without the timeout and I'll see what happens.

tvernick:
Backup is running but we also seem to get this error on the db host

[root@datawarehouse001 pg_log]# pgbackrest check --stanza data
ERROR [027]: : db and backup cannot both be configured as remote

config is below. The IP for db-host is the local db and the backup host is the remote host.

[data]
db-host=10.100.73.20
db-path=/dbdata/9.6/data/
db-user=postgres
db-port=5433
thread-max=8
resume=y
protocol-timeout=35000
db-timeout=30000
buffer-size=8388608
db-socket-path=/tmp

[global]
archive-async=y
archive-timeout=3000
backup-host=10.100.131.137
backup-user=postgres
backup-standby=y
repo-path=/data/backups/postgres/pgbackrest_repo
spool-path=/dbdata/postgres/walarchive
compress-level=3
stop-auto=y
[...]

PR #357, opened 2017-05-16
Suppress warnings on remote (bug #347)

cmwshang:
Also consolidated configFileValidation tests.


dwsteele:
Committed at 7e982f0.

PR #359, opened 2017-05-19
New abstracted storage and protocol IO layers.

dwsteele:
Probably best to start with `Storage/Local.pm` and work out from there.  `Storage/Local.pm` replaces `File.pm` and `FileCommon.pm`.  `Protocol/Storage/Remote.pm` implements remote storage capabilities.


lib/pgBackRest/Common/Io/Base.pm
```diff
@@ -0,0 +1,126 @@
...
+####################################################################################################################################
+# Package name constant
+####################################################################################################################################
+use constant COMMON_IO_BASE                                         => __PACKAGE__;
+    push @EXPORT, qw(COMMON_IO_BASE);
+
+####################################################################################################################################
+# Default buffer max
+####################################################################################################################################
+use constant COMMON_IO_BUFFER_MAX                                   => 4194304;
```
cmwshang: CSHANG Shouldn't this always be the same as OPTION_DEFAULT_BUFFER_SIZE? And visa-versa - so could we have a common file where these are set so they don't get out of sync?

lib/pgBackRest/Protocol/Helper.pm
```diff
@@ -95,15 +175,7 @@ sub protocolGet
```
cmwshang: CSHANG I don't see anywhere this funtion is called with $strRemoteType=NONE. Is NONE used anymore? If so, what is it's purpose?

lib/pgBackRest/Storage/Helper.pm
```diff
@@ -0,0 +1,129 @@
...
+        logDebugParam
+        (
+            __PACKAGE__ . '::storageLocal', \@_,
+            {name => 'strPath', default => '/', trace => true},
+        );
+
+    # Create storage if not defined
+    if (!defined($hStorage->{&STORAGE_LOCAL}{$strPath}))
+    {
+        # Create local storage
```
cmwshang: CSHANG If local is always read only, then would bAllowTemp need to be false?

lib/pgBackRest/Storage/Local.pm
```diff
@@ -0,0 +1,775 @@
...
+    $self->{strPathBase} = $strPathBase;
+    $self->{oDriver} = $oDriver;
+    $self->{hRule} = $hRule;
+    $self->{bAllowTemp} = $bAllowTemp;
+    $self->{strTempExtension} = $strTempExtension;
+    $self->{strDefaultPathMode} = $strDefaultPathMode;
+    $self->{strDefaultFileMode} = $strDefaultFileMode;
+
+    # Set temp extension in driver
+    if (defined($self->{oDriver}))
```
cmwshang: CSHANG Why is this check necessary? oDriver is a required parameter, if not defined it would blow up when the function is called, no? I don't see how we could test this.

lib/pgBackRest/Storage/Local.pm
```diff
@@ -0,0 +1,775 @@
...
+        {
+            $strPath = $self->{strPathBase};
+            $strFile = $strPathExp;
+        }
+    }
+
+    # Make sure a temp file is valid for this type and file
+    if ($bTemp)
+    {
+        # Only allow temp files for PATH_REPO_ARCHIVE, PATH_SPOOL_ARCHIVE_OUT, PATH_REPO_BACKUP_TMP and any absolute path
```
cmwshang: CSHANG Is this comment correct?  These constants don't exist - the PATH has been replaced with STORAGE - even so, how do we know if this boolean is set correctly if we're not checking those settings? And if we determined it was an absolute path, does that mean bAllowTemp should always be true?

lib/pgBackRest/Common/Io/Base.pm
dwsteele: They are not directly related, but it does make sense to set OPTION_DEFAULT_BUFFER_SIZE to COMMON_IO_BUFFER MAX.

lib/pgBackRest/Protocol/Helper.pm
dwsteele: This code is vestigial and has been removed completely now.

lib/pgBackRest/Storage/Helper.pm
dwsteele: Oops, actually there is one case where it is not read-only so I amended the comment.

lib/pgBackRest/Storage/Local.pm
dwsteele: oDriver used to be optional before I did some refactoring.  Fixed.

lib/pgBackRest/Storage/Local.pm
dwsteele: Comment has been fixed.

dwsteele:
Need to close this a reopen since the branch name has changed.

dwsteele:
Committed at de7fc37f88f670a1b4.

PR #360, opened 2017-05-22
Maintain two copies of .manifest and .info files.

dwsteele:
Previously these files were made atomic by moving them from a temp file to their final location.  This will not work on file stores (like S3) that do not have an atomic move command.

Instead, maintain two copies and reconcile them on load to find any errors that occurred when saving.

[Ignore the changes in FileCommon.pm as these are now made obsolete by the later storage branch.]

### DIFF
--- a/lib/pgBackRest/Backup/Backup.pm
+++ b/lib/pgBackRest/Backup/Backup.pm
@@ -95,7 +95,7 @@ sub fileNotInManifest
     foreach my $strName (sort(keys(%{$hFile})))
     {
         # Ignore certain files that will never be in the manifest
-        if ($strName eq FILE_MANIFEST ||
+        if ($strName eq FILE_MANIFEST_COPY ||
             $strName eq '.')
         {
             next;
@@ -835,7 +835,7 @@ sub process
     }
 
     # Save the backup manifest
-    $oBackupManifest->save();
+    $oBackupManifest->saveCopy();
 
     # Perform the backup
     my $lBackupSizeTotal =
@@ -900,8 +900,8 @@ sub process
     # will be consistent - at least not here.
     if (optionGet(OPTION_ONLINE) && optionGet(OPTION_BACKUP_ARCHIVE_CHECK))
     {
-        # Save the backup manifest a second time - before getting archive logs in case that fails
-        $oBackupManifest->save();
+        # Save the backup manifest before getting archive logs in case of failure
+        $oBackupManifest->saveCopy();
 
         # Create the modification time for the archive logs
         my $lModificationTime = time();
--- a/lib/pgBackRest/Common/Ini.pm
+++ b/lib/pgBackRest/Common/Ini.pm
@@ -40,8 +40,16 @@ use constant INI_KEY_CHECKSUM                                       => 'backrest
     push @EXPORT, qw(INI_KEY_CHECKSUM);
 use constant INI_KEY_FORMAT                                         => 'backrest-format';
     push @EXPORT, qw(INI_KEY_FORMAT);
+use constant INI_KEY_SEQUENCE                                       => 'backrest-sequence';
+    push @EXPORT, qw(INI_KEY_SEQUENCE);
 use constant INI_KEY_VERSION                                        => 'backrest-version';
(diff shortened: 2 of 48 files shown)


cmwshang:
I think we need more stanza tests (I have a branch that has them, but it also required modifications to Stanza.pm). For example, what happens if archive.info.copy exists but the archive.info file does not?

dwsteele:
Well, that case is covered in the Ini unit tests.  Is there a specific reason you think it should be in the stanza test as well?  I'm hoping to have as little duplicate coverage as possible, though of course it can't always be avoided.

cmwshang:
The stanza create tests for files/directories in the archive/backup directories. If info is missing, then .copy is considered to be "something in the directory" which requires the user to specify --force. So the question becomes Is that OK? or do we ignore it? I decided on the former - that if anything exists in the directory but the info file doesn't we require force. Remember, this branch does not have any stanza-create tests so I felt we should at least bring in some that are directly related to this.

dwsteele:
I agree that the presence of a .copy file counts as "something" in the directory.

lib/pgBackRest/Common/Ini.pm
```diff
@@ -323,13 +460,15 @@ sub save
 {
     my $self = shift;
 
+    # Save only if modified
```
dwsteele: Not always the best idea, as we discussed.

lib/pgBackRest/Common/Ini.pm
```diff
@@ -431,17 +586,18 @@ sub hash
...
 
-    # Calculate the checksum
-    my $oChecksumContent = dclone($self->{oContent});
     my $oSHA = Digest::SHA->new('sha1');
     my $oJSON = JSON::PP->new()->canonical()->allow_nonref();
     $oSHA->add($oJSON->encode($self->{oContent}));
 
     # Set the new checksum
     $self->{oContent}{&INI_SECTION_BACKREST}{&INI_KEY_CHECKSUM} = $oSHA->hexdigest();
+    $self->{oContent}{&INI_SECTION_BACKREST}{&INI_KEY_SEQUENCE} = $iSequence + 0;
```
dwsteele: Yes.

lib/pgBackRest/Backup/Backup.pm
```diff
@@ -95,7 +95,7 @@ sub fileNotInManifest
     foreach my $strName (sort(keys(%{$hFile})))
     {
         # Ignore certain files that will never be in the manifest
-        if ($strName eq FILE_MANIFEST ||
+        if ($strName eq FILE_MANIFEST_COPY ||
             $strName eq '.')
```
dwsteele: backup.manifest should never be present for a resumed backup.  If it is it should be removed.

lib/pgBackRest/Common/Ini.pm
```diff
@@ -120,7 +129,105 @@ sub load
...
+            {
+                $self->{oContent} = $oContent;
+            }
+            # Else only one header is valid and must be considered authoritative
+            else
+            {
+                if (!$bContentHeader)
+                {
+                    $self->{oContent} = $oContentCopy;
+                }
```
dwsteele: There is no else here - this is to assign the copy if the main header did not validate.

lib/pgBackRest/Common/Ini.pm
```diff
@@ -106,6 +114,7 @@ sub new
     # Else initialize
     else
     {
+        $self->{oContent}{&INI_SECTION_BACKREST}{&INI_KEY_SEQUENCE} = 0 + 0;
```
dwsteele: Just being extra paranoid?  ;) Probably a copy-pasto.  Fixed.

dwsteele:
This will be committed with the storage branch so closing this PR.

lib/pgBackRest/Backup/Backup.pm
dwsteele: I updated the comment to be clearer, however.

dwsteele:
Committed at de7fc37f88f670a1b4.

PR #370, opened 2017-06-07
S3 Repository

dwsteele:


lib/pgBackRest/Storage/S3/Request.pm
```diff
@@ -0,0 +1,202 @@
...
+    my $oHttpClient = new pgBackRest::Common::Http::Client(
+        $self->{strHost}, $strVerb,
+        {strUri => $strUri, hQuery => $hQuery, hRequestHeader => $hHeader, rstrRequestBody => $rstrBody,
+            bVerifySsl => $self->{bVerifySsl}, lBufferMax => $self->{lBufferMax}});
+
+    # Check response code
+    my $iReponseCode = $oHttpClient->responseCode();
+    my $oResponse;
+
+    if ($oHttpClient->responseCode() == 200)
```
cmwshang: Why is this not $iResponseCode? Why set up a variable and not use it?

lib/pgBackRest/Storage/S3/Request.pm
```diff
@@ -0,0 +1,202 @@
...
+        }
+        # An IO object is expected for file responses
+        elsif ($strResponseType eq S3_RESPONSE_TYPE_IO)
+        {
+            $oResponse = $oHttpClient;
+        }
+    }
+    else
+    {
+        if ($oHttpClient->responseCode() == 404)
```
cmwshang: Why is this not $iResponseCode? Why set up a variable and not use it?

lib/pgBackRest/Storage/S3/Request.pm
```diff
@@ -0,0 +1,202 @@
...
+            {
+                confess &log(ERROR, "unable to open '${strUri}': No such file or directory", ERROR_FILE_MISSING);
+            }
+        }
+        else
+        {
+            my $rstrResponseBody = $oHttpClient->responseBody();
+
+            confess &log(ERROR,
+                "S3 request error [$iReponseCode] " . $oHttpClient->responseCode() .
```
cmwshang: Won't this just read "S3 request error [123] 123 some body text"? why do you need to duplicate the response code?

lib/pgBackRest/Storage/S3/Driver.pm
```diff
@@ -0,0 +1,512 @@
...
+                if (defined($strPrefix))
+                {
+                    $strName = substr($strName, length($strPrefix));
+                }
+
+                # Strip off final /
+                $strName = substr($strName, 0, length($strName) - 1);
+
+                $hManifest->{$strName}->{type} = 'd';
+                $iPathTotal++;
```
cmwshang: What is $iPathTotal ever actually used for? Was it intended to be part of the while condition?

lib/pgBackRest/Storage/S3/Driver.pm
```diff
@@ -0,0 +1,512 @@
...
+sub list
+{
+    my $self = shift;
+
+    # Assign function parameters, defaults, and log debug info
+    my
+    (
+        $strOperation,
+        $strPath,
+    ) =
```
cmwshang: In the Posix drive, bIgnoreMissing is an option and appears to be used (e.g. lines 306, 318 ArchiveCommon.pm, 319, 337, 349 ArchiveInfo.pm, 178 ArchivePush.pm, 242 Expire.pm, 339 Info.pm, etc) so I'm wondering why it is not an option here - is it only because this list function is for the manifest? Shouldn't this function then be called manifestList?

lib/pgBackRest/Storage/S3/Driver.pm
```diff
@@ -0,0 +1,512 @@
...
+sub info
+{
+    my $self = shift;
+
+    # Assign function parameters, defaults, and log debug info
+    my
+    (
+        $strOperation,
+        $strFile,
+    ) =
```
cmwshang: Posix driver info command allows bIgnoreMissing as option and will not throw an error if set and missing - why not here?

lib/pgBackRest/Storage/S3/Request.pm
dwsteele: No good reason.  Fixed.

lib/pgBackRest/Storage/S3/Request.pm
dwsteele: Fixed.

lib/pgBackRest/Storage/S3/Request.pm
dwsteele: Fixed.

lib/pgBackRest/Storage/S3/Driver.pm
dwsteele: This was originally part of the debug code when writing this function.  Removed.  Also removed $iFileTotal.

lib/pgBackRest/Storage/S3/Driver.pm
dwsteele: There are no directories in S3 so it's not possible to ignore missing.  It does beg the question of whether we should disallow `bIgnoreMissing = false` for S3 but I'm not inclined to worry about that right now.

lib/pgBackRest/Storage/S3/Driver.pm
dwsteele: That's an important option when building manifests from the database directory but not useful in a repo where files are not supposed to disappear suddenly.  We could include it for the sake on completeness but it would probably never be used.

We could also handle cases like this with capabilities - i.e. ignore missing on info is not allowed, but I don't think it's worth the extra code.  Or maybe that's a refinement.

lib/pgBackRest/Storage/S3/Request.pm
```diff
@@ -0,0 +1,202 @@
...
+            {name => 'strResponseType', optional => true, default => S3_RESPONSE_TYPE_NONE, trace => true},
+            {name => 'bIgnoreMissing', optional => true, default => false, trace => true},
+        );
+
+    # Get datetime to be used for auth requests
+    my $strDateTime = s3DateTime();
+
+    # Set content length and hash
+    $hHeader->{&S3_HEADER_CONTENT_SHA256} = defined($rstrBody) ? sha256_hex($$rstrBody) : PAYLOAD_DEFAULT_HASH;
+    $hHeader->{&S3_HEADER_CONTENT_LENGTH} = defined($rstrBody) ? length($$rstrBody) : 0;
```
dwsteele: I read that to mean you should send it when there is a message.  If there's no message there is no need.

It's definitely not calculated in HttpClient - and can't be since content-length if present must be part of Signed Headers.

lib/pgBackRest/Common/Http/Client.pm
```diff
@@ -0,0 +1,284 @@
...
+        my $strHeaderKey = lc(substr($strHeader, 0, $iColonPos));
+        my $strHeaderValue = trim(substr($strHeader, $iColonPos + 1));
+
+        # Store the header
+        $self->{hResponseHeader}{$strHeaderKey} = $strHeaderValue;
+
+        # Process content length
+        if ($strHeaderKey eq HTTP_HEADER_CONTENT_LENGTH)
+        {
+            $self->{iContentLength} = $strHeaderValue + 0;
```
dwsteele: Not sure I follow. This header contains the length as an integer of the content that will follow after the header block.  It should be read as an integer.

lib/pgBackRest/Common/Http/Client.pm
```diff
@@ -0,0 +1,284 @@
...
+        $strOperation,
+        {name => 'rstrResponseBody', value => \$strResponseBody, trace => true}
+    );
+}
+
+####################################################################################################################################
+# Properties.
+####################################################################################################################################
+sub contentLength {shift->{iContentLength}}                         # Content length if available (-1 means not known yet)
+sub requestHeader {shift->{hResponseHeader}}
```
dwsteele: Indeed, but since it's not used I removed it instead.

lib/pgBackRest/Storage/S3/FileWrite.pm
```diff
@@ -0,0 +1,194 @@
...
+}
+
+####################################################################################################################################
+# close - close the file
+####################################################################################################################################
+sub close
+{
+    my $self = shift;
+
+    # Only close need to do close if something was written
```
dwsteele: Fixed.

lib/pgBackRest/Storage/S3/Auth.pm
```diff
@@ -0,0 +1,262 @@
...
+            {name => 'strUri', trace => true},
+            {name => 'strQuery', trace => true},
+            {name => 'hHeader', trace => true},
+            {name => 'strPayloadHash', trace => true},
+        );
+
+    # Create the canonical request
+    my $strCanonicalRequest =
+        "${strVerb}\n${strUri}\n${strQuery}\n";
+    my $strSignedHeaders;
```
dwsteele: Unusual, I know. I was trying to stay as close as possible the AWS terminology to make this code easier to parse/update later.

lib/pgBackRest/Storage/S3/Auth.pm
```diff
@@ -0,0 +1,262 @@
...
+        $strQuery,
+        $strDateTime,
+        $hHeader,
+        $strAccessKeyId,
+        $strSecretAccessKey,
+        $strPayloadHash,
+    ) =
+        logDebugParam
+        (
+            __PACKAGE__ . '::s3CanonicalRequest', \@_,
```
dwsteele: Fixed.

dwsteele:
Committed at 051c961151e75.

PR #377, opened 2017-07-10
Added archive min max per DB to info command.

cmwshang:
Also fixed test.pl missing debian package.


dwsteele:
Committed at cda3b9821b83b3cc96.

PR #389, opened 2017-08-02
Bugfixes for Backup/Info.pm and increased unit test coverage.

cmwshang:
Fixed 2 bugs in Backup/Info.pm: 
1) backupArchiveDbHistoryId (missing conditional for histories that no longer exist (- likely due to reconstruction on a create or update --force) preventing expiration from removing orphaned directories) and 
2) dbHistoryIdGet (missing \@_ so the FileRequired parameter was never being set neither by passing nor by the default).
Increased unit test coverage for Backup/Info.pm.


dwsteele:
Committed at f9bdc092bac8cdfe1.

PR #398, opened 2017-08-16
Added ability to configure up to 7 standby replicas.

cmwshang:


lib/pgBackRest/Db.pm
```diff
@@ -90,7 +92,12 @@ sub new
...
         if (!isDbLocal({iRemoteIdx => $self->{iRemoteIdx}}))
         {
-            $self->{oProtocol} = protocolGet(CFGOPTVAL_REMOTE_TYPE_DB, $self->{iRemoteIdx});
+            eval
+            {
+                $self->{oProtocol} = protocolGet(CFGOPTVAL_REMOTE_TYPE_DB, $self->{iRemoteIdx}, {bWarnOnError => $bWarnOnError});
+                return true;
+            }
+            or do {};
         }
```
cmwshang: The eval is in "if (!isDbLocal" condition, so I'm not sure I'm following when you say "it will try to connect to local host"

lib/pgBackRest/Db.pm
```diff
@@ -1022,16 +1037,18 @@ sub dbObjectGet
...
 
-        # Make sure the standby database is defined when backup from standby requested
+        # Make sure a standby database is defined when backup from standby option is set
         if (cfgOption(CFGOPT_BACKUP_STANDBY) && !defined($oDbStandby))
         {
-            confess &log(ERROR, 'unable to find standby database - cannot proceed');
+            # Throw an error that is distinct from connecting to the master for testing purposes
+            confess &log(ERROR, 'unable to find standby database - cannot proceed', ERROR_HOST_CONNECT);
         }
```
cmwshang: New tests in RealAllTest make sure each of these separate code blocks are hit. Would a new error code of ERROR_DB_STANDBY_CONNECT be better? Or maybe I can create a DB Unit Test module to test the dbObjectGet function given the error messages are different, but that would still mean in the Real Tests we can't be sure which code block was hit.

lib/pgBackRest/Db.pm
dwsteele: If the protocol object is undefined ->connect will connect to any server that happens to be running on the local host.  This is probably not what you intend.

lib/pgBackRest/Common/Io/Process.pm
```diff
@@ -89,10 +89,12 @@ sub error
     my $strMessage = shift;
     my $strDetail = shift;
     my $bClose = shift;
+    my $bWarnOnError = shift;
 
     if (defined($self->{iProcessId}))
     {
         my $oWait = waitInit(defined($iCode) ? IO_ERROR_TIMEOUT : 0);
+        my $bWarnInstead = defined($bWarnOnError) ? $bWarnOnError : false;
```
cmwshang: Fixed.

lib/pgBackRest/Common/Io/Process.pm
```diff
@@ -122,10 +124,20 @@ sub error
...
+                                ($iExitStatus != 255 ?  sprintf(' [%03d]', $iExitStatus) : ''),
+                            $strError);
+                    }
+                    # Confess a warning instead of an error - useful for looping through servers where it is not important if they
+                    # are unavailable
+                    else
+                    {
+                        confess &log(WARN, "[$iErrorCode] " . 'process ' . $self->id() . ' terminated unexpectedly' .
+                            ($iExitStatus != 255 ?  sprintf(' [%03d]', $iExitStatus) : '') . " : $strError");
+                    }
```
cmwshang: Fixed.

test/lib/pgBackRestTest/Env/Host/HostBackupTest.pm
```diff
@@ -1037,16 +1037,24 @@ sub configCreate
...
         {
-            $oParamHash{$strStanza}{$self->optionIndexName(CFGOPT_DB_HOST, 2)} = $oHostDb2->nameGet();
+            $oParamHash{$strStanza}{$self->optionIndexName(CFGOPT_DB_HOST, 2)} = BOGUS;
             $oParamHash{$strStanza}{$self->optionIndexName(CFGOPT_DB_USER, 2)} = $oHostDb2->userGet();
             $oParamHash{$strStanza}{$self->optionIndexName(CFGOPT_DB_CMD, 2)} = $oHostDb2->backrestExe();
             $oParamHash{$strStanza}{$self->optionIndexName(CFGOPT_DB_CONFIG, 2)} = $oHostDb2->backrestConfig();
             $oParamHash{$strStanza}{$self->optionIndexName(CFGOPT_DB_PATH, 2)} = $oHostDb2->dbBasePath();
 
+            # Add an invalid replica to simulate more than one replica. A warning should be thrown by dbObjectGet when a stanza is
+            # created and a valid replica should be chosen.
```
cmwshang: Fixed.

Issue #421, opened 2017-10-10
Info Command (json): History ID incorrect datatype


jasonodonnell:
When you output backup info using `json` format, the history ID is returned as a string:

```json
"db" : [
    {
	"id" : "1",
	"system-id" : 6475061310255132704,
        "version" : "9.6"
    }
],
```

The [source](https://github.com/pgbackrest/pgbackrest/blob/1e0ed074555df7d9079fdf52c4b6b5b049b6d155/lib/pgBackRest/Info.pm#L505) suggests this should be an integer but I cannot locate where its being casted to a string.

jasonodonnell:
Possibly the same issue with the `archive` blob as well:

```json
"archive" : [
	 {
	     "database" : {
	          "id" : "1"
	      },
	      "id" : "9.6-1",
	     "max" : "00000001000000000000000A",
             "min" : "000000010000000000000001"
    }
],
```

dwsteele:
Well, I'm not completely sure which one is right but it should definitely be all one way or the other.  @cmwshang, what do you think?

cmwshang:
Agreed it should all be one or the other, but it appears to always be a string in the output even though the variable prefix in the code is $i.  At this point the data is read from the file so it's just character strings - we should be able to type it truly to an integer if we do the + zero trick. Is this causing an error somewhere?

Cynthia Shang
www.crunchydata.com

jasonodonnell:
Agreed, if its an integer, it should be an integer.

I noticed this inconsistency when trying to auto generate golang structures of the json output.  The mix-match of data types resulted in a more complex object.

I've seen no errors related to this, just pointing it out for the sake of consistency.

cmwshang:
Sounds good. I'll make the change for next release.

Cynthia Shang
www.crunchydata.com

dwsteele:
Fixed at 2d56de00fcb04f2f949c18.

PR #428, opened 2017-10-17
Bug #421 - fixed json output of info command to display an integer vs string for db id

cmwshang:


dwsteele:
Committed at 2d56de00fcb04f2f949c18.

PR #435, opened 2017-10-27
Updated user-guide and reference to clarify pgbackrest archive-timeout vs. postgresql archive_timeout options.

cmwshang:


dwsteele:
Committed at 2be93036f1c8de734 with minor edits.

Issue #441, opened 2017-11-07
After upgrade from 9.4 to 9.6, archive-get fails on 9.4 versions restored from backup.


clad:
Upgrade process - 

- Create streaming 9.4 replica via pgbackrest restore
- Promote replica to master
- Upgrade new master to 9.6 via pg_upgrade
- upgrade pgbackrest

Now attempting to use a restore from a pgbackrest backup of a backup set prior to the upgrade fails during the log shipping -

2017-11-07 15:31:22.411 P00   INFO: archive-get command begin 1.25: --buffer-size=8388608 --compress --compress-level=8 --db1-path=/var/lib/pgsql/9.4/data --log-level-console=debug --log-level-file=info --log-level-stderr=warn --log-path=/var/lib/pgsql_wal/pgbackrest.log --repo-path=/mnt/nfs/backups/db_backups/db-05/pgbackrest --repo-type=posix --stanza=main
2017-11-07 15:31:22.462 P00   INFO: get WAL segment 00000001000025A300000076
2017-11-07 15:31:22.462 P00  DEBUG:     Archive::Get::Get->get(): strDestinationFile = pg_xlog/RECOVERYXLOG, strSourceArchive = 00000001000025A300000076
2017-11-07 15:31:22.462 P00  DEBUG:     Storage::Posix::Driver->new(): bFileSync = <true>, bPathSync = <true>
2017-11-07 15:31:22.462 P00  DEBUG:     Storage::Local->new(): bAllowTemp = <true>, hRule = [hash], lBufferMax = 8388608, oDriver = [object], strDefaultFileMode = <0640>, strDefaultPathMode = <0750>, strPathBase = /mnt/nfs/backups/db_backups/db-05/pgbackrest, strTempExtension = pgbackrest.tmp
2017-11-07 15:31:22.463 P00  DEBUG:     Archive::Base->getCheck(): strDbVersion = [undef], strWalFile = 00000001000025A300000076, ullDbSysId = [undef]
2017-11-07 15:31:22.463 P00  DEBUG:     Db::dbObjectGet(): bMasterOnly = true
2017-11-07 15:31:22.463 P00  DEBUG:     Db->new(): iRemoteIdx = 1
2017-11-07 15:31:22.463 P00  DEBUG:     Db::dbObjectGet=>: iDbMasterIdx = 1, iDbStandbyIdx = [undef], oDbMaster = [object], oDbStandby = [undef]
[...]

PR #446, opened 2017-11-16
Fixed issue #441 - unable to restore after upgrade to database version prior to upgrade

cmwshang:


dwsteele:
This looks good and I have pushed it up to integration.

I'm a bit concerned about the extra ->exists() even when there is only one directory to search.  This will introduce additional latency, especially for S3.  We should think about a follow up commit that removes the ->exists() call when only one directory matches.

dwsteele:
Committed at 6ef2167e972039e.

PR #461, opened 2017-12-19
Issue 460: improved info command to avoid outputting 'db (prior)' when no backups or archived WAL exist.

cmwshang:


dwsteele:
Committed at 67c8f0d (v1) and ef3ef85 (v2).

PR #479, opened 2018-02-01
Modify stanza-create so that it logs an INFO message instead of throwing an error

cmwshang:
…when info files exist and their DB sections match the current database.


dwsteele:
Committed to `master` at dc73abdb6e9612f7af8e and `release/1` at a592ecf18c804b8d2d7.

PR #483, opened 2018-02-13
Fix for setting value from default when --reset- prefix used.

cmwshang:


dwsteele:
Committed at 00e9aca872745d2552aa1b562.

PR #490, opened 2018-03-01
Remove sudo from src/Makefile.

disco-stu:
make should be called in the correct context not calling sudo itself.


dwsteele:
Agreed.  I'll get this into 2.01.  You'll patch this for 2.00?

disco-stu:
I think so. Otherwise I would need to add sudo to Build-Dependencies.

dwsteele:
Yeah, i'd rather not do that.

dwsteele:
Committed at 6ca6424fb1.

PR #500, opened 2018-03-08
Error when multiple arguments are set for a single argument option.

cmwshang:


dwsteele:
Committed at 34898ce471214e234.

Issue #543, opened 2018-06-13
Database restore failed with error.


nj3110:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version: 2.02

2. PostgreSQL version: 9.6.2

3. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each:
CentOS Linux release 7.4.1708 (Core)

4. Did you install pgBackRest from source or from a package?
package
5. Please attach the following as applicable:
    - `pgbackrest.conf` file(s)
    - `postgresql.conf` settings applicable to pgBackRest (`archive_command`, `archive_mode`, `listen_addresses`, `max_wal_senders`, `wal_level`, `port`)
    - errors in the postgresql log file before or during the time you experienced the issue
    - log file in `/var/log/pgbackrest` for the commands run (e.g. `/var/log/pgbackrest/mystanza_backup.log`)

7. Describe the issue:

Getting below mentioned error while trying to restore a database from backup.
----------------------------
[fsepuser@c7m1 inst2]$ pgbackrest --stanza=PD2 --delta --db-include=test restore
ERROR: [124]: process terminated due to an unhandled exception:
Use of uninitialized value $strDbFilter in concatenation (.) or string at /usr/share/perl5/pgBackRest/Restore.pm line 1224.
 at /usr/share/perl5/pgBackRest/Main.pm line 12.
        pgBackRest::Main::__ANON__('Use of uninitialized value $strDbFilter in concatenation (.) ...') called at /usr/share/perl5/pgBackRest/Restore.pm line 1224
        pgBackRest::Restore::process('pgBackRest::Restore=HASH(0x2941a28)') called at /usr/share/perl5/pgBackRest/Main.pm line 220
        eval {...} called at /usr/share/perl5/pgBackRest/Main.pm line 276
        pgBackRest::Main::main('restore') called at (eval 77) line 1
        eval '($result, $message) = pgBackRest::Main::main(\'restore\')
;' called at (eval 77) line 0

[pgbackrest.conf.txt](https://github.com/pgbackrest/pgbackrest/files/2096699/pgbackrest.conf.txt)
[postgresql.conf.txt](https://github.com/pgbackrest/pgbackrest/files/2096700/postgresql.conf.txt)

dwsteele:
Does your cluster only have one user database (excluding template* and postgres)? 
```
select datname, oid from pg_database;
```
[...]

PR #546, opened 2018-06-14
Fixed restore error with db-include when only one user DB - issue #543

cmwshang:


dwsteele:
Committed at 4e7692b74d229538f.

PR #554, opened 2018-06-29
change hour format for AWS x-amz-date header to %H from %k to match ISO-8601

trinchan:
The [AWS docs](https://docs.aws.amazon.com/general/latest/gr/sigv4-date-handling.html) specify ISO-8601 as the format for the `x-amz-date` header:

> The time stamp must be in UTC and in the following ISO 8601 format: YYYYMMDD'T'HHMMSS'Z'. For example, 20150830T123600Z is a valid time stamp. Do not include milliseconds in the time stamp.

The current `strftime` format string uses `%k` for the hour, 

> %k The hour (24-hour clock) as a decimal number (range 0 to 23); single digits are preceded by a blank. (See also %H.) (Calculated from tm_hour.)  (TZ)

which for single digit hours, inserts a space before the digit. Some S3 API implementations are strict about this format which causes failures during these hours. You can imagine the headaches debugging :).

The PR proposes changing the format string hour to `%H` to match ISO-8601

> %H The hour as a decimal number using a 24-hour clock (range 00 to 23). (Calculated from tm_hour.)

### DIFF
--- a/lib/pgBackRest/Storage/S3/Auth.pm
+++ b/lib/pgBackRest/Storage/S3/Auth.pm
@@ -62,7 +62,7 @@ sub s3DateTime
     return logDebugReturn
     (
         $strOperation,
-        {name => 'strDateTime', value => strftime("%Y%m%dT%k%M%SZ", gmtime($lTime)), trace => true}
+        {name => 'strDateTime', value => strftime("%Y%m%dT%H%M%SZ", gmtime($lTime)), trace => true}
     );
 }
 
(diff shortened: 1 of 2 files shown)


dwsteele:
> You can imagine the headaches debugging :).

No doubt.  Hard to believe this one lasted so long.

How much pain is this causing you?  Seems like it might be worth a release.

dwsteele:
Committed at 1bd98b61df7aeae.

I also updated the static date test to check for padded zeroes to prevent regression.

trinchan:
@dwsteele Thank you, new test case looks good too. When do you estimate the 2.0.4 release? Our current build pipeline pulls from a tagged release, but we could probably modify it to pull from a specific commit if it will be a while before the tag.

dwsteele:
It looks we'll be tagging a release this week.  Stay tuned.

PR #557, opened 2018-07-06
Exclude unlogged and temp files from manifest.

cmwshang:


dwsteele:
Committed at bec4c176dc478d5eff85.

PR #559, opened 2018-07-16
Require database catalog version to be passed to Manifest new() when not loading it from disk.

cmwshang:


dwsteele:
Committed at 0acf70541601d997 with minor whitespace/comment edits.

PR #563, opened 2018-08-01
Filters can now produce output that differs from input.

dwsteele:
This allows filters such as compression, encryption, etc. to be implemented.

This commit builds on the IO work already committed in 0ac176b722fe93301ec9f and 01aea0c067c2e667c47.  Have a look at everything in `src/common/io` even if the entire file is not added in this commit.  You can look at the other commits to see how the new IO layer was integrated with the existing storage code.


src/common/io/bufferRead.c
```diff
@@ -114,6 +101,21 @@ ioBufferReadMove(IoBufferRead *this, MemContext *parentNew)
     FUNCTION_TEST_RESULT(IO_BUFFER_READ, this);
 }
 
+/***********************************************************************************************************************************
+Have all bytes been read from the buffer?
+***********************************************************************************************************************************/
+bool
+ioBufferReadEof(IoBufferRead *this)
```
cmwshang: Should this be "const IoBufferRead *this"?

src/common/io/bufferWrite.c
```diff
@@ -63,6 +44,25 @@ ioBufferWriteNew(Buffer *buffer)
     FUNCTION_DEBUG_RESULT(IO_BUFFER_WRITE, this);
 }
 
+/***********************************************************************************************************************************
+Write to the buffer
+***********************************************************************************************************************************/
+void
+ioBufferWrite(IoBufferWrite *this, Buffer *buffer)
```
cmwshang: I don't see ioBufferWrite() ever being called. At a minimum there should be a unit test for it, no?

src/common/io/filter/group.c
```diff
@@ -45,29 +78,100 @@ Add a filter
...
+            {
+                // If this is the first input buffer found, store the index
+                if (lastOutputBuffer == NULL)
+                    this->firstOutputFilter = filterIdx;
+
+                // If this is not the last output filter then create a new output buffer for it.  The output buffer for the last
+                // filter will be provided to the process function.
+                if (filterIdx < lstSize(this->filterList) - 1)
+                {
+                    lastOutputBuffer = bufNew(ioBufferSize());
```
cmwshang: I know storage.c is not in this review, but I don't understand why there are 2 calls to ioBufferSize with parameters (ioBufferSize(source) and ioBufferSize(file)). I cannot find a corresponding function that takes a parameter. Can you help me understand?

dwsteele: This is because we did not have `-Wstrict-prototypes` enabled and I did a search and replace without properly checking my work.  A function declared `()` instead of `(void)` defaults to the old K&R style where params did not need to be explicitly defined.  I thought `-std=c99` forbid this style but apparently not.

I have enabled `-Wstrict-prototypes` and explicitly declared all functions as (void) where needed: 429a356e3323acf99f3d.  The examples you found were the only places where params were being passed in the old K&R way.

src/common/io/bufferRead.c
```diff
@@ -8,21 +8,50 @@ Buffer IO Read
...
+
+        FUNCTION_TEST_ASSERT(buffer != NULL);
+    FUNCTION_DEBUG_END();
+
+    IoBufferRead *this = NULL;
+
+    MEM_CONTEXT_NEW_BEGIN("IoBufferRead")
+    {
+        this = memNew(sizeof(IoBufferRead));
+        this->memContext = memContextCurrent();
```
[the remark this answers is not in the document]
dwsteele: The difference is that MEM_CONTEXT_NEW() is only valid in a MEM_CONTEXT_NEW* block but memContextCurrent() is always valid.  In this case they are the same so it doesn't matter which one you use.

I'm tempted to remove MEM_CONTEXT_NEW() but I like the symmetry with MEM_CONTEXT_OLD().  Thoughts?

src/common/io/filter/group.c
```diff
@@ -29,10 +60,12 @@ ioFilterGroupNew()
 
     IoFilterGroup *this = NULL;
 
-    MEM_CONTEXT_NEW_BEGIN("IoFilterGroup")
+    MEM_CONTEXT_NEW_BEGIN("IoFilterHash")
```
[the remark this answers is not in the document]
dwsteele: It's a typo.

src/common/io/filter/group.c
```diff
@@ -1,26 +1,57 @@
...
+#define FUNCTION_DEBUG_IO_FILTER_DATA_FORMAT(value, buffer, bufferSize)                                                            \
+    objToLog(value, "IoFilterData", buffer, bufferSize)
+
+/***********************************************************************************************************************************
+Object type
 ***********************************************************************************************************************************/
 struct IoFilterGroup
 {
     MemContext *memContext;                                         // Mem context
     List *filterList;                                               // List of filters to apply to read buffer
```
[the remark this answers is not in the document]
dwsteele: Comment left over from an old version or a straight-up typo.

src/common/io/bufferWrite.c
dwsteele: It is unit tested via the IoWrite interface so you won't see any direct calls to it.  Since this is the way it will be used in practice it's the best way to test it.  And the IoWrite code is tested at the same time.

src/common/io/filter/group.c
```diff
@@ -121,40 +329,101 @@ ioFilterGroupClose(IoFilterGroup *this)
 
         MEM_CONTEXT_TEMP_BEGIN()
         {
-            kvPut(this->filterResult, varNewStr(ioFilterType(filter)), filterResult);
+            kvPut(this->filterResult, varNewStr(ioFilterType(filterData->filter)), filterResult);
```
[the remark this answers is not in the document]
dwsteele: I changed this to a kvAdd() and added a test for getting multiple results from the same filter type.  Not sure why we would need it, but it was simpler to test and provides more functionality.

dwsteele:
Committed at e3ff6b209d9cf50.

PR #581, opened 2018-08-24
Allow manifest reference to zero-sized file regardless of timestamp.

cmwshang:


dwsteele:
Committed at eb30d88b6a6048353b26.  Renamed "bogus" test file to "zero_from_start".

PR #582, opened 2018-08-28
Directly check for ok/error status files

sfrost:
Prior to this commit, an expression was used to search the spool
directory for ok/error files for a specific WAL segment.  This involved
setting up a regular expression and using opendir/readdir.

Instead, directly probe for the status files, checking directly if a
'.ok' or '.error' file exists, avoiding the regular expression and
eliminating the directory scan.

Only the two files now probed for could have ever matched the regular
expression which had been provided and it's unlikely that many more
additional files will be added, so this is a good improvement, and
optimization, with little downside.


dwsteele:
Committed at 4498f726a59132421f with minor changes.

PR #594, opened 2018-09-17
Allow delta option to be specified in the config file.

cmwshang:


dwsteele:
Also, moved this to the "restore" section in the reference, in case we use a different option name for delta backup.

PR #602, opened 2018-09-29
Avoid calling strlen() twice when we know the length already.

sfrost:
Similar to whats done in bufNewC().

Entirely untested. ;)


dwsteele:
Committed at 3be9f2dee79c02e.

PR #603, opened 2018-09-29
Ignore the src/pgbackrest that gets built when doing make...

sfrost:
Would be nice to have this so that a 'make' doesn't make it look like there's something to commit. :)

### DIFF
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -1,0 +1,2 @@
+*.o
+src/pgbackrest


.gitignore
```diff
@@ -1,3 +1,4 @@
 **/*~
 *~
 *.swp
+src/pgbackrest
```
dwsteele: Wouldn't we want to ignore `.o` files, etc., as well?  The tests don't actually build in this directory which is why we haven't done this yet.

Also, I'd prefer to have these exclusions in `src/.gitignore`.

sfrost: Oh, perhaps...  My global gitignore includes:

tags
.*.swp
cscope.in.out
cscope.out
cscope.po.out
*.o

I didn't want to put this into my global one though.  I'm fine with adding *.o to the pgbackrest one.

I'm fine with putting the entry wherever works. :)

dwsteele: Yeah, add `.o` and move it to `src/.gitignore`.  The idea is if we get rid of dirs (not this one I hope) then all the exclusions go with it.

sfrost: ok, updated the branch with that change and did a force-push to it and seems to have correctly updated the PR. :)

dwsteele:
Committed at ed5d7a53de7cd.

PR #607, opened 2018-10-03
Automatically enable delta backup

cmwshang:
1) Delta backup automatically enabled for:
- detection of a timeline switch since the prior backup
- a file timestamp that is currently older than the prior backup
- a file size change that did not change the file timestamp
- a file timestamp is in the future compared to the start of the backup copying
- the online option has changed  since the prior backup 
2) Throw error when no files exist to be backed up.
3) Update user-guide to perform a proper PG upgrade for co6. 
4) Add 100% test coverage for resumeClean function.


lib/pgBackRest/Manifest.pm
```diff
@@ -556,6 +556,177 @@ sub isTargetTablespace
...
+    return logDebugReturn
+    (
+        $strOperation,
+        {name => 'bDelta', value => $bDelta, trace => true},
+    );
+}
+
+####################################################################################################################################
+# copyPriorChecksum
+#
```
dwsteele: What is the purpose of this function?  It causes code churn (which needs review) but is only called from one place.

lib/pgBackRest/Backup/Backup.pm
```diff
@@ -91,6 +99,56 @@ sub resumeClean
...
+                    if ($bCompressed)
+                    {
+                        $strFile = substr($strFile, 0, length($strFile) - 3);
+                    }
+
+                    # To be preserved the file must exist in the new manifest and not be a reference to a previous backup and must
+                    # have a checksum
+                    if ($oManifest->test(MANIFEST_SECTION_TARGET_FILE, $strFile) &&
+                        !$oManifest->test(MANIFEST_SECTION_TARGET_FILE, $strFile, MANIFEST_SUBKEY_REFERENCE) &&
+                        $oAbortedManifest->test(MANIFEST_SECTION_TARGET_FILE, $strFile, MANIFEST_SUBKEY_CHECKSUM))
```
dwsteele: Shouldn't we be checking size here as well?

PR #613, opened 2018-10-17
jsonToKv enhancement for parsing boolean and simple arrays.

cmwshang:


dwsteele:
Go ahead and submit this with the rest of the info code.  We'll pull this bit into a separate commit then.

PR #620, opened 2018-10-24
Change infoArchiveCheckPg() to display the PG version as a string when throwing an error.

cmwshang:


dwsteele:
Committed at d301720c581.

PR #625, opened 2018-11-20
Add TlsClient object.

dwsteele:
This is a simple, secure TLS client intended to allow access to services that are exposed via HTTPS.  We call it TLS instead of SSL because SSL methods are disabled so only TLS connections are allowed.

This object is intended to be used for multiple TLS connections against a service so tlsClientOpen() can be called each time a new connection is needed.  By default, an open connection will be reused for pipelining so the user must be prepared to retry their transaction on a read/write error if the server closes the connection before it can be reused.  If this behavior is not desirable then tlsClientClose() may be used to ensure that the next call to tlsClientOpen() will create a new TLS session.

Note that tlsClientRead() is non-blocking unless there are *zero* bytes to be read from the session in which case it will raise an error after the defined timeout.  In any case the tlsClientRead()/tlsClientWrite()/tlsClientEof() functions should not generally be called directly.  Instead use the read/write interfaces available from tlsClientIoRead()/tlsClientIoWrite().


src/common/io/tls/client.c
```diff
@@ -0,0 +1,526 @@
...
+        cryptoError(method == NULL, "unable to load TLS method");
+
+        // Create the SSL context
+        this->context = SSL_CTX_new(method);
+        cryptoError(this->context == NULL, "unable to create TLS context");
+
+        memContextCallback(this->memContext, (MemContextCallback)tlsClientFree, this);
+
+        // Exclude SSL versions to only allow TLS and also disable compression
+        SSL_CTX_set_options(this->context, (long)(SSL_OP_ALL | SSL_OP_NO_SSLv2 | SSL_OP_NO_SSLv3 | SSL_OP_NO_COMPRESSION));
```
sfrost: Do we really need SSL_OP_ALL here?  Doesn't look like PostgreSQL has that?

test/src/module/common/ioTlsTest.c
```diff
@@ -0,0 +1,218 @@
...
+
+        TEST_ERROR(
+            tlsClientHostVerifyName(
+                strNew("host"), strNewN("ab\0cd", 5)), CryptoError, "TLS certificate name contains embedded null");
+
+        TEST_ERROR(tlsClientHostVerify(strNew("host"), NULL), CryptoError, "No certificate presented by the TLS server");
+
+        TEST_RESULT_BOOL(tlsClientHostVerifyName(strNew("host"), strNew("**")), false, "invalid pattern");
+        TEST_RESULT_BOOL(tlsClientHostVerifyName(strNew("host"), strNew("*.")), false, "invalid pattern");
+        TEST_RESULT_BOOL(tlsClientHostVerifyName(strNew("a.bogus.host.com"), strNew("*.host.com")), false, "invalid host");
```
cmwshang: 1) Would like to see a couple tests where tlsClientVerifyName is true: www.host.com, *.host.com should return true.
2) Also, are we supporting Subject Alternative Names such that host.com would match *.host.com (see https://en.wikipedia.org/wiki/Wildcard_certificate and https://en.wikipedia.org/wiki/Subject_Alternative_Name)

src/common/io/tls/client.c
```diff
@@ -0,0 +1,544 @@
...
+    // Else check if a wildcard certificate matches the host name
+    //
+    // The rules are:
+    // 1. Only match the '*' character as wildcard
+    // 2. Only match wildcards at the start of the string
+    // 3. The '*' character does *not* match '.', meaning that we match only a single pathname component.
+    // 4. Don't support more than one '*' in a single pattern.
+    //
+    // This is roughly in line with RFC2818, but contrary to what most browsers appear to be implementing (point 3 being the
+    // difference)
```
cmwshang: I think 3. does not allow for Subject Alternative names - and maybe we should? (https://en.wikipedia.org/wiki/Wildcard_certificate#Example)

cmwshang: I see you do mention SAN support below, so my only comment is if this means the "naked" domain should be considered valid.

src/common/io/tls/client.c
```diff
@@ -0,0 +1,544 @@
...
+    const String *host, unsigned int port, TimeMSec timeout, bool verifySsl, const String *caFile, const String *caPath)
+{
+    FUNCTION_DEBUG_BEGIN(logLevelDebug)
+        FUNCTION_DEBUG_PARAM(STRING, host);
+        FUNCTION_DEBUG_PARAM(UINT, port);
+        FUNCTION_DEBUG_PARAM(TIME_MSEC, timeout);
+        FUNCTION_DEBUG_PARAM(BOOL, verifySsl);
+        FUNCTION_DEBUG_PARAM(STRING, caFile);
+        FUNCTION_DEBUG_PARAM(STRING, caPath);
+    FUNCTION_DEBUG_END();
```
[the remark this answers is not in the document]
dwsteele: host is the only param that must not be bull.  Added.

test/src/module/common/ioTlsTest.c
dwsteele: There are more tests below in run 2.  It doesn't seem worth it to duplicate.  As the comment says, `Additional coverage not provided by testing with actual certificates`.

We are following RFC2818 here -- as Postgres does.  In fact this code is pulled from Postgres.  The wiki seems to say the same thing as I read it: `*.domain.com is OK. It will match www.domain.com but not domain.com and not zzz.www.domain.com`

src/common/io/tls/client.c
dwsteele: The naked domain is valid if there is an exact match for it.

test/src/module/common/ioTlsTest.c
dwsteele: The wiki and RFC also mention partial wildcards but we don't support them.

src/common/io/tls/client.c
```diff
@@ -0,0 +1,550 @@
...
+void
+tlsClientWrite(TlsClient *this, const Buffer *buffer)
+{
+    FUNCTION_DEBUG_BEGIN(logLevelTrace);
+        FUNCTION_DEBUG_PARAM(TLS_CLIENT, this);
+        FUNCTION_DEBUG_PARAM(BUFFER, buffer);
+
+        FUNCTION_TEST_ASSERT(this != NULL);
+        FUNCTION_TEST_ASSERT(this->session != NULL);
+    FUNCTION_DEBUG_END();
```
cmwshang: Shouldn't buffer be required and not NULL, so is there a missing ASSERT here?

src/common/io/tls/client.h
```diff
@@ -0,0 +1,63 @@
+/***********************************************************************************************************************************
+TLS Client
+
+This is a simple, secure TLS client intended to allow access to services that are exposed via HTTPS.  We call is TLS instead of SSL
```
cmwshang: Typo: "We call is TLS" should be "We call it TLS"

test/certificate/README.md
```diff
@@ -0,0 +1,48 @@
+# pgBackRest Test Certificates
+
+The certificates in this directory are used for testing purposes only and are not used for actual services.  They are used only by the unit an integration tests and there should be no reason to modify them unless new tests are required.
```
cmwshang: Typo: "the unit and integration" should be "the unit and integration"

src/common/io/tls/client.c
dwsteele: Yes, fixed.

src/common/io/tls/client.h
dwsteele: Fixed.

test/certificate/README.md
dwsteele: Fixed.

dwsteele:
I saw three comments in your final round.  Thanks!

dwsteele:
Committed at 1dd06a6e460049ff8.

Issue #630, opened 2018-12-03
Info output breaks visual backup chronology after --stanza-upgrade


rustprooflabs:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version:  2.07

2. PostgreSQL version:  Upgraded 9.5.15 to 11.1

3. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each:  Ubuntu Linux 16.04

4. Did you install pgBackRest from source or from a package?  Package

5. Please attach the following as applicable:
    - `pgbackrest.conf` file(s)
    - `postgresql.conf` settings applicable to pgBackRest (`archive_command`, `archive_mode`, `listen_addresses`, `max_wal_senders`, `wal_level`, `port`)
    - errors in the postgresql log file before or during the time you experienced the issue
    - log file in `/var/log/pgbackrest` for the commands run (e.g. `/var/log/pgbackrest/mystanza_backup.log`)

Output from command in question:

```
~$ sudo -u pgbackrest pgbackrest --stanza=pgdev01 info
stanza: pgdev01
    status: ok
    cipher: none

    db (current)
        wal archive min/max (11-2): 0000000100000000000000CA / 0000000100000000000000CC

        full backup: 20181202-132211F
            timestamp start/stop: 2018-12-02 13:22:11 / 2018-12-02 13:23:37
            wal start/stop: 0000000100000000000000CA / 0000000100000000000000CA
            database size: 190.8MB, backup size: 190.8MB
            repository size: 36.8MB, repository backup size: 36.8MB

        incr backup: 20181202-132211F_20181202-132524I
            timestamp start/stop: 2018-12-02 13:25:24 / 2018-12-02 13:25:34
            wal start/stop: 0000000100000000000000CC / 0000000100000000000000CC
            database size: 190.8MB, backup size: 18.7MB
            repository size: 36.8MB, repository backup size: 16.3MB
            backup reference list: 20181202-132211F

    db (prior)
        wal archive min/max (9.5-1): 00000001000000000000001C / 000000040000000000000062

        full backup: 20180622-114546F
            timestamp start/stop: 2018-06-22 11:45:46 / 2018-06-22 11:47:03
            wal start/stop: 00000001000000000000001C / 00000001000000000000001C
            database size: 49.8MB, backup size: 49.8MB
            repository size: 20.1MB, repository backup size: 20.1MB

        full backup: 20180622-140134F
[...]

PR #638, opened 2018-12-14
Reorder info command text output to always displays backups oldest to newest.

cmwshang:
Addresses issue #630


dwsteele:
Committed at 35bbb5bd6.

PR #668, opened 2019-02-07
Addition of some config parameters to backup.manifest

blogh:
Hi,
I would like to propose the following PR.

Commit msg:
```
The following config options have been added to the backup manifest. It helps to explain
changes in volume, backup duration etc.. when the analysis is done after the fact using
the backup.history manifest files.

* CFGOPT_PROCESS_MAX
* CFGOPT_COMPRESS
* CFGOPT_COMPRESS_LEVEL
* CFGOPT_COMPRESS_LEVEL_NETWORK
* CFGOPT_BUFFER_SIZE
```
I wasn't sure about which option to add so I included only the one I would use for my purpose.
I hope it's acceptable.

Benoit.


dwsteele:
This seems reasonable to me.  We'll look at it for the next release.

dwsteele:
Committed at e4e2606fce with some tweaks to get the tests working.

PR #674, opened 2019-02-11
Change execRead to return a size_t

sfrost:
execRead should be returning a size_t, not a void.  Thankfully, this
isn't actually used much and therefore shouldn't be an issue, but we
should fix it anyway.


dwsteele:
Committed at 2cd204f38037.

PR #675, opened 2019-02-12
Info command : Check that the wal segments needed for the backups exist on disk

pgstef:
Hi,

I would like to suggest the following PR.

Commit msg:

```
check that the wal segments needed for the backups exist on disk 
```

In the info command, archiveDbList only look at the oldest and newest archived wal segment found on disk.

The idea here is to keep track of the first and last needed archive for the consistency of the backups given by the backupList and check if it exists on disk.

A next step would be to validate the archive chain to make sure there's nothing missing. Maybe in the info command itself with some "--validate-archive-chain" parameter or in a new separated command. Maybe should I open a separated issue to discuss this idea?

Kind regards

### DIFF
--- a/src/command/info/info.c
+++ b/src/command/info/info.c
@@ -69,6 +69,8 @@ STRING_STATIC(INFO_STANZA_STATUS_MESSAGE_MISSING_STANZA_PATH_STR,   "missing sta
 STRING_STATIC(INFO_STANZA_STATUS_MESSAGE_NO_BACKUP_STR,             "no valid backups");
 #define INFO_STANZA_STATUS_CODE_MISSING_STANZA_DATA                 3
 STRING_STATIC(INFO_STANZA_STATUS_MESSAGE_MISSING_STANZA_DATA_STR,   "missing stanza data");
+#define INFO_STANZA_STATUS_CODE_MISSING_WAL_SEG                     4
+STRING_STATIC(INFO_STANZA_STATUS_MESSAGE_MISSING_WAL_SEG_STR,       "missing wal segment(s)");
 
 /***********************************************************************************************************************************
 Set error status code and message for the stanza to the code and message passed.
@@ -82,7 +84,7 @@ stanzaStatus(const int code, const String *message, Variant *stanzaInfo)
         FUNCTION_TEST_PARAM(VARIANT, stanzaInfo);
     FUNCTION_TEST_END();
 
-    ASSERT(code >= 0 && code <= 3);
+    ASSERT(code >= 0 && code <= 4);
     ASSERT(message != NULL);
     ASSERT(stanzaInfo != NULL);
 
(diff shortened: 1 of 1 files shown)


dwsteele:
This looks pretty good in principle.  I won't be able to have a close look at it until next week, though.

We definitely need to make the more in-depth check optional.  We used to scan all the files and it caused problems for users with millions of WAL segments.  `info` needs to return very quickly, at least by default.

This patch will break coverage if the unit tests are run so you'll need to add tests.  I'll get information about that into `test/README.md` next week.  You can also test through travis-ci though I like to do local tests first.

dwsteele:
Is any part of this PR intended to fix the missing min/max WAL in 2.10 that you mention in the docs for `check_pgbackrest`?  If so, I'd I like to get that committed first.  If not, please submit an issue for that (or I can do so if you prefer).

pgstef:
Hi,

Indeed, a line in the last commit should solve that issue.

In src/command/info/info.c, archiveDbList function, line 122 : replace STORAGE_REPO_ARCHIVE by STORAGE_PATH_ARCHIVE.

Do you wish I create another PR for that?

dwsteele:
OK, I thought that might be it.  No need for another PR, just wanted to make sure we were addressing the issue.

dwsteele:
The min/max issue has been fixed at 80df1114bdc9.

dwsteele:
Stefan,

Sorry I have let this sit for so long.  Partly I have not found the time to add more documentation about testing, and partly I'm not sure that this patch is the right way to go.

I'm concerned that adding this much IO to the info command will be an issue.  We used to do something similar and it was a real problem for users that generate very large quantities of WAL.  Of course we can make it optional but I think most users would want to run it.

What we want, I think, is to integrate this logic with archive-push and create summaries that the info command can use to quickly report on missing WAL files.  Thoughts?

pgstef:
Hi David,

I've implemented this check directly in check_pgbackrest. I understand we want to keep the info command as fast as possible. That check can take some time with the amout of WAL's generated. I hesitated to create some new "validation" command in pgbackrest itself for that but found it easier outside to integrate with the monitoring tools.

Archive-push might get a list of the copied wal's and flag the "holes" for "archive-*-queue-max" but how would you detect that someone removed manually an archive (for example) ? At some point we have to go on disk and check that the files are there. Ask archive-push to check the previous archive? (might not work with async mode and slow the push)

As you mentionned in some other thread, we might work to strengthen the link between pgbackrest and check_pgbackrest to avoid slowing pgbackrest itself or integrate those checks into a new command ?

We might still discuss the topic but I agree with you, this patch can be dropped.

Thanks for your thoughts,
Kind regards

dwsteele:
I think that implementing this logic in check_pgbackrest will get more burdensome as we add storage drivers, so that's something to consider.

I agree that some sort of verification function would be needed in addition to having archive-push doing the work as it goes.  But verification could be done perhaps once a day -- or be run after a backup if it has not run in at least a day.

I also think it is OK to have info do a little bit of work -- just the files at the end that archive-push has not written out status files for yet.  I just don't want info going out and sweeping across the entire archive every time it is run.

I'll close this patch but let's keep it in mind as we go.  Perhaps we can talk about it in Milan if we haven't done something before then.

Issue #636, opened 2018-12-13
archive-get command can block indefinitely


rdunklau:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version:
pgBackRest 2.06

2. PostgreSQL version: 
9.4, 9.6

3. Operating system/version
Debian Stretch

4. Did you install pgBackRest from source or from a package?
From the PGDG packages

7. Describe the issue:

We have experienced the same problem on two different servers, with four different stanzas. Sometimes, pgbackrest archive-get can block undefinitely trying to read from a socket while trying to fetch WALs from S3. The socket is connected to S3. The only solution for us was to kill the pgbackrest master process.

Maybe  there is a timeout missing while trying to read at this point ?

Our relevant configuration information:

```
[global]
archive-async=y
archive-get-queue-max=4294967296
process-max=8
repo1-cipher-type=aes-256-cbc
repo1-type=s3
```

rdunklau:
Thanks for your answer.
We don't have a proxy, but our hosting provider most certainly do have a firewall, which may not send any RST packet.
Does pgbackrest use tcp keepalive ?

dwsteele:
Not sending RST packets would be awkward in this case.  Any chance you can convince the firewall to do so?

We have a new implementation in C that uses select() for all reads so it should timeout in this case.   At the moment though, that only works with synchronous archive-get and will come out with 2.08.

I'm not sure that I want to teach this same trick to Perl since we'll be replacing that code soon.

rdunklau:
Unfortunately, that's how a wide range of firewalls work.
Investigating the issue further, I've noticed that most socket opened by pgbackrest do use tcp keepalive, but the one currently blocking on reads does not. A simple workaround that would probably do the trick for us would be to make sure those sockets use tcp keealive, too.

dwsteele:
Added TCP keepalives at 23b583336f30.

The C code also sets TCP_KEEPIDLE, TCP_KEEPINTVL, and TCP_KEEPCNT (when available) to sensible values based on the timeout.  For Perl, system defaults will be used since these settings are not supported on a per-connection basis.

BTW, the C version of archive-get is about twice as fast since we added pipe-lining to that implementation.  Either way, you'll be getting something in 2.08.

So, how much pain is this causing you?  We were planning to release after the holidays.  There's a lot of new code in 2.08 and this seems like a bad time of year to break things.

rdunklau:
Thank you !

[...]

PR #679, opened 2019-02-20
Fix keepalive on the SSL connection

marco44:
Only perl 5.26+ supports passing sockopts directly to the instantiation
method

Should solve #636

### DIFF
--- a/lib/pgBackRest/Common/Http/Client.pm
+++ b/lib/pgBackRest/Common/Http/Client.pm
@@ -99,9 +99,11 @@ sub new
 
             eval
             {
-                $oSocket = IO::Socket::SSL->new(
+                $oSocket = IO::Socket::IP->new(PeerHost => $strHost, PeerPort => $iPort);
+                setsockopt($oSocket,SOL_SOCKET,SO_KEEPALIVE,1);
+                IO::Socket::SSL->start_SSL($oSocket,
                     PeerHost => $strHost, PeerPort => $iPort, SSL_verify_mode => $bVerifySsl ? SSL_VERIFY_PEER : SSL_VERIFY_NONE,
-                    SSL_ca_path => $strCaPath, SSL_ca_file => $strCaFile, Sockopts => [[SOL_SOCKET, SO_KEEPALIVE]]);
+                    SSL_ca_path => $strCaPath, SSL_ca_file => $strCaFile);
 
                 return 1;
             }
--- a/lib/pgBackRest/Common/Http/Client.pm
+++ b/lib/pgBackRest/Common/Http/Client.pm
@@ -96,10 +96,18 @@ sub new
         {
             # Connect to the server
             my $oSocket;
-
+            # find out what's available as a lib
             eval
             {
-                $oSocket = IO::Socket::IP->new(PeerHost => $strHost, PeerPort => $iPort);
+                if(eval{require IO::Socket::IP})
+                {
+                    $oSocket = IO::Socket::IP->new(PeerHost => $strHost, PeerPort => $iPort);
+                }
+                else
+                {
+                    require IO::Socket::INET;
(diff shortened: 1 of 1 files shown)


dwsteele:
I played with this patch for a bit and with some tweaking was able to get it working on Perl 5.26 but it fails for 5.10, 5.14, and 5.18 as far as I can see.  Plus, it requires the `IO::Socket::IP` module which may not be a bad thing but will cause some work for the packagers.

I'm not sure how much effort I want to put into code that we are actively trying to get rid of.  If you can get me a patch what works across our test suite I'll be happy to look at it.

marco44:
Ok, I probably can work around that… if there is a way I can view the CI reports

dwsteele:
Also -- our expectation is to have the `archive-get` and `archive-push` commands migrated entirely to C in the next release.  My understanding is that this is where you are seeing the issue.

marco44:
Anyway I don't think the code needs IO::Socket::IP, We could use IO::Socket::INET. I just chose to use IO::Socket::IP because that's what IO::Socket::SSL would have used behind the scenes on a Perl 5.24 or 5.26.

Anyway, yes, as soon as this code is removed, what we do here is of no use. But we're having this problem often these days, and it means having offsite replicas outdated for us, so I'm still trying to quick fix this in the meantime.

Do you want me to try to test a version using INET and then try to push it here ? What kind of tweaking did you need with Perl 5.26 ?

dwsteele:
We can't directly use `IO::Socket::INET` because that means no ipv6 support for those who do have a ipv6 compatible module installed, like `IO::Socket::IP` or `IO::Socket::INET6`.

I was trying something along these lines:
```
my $sock = IO::Socket::SSL->new(
    PeerAddr => ...
    SSL_startHandshake => 0,
    %sslargs
) or die $!;
... exchange plain data on $sock until starttls command ...
$sock->connect_SSL or die $SSL_ERROR;
... now $sock is an IO::Socket::SSL object ...
... exchange data with SSL on $sock until stoptls command ...
$sock->stop_SSL or die $SSL_ERROR;
... $sock is still an IO::Socket::SSL object ...
... but data exchanged again in plain ...
```
As recommended in https://metacpan.org/pod/IO::Socket::SSL.

But on older versions of Perl I was getting `invalid socket` errors when calling connect_SSL/start_SSL   even though  `IO::Socket::SSL->new()` was successful.

marco44:
Ok, then maybe I can try the 3 modules in order in my patch ?

dwsteele:
Perhaps just `IO::Socket::INET` and `IO::Socket::IP` would be OK.  `IO::Socket::INET` comes with stock Perl so that's safe, and `IO::Socket::IP` can be the optional package.  Up to you, though.

marco44:
Ok, a try here, but it's a complete shot in the dark, as I don't have the travis output :)

dwsteele:
That worked pretty well: https://travis-ci.org/dwsteele/pgbackrest-dev/builds/497178406

The only failures are in tests that check specific error messages.  Not too surprising that those would change based on the modules that got loaded.  Easy to fix, but first...

You should apply this patch to the 2.10 release to make sure it does what you are expecting.  @rdunklau thought this was fixed in 2.08 but it wasn't quite right.  Let's be sure this time.

[socket-patch.txt](https://github.com/pgbackrest/pgbackrest/files/2895579/socket-patch.txt)

marco44:
Hi, we'll put this on our test server on monday...

marco44:
Hi,

We just applied this patch. The socket now has keepalive. So this works...

dwsteele:
This took some tweaking of the tests due to the different modules used, but looks good now.  Committed at cb3b4fa24bbe271a517.

Note that the `archive-get` command has been migrated entirely into C for the next release, see db4b447be89878496f0a.

Issue #662, opened 2019-01-31
correct the usage of DESTDIR in Makefile


hunleyd:
Was compiling 2.09 this morning and trying not to edit the Makefile but use env overrides as you suggested in my last issue, and I realized that I still have to edit the Makefile since it appears DESTDIR is used incorrectly.

Existing Makefile:
line 54: `DESTDIR =`
line 169: `install -d $(DESTDIR)/usr/bin`
line 170: `install -m 755 pgbackrest $(DESTDIR)/usr/bin`

Which means if I set DESTDIR to anything it gets prepended to `/usr/bin` (e.g. setting it to `/usr/local` would get me `/usr/local/usr/bin`). This is incorrect. It should be:

line 54: `DESTDIR = /usr/bin`
line 169: `install -d $(DESTDIR)`
line 170: `install -m 755 pgbackrext $(DESTDIR)`

Which would give `/usr/bin/pgbackrest` by default, and allow me to set `make DESTDIR=/usr/local/bin -C src install` and have it install to `/usr/local/bin/pgbackrest` correctly.

You want a PR for this?

dwsteele:
Committed at 786e77a9.

PR #689, opened 2019-03-07
Makefile misc

fluca1978:
A few minor changes to the makefile with the hope of making it a little more usable.


dwsteele:
The clean and uninstall targets are easy enough.  

Not sure about DESTDIR, though.  I know that making the change you or @hunleyd (#662) have suggested will break the Debian package build.  But, looking at the package source, it seems like we just need to update:

https://salsa.debian.org/postgresql/pgbackrest/blob/master/debian/rules#L37

@disco-stu Does that look right to you?

dwsteele:
Committed clean and uninstall targets at 3d7edc4c.

disco-stu:
The purposed patch looks good. I'll update the Debian package accordingly.

dwsteele:
@disco-stu I'll need to write a patch anyway or the integration tests won't pass (since they build packages).  I'll provide that for the release (coming soon), though I doubt you'll need it.

dwsteele:
The DESTDIR part of this was commited at 786e77a based on Doug's patch in #662.

Issue #686, opened 2019-03-04
Cannot compile on FreeBSD 12


fluca1978:
On FreeBSD 12-RELEASE amd64, I've tried to compile version 2.10 (release) and HEAD, both failing:

```
% gmake -s -C src 
common/io/tls/client.c: In function 'tlsClientOpen':
common/io/tls/client.c:298:41: error: 'IPPROTO_TCP' undeclared (first use in this function); did you mean 'EPROTOTYPE'?
hints.ai_protocol = IPPROTO_TCP;
^~~~~~~~~~~
EPROTOTYPE
compilation terminated due to -Wfatal-errors.
gmake: *** [Makefile:284: common/io/tls/client.o] Error 1
```

I've installed libressl-devel, if that matters.
The `gcc` I'm using is:

```
% gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/local/libexec/gcc8/gcc/x86_64-portbld-freebsd12.0/8.2.0/lto-wrapper
Target: x86_64-portbld-freebsd12.0
Configured with: /wrkdirs/usr/ports/lang/gcc8/work/gcc-8.2.0/configure --with-build-config=bootstrap-debug --disable-nls --enable-gnu-indirect-function --libdir=/usr/local/lib/gcc8 --libexecdir=/usr/local/libexec/gcc8 --program-suffix=8 --with-as=/usr/local/bin/as --with-gmp=/usr/local --with-gxx-include-dir=/usr/local/lib/gcc8/include/c++/ --with-ld=/usr/local/bin/ld --with-pkgversion='FreeBSD Ports Collection' --with-system-zlib --enable-languages=c,c++,objc,fortran --prefix=/usr/local --localstatedir=/var --mandir=/usr/local/man --infodir=/usr/local/share/info/gcc8 --build=x86_64-portbld-freebsd12.0
Thread model: posix
gcc version 8.2.0 (FreeBSD Ports Collection)
```

Am I missing something?

fluca1978:
Apparently, with the following diff I can compile and it seems to start.

```
% git diff
diff --git a/src/Makefile b/src/Makefile
index 73672bff..b2425b55 100644
--- a/src/Makefile
+++ b/src/Makefile
@@ -8,7 +8,7 @@
CC=gcc
# Compile using C99 and Posix 2001 standards (also _DARWIN_C_SOURCE for MacOS)
[...]

PR #690, opened 2019-03-07
Detect if the running system is Linux or not.

fluca1978:
Borrowed from pgTAP, there is a new script that can detect the OS name
(lowercase) and therefore allows the Makefile to change compilation flags
depending on the OS name,

This however does not allow for a plain compilation on FreeBSD, because of a missing dependency in the includes. See <https://github.com/pgbackrest/pgbackrest/issues/686>. However, I believe this could be a good starting point.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -2,13 +2,21 @@
 # pgBackRest Makefile
 ####################################################################################################################################
 
+# Determine the OS. Borrowed from pgTAP Makefile
+OSNAME := $(shell $(SHELL) ./getos.sh)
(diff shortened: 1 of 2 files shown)


dwsteele:
I'd prefer to use autoconf instead.  This could also solve your problem with the default config directories.

dwsteele:
We added autoconf support at 027c2638719dffa9ba99250085c403e89a2a8a9a -- feel free to adjust `src/configure.ac` to work with *BSD.

PR #728, opened 2019-05-13
Allow stanza-delete force to by-pass database checks so a stanza can be deleted without spinning up a database.

cmwshang:


dwsteele:
Committed at 18d4cb574180d2.

PR #740, opened 2019-06-07
src/Makefile.in: Add CPPFLAGS

df7cb:
Debian's build log hardening check (blhc) noticed that
-D_FORTIFY_SOURCE=2 is missing in the gcc invocations. This is from
CPPFLAGS, add it to the CMAKE variable.

Cc: @disco-stu

### DIFF
--- a/src/Makefile.in
+++ b/src/Makefile.in
@@ -23,7 +23,7 @@ CINCLUDE = @CINCLUDE@
 CLIBRARY = @CLIBRARY@
 
 # Concatenate options for easy usage
-CMAKE = $(CSTANDARD) $(COPTIMIZE) $(CWARNING) $(CINCLUDE) $(CLIBRARY)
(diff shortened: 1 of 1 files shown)


dwsteele:
We already added this at 12bca3c43eb for the next release.  Does that look OK?

df7cb:
Yeah that looks like the same thing. Thanks, and sorry for the noise!

Btw, you should probably look into separating the compile rules from the dependencies. Doesn't make much sense to repeat the same CC command 50 times.

dwsteele:
> Btw, you should probably look into separating the compile rules from the dependencies. Doesn't make much sense to repeat the same CC command 50 times.

We could definitely combine some stuff (all the flags into CMAKE) -- but some files do have additional custom params.  Since all of this is auto-generated I thought it would be better to be verbose.

PR #746, opened 2019-06-14
Rename infoArchiveNew and infoBackupNew & remove unused param.

cmwshang:
Rename infoArchiveNew and infoBackupNew to infoArchiveNewLoad and infoBackupNewLoad. Remove ignoreMIssing parameter as it is never used and should not be moving forward.


dwsteele:
Committed at 44bafc127d9d4e38.

PR #754, opened 2019-06-21
Update expire.c to error if archive.info history does not have an exact correspondence to backup.info history.

cmwshang:
If there is a mismatch then something may be amiss with the repo so it is better not to expire any archive until the issue can be remediated.


dwsteele:
Committed at b498188f01f8d.

PR #760, opened 2019-06-26
Reverse loop in infoPgSave for consistency and readability with infoPgNewLoad.

cmwshang:
Reverse loop in infoPgSave for consistency and readability with infoPgNewLoad.


dwsteele:
Committed at 33e3d316d63a41.

PR #789, opened 2019-07-25
Add PostgreSQL query client

dwsteele:
This direct interface to libpq allows simple queries to be run against PostgreSQL and supports timeouts.

Testing is performed using a shim that can use scripted responses to test all aspects of the client code.  The shim will be very useful for testing backup scenarios on complex topologies.


cmwshang:
Review complete. Only item found was discussed and corrected.

dwsteele:
Committed at 415542b4a3589ff.

PR #802, opened 2019-08-07
Implement start and stop commands in c.

cmwshang:


dwsteele:
This fixes the unit test failure: https://github.com/dwsteele/pgbackrest-dev/compare/3b479f80367fbfe79...2cd1ffc4a807

src/command/control/stop.c
```diff
@@ -0,0 +1,111 @@
...
+                    // lock file so remove it now.
+                    if (flock(fileHandle, LOCK_EX | LOCK_NB) == 0)
+                    {
+                        unlink(strPtr(lockFile));
+                        close(fileHandle);
+                        continue;
+                    }
+
+                    // The file is locked so that means there is a running process - read the process id and send it a term signal
+                    char contents[64] = "";
```
dwsteele: No need to initialize this size we can't rely on the contents after the read.

src/command/control/stop.c
```diff
@@ -0,0 +1,111 @@
...
+                    ssize_t actualBytes = read(fileHandle, &contents, 64);
```
dwsteele: contents should not be dereferenced here.  This causes the data to be loaded into the pointer rather than the array.

src/command/control/stop.c
```diff
@@ -0,0 +1,111 @@
...
+
+                    String *processId = NULL;
+                    if (actualBytes > 0)
+                        processId = strTrim(strNew(contents));
```
dwsteele: Using strNew() is not safe here because we have no idea if the data is zero-terminated.  Use strNewN() instead.

test/src/module/command/controlTest.c
```diff
@@ -51,12 +60,309 @@ testRun(void)
...
+            }
+            HARNESS_FORK_PARENT_END();
+        }
+        HARNESS_FORK_END();
+
+        // lock file with another process lock, processId is invalid
+        // -------------------------------------------------------------------------------------------------------------------------
+        TEST_RESULT_VOID(storageRemoveNP(storageTest, stanzaStopFile), "remove stop file");
+        TEST_RESULT_VOID(
+            storagePutNP(storageNewWriteNP(storageTest, strNewFmt("%s/badpid" LOCK_FILE_EXT, strPtr(lockPath))), BUFSTRDEF("32768")),
```
dwsteele: 32768 is a valid process id, so this may have unintended consequences, i.e. killing a random other process.  Let's use -32768 instead.

dwsteele:
Committed lock file extension constant at feec674b6ff88bdfa62.

dwsteele:
Be sure to remove as much of the Perl code from `Lock.pm` as possible.

dwsteele:
Committed at 382ed9282504b2ad.

PR #809, opened 2019-08-21
Convert stanza-* commands to c.

cmwshang:
All Perl code related to/required by the stanza commands has been converted to C. The caveat is that changes outside the scope of this conversion are required to enable run 3 of the Real module tests to run without error.


dwsteele:
Very nice!  Committed at c733319063bbc9d.

The only real change, other than formatting, was to the warning about --force.  It's not really deprecated if is has been removed.

Issue #793, opened 2019-07-30
(doc) db timeout does not reference protocol timeout relationship


jameschancojr:
Not version specific*

On https://pgbackrest.org/configuration.html#section-general/option-db-timeout section 4.11 (protocol timeout) states: "The protocol-timeout option must be greater than the db-timeout option.", however, there is no reciprocal comment for db-timeout. 

If the user experiences a db-timeout, backrest will inform them of the same, and if you go to look at section 4.6 in the documentation it doesn't tell you about the relationship with 4.11 (protocol-timeout) this can lead to the user setting them inappropriately (as I did).

cmwshang:
Thanks for pointing this out! We will attempt to target this fix for the next release.

dwsteele:
Committed at f57e119b25d from #824.

PR #824, opened 2019-09-09
Add note to db-timeout description to reference protocol-timeout.

cmwshang:
The relationship between db-timeout and protocol-timeout options was detailed only in the protocol-timeout reference section in the documentation which resulted in confusion. Define the relationship in the db-timeout section also to avoid confusion.


cmwshang:
Once committed, issue https://github.com/pgbackrest/pgbackrest/issues/793 can be closed.

dwsteele:
Committed at f57e119b25d.

PR #826, opened 2019-09-09
Add FAQ regarding time-based PITR.

cmwshang:
When setting a target time during a restore for PITR, pgBackRest chooses the lates backup which, if the target time is before that backup's stop time, the target time is ignored. This caused confusion in why the database was replayed after the target time.
Added clarification to the FAQ, user-guide and reference documents to more clearly explain how to perform time-based PITR effectively using the --set option.


dwsteele:
Committed at 8675699d08b91c57294bec.

PR #827, opened 2019-09-10
Add FAQ regarding backup from standby.

cmwshang:
Clarify why backups fail when backup-standby is set and no standbys are available and how to override.

### DIFF
--- a/doc/xml/faq.xml
+++ b/doc/xml/faq.xml
@@ -80,6 +80,14 @@ process-max=1
         <p>The <link url="https://apt.postgresql.org">apt.postgresql.org</link> repository maintains an <link url="https://atalia.postgresql.org/morgue/p/pgbackrest">archive of older versions</link>.  Debian also maintains <link url="https://snapshot.debian.org/binary/pgbackrest/">snapshots</link> of all test builds.</p>
     </section>
 
+    <section id="backup-standby">
+        <title>If the configuration option backup-standby=y, why does a backup attempt fail when the standby database is down?</title>
+
+        <p>Usually the reason for configuring backup from standby is to reduce the load on the primary (master), so why would it be OK to increase the load on the primary when the standby is down? Putting more load on the primary in a situation where there are already failures in the system is not recommended. Backups are not all that important as long as you have one that is fairly recent -- the important thing is to keep up with WAL archiving. There is plenty of time to get a backup when the system is stable again.</p>
+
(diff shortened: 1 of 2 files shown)


dwsteele:
Committed at 56bf9d0566bbb9708d1687,

dwsteele:
Committed at 09ef03b7ef69d2760.

mahomedh:
Would it be possible to also add this `--no-backup-standby` option to the section `4.1.4 Backup from Standby Option (--backup-standby)` in the [Command References Document](https://pgbackrest.org/command.html) please?

dwsteele:
Any boolean command line option can be negated, as noted in the introduction:
```
The no- prefix may be used to set a boolean option to false on the command-line.
```
I'm not sure we want to reiterate that for every option.

mahomedh:
I've now seen that in the intro. Unfortunately, when searching for commands and their options, or reading through the contents list, most people are unlikely to read intro.

For some users who interact with the software on a daily basis, they will probably remember the `no-` prefix or to re-read the intro. But for other users like myself (who I think are the majority based on the number of such questions and the complaints of a high learning curve), I think more verbose documentation (and even examples) would be more helpful.

But to get back on topic, would you maybe at least consider splitting the paragraphs in the Intro section with sub-topic headings? I'd recommend this:

---

### 1.1 Resetting or over-riding config and boolean options

Non-boolean options configured in pgbackrest.conf can be reset to default on the command-line by using the `reset-` prefix. This feature may be used to restore a backup directly on a repository host. 

Normally, pgBackRest will error because it can see that the database host is remote and restores cannot be done remotely. By adding --reset-pg1-host on the command-line, pgBackRest will ignore the remote database host and restore locally. It may be necessary to pass a new --pg-path to force the restore to happen in a specific path, i.e. not the path used on the database host.

The `no-` prefix may be used to set a boolean option to false on the command-line.

### 1.2 Environment variables and option precedence

Any option may be set in an environment variable using the PGBACKREST_ prefix and the option name in all caps replacing - with _, e.g. pg1-path becomes PGBACKREST_PG1_PATH and stanza becomes PGBACKREST_STANZA. Boolean options are represented as they would be in a configuration file, e.g. PGBACKREST_COMPRESS="n", and reset-* variants are not allowed. Options that that can be specified multiple times on the command-line or in a config file can be represented by separating the values with colons, e.g. PGBACKREST_DB_INCLUDE="db1:db2".

Command-line options override environment options which override config file options.

---

PR #829, opened 2019-09-10
Renamed parameter processId to processIdx in harnessFork.h macros

cmwshang:
For clarify, when passing the processId which is actually a processIdx to a harnessFork macros, the parameter has been renamed to processIdx.


dwsteele:
Committed at c5f496b903ab9a4fa.

Issue #830, opened 2019-09-11
cannot compile 2.17 on FreeBSD


fluca1978:
1. pgBackRest version: 2.17 (release)

2. PostgreSQL version: 11.5

3. Operating system/version: FreeBSD 12 release p9

Trying to compile the pgbackrest from sources, release 2.17, I got different errors related to the include.
I've tried to define an includedir at `configure` time, but it seems it is not honored.

```
 % ./configure --prefix=/usr/local --includedir=/usr/src/sys      
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking for gcc option to accept ISO C99... none needed
checking for perl_parse in -lperl... yes
checking for PQconnectdb in -lpq... yes
checking for EVP_get_digestbyname in -lcrypto... yes
checking for SSL_new in -lssl... yes
checking for xmlSaveToBuffer in -lxml2... yes
checking for deflate in -lz... yes
configure: creating ./config.status
config.status: creating Makefile
config.status: creating build.auto.h
```

and then when doing
```
% gmake
...
gcc   -std=c99 -D_POSIX_C_SOURCE=200112L -D_DARWIN_C_SOURCE -O2 -Wall -Wextra -Wno-clobbered -Wno-missing-field-initializers -I. -I../libc -I/usr/local/include -DHAS_FPSETMASK -DHAS_FLOATINGPOINT_H -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_FORTIFY_SOURCE=2  -I/usr/local/lib/perl5/5.28/mach/CORE  -I/usr/local/include/libxml2 -I/usr/include -c command/control/stop.c -o command/control/stop.o
In file included from command/control/stop.c:10:
/usr/include/sys/file.h:226:2: error: unknown type name 'u_int'
  u_int xf_flag; /* flags (see fcntl.h) */
  ^~~~~
command/control/stop.c: In function 'cmdStop':
command/control/stop.c:70:25: warning: implicit declaration of function 'flock'; did you mean 'fork'? [-Wimplicit-function-declaration]
                     if (flock(fileHandle, LOCK_EX | LOCK_NB) == 0)
                         ^~~~~
                         fork
[...]

PR #831, opened 2019-09-11
Compile 2.17 on FreeBSD

fluca1978:
With these changes it is possible to compile 2.17 and 2.18dev on FreeBSD 12-RELEASE-p9.
One change is in the includes, and it has already been reported, so I guess the solution with an `#ifdef` is good enough to allow compilation in a portable way.
The Makefile.in has been changed, but this should be carefully reviewed since we should come up woth `CSTANDARD` out of `configure` but I'm not able of doing it.

Close ticket #830


dwsteele:
I agree that we should build CSTANDARD in configure, but I'm also not sure how to do that yet.  The short-term answer is to override CSTANDARD at make time, i.e. `make CSTANDARD=c99`.

The header file change has been committed at f9c8bd31de89cc1.

PR #843, opened 2019-09-19
Info command improvements

cmwshang:
Add --set option to the info command to display database, symlink and tablespace information as requested in issues 774 and 818.


dwsteele:
Committed at f96c54c4ba0e5 with minor code reformatting and replacing `strCmp()` with `strEq()`.  Also removed commas from the symlink and tablespace lists because it seemed like they could be copy/paste hazards.

PR #845, opened 2019-09-20
Backup Manifest Handler

dwsteele:
The backup manifest stores a complete list of all files, links, and paths in a backup along with metadata such as checksums, sizes, timestamps, etc.  A list of databases is also included for selective restore.

The purpose of the manifest is to allow the restore command to confidently reconstruct the PostgreSQL data directory and ensure that nothing is missing or corrupt.  It is also useful for reporting, e.g. size of backup, backup time, etc.


src/info/manifest.c
```diff
@@ -0,0 +1,1986 @@
...
+            loadData.pathFoundList = lstNew(sizeof(ManifestLoadFound));
+        }
+        MEM_CONTEXT_END();
+
+        this->info = infoNewLoad(read, manifestLoadCallback, &loadData);
+
+        // Process file defaults
+        for (unsigned int fileIdx = 0; fileIdx < lstSize(this->fileList); fileIdx++)
+        {
+            ManifestFile *file = lstGet(this->fileList, fileIdx);
```
cmwshang: For consistency, would it be better to use manifestFileTotal(this) and manifestFile(this, fileIdx) instead of lstSize and lstGet? If so, this function should also change for link and path. Not a big deal if don't want to change.

src/info/manifest.c
```diff
@@ -0,0 +1,1986 @@
...
+struct Manifest
+{
+    MemContext *memContext;                                         // Context that contains the Manifest
+
+    Info *info;                                                     // Base info object
+    StringList *ownerList;                                          // List of users/groups
+    StringList *referenceList;                                      // List of file references
+
+    ManifestData data;                                              // Manifest data and options
+    List *targetList;                                               // List of paths
```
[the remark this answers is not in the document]
dwsteele: Only different in my head.  Fixed.

src/info/manifest.c
```diff
@@ -0,0 +1,1986 @@
...
+#define MANIFEST_KEY_DB_VERSION                                     "db-version"
+    STRING_STATIC(MANIFEST_KEY_DB_VERSION_STR,                      MANIFEST_KEY_DB_VERSION);
+#define MANIFEST_KEY_DESTINATION                                    "destination"
+    VARIANT_STRDEF_STATIC(MANIFEST_KEY_DESTINATION_VAR,             MANIFEST_KEY_DESTINATION);
+#define MANIFEST_KEY_FILE                                           "file"
+    VARIANT_STRDEF_STATIC(MANIFEST_KEY_FILE_VAR,                    MANIFEST_KEY_FILE);
+#define MANIFEST_KEY_GROUP                                          "group"
+    STRING_STATIC(MANIFEST_KEY_GROUP_STR,                           MANIFEST_KEY_GROUP);
+    VARIANT_STRDEF_STATIC(MANIFEST_KEY_GROUP_VAR,                   MANIFEST_KEY_GROUP);
+#define MANIFEST_KEY_PRIMARY                                        "ma" "st" "er"
```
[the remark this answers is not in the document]
dwsteele: I've been trying to expunge the word "master" from the repo and I don't want this to come up in searches.  We can't change it, of course, because it would break the repo format.

src/info/manifest.c
```diff
@@ -0,0 +1,1986 @@
...
+                MEM_CONTEXT_TEMP_RESET(1000);
+            }
+        }
+        MEM_CONTEXT_TEMP_END();
+    }
+
+    // -----------------------------------------------------------------------------------------------------------------------------
+    if (infoSaveSection(infoSaveData, MANIFEST_SECTION_TARGET_LINK_DEFAULT_STR, sectionNext))
+    {
+        if (lstSize(manifest->linkList) > 0)
```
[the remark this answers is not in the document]
dwsteele: Fixed -- and a bunch of other places, too.

src/info/manifest.c
dwsteele: Using *Total() in place of lstSize() everywhere now.

Using manifestFile(), etc. when possible but since that returns a const pointer we still need to use lstGet() in the cases where we need to modify the struct.  The case above is an example of this.

src/info/manifest.c
```diff
@@ -0,0 +1,1986 @@
...
+        infoSaveValue(
+            infoSaveData, MANIFEST_SECTION_TARGET_PATH_DEFAULT_STR, MANIFEST_KEY_USER_STR,
+            jsonFromVar(saveData->pathUserDefault, 0));
+    }
+
+    FUNCTION_TEST_RETURN_VOID();
+}
+
+void
+manifestSave(Manifest *this, IoWrite *write)
```
cmwshang: So where is infoManifestSaveFile (i.e. where we would add the cipherBlock filter and save the file and a copy (or not)? I'm guessing we'd have one infoManifestSaveFile or maybe infoManifestSaveFile and infoManifestSaveFileCopy (for aborted backups)?

dwsteele: Saving a manifest to file is a complex multi-step process that is done in backup.  A general function would not have enough info to do it safely.

dwsteele:
Thanks for the review!  Committed at c969137021da5dc0.

PR #851, opened 2019-10-02
Clarify docs with upper version bound for Auto Stop

bradnicholson:
This feature does not work with Postgres 9.6 or greater, but that is only mentioned in the internal comments in the code.  This seems like it should also be mentioned in the docs.


dwsteele:
Thanks!  It's always a challenge to keep documentation up to date as features change.

Committed at 887ec97e0743f758.

PR #863, opened 2019-10-11
Remove pretty printing for json output of the info command.

cmwshang:
The json output is intended to be system readable. Pretty printing the info output forced the user to remove all whites spaces and carriage returns. There are many external tools to pretty print if needed.
Given that the info command unit tests extensively test the info command output, testing the command in the Mock tests was redundant and has been removed.


keithf4:
Any chance there could be an option left in backrest to pretty print the json output if desired? Have 3 modes: text, json, jsonpretty?

dwsteele:
We though about that but just aren't interested in maintaining the code.  There are a number of utilities out there for JSON formatting which are all better than what we had.

dwsteele:
Committed at 29725805668, d90b2724f883, db1dc4f2756.

PR #873, opened 2019-11-12
Fix time datatype in InfoBackupData structure 

cmwshang:
InfoBackupData structure had backup start and stop time members declared as uint64_t instead of time_t which was inconsistent with other timestamp declarations.


dwsteele:
Committed at c5b76d213bc24 with minor comment updates.

PR #880, opened 2019-11-22
Add manifestNewBuild() and supporting functions.

dwsteele:
This code should behave just like the Perl code, with one major exception -- `pg-path` can be a link to another directory.  We don't need to store the destination of the link in this case so we just follow it.


src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+    FUNCTION_TEST_BEGIN();
+        FUNCTION_TEST_PARAM_P(VOID, data);
+        FUNCTION_TEST_PARAM(STORAGE_INFO, *storageInfo);
+    FUNCTION_TEST_END();
+
+    ASSERT(data != NULL);
+    ASSERT(info != NULL);
+
+    // Skip all . paths because they have already been recorded on the previous level of recursion
+    if (strEq(info->name, DOT_STR))
```
cmwshang: Really? Where did we "record" this? I'm not sure I see where...

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+            // Skip the contents of these paths if they exist in the base path since they won't be reused after recovery
+            if (strEq(buildData.manifestParentName, MANIFEST_TARGET_PGDATA_STR))
+            {
+                if (strEqZ(info->name, PG_PATH_PGDYNSHMEM) && pgVersion >= PG_VERSION_94)
+                {
+                    FUNCTION_TEST_RETURN_VOID();
+                    return;
+                }
+
+                if (strEqZ(info->name, PG_PATH_PGNOTIFY))
```
cmwshang: In perl we check for the version 90 so why not here in C:

Skip pg_notify/* since these files cannot be reused on recovery
        next if $strFile =~ ('^' . MANIFEST_PATH_PGNOTIFY . '\/') && $self->dbVersion() >= PG_VERSION_90;

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+                    return;
+                }
+
+                if (strEqZ(info->name, PG_PATH_PGSNAPSHOTS) && pgVersion >= PG_VERSION_92)
+                {
+                    FUNCTION_TEST_RETURN_VOID();
+                    return;
+                }
+
+                if (strEqZ(info->name, PG_PATH_PGSTATTMP))
```
cmwshang: Have we stopped support for 8.3? In Perl this had:

Skip temporary statistics in pg_stat_tmp even when stats_temp_directory is set because PGSS_TEXT_FILE is always created
        # there.
        next if $strFile =~ ('^' . MANIFEST_PATH_PGSTATTMP . '\/') && $self->dbVersion() >= PG_VERSION_84;

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+                if (strEqZ(info->name, PG_PATH_PGSUBTRANS))
+                {
+                    FUNCTION_TEST_RETURN_VOID();
+                    return;
+                }
+            }
+
+            // Skip the contents of archive_status when online
+            if (buildData.online && strEq(buildData.manifestParentName, buildData.manifestWalName) &&
+                strEqZ(info->name, PG_PATH_ARCHIVE_STATUS))
```
cmwshang: I don't see this in the Perl code. In Perl we specifically did not skip it (note !~):

        next if ($bOnline && $strFile =~ (qw{^} . MANIFEST_TARGET_PGDATA . qw{/} . $self->walPath() . '\/') &&
                 $strFile !~ ('^' . MANIFEST_TARGET_PGDATA . qw{/} . $self->walPath() . qw{/} . DB_PATH_ARCHIVESTATUS . '$'));

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+        };
+
+        // We won't identify db paths for PostgreSQL < 9.0.  This means that temp relations will not be excluded but it doesn't seem
+        // worth supporting this feature on such old versions of PostgreSQL.
+        // -------------------------------------------------------------------------------------------------------------------------
+        if (pgVersion >= PG_VERSION_90)
+        {
+            ASSERT(buildData.tablespaceId != NULL);
+
+            buildData.dbPathExp = regExpNew(
```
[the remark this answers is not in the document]
dwsteele: dbPath is a path within the pg/repo directory that contains relations for a specific database.  In this case we are using repo paths to identify which ones contain database relations.

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+                strNewFmt("^" DB_PATH_EXP "$", strPtr(buildData.tablespaceId)));
```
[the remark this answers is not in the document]
dwsteele: 1.  This regex is being applied to `manifestName`, e.g. `pg_data/base` or `pg_tblspc/...`.
2. I'm not too worried about it because we are only recursing into directories with the exact name (see line 720 for where the path is built) so `PG_9D4_201409291` would never get to the regexp.
3. Reg exp does not require this char to be escaped, but Perl did in some places as you note.  C does not have this restriction so it is simpler not to escape.

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+
+            // Expression to find temp relations
+            buildData.tempRelationExp = regExpNew(STRDEF("^t[0-9]+_" RELATION_EXP "$"));
```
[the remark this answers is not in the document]
dwsteele: `_` does not need to be escaped, and I don't think Perl cares.  I think we were just over-escaping.

src/info/manifest.c
dwsteele: `pg_data` is added to the manifest at line 890 and each path below is added at 472.

src/info/manifest.c
dwsteele: Yeah, missed that one.  I'll add it after review or you can.

src/info/manifest.c
dwsteele: No, it just looks like this one was missed as well.  Same as above.

src/info/manifest.c
dwsteele: The logic is now split between line 540 and 597.  At line 540 we don't exclude pg_xlog/pg_wal because we later may want to preserve the contents (file at 597) or at least `archive_status`.  We don't have all that info at 540 so we just exclude the contents of `archive_status` when online.

src/info/manifest.c
```diff
@@ -383,6 +386,809 @@ manifestNewInternal(void)
...
+            {
+                FUNCTION_TEST_RETURN_VOID();
+                return;
+            }
+
+            // Skip files in the root data path
+            if (strEq(buildData.manifestParentName, MANIFEST_TARGET_PGDATA_STR))
+            {
+                // Skip recovery files
+                if (((strEqZ(info->name, PG_FILE_RECOVERYSIGNAL) || strEqZ(info->name, PG_FILE_RECOVERYSIGNAL)) &&
```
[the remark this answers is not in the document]
dwsteele: Yes, that sounds correct.

src/info/manifest.c
```diff
@@ -383,6 +386,809 @@ manifestNewInternal(void)
...
+                        pgVersion >= PG_VERSION_12) ||
+                    ((strEqZ(info->name, PG_FILE_RECOVERYCONF) || strEqZ(info->name, PG_FILE_RECOVERYDONE)) &&
+                            pgVersion < PG_VERSION_12) ||
+                    // Skip temp file for safely writing postgresql.auto.conf
+                    (strEqZ(info->name, PG_FILE_POSTGRESQLAUTOCONFTMP) && pgVersion >= PG_VERSION_94) ||
```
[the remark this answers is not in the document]
dwsteele: Yes, this was an oversight in the Perl code.  Not a big deal, but it's nice to be precise.

src/info/manifest.c
```diff
@@ -383,6 +386,809 @@ manifestNewInternal(void)
...
+                    // Skip backup_label in versions where non-exclusive backup is supported
+                    (strEqZ(info->name, PG_FILE_BACKUPLABEL) && pgVersion >= PG_VERSION_96) ||
```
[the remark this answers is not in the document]
dwsteele: It was an oversight.  If somebody is running an exclusive backup side-by-side with our non-exclusive backup, we certainly don't want to copy their `backup_label`.  I think in Perl it would just get overwritten with our `backup_label` at the end of the backup, but C definitely won't work that way.

It's an edge case, but worth handling.

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+
+    // Process storage types
+    switch (info->type)
+    {
+        // Add paths
+        // -------------------------------------------------------------------------------------------------------------------------
+        case storageTypePath:
+        {
+            // There should not be any paths in pg_tblspc
+            if (strEqZ(buildData.manifestParentName, MANIFEST_TARGET_PGDATA "/" MANIFEST_TARGET_PGTBLSPC))
```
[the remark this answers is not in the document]
dwsteele: Good point, we should add that.

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+            ManifestTarget target =
+            {
+                .name = manifestName,
+                .type = manifestTargetTypeLink,
+            };
+
+            // Make a copy of the link name because it will need to be modified when there are tablespace ids
+            const String *linkName = info->name;
+
+            // Is this a tablespace?
```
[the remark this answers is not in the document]
dwsteele: This is part of online backup which has not been coded yet.  I have a note to add it.

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+        return;
+    }
+
+    // Get build data
+    ManifestBuildData buildData = *(ManifestBuildData *)data;
+    unsigned int pgVersion = buildData.manifest->data.pgVersion;
+
+    // Contruct the name used to identify this file/link/path in the manifest
+    strPtr(info->name);
+    strPtr(buildData.manifestParentName);
```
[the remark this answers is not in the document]
dwsteele: Probably I had them there to avoid compiler errors while writing code, kind of like `(void)info`.  Probably.

But they are clearly vestigial now and should be removed.

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+            if (strEq(buildData.manifestParentName, STRDEF(MANIFEST_TARGET_PGDATA "/" MANIFEST_TARGET_PGTBLSPC)))
+            {
+                // Strip pg_data off the manifest name so it begins with pg_tblspc instead.  This reflects how the files are stored
+                // in the backup directory.
+                manifestName = strSub(manifestName, sizeof(MANIFEST_TARGET_PGDATA));
+
+                // Identify this target as a tablespace
+                target.name = manifestName;
+                target.tablespaceId = cvtZToUInt(strPtr(info->name));
+                target.tablespaceName = strNewFmt("ts%s", strPtr(info->name));
```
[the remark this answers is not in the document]
dwsteele: `ts` prefix is used in offline backups.  These names will be updated by the online backup code I'm working on now.

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+        this->data.backupOptionDelta = varNewBool(delta);
+
+        // If online then add one second to the copy start time to allow for database updates during the last second that the
+        // manifest was being built.  It's up to the caller to actually wait the remainder of the second, but for comparison
+        // purposes we want the time when the waiting started.
+        this->data.backupTimestampCopyStart = copyStart + (this->data.backupOptionOnline ? 1 : 0);
+    }
+    MEM_CONTEXT_END();
+
+    // Check the manifest for timestamp anomolies that require a delta backup (if delta is not already specified)
```
cmwshang: typo: anomolies -> anomalies

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+        else if (manifestData(manifestPrior)->backupOptionOnline != this->data.backupOptionOnline)
+        {
+            LOG_WARN(
+                "the online option has changed since the %s backup, enabling delta checksum",
+                strPtr(manifestData(manifestPrior)->backupLabel));
+
+            this->data.backupOptionDelta = BOOL_TRUE_VAR;
+        }
+
+        // Check for anomolies between manifests if delta is not already enabled.  This can't be combined with the main comparison
```
cmwshang: typo: anomolies -> anomalies

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+        lstSort(this->pathList, sortOrderAsc);
+        lstSort(this->targetList, sortOrderAsc);
+
+        // Remove unlogged relations from the manifest.  This can't be done during the initial build because of the requirement to
+        // check for _init files which will sort after the vast majority of the relation files.  We could check storage for each
+        // _init file but that would be expensive.
+        // -------------------------------------------------------------------------------------------------------------------------
+        if (pgVersion >= PG_VERSION_91)
+        {
+            RegExp *relationExp = regExpNew(STRDEF("^" DB_PATH_EXP "/" RELATION_EXP "$"));
```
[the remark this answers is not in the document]
dwsteele: Yeah, this needs to be fixed.

src/info/manifest.c
dwsteele: Whoops!

src/info/manifest.c
dwsteele: Double whoops!

src/info/manifest.c
```diff
@@ -489,9 +1236,14 @@ manifestLoadCallback(void *callbackData, const String *section, const String *ke
 
             file.sizeRepo = varUInt64(kvGetDefault(fileKv, MANIFEST_KEY_SIZE_REPO_VAR, VARUINT64(file.size)));
```
cmwshang: I'm a little confused about the repoSize:
1) repo-size is not always present, for example:
pg_data/backup_label={"checksum":"0aaa9318dc6a0d461f28f5e649c8f9e18a623220","master":true,"size":229,"timestamp":1574780855}

2) repo-size is not the same as "size" so why defaulting it to size here?

src/info/manifest.c
```diff
@@ -383,6 +386,750 @@ manifestNewInternal(void)
...
+
+            // Add file to manifest
+            ManifestFile file =
+            {
+                .name = manifestName,
+                .mode = info->mode,
+                .user = info->user,
+                .group = info->group,
+                .size = info->size,
+                .sizeRepo = info->size,
```
cmwshang: Why are we defaulting repoSize to fileSize - it is not the same thing. Will this be updated by some other code somewhere?

src/info/manifest.c
```diff
@@ -1552,6 +2322,55 @@ manifestFileTotal(const Manifest *this)
...
+        if (reference != NULL)
+            file->reference = strLstAddIfMissing(this->referenceList, reference);
+
+        // Update checksum if set
+        if (checksumSha1 != NULL)
+            memcpy(file->checksumSha1, checksumSha1, HASH_TYPE_SHA1_SIZE_HEX + 1);
+
+        // Update repo size
+        file->size = size;
+        file->sizeRepo = sizeRepo;
```
cmwshang: The file may not have a repo-size (see below) so what are we really passing to this function if it doesn't? zero?:

pg_data/backup_label={"checksum":"0aaa9318dc6a0d461f28f5e649c8f9e18a623220","master":true,"size":229,"timestamp":1574780855}

src/info/manifest.c
```diff
@@ -1552,6 +2322,55 @@ manifestFileTotal(const Manifest *this)
...
+
+        // Update checksum page info
+        file->checksumPage = checksumPage;
```
cmwshang: checksum-page may not even exist which is not the same as setting it to false, so how/where is that being handled?  
Example from same manifest:
```
pg_data/PG_VERSION={"checksum":"184473f470864e067ee3a22e64b47b0a1c356f29","master":true,"reference":"[BACKUP-FULL-2]","size":[SIZE],"timestamp":[TIMESTAMP-1]}
pg_data/base/1/12000={"checksum":"22c98d248ff548311eda88559e4a8405ed77c003","checksum-page":true,"reference":"[BACKUP-FULL-2]","size":[SIZE],"timestamp":[TIMESTAMP-1]}
pg_data/base/16384/17000={"checksum":"e0101dd8ffb910c9c202ca35b5f828bcb9697bed","checksum-page":false,"checksum-page-error":[1],"reference":"[BACKUP-FULL-2]","size":[SIZE],"timestamp":[TIMESTAMP-1]}
pg_tblspc/1/[TS_PATH-1]/16384/tablespace1.txt={"checksum":"d85de07d6421d90aa9191c11c889bfde43680f0f","checksum-page":false,"size":[SIZE],"timestamp":[TIMESTAMP-1]}

```

src/info/manifest.c
```diff
@@ -1552,6 +2322,55 @@ manifestFileTotal(const Manifest *this)
...
+    FUNCTION_TEST_END();
+
+    ASSERT(this != NULL);
+    ASSERT(name != NULL);
+    ASSERT(
+        (!checksumPage && !checksumPageError && checksumPageErrorList == NULL) ||
+        (checksumPage && !checksumPageError && checksumPageErrorList == NULL) || (checksumPage && checksumPageError));
+
+    ManifestFile *file = (ManifestFile *)manifestFileFind(this, name);
+
```
cmwshang: Do we no longer care about copying the "master" setting from the prior manifest for the file?

src/info/manifest.c
dwsteele: If it is not stored in the manifest that means it equals size -- i.e. an uncompressed backup.

src/info/manifest.c
dwsteele: It is updated by manifestFileUpdate() when the file is copied/resumed/referenced.

src/info/manifest.c
dwsteele: The file always has a repo size, we just don't store it if it is the same as size to save space.

src/info/manifest.c
dwsteele: Search for `kvPut(fileKv, MANIFEST_KEY_CHECKSUM_PAGE_VAR, VARBOOL(!file->checksumPageError));` to find the save logic.

Basically, if it is not in the manifest checksumPage is false, if in manifest then true means checksum errors, and false means no checksum errors.  checksumPageErrorList is present if there are page errors (rather than a validation error).  Yeah, it's complicated, but that's Perl for ya.

src/info/manifest.c
dwsteele: We do not.  We just recalculate master for each manifest.  It's doubtful that we should even waste space storing it, but that's the format.  This was only *really* being saved for expect log testing purposes.

src/info/manifest.c
cmwshang: Fixed (commit 1fe22c3bb51dbb7070d22d8329ee999290ea8560)

src/info/manifest.c
cmwshang: Fixed (commit 1fe22c3bb51dbb7070d22d8329ee999290ea8560)

src/info/manifest.c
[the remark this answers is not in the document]
cmwshang: Code changed to (commit 1fe22c3bb51dbb7070d22d8329ee999290ea8560):
`RegExp *relationExp = regExpNew(strNewFmt("^" DB_PATH_EXP "/" RELATION_EXP "$", strPtr(buildData.tablespaceId)));`

src/info/manifest.c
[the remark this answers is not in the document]
cmwshang: Removed (commit 1fe22c3bb51dbb7070d22d8329ee999290ea8560)

src/info/manifest.c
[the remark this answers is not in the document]
cmwshang: Added to case storageTypeFile (commit 1fe22c3bb51dbb7070d22d8329ee999290ea8560):

```
            // There should not be any files in pg_tblspc
            if (strEqZ(buildData.manifestParentName, MANIFEST_TARGET_PGDATA "/" MANIFEST_TARGET_PGTBLSPC))
            {
                THROW_FMT(
                    LinkExpectedError, "'%s' is not a symlink - " MANIFEST_TARGET_PGTBLSPC " should contain only symlinks",
                    strPtr(manifestName));
            }
```

src/info/manifest.c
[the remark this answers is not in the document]
cmwshang: Code has been changed to (commit 1fe22c3bb51dbb7070d22d8329ee999290ea8560):
`if (((strEqZ(info->name, PG_FILE_RECOVERYSIGNAL) || strEqZ(info->name, PG_FILE_STANDBYSIGNAL)) &&`

src/info/manifest.c
cmwshang: Code has been changed to (commit 1fe22c3bb51dbb7070d22d8329ee999290ea8560):
`if (strEqZ(info->name, PG_PATH_PGSTATTMP) && pgVersion >= PG_VERSION_84)`

src/info/manifest.c
cmwshang: Code has been changed to (commit 1fe22c3bb51dbb7070d22d8329ee999290ea8560):
`if (strEqZ(info->name, PG_PATH_PGNOTIFY) && pgVersion >= PG_VERSION_90)`

dwsteele:
Committed at d3132dae26a.  Thank you for the review and tests!

PR #886, opened 2019-12-09
C migration of backup command

dwsteele:


src/command/backup/backup.c
```diff
@@ -0,0 +1,1911 @@
...
+    result->walSegmentSize = pgControl.walSegmentSize;
+
+    // Validate pg_control info against the stanza
+    if (result->version != infoPg.version || pgControl.systemId != infoPg.systemId)
+    {
+        THROW_FMT(
+            BackupMismatchError,
+            PG_NAME " version %s, system-id %" PRIu64 " do not match stanza version %s, system-id %" PRIu64,
+            strPtr(pgVersionToStr(pgControl.version)), pgControl.systemId, strPtr(pgVersionToStr(infoPg.version)),
+            infoPg.systemId);
```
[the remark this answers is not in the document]
dwsteele: Looks like an oversight to me.  Yeah, we should centralize this one we have an idea of all the different ways it is being used.  I'm not worried about that right now, though.

lib/pgBackRest/Db.pm
```diff
@@ -527,157 +465,6 @@ sub versionGet
...
-    {
-        # Running backups can only be detected in PostgreSQL >= 9.3
-        if ($self->{strDbVersion} >= PG_VERSION_93)
-        {
-            # If a backup is currently in progress emit a warning and then stop it
-            if ($self->executeSqlOne('select pg_is_in_backup()'))
-            {
-                &log(WARN, 'the cluster is already in backup mode but no ' . PROJECT_NAME . ' backup process is running.' .
-                           ' pg_stop_backup() will be called so a new backup can be started.');
-                $self->backupStop();
```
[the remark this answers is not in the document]
dwsteele: It's at `db/db.c` line 316 now.

src/command/backup/backup.c
```diff
@@ -0,0 +1,1911 @@
...
+        // backup from standby only queue 0 will be used.
+        unsigned int queueOffset = jobData->backupStandby && clientIdx > 0 ? 1 : 0;
+        int queueIdx = jobData->backupStandby && clientIdx == 0 ?
+            0 : (int)(clientIdx % (lstSize(jobData->queueList) - queueOffset));
+        int queueEnd = queueIdx;
+
+        do
+        {
+            List *queue = *(List **)lstGet(jobData->queueList, (unsigned int)queueIdx + queueOffset);
+
```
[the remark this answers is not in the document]
dwsteele: This was a bug in Perl.  We were getting the copyStart from the remote system but Start and Stop from the local system when online.  If the clocks are not synced then it could get weird.

All timestamps should come from the database host (which could be local) when online and directly from time() when offline.

test/src/module/command/backupTest.c
```diff
@@ -465,5 +691,1356 @@ testRun(void)
...
+        strLstAddZ(argList, "--no-" CFGOPT_ONLINE);
+        strLstAddZ(argList, "--" CFGOPT_COMPRESS);
+        strLstAddZ(argList, "--" CFGOPT_REPO1_HARDLINK);
+        strLstAddZ(argList, "--" CFGOPT_TYPE "=" BACKUP_TYPE_DIFF);
+        harnessCfgLoad(cfgCmdBackup, argList);
+
+        TEST_ERROR(cmdBackup(), FileMissingError, "no files have changed since the last backup - this seems unlikely");
+
+        TEST_RESULT_LOG(
+            "P00   INFO: last backup label = [FULL-1], version = " PROJECT_VERSION "\n"
```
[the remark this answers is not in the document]
dwsteele: We do in the mock/all tests.  We can make a specific note if you like so we don't forget to move it.

src/command/backup/backup.c
```diff
@@ -0,0 +1,1911 @@
...
+        }
+        // Else start the backup normally
+        else
+        {
+            // Check database configuration
+            checkDbConfig(backupData->version, backupData->pgIdPrimary, backupData->dbPrimary, false);
+
+            // Start backup
+            LOG_INFO_FMT(
+                "execute %sexclusive pg_start_backup(): backup begins after the %s checkpoint completes",
```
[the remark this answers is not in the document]
dwsteele: In the C code we don't know the backup label yet.  Perl would look for resumable backups before start backup but C doesn't do that in order to save memory (we don't want the prior manifest and resumable manifest loaded at the same time).

src/command/backup/backup.c
```diff
@@ -0,0 +1,1911 @@
...
+                backupData->version >= PG_VERSION_96 ? "non-" : "",
+                cfgOptionBool(cfgOptStartFast) ? "requested immediate" : "next regular");
+
+            DbBackupStartResult dbBackupStartResult = dbBackupStart(
```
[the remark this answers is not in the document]
dwsteele: 1a) Weird -- what test are you running to get this error?  I can't reproduce it.  I think the error may be in a previous statement.
2b) I'm not worried about these times being synced since this time is not used for anything.
3) Yes -- backup.c, line 225.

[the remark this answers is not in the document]
cmwshang: 1a) the error was my fault so ignore.
2b) I'd like to understand more of how we are using the timestamps now so let's discuss
3) OK

src/command/backup/backup.c
```diff
@@ -0,0 +1,1906 @@
...
+            else if (fileResume->reference != NULL)
+                removeReason = "reference in resumed manifest";
+            else if (fileResume->checksumSha1[0] == '\0')
+                removeReason = "no checksum in resumed manifest";
+            else if (file->size != fileResume->size)
+                removeReason = "mismatched size";
+            else if (!resumeData->delta && file->timestamp != fileResume->timestamp)
+                removeReason = "mismatched timestamp";
+            else if (file->size == 0)
+                // ??? don't resume zero size files because Perl wouldn't -- this can be removed after the migration)
```
cmwshang: I don't see where this is true in the Perl code. Can you point me to where this was in the Perl code because as far as I can see in the Perl code was checking for file size equality.

src/command/backup/backup.c
```diff
@@ -0,0 +1,1906 @@
...
+
+            // Set the backup label to the resumed backup
+            manifestBackupLabelSet(manifest, manifestData(manifestResume)->backupLabel);
+
+            LOG_WARN_FMT(
+                "resumable backup %s of same type exists -- remove invalid files and resume",
+                strPtr(manifestData(manifest)->backupLabel));
+
+            // If resuming a full backup then copy cipher subpass since it was used to encrypt the resumable files
+            if (manifestData(manifest)->backupType == backupTypeFull)
```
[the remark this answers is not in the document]
dwsteele: If incr/diff we need to get cipher pass from the prior backup (which has already been done).  If full, we need it from the resumed manifest.  If we exclude this check it would still work, but we'd set the cipher pass twice for diff/incr backups which seems weird.

src/command/backup/backup.c
dwsteele: In Perl zero-length files did not have checksums in the manifest in memory.  So, checking the reference is defined (https://github.com/pgbackrest/pgbackrest/blob/master/lib/pgBackRest/Backup/Backup.pm#L194) implies the file is also non-zero.

In C we store all checksums in the manifest in memory, and only skip them for zero files when writing to disk.  Since the checksum is always defined we need to check size to mimic the Perl behavior.

cmwshang: Ah, ok, I did not realize checksums in C were always defined so I thought the 'removeReason = "no checksum in resumed manifest";' would simply be hit.

src/command/backup/backup.c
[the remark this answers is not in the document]
cmwshang: Fixed in commit https://github.com/pgbackrest/pgbackrest/pull/886/commits/aa3b830f622ceaf7068ec41b93b4eaf01dcf5242

src/command/backup/backup.c
```diff
@@ -0,0 +1,1906 @@
...
+            {
+                const ManifestFile *file = *(ManifestFile **)lstGet(queue, 0);
+
+                // Create backup job
+                ProtocolCommand *command = protocolCommandNew(PROTOCOL_COMMAND_BACKUP_FILE_STR);
+
+                protocolCommandParamAdd(command, VARSTR(manifestPathPg(file->name)));
+                protocolCommandParamAdd(
+                    command, VARBOOL(!strEq(file->name, STRDEF(MANIFEST_TARGET_PGDATA "/" PG_PATH_GLOBAL "/" PG_FILE_PGCONTROL))));
+
```
cmwshang: Why is this check for determining to save a copy not inside the job loop above? If we go through 100 jobs and at the 50th job the sizeCopied is already > manifestSaveSize, we won't save until we've completed the 100th job.

cmwshang: Are all jobs completed at this point or would we ever have a partial (say only 3 of the 4 jobs completed so far)?

cmwshang: Do we need to worry about a memory reset if we haven't yet saved the manifest?

dwsteele: You are right, but it doesn't seem worth doing this in the loop since we are only logging.  We'll save the manifest before we go wait for more completed jobs.

dwsteele: Yes, this if for variables used in this loop -- it's not there for the manifest.

src/command/backup/backup.c
```diff
@@ -0,0 +1,1904 @@
...
+    if (cfgOptionBool(cfgOptOnline))
+    {
+        MEM_CONTEXT_TEMP_BEGIN()
+        {
+            // Stop the backup
+            LOG_INFO_FMT(
+                "execute %sexclusive pg_stop_backup() and wait for all WAL segments to archive",
+                backupData->version >= PG_VERSION_96 ? "non-" : "");
+
+            DbBackupStopResult dbBackupStopResult = dbBackupStop(backupData->dbPrimary);
```
cmwshang: The dbBackupStop query used to have a timestamp in it like the backupStart query but in the new db.c code it does not. Form consistency maybe we should have it for both start and stop or for neither.

src/command/backup/backup.c
```diff
@@ -0,0 +1,1904 @@
...
+                const ManifestFile *file = *(ManifestFile **)lstGet(queue, 0);
+
+                // Create backup job
+                ProtocolCommand *command = protocolCommandNew(PROTOCOL_COMMAND_BACKUP_FILE_STR);
+
+                protocolCommandParamAdd(command, VARSTR(manifestPathPg(file->name)));
+                protocolCommandParamAdd(
+                    command, VARBOOL(!strEq(file->name, STRDEF(MANIFEST_TARGET_PGDATA "/" PG_PATH_GLOBAL "/" PG_FILE_PGCONTROL))));
+                protocolCommandParamAdd(command, VARUINT64(file->size));
+                protocolCommandParamAdd(command, file
```
cmwshang: (Line 1795?) The sync backup path at the end of the backup in Perl used to sync the backup LABEL path - which in C it appears we are syncing everything in the backup repo? There were also some conditionals in Perl (e.g. backup=full or pathExists) which I don't see here.  Also the sync was performed before we saved the manifest (which did not seem like that was right anyway).
```
    # Sync backup path if supported
    if ($oStorageRepo->capability(STORAGE_CAPABILITY_PATH_SYNC))
    {
        # Sync all paths in the backup
        $oStorageRepo->pathSync(STORAGE_REPO_BACKUP . "/${strBackupLabel}");

        foreach my $strPath ($oBackupManifest->keys(MANIFEST_SECTION_TARGET_PATH))
        {
            my $strPathSync = $oStorageRepo->pathGet(STORAGE_REPO_BACKUP . "/${strBackupLabel}/$strPath");

            # Not all paths are created for diff/incr backups, so only sync if this is a full backup or the path exists
            if ($strType eq CFGOPTVAL_BACKUP_TYPE_FULL || $oStorageRepo->pathExists($strPathSync))
            {
                $oStorageRepo->pathSync($strPathSync);
            }
        }
    }
```

src/command/backup/backup.c
dwsteele: Hmmm, dbBackupStart() doesn't have a timestamp.  I'm not sure what you mean.

cmwshang: String *result = strNewFmt(
        "select lsn::text as lsn,\n"
        "       pg_catalog.pg_%sfile_name(lsn)::text as wal_segment_name\n"
        "  from pg_catalog.pg_start_backup('" PROJECT_NAME " backup started at ' **|| current_timestamp**",
        strPtr(pgWalName(pgVersion)));

src/command/backup/backup.c
dwsteele: The backup label path is synced in this code.  There are also conditionals for full and exists.  Are we looking at the same code?

Yes -- these syncs used to be done later.  They have been moved up partly to be modular and partly to make sure the paths are synced for a possible resume if stop backup or archive-check fails.  Mostly just to be earlier.

cmwshang: Super annoying with the wrong line numbers - I was hoping 1795 might get you close to the spot. Can you point me to the line number you're looking at that has the sync conditionals? I was in backupComplete()

dwsteele:
Committed at 1f2ce45e.  Thank you for the review!

PR #893, opened 2019-12-18
Skip vagrant disksize option if no plugin

mpalmi:
Previously, `vagrant up` would bail if no `vagrant-disksize` plugin was
installed. This option is just a nice-to-have, so skip it rather than
bailing.


dwsteele:
Great idea!  Committed at d89d9f1c52b9fc.

PR #894, opened 2019-12-18
Add pg-user config option

mpalmi:
Allow a pg-user config option to be specified for backrest commands. If
no user is specified, default to NULL.


dwsteele:
Committed at dc1e7ca22dbda with some minor formatting and doc changes.  Thank you!

PR #904, opened 2020-01-09
Remove internal commands and replace with roles.

dwsteele:
The local, remote, archive-get-async, and archive-push-async commands were used to run functionality that was not directly available to the user.  Unfortunately that meant they would not pick up options from the command that the user expected, e.g. backup, archive-get, etc.

Remove the internal commands and add roles which allow pgBackRest to determine what functionality is required without implementing special commands.  This way the options are loaded from the expected command section.

Since remote is no longer a specific command with its own options, more manipulation is required when calling remote.  This might be something we can improve in the future in the config system, but it may be worth leaving as is because it is a one-off, for now at least.


dwsteele:
Committed at 8d3710b2fe94c90.  Thanks for the review!

PR #908, opened 2020-01-17
Fix detection of keepalive options on Linux

marco44:
Hi,

keepalive was at 2 hours on Linux. That's because TCP_KEEPIDLE wasn't defined, as netinet/tcp.h wasn't included.

By the way, the SO_KEEPIDLE and SO_KEEPINTVL hard set at 3 seem weird, though they'll probably be ok most of the time.

### DIFF
--- a/src/common/io/tls/client.c
+++ b/src/common/io/tls/client.c
@@ -16,6 +16,11 @@ TLS Client
 #include <netinet/in.h>
 #endif
 
+#ifdef __linux__
+#include <netinet/in.h>
+#include <netinet/tcp.h>
+#endif
+
 #include <openssl/conf.h>
 #include <openssl/ssl.h>
 #include <openssl/x509v3.h>
--- a/src/common/io/tls/client.c
+++ b/src/common/io/tls/client.c
@@ -619,7 +619,7 @@ tlsClientOpen(TlsClient *this)
                     int socketValue = 1;
 
                     THROW_ON_SYS_ERROR(
-                        setsockopt(this->socket, SOL_SOCKET, SO_KEEPALIVE, &socketValue, sizeof(int)) == -1, ProtocolError,
+                        setsockopt(this->socket, IPPROTO_TCP, SO_KEEPALIVE, &socketValue, sizeof(int)) == -1, ProtocolError,
                          "unable set SO_KEEPALIVE");
 
                     // Set per-connection keepalive options if they are available
@@ -627,18 +627,18 @@ tlsClientOpen(TlsClient *this)
                     socketValue = 3;
 
                     THROW_ON_SYS_ERROR(
-                        setsockopt(this->socket, SOL_SOCKET, TCP_KEEPIDLE, &socketValue, sizeof(int)) == -1, ProtocolError,
+                        setsockopt(this->socket, IPPROTO_TCP, TCP_KEEPCNT, &socketValue, sizeof(int)) == -1, ProtocolError,
+                         "unable set SO_KEEPCNT");
+
+		    socketValue = this->timeout / 4; /* first try + 3 failures to get to the timeout */
+
+		    THROW_ON_SYS_ERROR(
+                        setsockopt(this->socket, IPPROTO_TCP, TCP_KEEPIDLE, &socketValue, sizeof(int)) == -1, ProtocolError,
                          "unable set SO_KEEPIDLE");
 
                     THROW_ON_SYS_ERROR(
-                        setsockopt(this->socket, SOL_SOCKET, TCP_KEEPINTVL, &socketValue, sizeof(int)) == -1, ProtocolError,
+                        setsockopt(this->socket, IPPROTO_TCP, TCP_KEEPINTVL, &socketValue, sizeof(int)) == -1, ProtocolError,
                          "unable set SO_KEEPINTVL");
-
-                    socketValue = this->timeout / socketValue;
-
-                    THROW_ON_SYS_ERROR(
-                        setsockopt(this->socket, SOL_SOCKET, TCP_KEEPCNT, &socketValue, sizeof(int)) == -1, ProtocolError,
-                         "unable set SO_KEEPCNT");
 #endif
 
                     // Negotiate TLS


dwsteele:
> By the way, the SO_KEEPIDLE and SO_KEEPINTVL hard set at 3 seem weird, though they'll probably be ok most of the time.

I think I may have flipped these.  Seems like TCP_KEEPCNT should be 3 and the other two should be `this->timeout / 3`.  Does that make more sense?

dwsteele:
Hmm, now that this code is "live" it doesn't seem to like TCP_KEEPIDLE.  If I set that to any value I get `[111] Connection refused` in testing.

marco44:
Don't know why you should get that. I have no way of testing it right now anyway. For

> I think I may have flipped these. Seems like TCP_KEEPCNT should be 3 and the other two should be `this->timeout / 3`. Does that make more sense?
Yeah, it would be more logical. Or even by… 4, as you'll first wait TCP_KEEPINTVL, then do 3 TCP_KEEPCNT, so technically, you'll wait 4 times.

marco44:
Ok, I could do some testing now. What can I do to help ?

dwsteele:
Thanks, Marc.  The error I'm getting is:
```
ERROR: [039]: unable set TCP_KEEPIDLE: [92] Protocol not available
```
I'm also getting `[111] Connection refused` in the unit tests, but I think this is because our test TLS server is forked off from a process that has already set TCP_KEEPIDLE.  I can fix the second problem if that's the cause, but I'm not sure what is causing the `[92] Protocol not available` error.

The branch with your PR and my changes: https://github.com/pgbackrest/pgbackrest/compare/dev-keepalive

marco44:
Mmm, I think the error comes from using setsockopt with keepalives with SOL_SOCKET instead of IPPROTO_TCP.  I'll try and fix this in the next hours.

marco44:
Ok, reproduced the problem. FreeBSD accepts SOL_SOCKET even for keepalive (which only have meaning for TCP sockets). From the tcp manpage, https://www.freebsd.org/cgi/man.cgi?query=tcp, I think we could use IPPROTO_TCP. Let's first try this, it's much simpler. I don't have a FreeBSD at hand, do you have an easy way to check I'm not breaking things for it, or do you need me to test it ?

marco44:
Ok, just did that and tested. And while at it, I added the correct computation of the keepalive values. Works perfectly under Linux. If necessary, we can ifdef the setsockopt, but I think this should work with FreeBSD too.

dwsteele:
@fluca1978 Any thoughts on whether this looks good for *BSD?

dwsteele:
OK, this passes all our Linux tests.  Since we don't test on BSD I can't say what will happen there.

It looks like cirrus-ci supports FreeBSD so we'll look into adding.

fluca1978:
> @fluca1978 Any thoughts on whether this looks good for *BSD?

Apparently it cannot compile due to header mismatch (see <https://github.com/pgbackrest/pgbackrest/issues/830>).

```
% gmake 
...
In file included from command/control/stop.c:10:
/usr/include/sys/file.h:229:2: error: unknown type name 'u_int'
  229 |  u_int xf_flag; /* flags (see fcntl.h) */
      |  ^~~~~
command/control/stop.c: In function 'cmdStop':
command/control/stop.c:70:25: warning: implicit declaration of function 'flock'; did you mean 'clock'? [-Wimplicit-function-declaration]
   70 |                     if (flock(fileHandle, LOCK_EX | LOCK_NB) == 0)
      |                         ^~~~~
      |                         clock
command/control/stop.c:70:43: error: 'LOCK_EX' undeclared (first use in this function)
   70 |                     if (flock(fileHandle, LOCK_EX | LOCK_NB) == 0)
      |                                           ^~~~~~~
command/control/stop.c:70:43: note: each undeclared identifier is reported only once for each function it appears in
command/control/stop.c:70:53: error: 'LOCK_NB' undeclared (first use in this function)
   70 |                     if (flock(fileHandle, LOCK_EX | LOCK_NB) == 0)
      |                                                     ^~~~~~~
gmake: *** [Makefile:267: command/control/stop.o] Error 1
```

I'm running FreeBSD 12, against this revision:

```
 % git show HEAD
commit c90eff080a837c8f9355a7150f67dbc763f60b15 (HEAD -> fix_keepalive_linux)
Author: Marc Cousin <marc.cousin@people-doc.com>
Date:   Tue Jan 21 15:42:36 2020 +0100

    Fix socket type for keepalive and compute values in the correct order
```

dwsteele:
@fluca1978 Hmmm, seems like that issue is probably in master.  Could you please check master?

marco44:
Starting to work on the FreeBSD issue. I was suprised that configure could not find --libpq (the postgresql lib isn't in the standard library path in my test). I add to add 
```
LDFLAGS="$LDFLAGS -L`pg_config --libdir`"
```
before performing the
```
AC_CHECK_LIB([pq], [PQconnectdb], [], [AC_MSG_ERROR([library 'pq' is required])])
```

marco44:
Removing -D_POSIX_C_SOURCE=200809L from Makefile.in solves @fluca1978 's error. And it links at the end. But removing it seems to break Linux.

dwsteele:
`-D_POSIX_C_SOURCE=200809L` is required for symlinks in Linux (or `GNU_SOURCE`).  The question is how to get configure to leave it off for BSD.  Might be easier to *include* it for Linux instead.

marco44:
after a bit more digging, -D_POSIX_C_SOURCE=200809L seems to be only required for symlinks on linux if you use -D_DARWIN_C_SOURCE (which isn't for linux from what I understand of its name :) )

dwsteele:
So it looks like it is really MacOS we need to detect.  Thanks for figuring this out!

marco44:
Yes, maybe this could be fixed only for those occasional files that have problem with an ifdef in the specific files on mac os. I don't have any mac to test it though :)

dwsteele:
I get tons of failures on Linux without `-D_POSIX_C_SOURCE=200809L` even after removing `-D_DARWIN_C_SOURCE`.

So it looks like we probably need `-D_POSIX_C_SOURCE=200809L` for Linux, `-D_DARWIN_C_SOURCE` for MacOS, and neither for BSD.

marco44:
Depends on #957

dwsteele:
Committed at 4b24a74afb5c.  Thanks!

PR #913, opened 2020-01-23
Remove TRY blocks from MEM_CONTEXT_* macros

dwsteele:
Rather that doing cleanup in CATCH blocks, keep a stack of all current and new contexts so they can be cleaned up when a error occurs without the need for a TRY block.

This simplication results in faster compiles, a smaller binary, and faster execution.


libc/LibC.h
```diff
@@ -67,7 +67,7 @@ Core context handling macros, only intended to be called from other macros
 ***********************************************************************************************************************************/
 #define MEM_CONTEXT_XS_CORE_BEGIN(memContext)                                                                                      \
     /* Switch to the new memory context */                                                                                         \
-    MemContext *MEM_CONTEXT_memContextPrior = memContextSwitch(memContext);                                                        \
+    memContextPush(memContext);                                                                                                    \
```
cmwshang: If MEM_CONTEXT_XS_CORE_BEGIN does memContextPush then why would MEM_CONTEXT_XS_CORE_END no do a memContextPop?

libc/LibC.h
```diff
@@ -114,10 +109,10 @@ Simplifies creation of the memory context in constructors and includes error han
...
     if (MEM_CONTEXT_XS_croak)                                                                                                      \
     {                                                                                                                              \
-        memContextFree(MEM_CONTEXT_XS_memContext);                                                                                 \
-                                                                                                                                   \
         ERROR_XS()                                                                                                                 \
     }                                                                                                                              \
+    else                                                                                                                           \
+        memContextKeep();                                                                                                          \
```
cmwshang: In the C code, MEM_CONTEXT_NEW_BEGIN does memContextNew() and memContextPush() and then MEM_CONTEXT_NEW_END does memContextPop() and memContextKeep().  So here, in MEM_CONTEXT_XS_NEW_BEGIN there is also a New and a Push but the MEM_CONTEXT_XS_NEW_END does not do a Pop() only a Keep() - why?

libc/LibC.h
dwsteele: Good catch. It should, but this macro is dead code so it doesn't really matter what the contents are.  It must have become dead during the cleanup after the Perl migration was complete.

I have removed the code from the master branch in this commit: https://github.com/dwsteele/pgbackrest-dev/commit/7866861b9f61c12b63e0b1cb83ce0b78e0611780.  Since MEM_CONTEXT_XS_CORE_BEGIN() is now only used in one place I merged it into MEM_CONTEXT_XS_TEMP_BEGIN().  If that looks OK I'll push to master and update the trymem branch accordingly.

libc/LibC.h
dwsteele: Same reason as above.

src/common/memContext.h
```diff
@@ -198,25 +198,26 @@ nearest try block.
...
 
-// Push a switchable type mem context onto the stack which will then become the current mem context. Only context of type switch can
-// become the current mem context.
+// Switch to a context making it the current mem context
 void memContextSwitch(MemContext *this);
 
-// Pop a mem context off the stack and make the prior (first switchable mem context) the current mem context. If the top of the
-// stack is a type new mem context (new context must be kept or discarded) then an error will be thrown in debug builds.
+// Switch back to the context that was current before the last switch.  This is the same context that will be returned from
+// memContextPrior().  If the last function called was memContextNew(), then memContextKeep()/memContextDiscard() must be called
```
cmwshang: If we say it's the same as Prior than it makes me wonder why we even have SwitchBack. I'd rather not confuse the issue and would prefer we just remove that sentence.

src/common/memContext.h
```diff
@@ -198,25 +198,26 @@ nearest try block.
...
-// an error will be thrown in debug builds.
+// Keep a context created by memContextNew() so that it will not be automatically freed if an error occurs.  If the context was
+// switched after the call to memContextNew(), then it must be switched back before calling memContextKeep() or an error will occur
+// in debug builds and the behavior is undefined in production builds.
 void memContextKeep(void);
 
-// Discard a context created by memContextNew(). Discarding will free the mem context and then remove it from the top of the stack.
-// If the top of the stack is not a type new context then an error will be thrown in debug builds.
+// Discard a context created by memContextNew(). If the context was switched after the call to memContextNew(), then it must be
+// switched back before calling memContextKeep() or an error will occur in debug builds and the behavior is undefined in production
```
cmwshang: Did you mean memContextDiscard (vs memContextKeep)?

dwsteele:
Committed at 5afd950ed9.  Thank you for the review!

PR #915, opened 2020-01-26
Fix missing files corrupting the manifest

dwsteele:


dwsteele:
Thank you for the review! Committed at 45ec694af22dda.

PR #916, opened 2020-01-26
Use designated initializers to initialize structs.

dwsteele:
Previously memNew() used memset() to initialize all struct members to 0, NULL, false, etc.  While this appears to work in practice, it is a violation of the C specification.  For instance, NULL == 0 must be true but neither NULL nor 0 must be represented with all zero bits.

Instead use designated initializers to initialize structs.  These guarantee that struct members will be properly initialized even if they are not specified in the initializer.  Note that due to a quirk in the C99 specification at least one member must be explicitly initialized even if it needs to be the default value.

Since pre-zeroed memory is no longer required, adjust memAllocInternal()/memReallocInternal() to return raw memory and update dependent functions accordingly.  All instances of memset() have been removed except in debug/test code where needed.

Add memMewPtrArray() to allocate an array of pointers and automatically set all pointers to NULL.

Rename memGrowRaw() to the more logical memResize().

### DIFF
--- a/src/db/db.c
+++ b/src/db/db.c
@@ -76,10 +76,11 @@ dbNew(PgClient *client, ProtocolClient *remoteClient, const String *applicationN
         *this = (Db)
         {
             .memContext = memContextCurrent(),
-            .client = pgClientMove(client, memContextCurrent()),
             .remoteClient = remoteClient,
             .applicationName = strDup(applicationName),
         };
+
+        this->client = pgClientMove(client, this->memContext);
     }
     MEM_CONTEXT_NEW_END();
 
--- a/src/db/db.c
+++ b/src/db/db.c
@@ -72,11 +72,14 @@ dbNew(PgClient *client, ProtocolClient *remoteClient, const String *applicationN
     MEM_CONTEXT_NEW_BEGIN("Db")
     {
         this = memNew(sizeof(Db));
-        this->memContext = memContextCurrent();
 
-        this->client = pgClientMove(client, this->memContext);
-        this->remoteClient = remoteClient;
-        this->applicationName = strDup(applicationName);
+        *this = (Db)
+        {
+            .memContext = memContextCurrent(),
+            .client = pgClientMove(client, memContextCurrent()),
+            .remoteClient = remoteClient,
+            .applicationName = strDup(applicationName),
+        };
     }
     MEM_CONTEXT_NEW_END();
 
--- a/src/protocol/client.c
+++ b/src/protocol/client.c
@@ -80,13 +80,16 @@ protocolClientNew(const String *name, const String *service, IoRead *read, IoWri
     MEM_CONTEXT_NEW_BEGIN("ProtocolClient")
     {
         this = memNew(sizeof(ProtocolClient));
-        this->memContext = memContextCurrent();
 
-        this->name = strDup(name);
-        this->errorPrefix = strNewFmt("raised from %s", strPtr(this->name));
-        this->read = read;
-        this->write = write;
-        this->keepAliveTime = timeMSec();
+        *this = (ProtocolClient)
+        {
+            .memContext = memContextCurrent(),
+            .name = strDup(name),
+            .errorPrefix = strNewFmt("raised from %s", strPtr(name)),
+            .read = read,
+            .write = write,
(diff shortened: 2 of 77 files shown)


src/db/db.c
```diff
@@ -72,11 +72,14 @@ dbNew(PgClient *client, ProtocolClient *remoteClient, const String *applicationN
...
-        this->memContext = memContextCurrent();
 
-        this->client = pgClientMove(client, this->memContext);
-        this->remoteClient = remoteClient;
-        this->applicationName = strDup(applicationName);
+        *this = (Db)
+        {
+            .memContext = memContextCurrent(),
+            .client = pgClientMove(client, memContextCurrent()),
+            .remoteClient = remoteClient,
```
cmwshang: In dbNew, it might be clearer to set .memContext = MEM_CONTEXT_NEW() and then pull out .client below the initializer and have it as `this->client = pgClientMove(client, this->memContext)`

src/protocol/client.c
```diff
@@ -80,13 +80,16 @@ protocolClientNew(const String *name, const String *service, IoRead *read, IoWri
...
-        this->memContext = memContextCurrent();
 
-        this->name = strDup(name);
-        this->errorPrefix = strNewFmt("raised from %s", strPtr(this->name));
-        this->read = read;
-        this->write = write;
-        this->keepAliveTime = timeMSec();
+        *this = (ProtocolClient)
+        {
+            .memContext = memContextCurrent(),
```
cmwshang: In protocolClientNew, shouldn't we be using MEM_CONTEXT_NEW() instead of memContextCurrent()?

src/protocol/command.c
```diff
@@ -47,9 +47,12 @@ protocolCommandNew(const String *command)
     MEM_CONTEXT_NEW_BEGIN("ProtocolCommand")
     {
         this = memNew(sizeof(ProtocolCommand));
-        this->memContext = memContextCurrent();
 
-        this->command = strDup(command);
+        *this = (ProtocolCommand)
+        {
+            .memContext = memContextCurrent(),
```
cmwshang: In protocolCommandNew, shouldn't we be using MEM_CONTEXT_NEW() instead of memContextCurrent()?

src/protocol/parallelJob.c
```diff
@@ -46,11 +46,14 @@ protocolParallelJobNew(const Variant *key, ProtocolCommand *command)
...
     {
         this = memNew(sizeof(ProtocolParallelJob));
-        this->memContext = memContextCurrent();
-        this->state = protocolParallelJobStatePending;
 
-        this->key = varDup(key);
-        this->command = protocolCommandMove(command, memContextCurrent());
+        *this = (ProtocolParallelJob)
+        {
+            .memContext = memContextCurrent(),
```
cmwshang: In protocolParallelJobNew, shouldn't we be using MEM_CONTEXT_NEW() instead of memContextCurrent()?

src/protocol/parallelJob.c
```diff
@@ -46,11 +46,14 @@ protocolParallelJobNew(const Variant *key, ProtocolCommand *command)
...
+            .state = protocolParallelJobStatePending,
+            .key = varDup(key),
+            .command = protocolCommandMove(command, memContextCurrent()),
```
cmwshang: Maybe we could follow the same convention here and pull `.command = ` out below the intializer and then set it to `this.command = protocolCommandMove(command, this->memContext);`

src/protocol/server.c
```diff
@@ -56,13 +56,15 @@ protocolServerNew(const String *name, const String *service, IoRead *read, IoWri
...
-        this->memContext = memContextCurrent();
 
-        this->name = strDup(name);
-        this->read = read;
-        this->write = write;
-
-        this->handlerList = lstNew(sizeof(ProtocolServerProcessHandler));
+        *this = (ProtocolServer)
+        {
+            .memContext = memContextCurrent(),
```
cmwshang: In protocolServerNew, shouldn't we be using MEM_CONTEXT_NEW() instead of memContextCurrent()?

dwsteele:
We'll do the memContext()/MEM_CONTEXT_NEW() cleanup in a separate commit once we have decided which direction we want to go.

PR #922, opened 2020-01-28
Fix detection of keepalive and switch to CMake

PierreDucroquet:
Hello

I'm a colleague of @marco44 and I noticed the issues regarding keepalive detection in your current build system.
Since I've worked on quite a few projects using CMake, and your usecase being easily fixed there, I wrote the attached CMakeLists that is able to build your whole project while also fixing the keepalive detection troubles. We tested it on both Linux and FreeBSD with no trouble. While on it, I also fixed a warning regarding compiler flags not being compatible with clang.
If you are interested in this, I can help further to integrate install command or other additional features in the build process.

Regards


dwsteele:
I like the look of cmake much more that configure, but not sure I want to make a change until I get some buy in from the packagers, and pgbackrest has been stable for a release or two.  If we need a quick release we can't be worried about updating packages to get it out.

Also, I tried the patch and it did not work on Linux.  It seems like `${CMAKE_SYSTEM_NAME} EQUAL "Linux"` should be `${CMAKE_SYSTEM_NAME} MATCHES "Linux"`?

dwsteele:
As much as we like the idea of CMake we don't think this is a good time to change make systems.  We also depend on distros for testing (e.g. Ubuntu 12.04) that do not ship with CMake 3.

We are going to use #957 to fix the issue with the build flags.

PR #936, opened 2020-02-12
Improve fork and exec for async archiving.

dwsteele:
The main improvement here is a double-fork to prevent zombie processes if the parent process exits after the (child) async process.  This is a real possibility since the parent process sticks around to monitor the results of the async process.

In the first fork, ignore SIGCHLD in the very unlikely case that the async process exits before the first fork.  This is probably only possible if the async process exits immediately, perhaps due to a chdir() failure.  Set SIGCHLD back to default in the async process so waitpid() will work as expected.

Also update the comment on chdir() to more accurately reflect what is happening.

Finally, add a test in certain debug builds to ensure the first fork exits very quickly.  This only works when valgrind is not in use because valgrind makes forking so slow that it is hard to tell if the async process ran or not.


src/common/fork.c
```diff
@@ -33,12 +35,27 @@ forkDetach(void)
...
+    // The process should never receive a SIGHUP but ignore it just in case
+    signal(SIGHUP, SIG_IGN);
+
+    // There should be no way the child process can exit first (after the next fork) but just in case ignore SIGCHLD.  This means
+    // that the child process will automatically be reaped by the kernel should it finish first rather than becoming defunct.
+    signal(SIGCHLD, SIG_IGN);
+
+    // Fork again and let the parent process terminate to ensure that we get rid of the session leading process. Only session
+    // leaders may get a TTY again.
+    if (forkSafe() > 0)
```
sfrost: This is probably fine but it made me think for a bit about if forkSafe() could possibly return a less-than-zero value.  Seems like it might draw fewer questions to make this the exact opposite of the regular if (forkSafe() == 0) check, ie:

 if (forkSafe() != 0)

Instead, since forkSafe() is defined to return 0 for child, non-zero for parent, and throw an exception on error.  In other words, the interface for forkSafe() isn't quite the same as that of fork() itself.  Anyway, relatively minor point, either way should work.

dwsteele:
Thanks for the review!  Committed at e2c304d47.

src/common/fork.c
dwsteele: Works for me.

PR #941, opened 2020-02-19
Exclude pg_internal.ini.[pid] from backups

dwsteele:
It came up this this thread (https://www.postgresql.org/message-id/flat/20200131045352.GB2631%40paquier.xyz#7700b9481ef5b0dd5f09cc410b4750f6) that PostgreSQL can leave behind temp files for pg_internal.init that have the creating process id as the extension.  Furthermore, Postgres does not seem to clean these files up so they could be left after a crash at just the right time.

While it seems a bug that PostgreSQL does not cleanup these files, it's probably a good idea for us to exclude them from backups just in case.

### DIFF
--- a/src/info/manifest.c
+++ b/src/info/manifest.c
@@ -579,15 +579,16 @@ manifestBuildCallback(void *data, const StorageInfo *info)
                     strPtr(manifestName));
             }
 
-            // Skip pg_internal.init since it is recreated on startup
-            if ((pgVersion <= PG_VERSION_84 || buildData.dbPath) && strBeginsWithZ(info->name, PG_FILE_PGINTERNALINIT))
+            // Skip pg_internal.init since it is recreated on startup.  It's also possible, (though unlikely) that a temp file with
+            // the creating process id as the extension can exist so skip that as well.  This seems to be a bug in PostgreSQL since
+            // the temp file should be removed on startup.  Use regExpMatchOne() here instead of preparing a regexp in advance since
+            // the likelihood of needing the regexp should be very small.
+            if ((pgVersion <= PG_VERSION_84 || buildData.dbPath) && strBeginsWithZ(info->name, PG_FILE_PGINTERNALINIT) &&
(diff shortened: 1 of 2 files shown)


src/info/manifest.c
```diff
@@ -579,8 +579,13 @@ manifestBuildCallback(void *data, const StorageInfo *info)
...
 
-            // Skip pg_internal.init since it is recreated on startup
-            if (strEqZ(info->name, PG_FILE_PGINTERNALINIT))
+            // Skip pg_internal.init since it is recreated on startup.  It's also possible, (though unlikely) that a temp file with
+            // the creating process id as the extension can exist so skip that as well.  This seems to be a bug in PostgreSQL since
+            // the temp file should be removed on startup.  Use regExpMatchOne() here instead of preparing a regexp in advance since
+            // the likelihood of needing the regexp should be very small.
+            if ((pgVersion <= PG_VERSION_84 || buildData.dbPath) && strBeginsWithZ(info->name, PG_FILE_PGINTERNALINIT) &&
+                (strSize(info->name) == sizeof(PG_FILE_PGINTERNALINIT) - 1 ||
+                    regExpMatchOne(STRDEF("\\.[0-9]+"), strSub(info->name, sizeof(PG_FILE_PGINTERNALINIT) - 1))))
```
cmwshang: I think strlen(PG_FILE_PGINTERNALINIT) vs sizeof(PG_FILE_PGINTERNALINIT) - 1 is much more readable. Is there a good reason why we need to use sizeof -1 vs strlen?

dwsteele: Since `PG_FILE_PGINTERNALINIT` is a constant its length can be determined using `sizeof()` at compile time and that value is used instead of a function call.  `strlen()` is a function that is executed each time the code is run.

So, yeah, it's a little less readable but it is more efficient so generally that's the way it is done, when possible.

dwsteele:
Thanks!  Committed at cc743f2e04d.

PR #943, opened 2020-02-25
Attempt to fix flapping test in PITR section of user-guide doc

cmwshang:


dwsteele:
Unfortunately this still failed after a few runs.

PR #947, opened 2020-02-28
Fix user guide pitr flapping test

cmwshang:
Added a 1 second sleep before and after retrieving the time-recovery-timestamp. Adding a sleep before was necessary since adding only a sleep after did not always work. This helps to ensure the backup stop time for the previous backup does not equal to the time-recovery-timestamp. The sleep after allows there to be enough time between the time retrieval and the actual dropping of the table.


dwsteele:
Thanks.  I've run this a number of times now without failure.  Committed at 089049ec563b021a.

PR #948, opened 2020-02-28
Fix real test flapping

cmwshang:
The restore test function was passing the strBackup to the restoreCompare function but when the restore is expected to pick a backup based on a timestamp, the strBackup may not be the one chosen. Modified the code so that the strBackupExpected is set based on the parameters passed to the function and this is then passed to restoreCompare.


dwsteele:
Thanks.  I've run this a number of times now without failure.  Committed at 
ceb050e950c596835a9.

PR #957, opened 2020-03-10
Fix bsd compilation, autodetect operating system

marco44:
The PR may not be 100% correct (the flags may not be correct), but this adds operating system detection. It seems to work on linux machines. The real modifications are in configure.ac and makefile.in.

While working on this, I added the use of pg_config to detect the postgresql library paths, as it didn't compile on my environment, due to unusual path for these libraries.


dwsteele:
Committed at d6ef6c57c33, minus the pg_config changes.  We tried committing the pg_config changes like this before (b1c5885017c035) but it Broke SuSE and RHEL.  Instead, add a PG_CONFIG option and ignore when pg_config doesn't exist, committed as part of 237a3da4d6.

PR #960, opened 2020-03-11
Implement a backup-retention-days option in pgbackrest

PierreDucroquet:
The current backup retention rules in pgbackrest make it quite complicated to implement a date based policy, which are often required regarding the latest privacy ruling.
For instance, if you want to keep 12 months worth of backups, with the current system, you would have to implement a full backup retention limit with a corresponding full backup periodicity, but this would make it impossible later to change any of these settings without breaking your 12 months rules one way or another.
This proposed patch thus adds a second backup retention mode, based on the backups age. It is an addition, not a replacement, and the strictest rule apply.

I have tested this on development systems with an S3 based backup.


src/command/expire/expire.c
```diff
@@ -74,6 +74,72 @@ expireBackup(InfoBackup *infoBackup, String *removeBackupLabel, String *backupEx
     FUNCTION_LOG_RETURN_VOID();
 }
 
+/***********************************************************************************************************************************
+Expire backups based on dates
+***********************************************************************************************************************************/
+static unsigned int
+expireTimeBackup(InfoBackup *infoBackup)
```
cmwshang: Firstly, the code is nicely written and commented! Really well done!
Unfortunately, I don't think this is doing exactly what you expected since index 0 will always be retained and the dependencies are not being expired.  But if you will bear with me, we can work through this together.

Let's take an example. If we start implementing repo-retention-days on a system that has the following backups:

F1 - 40 days old
D1 - diff 1 depends on F1
I1 - incremental 1 depends on D1
F2 - 30 days old
D2 - diff 2 depends on F2
I2 - incremental 2 depends on D2
F3 - 20 days old
F4 - 10 days old

The current code would delete F2 but not F1. It also would not delete D2 or I2 so they would be left on disk and in the backup.info file. The reason for this is that the expireBackup() only expires the label you pass it and not dependencies.  Right now this code is spread in expireBackupFull and expireBackupDiff because combining it is convoluted and didn't make sense before, but now it looks like we have a need. Rather than you trying to do that, I can work on a function and call it expireBackupSet. 

However, let's step back for a minute and define some requirements:
1. Expire each Full backup and all its dependents that are older than the retention days while retaining at least one Full backup that satisfies the number of retention days.
2. Retention days will (for now) only be applied to Full backups.
3. Retention days will supersede repo-retention-full setting.

Given these requirements, and the example backup list from above, if we set repo-retention-days = 28 that means we must have at least one Full backup that is at least 28 days old. So in this scenario, we would only want to delete F1 and its dependencies since F3 is only 20 days old and we mush have a backup that is at least 28 days old.

To accomplish this:
* In cmdExpire, we would check for repo-retention-full-days (let's add "full" to the option name) and call expireTimeBackup else call expireFullBackup. Leave the rest of cmdExpire as is.
* In expireTimeBackup, call infoBackupDataLabelList(infoBackup, backupRegExpP(.full = true)); -- this will get only the list of full backups to check their modification times. For example, from the above backup list, it would return F1, F2, F3, F4 (index 0-3).  Then find the appropriate backups to expire given the requirements.

I am planning to work on the expireBackupSet function next week (March 16...) so if you want to assume that function will, given a backup label, delete all dependencies, that would be fine.

Thanks again, so much, and we look forward to your update!

PierreDucroquet: Thank you for your review. It seems my comments were not enough to explain my code :)

If I take your example:
F1 - 40 days old
D1 - diff 1 depends on F1
I1 - incremental 1 depends on D1
F2 - 30 days old
D2 - diff 2 depends on F2
I2 - incremental 2 depends on D2
F3 - 20 days old
F4 - 10 days old

And the expiration is set at, say, 25 days.
The do-while loop lines 111/121 will go from latest backup to oldest, so F4, F3, I2, D2, and stop at F2 (full backup, older than now - 25 days).
The purge loop lines 127/131 will then start at oldest backup and delete until it meets F2, no matter what type. So it will remove F1, D1, I1 and stop.
This is why the first loop finds the first full backup *older* than retention days. This way we know that everything older, no matter what type, can be purged.

cmwshang: Yes, I see that now (a lesson to me not to rush a review on a Friday!). Going back to the requirements: number 1 and 2 are met so we need to address number 3, tests and, after discussing with David, a slight name change for the option. So, if you could:

1. Rename the option to repo-retention-full-period.  The default will be days, but we want the ability to use weeks, months etc (this will be at a later date so for now, the number will be interpreted as days) so using the word "period" instead of "days" allows that flexibility.
2. In expireTimeBackup, call infoBackupDataLabelList(infoBackup, backupRegExpP(.full = true)); -- this will get only the list of full backups to check their modification times. From the earlier example,  it would return F1, F2, F3, F4 (index 0-3). The check for && (backupType(info.backupType) == backupTypeFull) can then be removed.
3. In cmdExpire, check if repo-retention-full-period option is set and if so, call expireTimeBackup else call expireFullBackup. 
4. Finally, add some unit tests to expireTest.c. See https://github.com/pgbackrest/pgbackrest/blob/master/CONTRIBUTING.md: the section Testing describes writing tests and you should be able to get away with a minimum test environment (which you probably already have but is described in the document as well) and running `pgbackrest/test/test.pl --vm=none --module=command --test=expire`. If you have a fuller environment, you could use --coverage-only testing instead of --vm=none to provide a test/output coverage.html report to help you confirm coverage.

Thanks again.

PierreDucroquet: I will do these patches, but I don't really understand the benefits of calling infoBackupDataLabelList instead of iterating on infoBackup. infoBackupDataLabelList will return a list of string, for each one I will have to call a function to get the corresponding backup information and timestamp... It seems much heavier that simply iterating.

cmwshang: True. We don't actually have a "get backup data from label" so skip number 2.

PierreDucroquet: Hi

I've done the requested changes, but I am struggling with the tests.
`./test/test.pl --vm=none --module=command --test=expire` fails on master with the following error:
̀```
2020-03-18 08:59:02.270 P00  ERROR: [125]: module 'command/expire/expire' is marked 'code' but has no code
module 'command/expire/expire' is marked 'code' but has no code
 at test/lib/pgBackRestTest/Common/JobTest.pm line 737.
        pgBackRestTest::Common::JobTest::end(undef) called at ./test/test.pl line 1054
        eval {...} called at ./test/test.pl line 1325
̀```

If it fails on master, I don't expect it to succeed on my branch... I have lcov 1.14 installed, patched by debian for GCC-9 compatibility so it should not be an issue. Any suggestion regarding that issue?

dwsteele: Well, to start with you can use `--no-coverage` to disable coverage so you can at least work on tests.  Not ideal since you won't know if you have complete coverage.

What version of Debian are you on? I can't repro with any of our environments.  What's the lcov package version?

PierreDucroquet: I am using debian sid with lcov 1.14-2. I also have the exact same issue on a verbatim Ubuntu 18.04 I just installed for this. With --no-coverage, the tests run so I will be able to work on it in the next days, thanks.

dwsteele: That's odd, we use Ubuntu 18.04 to do our coverage testing and haven't see any issues.  Here's the environment we use: https://github.com/pgbackrest/pgbackrest/blob/f538da0571e9e6b224b208fa5f2b3bc3025f5f02/test/Vagrantfile

I'm starting to suspect this is an env/configuration issue that our test code is not handling well.  I have done some work in this area recently that might help.  Can you rebase against master and see if that helps?

dwsteele: Hmmm -- are you also running gcc9 on Ubuntu 18.04?  If so, how did you install it?

PierreDucroquet: @dwsteele indeed much better now after a rebase, thanks! on ubuntu I was using the default gcc 7 provided with ubuntu, but had the same issues.

dwsteele: That's good news!  A lot of a working data was stored in the `.vagrant` directory due to some old (bad) decisions when there were only a few of us doing development.  All of the working data has been moved to other places now.

Do you get full coverage on `master` now?  I've been playing with gcc9 on Fedora 30 and I don't get full coverage on `--module=command --test=expire`.

dwsteele: The issue was Fedora 30 on gcc9 has been fixed in master.  It was due to a slight change in the lcov output format on gcc9.

cmwshang: Hey Pierre - Just touching base to see how things are going. Have you had any luck writing test cases? Do you have an estimated completion date? Please do not hesitate to reach out if you need help with anything.
Thank you for your contribution to pgBackRest!

dwsteele:
Hi Pierre, Cynthia has reworked this patch in #1027 and this is what we are intending to commit. Feel free to review that PR.

PR #961, opened 2020-03-11
Implement dry-run option for Expire command

cmwshang:
When running the Expire command from the command line, the new --dry-run option can be passed to provide the same logging information (especially if logging at 'detail' level) as if the command were going to actually perform expiration but in reality, all writes will be bypassed so that the system will remain the same before and after command execution. 

NOTE: This PR, per original author request, was completed based on https://github.com/pgbackrest/pgbackrest/pull/853.


dwsteele:
Committed at 2fa69af8da40.  Thanks!

PR #968, opened 2020-03-18
Limit reads during backup to file size recorded in manifest.

dwsteele:
Files can grow or shrink during backups which is normal. However, we would prefer not to copy the portion of files that have grown since the manifest was built because that portion will contain all-new blocks that might be torn and will need to be tested to be sure they are in a valid range for the backup. So, we will limit reads to the size reported when the manifest was build and skip all the new blocks. The blocks will be rewritten by Postgres during recovery so storing them also wastes space.

Also have a look at f2548f4 since that's where the feature was added to the storage layer.


dwsteele:
Committed at dcddf3a58bc6.  Thank you for the review!

Issue #940, opened 2020-02-18
intermittent 15 minute slowdown in S3 uploads after system update


slava-pagerduty:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version:
2.23
2. PostgreSQL version:
11.6

3. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each:
Ubuntu 18.04.4 LTS (GNU/Linux 4.15.0-1058-aws x86_64) running on AWS i3.8xlarge (32 CPU 256GB RAM)

4. Did you install pgBackRest from source or from a package?
package

5. Please attach the following as applicable:
    - `pgbackrest.conf` file(s)
[global]
process-max=12
start-fast=y
archive-async=y
spool-path=/var/spool/pgbackrest
repo1-type=s3
repo1-path=/pgbackrest-repo
repo1-retention-full=8
repo1-retention-diff=20
repo1-s3-endpoint=s3.amazonaws.com
repo1-s3-region=us-west-2
repo1-s3-bucket=pd-pg-warehouse
repo1-s3-key=A****************
repo1-s3-key-secret=O**********************
[pg-warehouse]
db-path=/data/pg-warehouse/

    - `postgresql.conf` settings applicable to pgBackRest (`archive_command`, `archive_mode`, `listen_addresses`, `max_wal_senders`, `wal_level`, `port`)
(no problems here)
    - errors in the postgresql log file before or during the time you experienced the issue
(no errors here)
    - log file in `/var/log/pgbackrest` for the commands run (e.g. `/var/log/pgbackrest/mystanza_backup.log`)
see below for a portion of log with relevant information

7. Describe the issue:
probably this is similar to previously reported issue: https://github.com/pgbackrest/pgbackrest/issues/788

After installing update on 2020-02-05 we are seeing intermittent slowdowns in S3 transfer.
Backup transfer would run for few minutes and then sit for approximately 15 minutes after which I see the following messages:
```
common/io/http/client::httpClientRequest: retry KernelError: tls failed syscall: [110] Connection timed out
storage/s3/storage::storageS3Request: retry RequestTimeTooSkewed: The difference between the request time and the current time is too large.
```
 
Here are the log records around such slowdown.
```
[...]

PR #973, opened 2020-03-23
Show local backup pid in the info command

pgstef:
Look for local backup lock and show it in the info command.

Since I'm not sure we can safely detect that a backup is running from any storage file, I try to locate a lock file. It's maybe something we could improve for the future ?


dwsteele:
This patch will only show the backup as running on the host where the backup is running or the pg primary host if the backup is running on a repo host.  That's definitely a limitation, though we might be able to live with it.

It will also give the wrong result if it is run on the pg primary host and the backup is running on the repo host.  In that case it will show the pid of the remote process instead of the backup process.

We are planning a feature to show that a backup is running in the info command, but were not planning to include to pid in that info.  Is there a particular reason you need the pid?

pgstef:
Hi,

Thanks for the review! I wanted to propose the patch so we could discuss the implementation a bit further :-)

Indeed, the pid info will be shown on the pg-host side and give the pid for that server :

```
-bash-4.2$ cat my_stanza-backup.lock
7483
-bash-4.2$ ps -ef | grep 7483
postgres  7483  7482  6 16:19 ?        00:00:00 pgbackrest --delta --log-level-console=off --log-level-file=off --log-level-stderr=error --pg1-path=/var/lib/pgsql/12/data --process=0 --process-max=2 --remote-type=pg --stanza=my_stanza --start-fast backup:remote
```

That's imho an interesting information. It could of course be better to also be able to detect the running process on the repo host and there, give the good pid too.

However, I don't use that info (yet). If you think just saying "false/true" is enough, I can modify this patch, no problem.

What should be the message then ? "local backup running: true/false" ? 

Would this be acceptable with that change ?

How did you imagine being able to detect the running backup command on the repo-host ?

Kind regards,

src/common/lock.c
```diff
@@ -284,3 +284,32 @@ lockRelease(bool failOnNoLock)
...
+    String *backupProcessId = NULL;
+    IoRead *source = storageReadIo(storageNewReadP(storageLocal(), lockFileToFind, .ignoreMissing = 1));
+    if (ioReadOpen(source))
+    {
+        backupProcessId = ioReadLine(source);
+        ioReadClose(source);
+    }
+
+    FUNCTION_LOG_RETURN(STRING, backupProcessId);
+}
```
dwsteele: Since we are not going to capture the pid I think we remove this and just use `lockAcquire()` and `lockRelease()`.

src/common/lock.h
```diff
@@ -29,5 +29,6 @@ Functions
 bool lockAcquire(const String *lockPath, const String *stanza, LockType lockType, TimeMSec lockTimeout, bool failOnNoLock);
 bool lockClear(bool failOnNoLock);
 bool lockRelease(bool failOnNoLock);
+String *lockTypeBackupFound(const String *lockPath, const String *stanza);
```
dwsteele: So this can be removed.

src/command/info/info.c
```diff
@@ -792,6 +795,13 @@ infoRender(void)
...
                         strPtr(varStr(kvGet(stanzaInfo, STANZA_KEY_CIPHER_VAR))));
 
+                    // Is a local backup command running?
+                    if (varStr(kvGet(stanzaInfo, STANZA_KEY_LOCAL_BACKUP_RUNNING_PID_VAR)) != NULL)
+                    {
+                        strCatFmt(resultStr, "    local backup command running pid: %s\n",
+                            strPtr(varStr(kvGet(stanzaInfo, STANZA_KEY_LOCAL_BACKUP_RUNNING_PID_VAR))));
+                    }
+
                     formatTextDb(stanzaInfo, resultStr, backupLabel);
```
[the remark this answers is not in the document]
pgstef: Okay. Pushed a new proposal :-)

pgstef:
Hi David,

Thanks for your feedback. I've pushed a new proposal including the "running" status based on lockAcquire / Release.

If that's better for you, I'll have a look at the documentation.

What kind of tests would you imagine here?

Kind regards,

dwsteele:
OK, the locking code looks good now.

I'm not sure we want to add a new status code, though.  I'm concerned that it might break monitoring systems that are expecting to see 0 for ok.  Perhaps we leave the status code at 0 but update the message to running?

Or do we need a new field in the json to handle running? Perhaps:
`"status":{"code":0,"message":"ok","running":true}`

Any thoughts @cmwshang ?

pgstef:
I think you're absolutely right regarding monitoring systems. Having a new status "key" would be better.

Something like:

```
"status":{"code":0,"message":"ok","running":true}
```

```
stanza: my_stanza
    status: ok
    running: true
```

```
stanza: my_stanza
    status: error (no valid backups)
    running: true
```

This would give us the possibility to say: "ok, there's no valid backup YET, but a new one is running" and raise WARNING alerts rather than CRITICAL alerts (i. e.).

That wouldn't be too difficult to change. I'll modify the code accordingly.

Best regards,

dwsteele:
Perhaps it would be better to report the status of the repo as `ok` if a backup is running even if no completed backups currently exist.  If the first backup errors then it would go back to the error state.  Thoughts?

Also, for the text output I think we should make the output more compact:
```
stanza: my_stanza
    status: ok (running)
```

dwsteele:
Cynthia and I spent some time discussing this today and we identified a few problems with this approach.

1) There's no way to know exactly which command is running when there is a backup lock.  That's because that lock really just means that something is being written to the backup path in the repo.  Currently the commands that do that are `backup`, `expire`, and `stanza-*`. So we know something is happening, just not quite what.  If expire runs after backup then it keeps the same lock, so even if we wrote the command in we would still not be sure it if was backup or expire that was running.

2) Partly because of the uncertainty of #1 we're not sure how best to represent this in the JSON.  We guarantee that the JSON won't change so it is pretty important to get it right the first time.

What we would like to do is take a few days to step back from this and think about the best way to proceed.  I don't want to be giving you new direction about this every day or so.

Of course, we're happy to have your input if you have ideas on how to solve either problem.

pgstef:
As you mentioned, by checking the lock file, we only know "something" is running. That's why we can't set the status "ok" if a backup is running even if no completed backups currently exist. That's also why, IMHO we should keep the "running" info in a separated line of the text output. The current JSON output will allow monitoring systems to raise different types of alerts combining the real status with the running flag. For those reasons, I really like the current patch.

Regarding future improvement, we could find a way to know precisely *what* is running and *where*. But that would then be another part of the info text/JSON. So, having the "running" flag would tell us "Oh, something is running. Go see that info part to know more precisely...". IMHO, the running flag would then still be interesting and wouldn't move. What triggers it would change but not the output itself.

TBH, I think the current patch is a very good basis for the future and would already be useful now.

dwsteele:
Cynthia and I discussed this and we think it is important to be able to differentiate between backup and expire so we don't need to keep changing the JSON output as we improve this command. We believe we have a method for that, but for now nailing down the JSON format is more important.

I've invited some colleagues to comment so hopefully we'll get some fresh ideas from them.

dwsteele:
I spent some time talking to Stephen and Cynthia about this and we agree that we should only report what we know. What we know is that we have a backup lock so that's what we should report in the JSON. I'm thinking `lock:{backup:{held:true/false}}` (in the `status` key) to get started. That way we have a place to add `pid`, `command`, main `pid`, etc., later.

For the text output I think `ok (backup/expire running)` is fine for now. We can improve that in the future by adding `command` to the lock file.

So, if you agree, go ahead and update the JSON and the text output and let us have a look. Then you'll be ready to move on to tests.

pgstef:
Hi,

Thanks very much for the feedback.

This proposal seems very reasonable and make sense. I'll update the patch asap.

Kind regards

src/command/info/info.c
```diff
@@ -786,7 +811,17 @@ infoRender(void)
...
+                        // Get the specific lock part
+                        KeyValue *lockKv = varKv(kvGet(stanzaStatus, STATUS_KEY_LOCK_VAR));
+                        KeyValue *backupLockKv = varKv(kvGet(lockKv, STATUS_KEY_LOCK_BACKUP_VAR));
+
+                        // Change displayed status if lock is found
+                        if (varBool(kvGet(backupLockKv, STATUS_KEY_LOCK_BACKUP_HELD)))
+                            strCatFmt(resultStr, "%s (%s)\n", INFO_STANZA_STATUS_OK, strPtr(varStr(STATUS_KEY_LOCK_BACKUP_STR)));
+                        else
+                            strCatFmt(resultStr, "%s\n", INFO_STANZA_STATUS_OK);
+                    }
```
dwsteele, line 824: This looks pretty good, but shouldn't we also get here if there are no backups but a backup is running? So instead of `status: error (no valid backups)` you'd get `status: ok (backup/expire running`?

At line 769, perhaps `if (!varBool(kvGet(backupLockKv, STATUS_KEY_LOCK_BACKUP_HELD)) && statusCode != INFO_STANZA_STATUS_CODE_OK))`?

pgstef: I'm not really feeling confortable with saying "ok" if we don't know (yet) that the backup will be correct. It would also give inconsistency between the JSON and the text output regarding the status.

I would rather "add" the information to the error message. I'll patch the code in that way.

If you prefer having "ok" even with JSON saying "error", I'll adjust the IF as you suggest.

dwsteele: I like the way you did it, so time for tests!

Basically, what we need is to create a lock in infoTest.c at line 182 and 359. Then rerun the previous two tests at each location and make sure the JSON and text have been updated appropriately. Yes, this means recopying a lot of text, but that's how the tests work. Be sure to remove the lock right after your tests in each section. When the coverage report says 100% then you are done!

After the tests are done we'll talk about documentation.

pgstef: Great, Thanks for the review !

Should I rebase my branch before the tests part ?

dwsteele: Rebasing often is generally a good idea, but I don't think this area of the code has changed in a while.

pgstef: Rebased, thanks.
I'll now have a look at the tests part :-)

pgstef:
Hi,

Coverage tests has been updated based on your advice.

The coverage summary seems fine:

```xml
<table-row>
    <table-cell>command/info</table-cell>
    <table-cell>7/7 (100.0%)</table-cell>
    <table-cell>168/168 (100.0%)</table-cell>
    <table-cell>360/360 (100.0%)</table-cell>
</table-row>
```

And the coverage test also seems good:

```
INFO: P1-T1/1 - vm=u18, module=command, test=info (9.85s)
INFO: writing C coverage report
INFO: TESTS COMPLETED SUCCESSFULLY (12s)
```

Does it looks good for you too?

test/src/module/command/infoTest.c
```diff
@@ -180,6 +183,78 @@ testRun(void)
             "        wal archive min/max (9.4-3): none present\n",
             "text - single stanza, no valid backups");
 
+        // repeat those tests while a backup lock has been taken
+        HARNESS_FORK_BEGIN()
+        {
```
dwsteele, line 188: Hmmm, interesting, I didn't think these HARNESS_FORK() blocks would be required but now I can see that they are. I'm not very excited about the 1000ms sleeps, but that's not your problem to fix since we do it in other places.

Anyway, I'd like to make this pattern better, but I'm glad you were able to find examples and make it work. Nicely done.

src/command/info/info.c
```diff
@@ -444,16 +457,28 @@ stanzaInfoList(const String *stanza, StringList *stanzaList, const String *backu
         kvPut(varKv(stanzaInfo), STANZA_KEY_BACKUP_VAR, varNewVarLst(backupSection));
         kvPut(varKv(stanzaInfo), KEY_ARCHIVE_VAR, varNewVarLst(archiveSection));
 
+        // If a status has not already been set, check if there's a local backup running
+        static bool runningBackup = false;
+        if (kvGet(varKv(stanzaInfo), STANZA_KEY_STATUS_VAR) == NULL)
+        {
+            // Try to acquire a lock. If not possible, assume another backup is already running.
+            static LockType lockTypeToFind = lockTypeBackup;
```
dwsteele, line 465: This declaration isn't needed.  Just put `lockTypeBackup` in the function below.

src/command/info/info.c
```diff
@@ -444,16 +457,28 @@ stanzaInfoList(const String *stanza, StringList *stanzaList, const String *backu
...
+            runningBackup = lockAcquire(cfgOptionStr(cfgOptLockPath), stanzaListName, lockTypeToFind, 0, false) == true ? false : true;
```
dwsteele, line 466: Wouldn't this be simpler as:
```
runningBackup = !lockAcquire(cfgOptionStr(cfgOptLockPath), stanzaListName, lockTypeToFind, 0, false)
```

src/command/info/info.c
```diff
@@ -444,16 +457,28 @@ stanzaInfoList(const String *stanza, StringList *stanzaList, const String *backu
...
+
+            // Release immediately the lock acquired
+            lockRelease(false);
```
dwsteele, line 469: How about `lockRelease(!runningBackup);` here?

src/command/info/info.c
```diff
@@ -77,15 +78,22 @@ STRING_STATIC(INFO_STANZA_STATUS_MESSAGE_NO_BACKUP_STR,             "no valid ba
 #define INFO_STANZA_STATUS_CODE_MISSING_STANZA_DATA                 3
 STRING_STATIC(INFO_STANZA_STATUS_MESSAGE_MISSING_STANZA_DATA_STR,   "missing stanza data");
 
+// Locks
+VARIANT_STRDEF_STATIC(STATUS_KEY_LOCK_VAR,                          "lock");
+VARIANT_STRDEF_STATIC(STATUS_KEY_LOCK_BACKUP_VAR,                   "backup");
+VARIANT_STRDEF_STATIC(STATUS_KEY_LOCK_BACKUP_HELD,                  "held");
```
dwsteele, line 84: `STATUS_KEY_LOCK_BACKUP_HELD` should be `STATUS_KEY_LOCK_BACKUP_HELD_VAR`.

I also think these three keys should be moved up with the test of the keys.

test/src/module/command/infoTest.c
pgstef: Thanks.
Indeed, the "asynchronous" fork need to wait before the lockRelease. It could be really cool to have both fork and parent being able to communicate to wait each-other. But indeed, that's a bigger problem :-)

src/command/info/info.c
pgstef: I think I already tested that and it didn't work. I've just retested it now (post-rebase) and it works indeed. Modified. Thanks

src/command/info/info.c
pgstef: Good catch. Thanks.

src/command/info/info.c
pgstef: Very clever ! Thanks

src/command/info/info.c
pgstef: Okay. Modified.

pgstef:
Thanks for your careful review, and your patience.
I've modified the code according to your advices.

Should I now have a look at the documentation part ? (Not sure where to put that info)

Kind regards,

dwsteele:
Ah, right, documentation. Thanks for reminding me.

We should mention it in `user-guide.xml` around line 1161, I think. This just applied to the text format.

We don't really have any documentation for the JSON format at all. This a problem but it's not the job of this commit to fix it.

pgstef:
What would you then think about something simple like :

```xml
<p>Next to the '<id>status</id>' information, the specific '<i>backup/expire running</i>' message will appear if one of those commands are currently running on the host.</p>
```

dwsteele:
Sure, that seems reasonable.

pgstef:
Hi,

I've pushed that simple message in the user-guide. Is that enough?

I tried to build the doc (`doc.pl --out=html`) but I get errors:

```
E: The repository 'http://security.ubuntu.com/ubuntu disco-security Release' does not have a Release file.
E: The repository 'http://archive.ubuntu.com/ubuntu disco Release' does not have a Release file.
E: The repository 'http://archive.ubuntu.com/ubuntu disco-updates Release' does not have a Release file.
E: The repository 'http://archive.ubuntu.com/ubuntu disco-backports Release' does not have a Release file.
```

I'll try again later if needed.

Kind regards,

dwsteele:
> I've pushed that simple message in the user-guide. Is that enough?

I think so.

> I tried to build the doc (doc.pl --out=html) but I get errors:

Yeah, Ubuntu 19.04 expired out of the repos yesterday. We have fixed it in master so you can rebase/merge and try again.

pgstef:
The `doc.pl --out=html` command runs indeed correctly after refresh.
Nothing left to commit I think.

Is it good / complete for you now ?

Thanks again for the explanations and your patience.
Kind regards,

dwsteele:
Committed at c88684e2bf32. Thank you for the review!

dwsteele:
Sorry, I closed the wrong PR.

dwsteele:
BTW -- everything looks good with the PR but I won't be able to get back to it until early next week.

dwsteele:
We have pushed the non-blocking TLS code into master at c88684e2bf3. If you could test that over the weekend it would be greatly appreciated. That's our release candidate so if you give the thumbs up we'll release it on Monday.

Also, please let us know how you'd like to appear in the release notes.

Thanks again! We really could not have done this without your help.

pgstef:
Hello,

For the release notes, you can proceed like you usually do, no problem for me. You can use my contributor id (I think it's already there somewhere).

Regarding the TLS commit, do you have an issue with the original problem ? So I could know "what" to test and try to give you some feedback ? 
In any case, I can at least try to run all my regress/performance tests for check_pgbackrest with both 2.25 and master but I wouldn't be sure to know what to look at exactly ;-)

Thanks again for your help and patience in this PR.

Kind regards,

dwsteele:
> You can use my contributor id (I think it's already there somewhere).

This PR will be committed with you as the author. I have the info from your branch.

> Regarding the TLS commit, do you have an issue with the original problem

The issue is #940 but it may be hard for you to test unless you are already experiencing the problem. We have not been able to reproduce it despite numerous efforts.

Even so, any testing against master is always appreciated.

pgstef:
Okay, I see now that your comment (https://github.com/pgbackrest/pgbackrest/pull/973#issuecomment-614906444) was for that issue.
Since I don't really use S3 buckets very often, I indeed never had that issue.

I've just run the complete regression/performance test cases I have for check_pgbackrest comparing 2.25 and master. Didn't found any weird behaviour. Small performance improvement though.

Have a nice week-end,
Kind regards,

dwsteele:
OK, I've completed my final review and made some adjustments at 01882f9274. Mostly formatting changes/fixes, but also renamed one var for consistency and renamed/moved the backup/expire message.

If these changes look good to you then we are ready to commit.

pgstef:
Hello,
That looks great.
Thank you for the review !
Kind regards,

dwsteele:
Committed at e92eb709d6. Thank you!

PR #977, opened 2020-03-25
Expire test refactor to replace unnecessary disk writes/reads to speed testing.

cmwshang:
There are several tests in which we can now use ioRead objects instead of writing the info files to disk. This PR replaces those disk writes with ioRead objects in order to speed up testing and streamline the tests a bit.


dwsteele:
Committed at 480a11066db83.

PR #978, opened 2020-03-25
Fix backup-prior setting in backupCreate so DIFF has FULL as prior.

cmwshang:
In the ExpireEnvTest.pm backupCreate function, backup-prior was incorrectly set for DIFF backups to the previous backup regardless of what backup type the previous backup was. This did not cause any issues in the Mock Expire tests before because it was not being checked. However, in order to reduce churn in the expect logs for a new feature where the backup-prior is utilized, this is being fixed now.


dwsteele:
Committed at eb7f7dd5caf9f4.  Thanks!

PR #983, opened 2020-03-30
Remove/add dependent backups during backup.info reconstruct

cmwshang:
Remove/add dependent backups during backup.info reconstruct.

Prior to performing a backup or expiring backups, the backup.info file is validated by reconstructing it from the backups in the repository. When a backup had already been removed from the repo, it was removed from the backup.info file but its dependents were not. Now, the dependent backups will also be removed from backup.info and only backups in the repo that have their full dependency chain will be added to backup.info if they are missing.


dwsteele:
Committed with the addition of one linefeed at 3fbde30c6fac4.

PR #984, opened 2020-03-30
Simplify storage driver info and list functions.

dwsteele:
The storage driver requires two list functions to be implemented, `list` and `infoList`. But the former is a subset of the latter so implementing both in every driver is wasteful.  The reason both exist is that in Posix it is cheaper to get a list of names than it is to stat files to get size, time, etc.  In S3 these operations are equivalent.

Introduce `storageInfoLevelType` to determine the amount of information required by the caller.  That way Posix can work efficiently and all drivers can return only the data required which saves some bandwidth.  The `storageList()` and `storageInfoList()` functions remain in the storage interface since they are useful -- the only change is simplifying the drivers with no external impact.

Note that since `list()` accepted an expression `infoList()` must now do so. Checking the expression is optional for the driver but can be used to limit results or save IO costs.

Similarly, `exists()` and `pathExists()` are just specialized forms of `info()` so adapt them to call `info()` instead.


src/storage/posix/storage.c
```diff
@@ -341,7 +263,8 @@ storagePosixMove(THIS_VOID,  StorageRead *source, StorageWrite *destination, Sto
             // Determine which file/path is missing
             if (errno == ENOENT)
             {
-                if (!storageInterfaceExistsP(this, sourceFile))
+                // Check if the source is missing. Rename does not follow links so there is no need to set followLink.
+                if (!storageInterfaceInfoP(this, sourceFile, storageInfoLevelExists).exists)
```
cmwshang, line 267: But `storageInterfaceInfoP(this, sourceFile, storageInfoLevelExists).exists` does not check for whether this is a file or a directory and in the old code, `storageInterfaceExistsP(this, sourceFile)` would return false if it were a directory, so this is a change in behavior -was that intentional?

src/storage/info.h
```diff
@@ -6,6 +6,25 @@ Storage Info
...
+Specify the level of information required when calling functions that return StorageInfo
+***********************************************************************************************************************************/
+typedef enum
+{
+    // The info type is determined by driver capabilities.  This mimics the prior behavior where drivers would always return as
+    // much information as they could.
+    storageInfoLevelDefault = 0,
+
+    // Only test for existence.  All drivers support this type.
+    storageInfoLevelExists,
```
cmwshang, line 19: storageInfoLevelExists is only applicable to *InfoList function vs *Info function which makes this confusing. Why is it not implemented in the *Info function?

src/storage/posix/storage.c
```diff
@@ -42,55 +42,23 @@ struct StoragePosix
     MemContext *memContext;                                         // Object memory context
 };
 
-/**********************************************************************************************************************************/
-static bool
-storagePosixExists(THIS_VOID, const String *file, StorageInterfaceExistsParam param)
```
cmwshang, line 47: I think removing the Exists() interface functions is a bit of a mistake. These are more streamlined because checking existence should not care about type, size, timeModified but now we are setting these when we don't need to and making exceptions in the code. I think if you had implemented the storageInfoLevelExists in the *Info functions to replace the *Exists functions I'd feel differently, but I think all we've done is make things more confusing for no gain.

src/storage/storage.c
```diff
@@ -159,17 +157,18 @@ storageExists(const Storage *this, const String *pathExp, StorageExistsParam par
...
         // Create Wait object of timeout > 0
         Wait *wait = param.timeout != 0 ? waitNew(param.timeout) : NULL;
 
         // Loop until file exists or timeout
         do
         {
             // Call driver function
-            result = storageInterfaceExistsP(this->driver, path);
+            StorageInfo info = storageInfoP(
+                this, pathExp, .level = storageInfoLevelBasic, .ignoreMissing = true, .followLink = true);
```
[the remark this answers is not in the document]
dwsteele: I'll update this to fix the comment and explain why basic level required.

src/storage/posix/storage.c
dwsteele: This fixes an existing bug where a move on path with a missing destination would error on path missing rather than creating the destination.

Error should be `unable to move missing source`.

src/storage/storage.c
```diff
@@ -474,18 +489,20 @@ storageInfoList(
...
+                .expression = param.expression,
                 .sortOrder = param.sortOrder,
                 .recurse = param.recurse,
                 .path = path,
             };
 
-            if (param.expression != NULL)
-                data.expression = regExpNew(param.expression);
+            if (data.expression != NULL)
+                data.regExp = regExpNew(param.expression);
```
cmwshang, line 499: This is redundant for S3 but not for Posix. In s3 we pass the expression and then in storageS3ListInternal we create the regExp again there instead. So by the time we get back to storageInfoListCallback is it meaningless to apply the regExp because in s3 that was handled at the driver level but in Posix it was not.

src/storage/s3/storage.c
```diff
@@ -662,65 +647,14 @@ storageS3InfoList(
 
     MEM_CONTEXT_TEMP_BEGIN()
     {
-        StorageS3InfoListData data = {.callback = callback, .callbackData = callbackData};
-        storageS3ListInternal(this, path, false, false, storageS3InfoListCallback, &data);
+        StorageS3InfoListData data = {.level = level, .callback = callback, .callbackData = callbackData};
+        storageS3ListInternal(this, path, param.expression, false, storageS3InfoListCallback, &data);
```
cmwshang, line 651: I don't see any regExp tests in s3Test (maybe you can point them out?) - is it not possible/practical to have a test there?

src/storage/storage.c
```diff
@@ -517,23 +554,16 @@ storageList(const Storage *this, const String *pathExp, StorageListParam param)
...
-        String *path = storagePathP(this, pathExp);
-
-        // Get the list
-        result = storageInterfaceListP(this->driver, path, .expression = param.expression);
+        result = strLstNew();
 
-        // If the path does not exist
-        if (result == NULL)
+        // Build an empty list if the directory does not exist by default.  This makes the logic in calling functions simpler when
+        // the caller don't care if the path is missing.
```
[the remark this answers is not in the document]
dwsteele: Will fix.

src/storage/s3/storage.c
dwsteele: There are tests in s3Test.c at 935,-940. These are calling storageList() but storageList() calls storageInfoList().

dwsteele: Converted these to storageInfoList() tests to make the coverage more obvious.

src/storage/storage.c
```diff
@@ -343,7 +351,7 @@ storageInfoListSort(
         // If no sorting then use the callback directly
         if (sortOrder == sortOrderNone)
         {
-            result = storageInterfaceInfoListP(this->driver, path, callback, callbackData);
+            result = storageInterfaceInfoListP(this->driver, path, level, callback, callbackData, .expression = expression);
```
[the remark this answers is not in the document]
dwsteele: Use of the expression by the driver is meant to be optional and used only if the driver can avoid some expensive disk/network IO. This has now been documented in `storage.intern.h`.

That said, the S3 driver should not have been doing the full regexp check. It should only be using the prefix to limit the results that come back from S3.

The overall idea is to move this check up to the storage interface so the drivers don't need to worry about it unless they have a good reason to. The potential extra callbacks are not a big worry since the IO needed to get the list is many orders of magnitude more expensive.

src/storage/storage.c
dwsteele: Yeah, it should not have been fully applied in S3. This has been fixed. See comment above.

src/storage/storage.c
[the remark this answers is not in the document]
dwsteele: Also added this to the PR description which will become the commit comment.

src/storage/posix/storage.c
```diff
@@ -465,39 +388,46 @@ storagePosixPathCreate(
...
-    {
-        if (errno != ENOENT)
-            THROW_SYS_ERROR_FMT(PathOpenError, "unable to stat '%s'", strPtr(path));
+        // Rather than stat the file to discover what type it is, just try to unlink it and see what happens
+        if (unlink(strPtr(file)) == -1)
+        {
+            // These errors indicate that the entry is actually a path so we'll try to delete it that way
+            if (errno == EPERM || errno == EISDIR)              // {uncovered_branch - no EPERM on tested systems}
+            {
+                storageInterfacePathRemoveP(data->driver, strNewFmt("%s/%s", strPtr(data->path), strPtr(info->name)), true);
```
cmwshang, line 420: `String *file` is already set to `strNewFmt("%s/%s", strPtr(data->path), strPtr(info->name))` so why can't you use file in this function call? 
`storageInterfacePathRemoveP(data->driver, file, true);`

src/storage/remote/protocol.c
```diff
@@ -143,18 +142,22 @@ storageRemoteInfoWrite(ProtocolServer *server, const StorageInfo *info)
...
+    if (info->level >= storageInfoLevelDetail)
+    {
+        protocolServerWriteLine(server, jsonFromUInt(info->userId));
+        protocolServerWriteLine(server, jsonFromStr(info->user));
+        protocolServerWriteLine(server, jsonFromUInt(info->groupId));
+        protocolServerWriteLine(server, jsonFromStr(info->group));
+        protocolServerWriteLine(server, jsonFromUInt(info->mode));
+
+        if (info->type == storageTypeLink)
+            protocolServerWriteLine(server, jsonFromStr(info->linkDestination));
```
cmwshang, line 159: Can you help me to understand why storageTypeLink is only written to protocol server if the level >= storageLevelDetail. I know there is still a "default" level and why wouldn't storageInfoLevelExists not write out link?

src/storage/remote/storage.c
```diff
@@ -95,41 +66,47 @@ storageRemoteInfoParse(ProtocolClient *client, StorageInfo *info)
...
+    if (info->level >= storageInfoLevelDetail)
+    {
+        info->userId = jsonToUInt(protocolClientReadLine(client));
+        info->user = jsonToStr(protocolClientReadLine(client));
+        info->groupId = jsonToUInt(protocolClientReadLine(client));
+        info->group = jsonToStr(protocolClientReadLine(client));
+        info->mode = jsonToUInt(protocolClientReadLine(client));
+
+        if (info->type == storageTypeLink)
+            info->linkDestination = jsonToStr(protocolClientReadLine(client));
```
cmwshang, line 83: Same comment about storageTypeLink as above in src/storage/remote/protocol.c

test/src/module/storage/s3Test.c
```diff
@@ -200,7 +200,11 @@ testS3Server(void)
```
cmwshang, line 200: Just a question since I don't see any common expressions in regExpPrefix nor here, but a common expression would be `^[0-F]{16}$` so I just want to confirm that the regExpPrefix would return NULL but the s3Test would still filter the return based on the expression, yes?

src/storage/posix/storage.c
dwsteele: No reason that I can see. Fixed.

src/storage/remote/protocol.c
dwsteele: Links require Posix and Posix requires storageInfoLevelDetail so they kind of got tied together. No reason this can't be split out, though, if you think it will be clearer.

test/src/module/storage/s3Test.c
dwsteele: The filtering happens in storageInfoList(), but yes.

src/storage/remote/storage.c
dwsteele: Same response as above... :)

src/storage/remote/protocol.c
dwsteele: Actually, this is specified in the StorageInfo type as only being returned at level detail. I'm OK wit leaving it that way.

dwsteele:
Committed at 5e55d588507e794. Thank you for the review!

PR #994, opened 2020-04-09
Update contributing documentation and remove test/README.md

cmwshang:
When the Vagrant file was updated to use `pgbackrest/` vs `/backrest/` as the location for executing tests and building the documentation, parts of the contributing.xml (and hence the CONTRIBUTING.md) were not updated since parts of the document are not actually executed when the CONTRIBUTING.md is built from contributing.xml: those parts that are executed were updated but those parts that are not executed were not. 

This commit fixes the contributing.xml issue but also removes the test/README.md as its contents were out of date and redundant given that they are covered in the CONTRIBUTING.md.


dwsteele:
Committed at 63b9f46a34ce20c1. Thank you!

PR #1003, opened 2020-04-15
Non-blocking TLS implementation.

dwsteele:
The prior blocking implementation seemed to be prone to locking up on some (especially recent) kernel versions. Since we were unable to reproduce the issue in a development environment we can only speculate as to the cause, but there is a good chance that blocking sockets were the issue or contributed to the issue.

So move to a non-blocking implementation to hopefully clear up these issues. Testing in production environments that were prone to locking shows that the approach is promising and at the very least not a regression.

The main differences from the blocking version are the non-blocking connect() implementation and handling of WANT_READ/WANT_WRITE retries for all SSL*() functions.

Timeouts in the tests needed to be increased because socket connect() and TLS SSL_connect() were not included in the timeout before. The tests don't run any slower, though. In fact, all platforms but Ubuntu 12.04 worked fine with the shorter timeouts.

Note that options will be added for both blocking mode (sck-block) and I/O timeout (io-timeout) but they will be noisy so will be done in separate commits.


src/common/io/tls/session.c
```diff
@@ -144,77 +184,33 @@ tlsSessionRead(THIS_VOID, Buffer *buffer, bool block)
...
 
+    // If no TLS data pending then check the socket to reduce blocking
+    if (!SSL_pending(this->session))
+        sckSessionReadyRead(this->socketSession);
+
     // If blocking read keep reading until buffer is full
     do
     {
-        // If no tls data pending then check the socket
-        if (!SSL_pending(this->session))
```
cmwshang, line 151: Should we really move this out of the loop? Does postgres do it outside the loop?

src/common/io/socket/common.c
```diff
@@ -129,6 +141,52 @@ sckOptionSet(int fd)
...
+sckConnect(int fd, const String *host, unsigned int port, const struct addrinfo *hostAddress, TimeMSec timeout)
+{
+    FUNCTION_LOG_BEGIN(logLevelTrace);
+        FUNCTION_LOG_PARAM(INT, fd);
+        FUNCTION_LOG_PARAM(STRING, host);
+        FUNCTION_LOG_PARAM(UINT, port);
+        FUNCTION_LOG_PARAM_P(VOID, hostAddress);
+        FUNCTION_LOG_PARAM(TIME_MSEC, timeout);
+    FUNCTION_LOG_END();
+
```
cmwshang, line 161: Are we missing some ASSERTs here - can *host, or *host Address be NULL?

src/common/io/socket/common.c
```diff
@@ -129,6 +141,52 @@ sckOptionSet(int fd)
...
+    // Attempt connection
+    if (connect(fd, hostAddress->ai_addr, hostAddress->ai_addrlen) == -1)
+    {
+        // Save the error
+        int errNo = errno;
+
+        // The connection has started but since we are in non-blocking mode it has not completed yet
+        if (sckConnectInProgress(errNo))
```
cmwshang, line 169: calls to connect() for the same socket, before the connection is established, shall fail and set errno to [EALREADY]" meaning "a connection request is already in progress for the specified socket." - if this happens we will get to line 183 and error - is that really what we want?

src/common/io/socket/common.c
```diff
@@ -129,6 +141,52 @@ sckOptionSet(int fd)
...
+        {
+            // Wait for write-ready
+            if (!sckReadyWrite(fd, timeout))
```
cmwshang, line 172: I'm guessing choosing to sckReadyWrite vs sckReadyRead was arbitrary?

src/common/io/tls/session.c
dwsteele: Postgres has it in the loop and so should we.

src/common/io/socket/common.c
dwsteele: Yes, will add.

src/common/io/socket/common.c
dwsteele: We should never try to connect twice on the same socket, so yeah it should be a hard error.

src/common/io/socket/common.c
dwsteele: A socket won't be read ready until the other side writes something, but it will be write ready as soon as the connection is complete.

src/common/io/socket/client.c
```diff
@@ -128,9 +124,7 @@ sckClientOpen(SocketClient *this)
                     THROW_ON_SYS_ERROR(fd == -1, HostConnectError, "unable to create socket");
 
                     sckOptionSet(fd);
-
-                    if (connect(fd, hostAddress->ai_addr, hostAddress->ai_addrlen) == -1)
-                        THROW_SYS_ERROR_FMT(HostConnectError, "unable to connect to '%s:%u'", strPtr(this->host), this->port);
+                    sckConnect(fd, this->host, this->port, hostAddress, waitRemaining(wait));
```
cmwshang, line 127: So here we are passing what the remaining time is in the wait object created in this sckClientOpen function but then in the CATCH_ANY() (line 145 that you cannot see here) the wait object time is not affected by the calls made to sckReady functions via sckConnect. So sckConnect could use up a good portion (or all?) of the wait time from the wait object here, but we sould ignore that because the wait in the Ready function is a different object. This seems a little incorrect to me...

src/common/io/socket/common.c
```diff
@@ -129,6 +141,52 @@ sckOptionSet(int fd)
...
+                THROW_FMT(HostConnectError, "timeout connecting to '%s:%u'", strPtr(host), port);
+
+            // Check that the connection was successful
+            socklen_t errNoLen = sizeof(errNo);
+
+            THROW_ON_SYS_ERROR(
+                getsockopt(fd, SOL_SOCKET, SO_ERROR, &errNo, &errNoLen) == -1, HostConnectError, "unable to get socket error");
```
cmwshang, line 179: So here you are calling getsockopt with SO_ERROR to actually check for and clear an error at the same time? When I first read trough this section of the code, I thought we were always erroring because we are getting SO_ERROR. Maybe just a comment to say "call getsockopt with SO_ERROR to clear any errors if the connection was successful or report on an unsucessful connection" - or something. Dunno. Your choice. It's clever but it just took me a bit to figure out what was going on here.

src/common/io/tls/session.c
```diff
@@ -144,77 +184,33 @@ tlsSessionRead(THIS_VOID, Buffer *buffer, bool block)
...
-            sckSessionReadyRead(this->socketSession);
-
         // Read and handle errors
-        result = SSL_read(this->session, bufRemainsPtr(buffer), (int)bufRemains(buffer));
+        result = tlsSessionResult(this, (int)SSL_read(this->session, bufRemainsPtr(buffer), (int)bufRemains(buffer)), true);
```
cmwshang, line 195: Why did you feel it necessary to cast the SSL_read result to an (int) - the function is declared as returning an int and I see no where else in the code where a call to tlsSessionResult from an SSL_write, SSL_connect, SSL_accept (all returning an int) needed a cast, so why here?

src/common/io/socket/client.c
dwsteele: waitMore() is checking the system clock to find elapsed time so any time used in sckConnect() will be accounted for.

src/common/io/socket/common.c
dwsteele: Yeah, I can add to the comment, but we don't want two error codes because that makes to code overall more complicated.

dwsteele: Changed the comment to `Check for success or error. If the connection was successful this will set errNo to 0.`

src/common/io/tls/session.c
dwsteele: Hmmm, not sure. This may have been copied from `IoHandleWrite` which is using `ssize_t` or maybe I wanted to make it look more like size_t which is used by `Buffer`. 

Anyway, all that is useless so I have changed result to `int` and removed the cast.

dwsteele: Yes, and `ssize_t result` was changed to `int result` since that's what is actually returned.

src/common/io/tls/session.c
```diff
@@ -75,52 +75,92 @@ tlsSessionClose(TlsSession *this, bool shutdown)
...
+
+    ASSERT(this != NULL);
+    ASSERT(this->session != NULL);
+
+    // Process errors
+    if (result <= 0)
+    {
+        // Get TLS error and store errno in case of syscall error
+        int errorTls = SSL_get_error(this->session, result);
+        int errorSys = errno;
```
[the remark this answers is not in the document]
dwsteele: Yeah -- it's fine for the SSL_error but probably not for errno. I tried capturing errno right after the SSL_read() but it was zero in all error cases that I could cause with testing. People have complained about that on the interwebs, too.

I think we can improve this but it doesn't change the program flow. There aren't any recoverable syscall errors.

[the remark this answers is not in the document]
dwsteele: Also, it's not clear to. me when errno should be captured. PG does it after the SSL_get_error() and ERR_get_error() calls like here, but there's less in between.

Having said that, I don't think anything in our debug macros is resetting errno. We only do it in cvtZToUInt64Internal() and cvtZToInt64Internal(), which is not done in the debug macros.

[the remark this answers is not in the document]
sfrost: I was specifically worried that some of those macros might expand in certain cases to do things like snprintf like in typeToLog() which can set errno on some platforms.

The PG code does SSL_write, SSL_get_error, then calls ERR_get_error if SSL_write returned < 0 OR if SSL_get_error returned anything other than SSL_ERROR_NONE, and then looks at errno only if SSL_ERROR_SYSCALL was returned from SSL_get_error.

It seems like the SSL code is perhaps managing its own queue of errno results and setting it only when SSL_ERROR_SYSCALL is returned from SSL_get_error (and that's probably the only time we should actually be looking at errno after SSL_get_error...  maybe we should also explicit set errno to 0 before calling any of the above, like the PG code does), but I have to admit that I'm still not thrilled at having so much happening between these calls, which is quite unlike the PG code..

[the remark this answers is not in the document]
dwsteele: I haven't been able to get any joy from ERR_get_error() either, but it may have just been the test cases I was trying. I tried a bunch of different ways to get some value from errno while trying to debug this issue.

Agreed, this is area for improvement, but I don't think we should do it now. Historically the value of this information has been pretty much NIL and it's only available at debug level so it is mostly never seen. Errors from S3 are pretty common so we don't log them at a higher level.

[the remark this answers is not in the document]
dwsteele: Also, I do hope `snprintf` is not resetting errno on success. If that's happening on some platforms then they won't get detailed info in the error, not that we do on any platform as far as I can tell.

dwsteele:
Committed at c88684e. Thank you for the review!

Issue #999, opened 2020-04-14
check resulting in segmentation fault


jerome-peng:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version:  pgBackRest 2.25

2. PostgreSQL version: 12.2

3. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each: CentOS Linux release 7.6.1810 (Core) 

4. Did you install pgBackRest from source or from a package? source

5. Please attach the following as applicable:
    - `pgbackrest.conf` file(s)
[demo]
pg1-path=/opt/data6000
pg1-port=6000
pg1-user=postgres

[global]
repo1-path=/home/postgres/pgbackrest

    - `postgresql.conf` settings applicable to pgBackRest (`archive_command`, `archive_mode`, `listen_addresses`, `max_wal_senders`, `wal_level`, `port`)
port=6000
archive_command='/usr/bin/pgbackrest --stanza=demo archive-push /opt/data6000/%p'
archive_mode = on
listen_addresses = '*'
wal_level = logical
max_wal_senders=8

    - errors in the postgresql log file before or during the time you experienced the issue
    - log file in `/var/log/pgbackrest` for the commands run (e.g. `/var/log/pgbackrest/mystanza_backup.log`)

$  pgbackrest --stanza=demo \
> --repo-path=/home/postgres/pgbackrest \
> --config=/home/postgres/pgbackrest/etc/pgbackrest.conf \
> --log-path=/home/postgres/pgbackrest/log \
> --log-level-file=info --log-subprocess \
> check 
ERROR: [082]: WAL segment 0000000200000004000000F7 was not archived before the 60000ms timeout
       HINT: check the archive_command to ensure that all options are correct (especially --stanza).
       HINT: check the PostgreSQL server log for errors.

$ tailf -n5 /opt/data6000/pg_log/pg_log_2020-04-14_153907.csv 
2020-04-14 15:51:08.510 CST,,,6550,,5e95689b.1996,16,,2020-04-14 15:39:07 CST,,0,LOG,00000,"archiver process (PID 6609) exited with exit code 1",,,,,,,,,""
[...]

PR #1010, opened 2020-04-24
Fix archive-push when pg1-path is not set and WAL path is absolute.

dwsteele:
This issue was reported in #999. Setting `pg1-path` worked for the user, but it raises the question of how to fix the underlying issue.

The `archive-push` command does not require `pg1-path` to be set if the WAL to the archive has an absolute path. However, at some point we added a check against pg_control that requires `pg1-path`. Since it's NULL in this case, segfault.

The main question is, do we want to continue to allow this behavior? The original idea was that we should be able to push a segment adhoc into the repo as long as the version/system-id matches and it does not conflict with existing WAL. This still seems like a useful capability even if it has not really come up.

The same functionality is definitely useful for archive-get. i.e. get a WAL segment and place it anywhere without needing to have pg1-path set or check against pg_control. If nothing else it is great for testing and analysis.

If we have it I think it should work for both. Does that make sense? 

I fixed the issue to restore the old behavior and added a test to prevent a future regression. Also remove an unused struct member, `pgWalSegmentSize`.

If we decide to keep this feature I'll go add tests for archive-get to make sure it is working as expected.

Another thing -- we've had more than a few segfaults related to options being NULL unexpectedly. I think we should make `cfgOptionStr()` throw a clear error in this case and add a new function, `cfgOptionStrNullable()` or something, to signal that we expect that the value can be NULL. So most calls would stay the same and we change to `Nullable` where needed.

### DIFF
--- a/test/src/module/command/archivePushTest.c
+++ b/test/src/module/command/archivePushTest.c
@@ -134,7 +134,7 @@ testRun(void)
                 "1={\"db-id\":5555555555555555555,\"db-version\":\"9.4\"}\n"));
 
         TEST_ERROR(
-            archivePushCheck(cipherTypeNone, NULL), ArchiveMismatchError,
+            archivePushCheck(true, cipherTypeNone, NULL), ArchiveMismatchError,
             "PostgreSQL version 9.6, system-id 18072658121562454734 do not match stanza version 9.4, system-id 5555555555555555555"
                 "\nHINT: are you archiving to the correct stanza?");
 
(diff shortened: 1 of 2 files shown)


test/src/module/command/archivePushTest.c
```diff
@@ -292,12 +291,17 @@ testRun(void)
...
 
-        // Save it to a new file instead
-        argListTemp = strLstDup(argList);
-        strLstAddZ(argListTemp, "pg_wal/000000010000000100000002");
+        // -------------------------------------------------------------------------------------------------------------------------
+        TEST_TITLE("WAL with absolute path and no pg1-path");
+
+        argListTemp = strLstNew();
+        strLstAddZ(argListTemp, "--" CFGOPT_STANZA "=test");
+        // strLstAdd(argListTemp, strNewFmt("--" CFGOPT_PG1_PATH "=%s/pg", testPath()));
```
cmwshang, line 299: I don't think you need to leave this commented out code here - the title of the test explains what is happening.

dwsteele: That's just there for my testing. This isn't quite a finished commit, just wanted to float the general idea.

dwsteele:
Fixed at 816ba9244f3. I ran into some problems performing a similar fix on archive-get, but it seems like a good idea to get this in at least.

PR #1011, opened 2020-04-24
Throw user-friendly error if expire is not run on repository host.

cmwshang:
Running the expire command on anything other than the repository host caused a segmentation fault because the command is not yet configured for remote execution. Also disallow execution to proceed if a stop file is present.


dwsteele:
Committed at a5b2630d7c0 and 483838233f.

This is no longer a segfault after 987df62ec014 so update the comments accordingly.

PR #1015, opened 2020-04-28
Fix issue checking if file links are contained in path links.

dwsteele:
There is no conflict if the path containing a file link is a parent path of a path link. The Perl code apparently had this right but the migration to C missed it. Exclude this case when checking for link violations.


dwsteele:
Committed at baf8cb90684fe. Thank you for the review!

PR #1018, opened 2020-04-29
Add Zstandard support.

dwsteele:


src/common/compress/zst/common.c
```diff
@@ -0,0 +1,32 @@
...
+        FUNCTION_TEST_PARAM(SIZE, error);
+    FUNCTION_TEST_END();
+
+    if (ZSTD_isError(error))
+        THROW_FMT(FormatError, "zst error: [%zd] %s", (ssize_t)error, ZSTD_getErrorName(error));
+
+    FUNCTION_TEST_RETURN(error);
+}
+
+#endif // HAVE_LIBLZ4
```
cmwshang, line 32: Shouldn't this comment be HAVE_LIBZST

src/common/compress/zst/compress.c
```diff
@@ -0,0 +1,207 @@
...
+    FUNCTION_LOG_END();
+
+    ASSERT(this != NULL);
+    ASSERT(!(this->flushing && !this->inputSame));
+    ASSERT(this->context != NULL);
+    ASSERT(compressed != NULL);
+    ASSERT(!this->flushing || uncompressed == NULL);
+
+    // Initialize output buffer
+    ZSTD_outBuffer out = {.dst = bufRemainsPtr(compressed), .size = bufRemains(compressed)};
```
cmwshang, line 87: I don't think we have a coding standard written for this, but I thought when we declare and initialize a structure that it would be on separate lines - like `ZSTD_inBuffer in` below (line 99 - 103)

src/common/compress/zst/compress.c
```diff
@@ -0,0 +1,207 @@
...
+
+    // If input is NULL then start flushing
+    if (uncompressed == NULL)
+    {
+        this->flushing = true;
+        this->inputSame = zstError(ZSTD_endStream(this->context, &out));
```
cmwshang, line 93: zstError should either return a boolean instead of size_t (seems only time the return value is assigned is to a boolean) or maybe have a conditional here like `zstError(ZSTD_endStream(this->context, &out)) != 0`

src/common/compress/zst/decompress.c
```diff
@@ -0,0 +1,191 @@
...
+        if (!this->frameDone)
+            THROW(FormatError, "unexpected eof in compressed data");
+
+        this->done = true;
+    }
+    else
+    {
+        // Initialize input/output buffer
+        ZSTD_inBuffer in = {.src = bufPtrConst(compressed) + this->inputOffset, .size = bufUsed(compressed) - this->inputOffset};
+        ZSTD_outBuffer out = {.dst = bufRemainsPtr(decompressed), .size = bufRemains(decompressed)};
```
cmwshang, line 97: Same comment about formatting structures as before. I don't really have a preference but we should probably make a decision on the standard (currently the coding standard only discusses structure declaration not initialization).

src/common/compress/zst/common.c
dwsteele: Yes, fixed.

src/common/compress/zst/compress.c
dwsteele: Usually I have been leaving them on one line if it fits, just like a function call. At least for new stuff. Since a lot of our designated initializers were converted from a list of initializations I did not necessarily reformat them as I went.

I'm not set on that, but if we decided to split them out we'll need to run through the code and make sure it is done everywhere.

src/common/compress/zst/compress.c
dwsteele: Passing the size through is a lot more efficient in other places so I want to keep zstError as is. But I did update the syntax as suggested.

src/common/compress/zst/decompress.c
dwsteele: We have plenty of cases where we only init one field and I'd certainly hate to split those out. I'd rather leave the initialization on one line if it fits.

src/common/compress/zst/compress.c
cmwshang: No problem. I'll make a note to update the coding standard that if it is not able to fit on one line then each structure member should be on a separate line. Sound ok?

dwsteele: Sounds good to me.

dwsteele:
Committed at 47aa765375.  Thanks for the review!

Also updated the Debian package and pushed new container builds.

PR #1025, opened 2020-05-05
Add bzip2 support

sfrost:
Because, why not?

Code is basically the gzip code with appropriate replacements done.  The interface is just about identical between gzip and bzip2, so that made it pretty straight-forward.

Tested manually but not integrated into the test suite (though seems like it shouldn't be hard?)
Looked at the testing code that the zstd patch added but didn't see immediately how to apply what it did with testing here.


dwsteele:
Why not, indeed?

The unit tests should be easy enough to add to `test/src/module/common/compressTest.c`. There's a standard suite for most of it so it's just a copy/paste operation. You'll also need to update `test/define.yaml` to increment the test count for the compress test.

Integration tests require updating `ContainerTest.pm` to add the packages for the RHEL and Debian containers. Then bzip can be added to the test matrices in `MockAllTest.pm`, etc.  `RealAllTest.pm` does not have a matrix and the logic for selecting compression is a little hokey so it's a little harder to add there.

You'll really want to run `test.pl` to do even basic testing. You've manually updated a number of files that are automatically generated and don't match what would be generated. If you need to actually *add* options that's just not going to work.

PR #1028, opened 2020-05-06
Update what kind of assertions are fired for bzip2

sfrost:


dwsteele:
This looks good to me, but let's leave this PR open in case Cynthia spots any other issues.

sfrost:
I'm pretty flexible, if we want to change the default and flip things around, that's fine by me too, though not sure it makes much difference.

dwsteele:
Probably a good idea to flip it since there are only two format errors. In that case the same should be done for gz because the assert errors slightly outnumber the format errors.

sfrost:
Updated gzip also per discussion and add a 'break;' for the default clause because that makes sense.

PR #1041, opened 2020-05-18
Add local MD5 implementation so S3 works when FIPS is enabled.

dwsteele:
S3 requires the Content-MD5 header for many requests but MD5 is not available via OpenSSL when FIPS is enabled because it is considered to be insecure.

Even though our usage does not present any security risks a local M5 implementation is required to circumvent the over-broad FIPS restriction.

Vendorize the MD5 implementation found at https://openwall.info/wiki/people/solar/software/public-domain-source-code/md5 and add full coverage for the new module in the common/crypto unit tests.


test/src/module/common/cryptoTest.c
```diff
@@ -329,9 +329,42 @@ testRun(void)
...
+
         TEST_ASSIGN(hash, cryptoHashNew(strNew(HASH_TYPE_MD5)), "create md5 hash");
         TEST_RESULT_STR_Z(varStr(ioFilterResult(hash)), HASH_TYPE_MD5_ZERO, "    check empty hash");
 
+        // Exercise most of the conditions in the local MD5 code
+        // -------------------------------------------------------------------------------------------------------------------------
+        TEST_TITLE("md5 hash - mixed bytes");
+
+        TEST_ASSIGN(hash, cryptoHashNew(strNew(HASH_TYPE_MD5)), "create md5 hash");
+        TEST_RESULT_VOID(ioFilterProcessIn(hash, BUFSTRZ("1")), "    add 1");
```
cmwshang, line 342: You have a TEST_TITLE for each test from line 332 to 367 so why are you adding spaces in the TEST_RESULT_* comments? These all "look" like new tests so we should treat the formatting as such.

src/common/crypto/hash.c
```diff
@@ -46,6 +51,7 @@ typedef struct CryptoHash
     MemContext *memContext;                                         // Context to store data
     const EVP_MD *hashType;                                         // Hash type (sha1, md5, etc.)
     EVP_MD_CTX *hashContext;                                        // Message hash context
+    MD5_CTX md5Context;                                             // MD5 context (used to bypass FIPS restrictions)
```
cmwshang, line 54: Why is md5Context not a pointer? Why allocating statically vs dynamically?

test/src/module/common/cryptoTest.c
cmwshang, line 342: From lines 332 to 366, we're using TEST_TITLE so why are some of the TEST_RESULT_* comment parameters including leading spaces? Since these appear to be new tests, we should treat the formatting of the output the way we now format tests.

src/common/crypto/hash.c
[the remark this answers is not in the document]
dwsteele: Easier and safer to statically allocate, but it does use up a chunk of memory when not being used for MD5. Changed to dynamically allocate.

This also required using a designated initializer in MD5_Init() since this function does not do a complete initialization, i.e. it expects the passed struct to be pre-zeroed.

test/src/module/common/cryptoTest.c
dwsteele: Fixed.

src/common/crypto/hash.c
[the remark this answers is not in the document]
cmwshang: OK. Can you please push the changes to this branch when you are done so I can do a final review? Thanks.

PR #1047, opened 2020-05-21
PostgreSQL 13 beta1 support.

dwsteele:
No major changes from PostgreSQL 12.

### DIFF
--- a/src/postgres/interface/static.vendor.h
+++ b/src/postgres/interface/static.vendor.h
@@ -210,7 +210,7 @@ typedef PageHeaderData *PageHeader;
 // ---------------------------------------------------------------------------------------------------------------------------------
 /*
  * PageIsNew
- *		returns true if page has not been initialized (by PageInit)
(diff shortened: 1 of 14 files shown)


src/postgres/interface/static.vendor.h
```diff
@@ -210,7 +210,7 @@ typedef PageHeaderData *PageHeader;
 // ---------------------------------------------------------------------------------------------------------------------------------
 /*
  * PageIsNew
- *		returns true if page has not been initialized (by PageInit)
+ *		returns true iff page has not been initialized (by PageInit)
```
cmwshang, line 213: Is this truly "If and only if" now? I really don't like the "iff" but I'm assuming this is how it is now written in PostgreSQL code and we have not control over it?

dwsteele: Yeah, it's always been like that in the PG code. I'm not sure how it got be be "if" in our code. I must have reflexively spell-checked it. I think I'll commit this one separately, though.

PR #1054, opened 2020-05-27
Improve handling of invalid HTTP response status.

dwsteele:
A truncated HTTP response status could lead to an an unfriendly error message, which would be retried, but could be confusing if the error was persistent and required debugging.

Improve the error handling overall to catch more error cases explicitly and respond better to edge cases.

Also update the terminology in comments to align with the RFC. Variable and function names were not changed because a refactor is intended for HTTP response and it doesn't seem worth the additional code churn.


dwsteele:
Interesting -- do you have a link for that?

cmwshang:
Just a couple of questions (https://stackoverflow.com/questions/19858251/http-status-code-0-error-domain-nsurlerrordomain <https://stackoverflow.com/questions/19858251/http-status-code-0-error-domain-nsurlerrordomain> and https://stackoverflow.com/questions/872206/what-does-it-mean-when-an-http-request-returns-status-code-0 <https://stackoverflow.com/questions/872206/what-does-it-mean-when-an-http-request-returns-status-code-0>) 

But here they explain https://stackoverflow.com/questions/3825581/does-an-http-status-code-of-0-have-any-meaning/26451773#26451773 <https://stackoverflow.com/questions/3825581/does-an-http-status-code-of-0-have-any-meaning/26451773#26451773>

So I know it is not "valid" but at least we will catch it if, in the very unlikely scenario, it happens.  

Cynthia Shang
www.crunchydata.com

dwsteele:
I see -- yeah from our perspective it's an invalid header and would fail the minimum size check. I agree we don't need to be anymore specific than that.

dwsteele:
A truncated HTTP response status could lead to an an unfriendly error message, which would be retried, but could be confusing if the error was persistent and required debugging.

Improve the error handling overall to catch more error cases explicitly and respond better to edge cases.

Also update the terminology in comments to align with the RFC. Variable and function names were not changed because a refactor is intended for HTTP response and it doesn't seem worth the additional code churn.

PR #1064, opened 2020-06-09
Refactor HTTP response code into a separate object.

dwsteele:
Moving this code out of `HttpClient` allows for more flexibility and gets us closer to async request/response.

Another benefit is that there is no need to explicitly close a response that was partially read. The `HttpResponse` destructor takes care of this now.


dwsteele:
As discussed, trying to review this feature (async S3 upload) in parts is going to take more time than it is worth. I'll submit a new PR when the entire feature is done.

PR #1067, opened 2020-06-10
Fix expression when recursion enabled in storageInfoListP().

dwsteele:
Filters only worked at the first level of recursion because the filter was also being applied to paths so the path had to match the filter in order to recurse.

This is not considered a bug since it does not affect any existing code paths, but it is required for the general-purpose repo-ls command.


dwsteele:
Cleaned up the test formatting, should be good to go.

PR #1076, opened 2020-06-16
Rename postmaster where possible.

dwsteele:
Rename postmaster in cases where it is really used to mean "PostgreSQL".  This should make error messages clearer since the user may not even be aware of what the "postmaster" is.


test/src/module/command/restoreTest.c
```diff
@@ -403,12 +403,12 @@ testRun(void)
         storagePathCreateP(storagePgWrite(), NULL);
 
         // -------------------------------------------------------------------------------------------------------------------------
-        TEST_TITLE("error on postmaster.pid is present");
+        TEST_TITLE("error on when pg appears to be running");
```
cmwshang, line 406: Silly typo here (remove "on") - should just be "error when pg appears to be running"

dwsteele: Haha, way to propagate typo. Fixed.

PR #1081, opened 2020-06-22
Asynchronous S3 multipart upload.

dwsteele:
When uploading large files the upload is split into multiple parts which are assembled at the end to create the final file. Previously we waited until each part was acknowledged before starting on the processing (i.e. compression, etc.) of the next part.

Now, the request for each part is sent while processing continues and the response is read just before sending the request for the next part. This asynchronous method allows us to continue processing while the S3 server formulates a response.

Testing from outside AWS in a high-bandwidth, low-latency environment showed a 35% improvement in the upload time of 1GB files. The time spent waiting for multipart notifications was reduced by ~300% (this measurement included the final part which is not uploaded asynchronously).

There are still some possible improvements: 1) the creation of the multipart id could be made asynchronous when it looks like the upload will need to be multipart (this may incur cost if the upload turns out not to be multipart). 2) allow more than one async request (this will use more memory).

A fair amount of refactoring was required to make the HTTP responses asynchronous. This may seem like overkill but having well-defined request, response, and session objects will also be advantageous for the upcoming HTTP server functionality.

Another advantage is that the lifecycle of an HttpSession is better defined. We only want to reuse sessions that complete the request/response cycle successfully. Otherwise we consider the session to be in a bad state and would prefer to start clean with a new one. Previously, this required complex notifications to mark a session as "successfully done". Now, ownership of the session is passed to the request and then the response and only returned to the client after a successful response. If an error occurs anywhere along the way the session will be automatically closed by the object destructor when the request/response object is freed (depending on which one currently owns the session).


src/common/io/http/request.c
```diff
@@ -0,0 +1,310 @@
...
+                        // looks valid. There are a few errors that might be permanently fatal but they are rare and it seems best
+                        // not to try and pick and choose errors in this class to retry.
+                        if (httpResponseCode(result) / 100 == HTTP_RESPONSE_CODE_RETRY_CLASS)
+                            THROW_FMT(ServiceError, "[%u] %s", httpResponseCode(result), strPtr(httpResponseReason(result)));
+
+                        // Move response to prior context
+                        httpResponseMove(result, memContextPrior());
+                    }
+                }
+                MEM_CONTEXT_END();
```
cmwshang: Why is this a mem_context_end? why not mem_context_temp_end?

test/src/module/common/ioHttpTest.c
```diff
@@ -406,28 +427,59 @@ testRun(void)
                 hrnTlsServerAccept();
 
                 hrnTlsServerExpectZ("GET / HTTP/1.1\r\n\r\n");
-                hrnTlsServerReplyZ("HTTP/1.1 404 Not Found\r\ncontent-length:0\r\n\r\n");
```
cmwshang, line 409: removing content-length:0 is changing the test

src/common/io/http/response.c
```diff
@@ -0,0 +1,412 @@
...
+
+    // If close was requested by the server then free the session
+    if (this->closeOnContentEof)
+    {
+        httpSessionFree(this->session);
+
+        // Only update the close stats after a successful response so it is not counted if there was an error/retry
+        httpClientStat.close++;
+    }
+    // Else return it to the client do it can be reused
```
cmwshang, line 83: do -> so   (Else return it to the client **so** it can be reused)

src/common/io/http/response.c
```diff
@@ -0,0 +1,412 @@
...
+        {
+            do
+            {
+                // If chunked content and no content remaining
+                if (this->contentChunked && this->contentRemaining == 0)
+                {
+                    // Read length of next chunk
+                    MEM_CONTEXT_TEMP_BEGIN()
+                    {
+                        this->contentRemaining = cvtZToUInt64Base(strPtr(strTrim(ioReadLine(rawRead))), 16);
```
cmwshang, line 135: What is this hardcoded 16 for?

src/common/io/http/response.c
```diff
@@ -0,0 +1,412 @@
...
+    ASSERT(buffer != NULL);
+    ASSERT(!bufFull(buffer));
+    ASSERT(this->contentEof || this->session != NULL);
+
+    // Read if EOF has not been reached
+    size_t actualBytes = 0;
+
+    if (!this->contentEof)
+    {
+        IoRead *rawRead = httpSessionIoRead(this->session);
```
cmwshang, line 116: So I guess everything created in this function is part of the memContext of the calling function (except line 135)?

src/common/io/http/response.c
```diff
@@ -0,0 +1,412 @@
...
+
+        // Read reason phrase. A missing reason phrase will be represented as an empty string.
+        MEM_CONTEXT_BEGIN(this->memContext)
+        {
+            this->reason = strSub(status, (size_t)spacePos + 1);
+        }
+        MEM_CONTEXT_END();
+
+        // Read headers
+        MEM_CONTEXT_BEGIN(this->memContext)
```
cmwshang, line 266: Is there a reason when the "this->reason = " and the "this->header = " cannot be combined under the same MEM_CONTEXT_BEGIN(this->memContext)

test/src/module/common/ioHttpTest.c
dwsteele: Added `content-length:0` to the test at line 324.

src/storage/s3/storage.c
```diff
@@ -234,77 +234,97 @@ storageS3Auth(
 /***********************************************************************************************************************************
 Process S3 request
 ***********************************************************************************************************************************/
-StorageS3RequestResult
-storageS3Request(
-    StorageS3 *this, const String *verb, const String *uri, const HttpQuery *query, const Buffer *body, bool returnContent,
-    bool allowMissing)
+HttpRequest *storageS3RequestAsync(StorageS3 *this, const String *verb, const String *uri, StorageS3RequestAsyncParam param)
```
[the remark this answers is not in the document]
dwsteele: Added.

src/command/repo/create.c
```diff
@@ -19,10 +19,7 @@ cmdRepoCreate(void)
     MEM_CONTEXT_TEMP_BEGIN()
     {
         if (strEq(storageType(storageRepo()), STORAGE_S3_TYPE_STR))
-        {
-            storageS3Request(
-                (StorageS3 *)storageDriver(storageRepoWrite()), HTTP_VERB_PUT_STR, FSLASH_STR, NULL, NULL, true, false);
-        }
+            storageS3RequestP((StorageS3 *)storageDriver(storageRepoWrite()), HTTP_VERB_PUT_STR, FSLASH_STR);
```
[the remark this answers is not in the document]
dwsteele: Basically, this got flipped. Previously, if the caller was not going to read the response in chunks (like StorageS3Read does) they had to set `.returnContent = true` to make the connection reusable, whether the content was important to them or not.

Now the content is automatically read into a buffer to make the connection resuable *unless* the caller sets `.contentIo = true` to indicate they will do the read in chunks themselves. Since it's rare for the caller to want to read in chunks (only one place in the S3 driver) this seems less prone to coding error.

src/storage/s3/write.c
```diff
@@ -160,7 +186,7 @@ storageWriteS3(THIS_VOID, const Buffer *buffer)
         // If the part buffer is full then write it
         if (bufRemains(this->partBuffer) == 0)
         {
-            storageWriteS3Part(this);
```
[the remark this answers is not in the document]
dwsteele: Looks like this was always missing. Added.

src/storage/s3/write.c
```diff
@@ -160,7 +186,7 @@ storageWriteS3(THIS_VOID, const Buffer *buffer)
...
+            storageWriteS3PartAsync(this);
```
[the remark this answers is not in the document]
dwsteele: "this" is the `StorageWriteS3` object so it is owned by whoever is doing the writing. So, for instance, if the `backupFile()` function were writing to S3 then the `StorageWriteS3` object would be owned by the context that starts at line 70 in `backupFile()`.

src/common/io/http/request.c
```diff
@@ -0,0 +1,310 @@
...
+            // Assume there will be no retry
+            retry = false;
+
+            TRY_BEGIN()
+            {
+                MEM_CONTEXT_TEMP_BEGIN()
+                {
+                    HttpSession *session = NULL;
+
+                    // If a session is saved then the request was already sent
```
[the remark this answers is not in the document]
dwsteele: That is correct. Updated.

src/common/io/http/request.c
```diff
@@ -0,0 +1,310 @@
...
+OBJECT_DEFINE_GET(Verb, const, HTTP_REQUEST, const String *, verb);
+OBJECT_DEFINE_GET(Uri, const, HTTP_REQUEST, const String *, uri);
+OBJECT_DEFINE_GET(Query, const, HTTP_REQUEST, const HttpQuery *, query);
+OBJECT_DEFINE_GET(Header, const, HTTP_REQUEST, const HttpHeader *, header);
+
+/***********************************************************************************************************************************
+Process the request
+***********************************************************************************************************************************/
+static HttpResponse *
+httpRequestProcess(HttpRequest *this, bool requestOnly, bool contentCache)
```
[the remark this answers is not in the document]
dwsteele: Agreed, but the motivation for combining them is that the retry logic is as much code as all the rest and I didn't want to make two copies of it. Plus, the request might need to be re-run if the response indicates a retry so they need to be in the same loop together. So, if there were two functions they would be almost identical except that one version would not need to read the reponse.

src/common/io/http/response.c
dwsteele: Fixed.

src/common/io/http/response.c
dwsteele: It means base 16, i.e. we are converting from a hex string to uint64. We could add a constant for BASE16, BASE10, BASE8, BASE2, etc. to `common/convert` if you like but I'd rather do that in a separate commit.

src/common/io/http/response.c
dwsteele: That's the only place that memory is allocated in this function, so yes. Of course, that could change.

When this code was originally written (it was moved from httpClient in this commit) mem context blocks were much more expensive because they contained a TRY...FINALLY. Now that they are cheaper there seems to be little harm in wrapping the entire block in a temp mem context. So done.

src/common/io/http/response.c
dwsteele: No, but actually the mem contexts were not quite right here anyway. This code was copied from httpClient where it lived inside a temp context, but here we are already in the `this->memContext` context. What we need is a temp context to wrap it. And creating the header list can going into the object's designated initializer.

So, did all that.

The way it is currently the temp strings created while parsing the response would have the same lifetime as the response object. Not a big deal, perhaps, but not as efficient as it could be.

PR #1086, opened 2020-06-25
Disable query parallelism in PostgreSQL sessions used for backup control.

dwsteele:
There is no need to have parallelism enabled in a backup control session. In particular, 9.6 marks pg_stop_backup() as parallel-safe but an error will be thrown if pg_stop_backup() is run in a worker.


dwsteele:
Excellent, thank you!

PR #1093, opened 2020-07-07
Add shared access signature (SAS) authorization for Azure.

dwsteele:
See https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview for details.


doc/xml/user-guide.xml
```diff
@@ -626,6 +627,7 @@
             <backrest-config-option section="global" key="repo1-type">azure</backrest-config-option>
             <backrest-config-option section="global" key="repo1-path">/{[azure-repo]}</backrest-config-option>
             <backrest-config-option section="global" key="repo1-azure-account">{[azure-account]}</backrest-config-option>
+            <backrest-config-option if="'{[azure-local]}' ne 'shared'" section="global" key="repo1-azure-key-type">{[azure-key-type]}</backrest-config-option>
```
cmwshang, line 630: Does this `if="'{[azure-local]}' ne 'shared'"` make sense - I thought `azure-local` was a boolean variable? Shouldn't it be `if="'{[azure-key-type]}' ne 'shared'"`?

test/src/module/storage/azureTest.c
```diff
@@ -216,6 +244,25 @@ testRun(void)
...
+            (StorageAzure *)storageDriver(
+                storageAzureNew(
+                    STRDEF("/repo"), false, NULL, TEST_CONTAINER_STR, TEST_ACCOUNT_STR, storageAzureKeyTypeSas, TEST_KEY_SAS_STR,
+                    16, NULL, 443, 1000, true, NULL, NULL)),
+            "new azure storage - sas key");
+
+        query = httpQueryAdd(httpQueryNew(), STRDEF("a"), STRDEF("b"));
+        header = httpHeaderAdd(httpHeaderNew(NULL), HTTP_HEADER_CONTENT_LENGTH_STR, STRDEF("66"));
+
+        TEST_RESULT_VOID(storageAzureAuth(storage, HTTP_VERB_GET_STR, STRDEF("/path/file"), query, dateTime, header), "auth");
```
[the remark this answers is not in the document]
dwsteele: > Is there no way to test this through storageAzureRequestAsync()?

We do -- all the tests after line 714 do SAS auth via storageAzureRequestAsync() which is verified by testRequest().

My experience has been that auth issues are tricky to debug so we have separate tests for them. If something happens to break auth we'll know it immediately without having to analyze failed requests in the tests.

doc/xml/user-guide.xml
dwsteele: Whoops! Fixed.

Issue #1090, opened 2020-07-03
[refactoring] repo-put / repo-get


pgstef:
Hi,

I was trying to improve the repo-get command encrypted part. 

Until now, here are the encryption levels I found:
* repo cipher -> archive.info & backup.info
* archive cipher, stored in archive.info, applied to WAL archives (and .history)
* backup cipher, stored in backup.info, applied to manifests (and backup.history manifests)
* manifest cipher, stored in backup.manifest, applied to backed-up files

Am I missing something there?

In order to imagine all the possible cases for the repo-get command (like guessing which cipher to use), I think it will be easier to first work on the repo-put command.

ATM, the repo-put command can store pretty much anything with a specific `--cipher-pass`. This cipher won't be stored anywhere. Is that something we wish to keep ?

* make stanza parameter mandatory
* copy file in the according repo archive -> apply archive.info stored cipher (used to push a wal archive eg.)
* copy file in the according repo backup dir -> apply backup.info stored cipher

I'm then considering using `--type` to know where to put files (also meaning that absolute path would only work if it match all the options).

Then, to put a file in a specific backup set, I would use the `--type=backup` + `--set` to apply the manifest specific cipher.

It would give a pretty restricted `repo-put` command, allowing to apply the good cipher.

Does that seems good for you or did you imagine something else ?

dwsteele:
> Until now, here are the encryption levels I found

This looks right to me.

> I think it will be easier to first work on the repo-put command.

Actually, I think `repo-get` is a bit simpler and more useful overall. Also, it's not clear that we would ever make `repo-put` public.

> repo-put command can store pretty much anything with a specific --cipher-pass

This is needed for testing with `repo-put` and `repo-get` but the option will remain hidden from users even if we make the command public.

> I'm then considering using --type ... I would use the --type=backup + --set

This seems much more complicated than necessary and will make the commands very hard to use. All the info needed to is in the path, e.g.
```
backup/test/20200628-084515F/pg_data/backup_label
```
is a backup file which needs to be decrypted with the key in `backup/test/20200628-084515F/backup.manifest` which needs to be decrypted with the key in `backup/test/backup.info` which needs to be decrypted with `repo1-cipher-pass`.

> Does that seems good for you or did you imagine something else ?

In summary, I think the `repo-get` command will be a bit simpler to implement and generally more useful (especially to check_pgbackrest). `repo-get` is also less dangerous, i.e. it can't be used to corrupt the repo.

[...]

PR #1095, opened 2020-07-08
Automatically determine cipher passphrase in repo-get command.

pgstef:
Based upon feedback on #1090.

Given the following repo encryption design:

```
REPO / (repoCipherPass)
           / archive   / (repoCipherPass)
           / archive   / stanza / (archiveCipherPass)
           / backup    / (repoCipherPass)
           / backup    / stanza / (backupCipherPass)
           / backup    / stanza / set / (manifestCipherPass)
           / backup    / stanza / backup.history / (manifestCipherPass)
```

Coverage tests seem to run fine:

```
2020-07-08 15:48:03.965 P00   INFO: P1-T1/1 - vm=u18, module=command, test=repo (22.7s)
2020-07-08 15:48:04.748 P00   INFO: tested modules have full coverage
2020-07-08 15:48:04.748 P00   INFO: writing C coverage report
2020-07-08 15:48:04.964 P00   INFO: TESTS COMPLETED SUCCESSFULLY (28s)
```

Many thanks in advance for the review :-)


src/command/repo/get.c
```diff
@@ -46,13 +49,81 @@ storageGetProcess(IoWrite *destination)
...
 
-                // If not passed as a parameter use the repo passphrase
+                // If not passed as a parameter, find the passphrase
                 if (cipherPass == NULL)
+                {
                     cipherPass = cfgOptionStr(cfgOptRepoCipherPass);
 
+                    const String *relativeFile = strDup(file);
+                    if (strBeginsWith(relativeFile, FSLASH_STR))
+                        relativeFile = strSub(file, strSize(cfgOptionStr(cfgOptRepoPath)) + 1);
```
dwsteele, line 73: I think we need to check that an absolute path really does being with cfgOptRepoPath. We might get some pretty odd behavior here if it does not.

src/command/repo/get.c
```diff
@@ -46,13 +49,81 @@ storageGetProcess(IoWrite *destination)
...
+                                   const Manifest *manifest = manifestLoadFile(
+                                        storageRepo(), strNewFmt(STORAGE_PATH_BACKUP "/%s/%s/%s", strPtr(stanza), 
+                                        strPtr(strLstGet(filePathSplitLst, 2)), BACKUP_MANIFEST_FILE), repoCipherType, cipherPass);
+                                    cipherPass = strDup(manifestCipherSubPass(manifest));
+                                }
+                            }
+                        }
+                    }
+                }
+
```
dwsteele, line 126: I think there needs to be some sort of error here. Currently no files can be above the archive/backup directories so it's not clear what key would be used if one did exist.

src/command/repo/get.c
```diff
@@ -46,13 +49,81 @@ storageGetProcess(IoWrite *destination)
...
+                    // Encrypted files could be stored directly at the root of the repo if needed
+                    // We should then at least be able to determine the archive or backup directory
+                    StringList *filePathSplitLst = strLstNewSplit(relativeFile, FSLASH_STR);
+                    if (strLstSize(filePathSplitLst) > 1)
+                    {
+                        const String *stanza = strLstGet(filePathSplitLst, 1);
+
+                        // If stanza option is specified, it must match the given file path
+                        if (cfgOptionStrNull(cfgOptStanza) != NULL && ! strEq(stanza, cfgOptionStr(cfgOptStanza)))
```
dwsteele, line 83: No need for a space after !. Lots of other places this needs to be updated.

src/command/repo/get.c
```diff
@@ -46,13 +49,81 @@ storageGetProcess(IoWrite *destination)
...
+                            THROW_FMT(AssertError, "stanza name '%s' given in option doesn't match the given path",
```
dwsteele, line 84: If more than one line follows an if then add braces to make it clearer: https://github.com/pgbackrest/pgbackrest/blob/master/CODING.md#braces.

Also there should be a LF after `THROW_FMT(`.

src/command/repo/get.c
```diff
@@ -46,13 +49,81 @@ storageGetProcess(IoWrite *destination)
...
+                            // Find the archiveCipherPass
+                            if (! strEndsWithZ(relativeFile, INFO_ARCHIVE_FILE) &&
+                                ! strEndsWithZ(relativeFile, INFO_ARCHIVE_FILE".copy"))
+                            {
+                                InfoArchive *info = infoArchiveLoadFile(
+                                    storageRepo(), strNewFmt(STORAGE_PATH_ARCHIVE "/%s/%s", strPtr(stanza), INFO_ARCHIVE_FILE),
+                                    repoCipherType, cipherPass);
+                                cipherPass = strDup(infoArchiveCipherPass(info));
+                            }
+
```
dwsteele, line 98: Remove extra LF here.

src/command/repo/get.c
```diff
@@ -46,13 +49,81 @@ storageGetProcess(IoWrite *destination)
...
+                                    storageRepo(), strNewFmt(STORAGE_PATH_BACKUP "/%s/%s", strPtr(stanza), INFO_BACKUP_FILE),
+                                    repoCipherType, cipherPass);
+                                cipherPass = strDup(infoBackupCipherPass(info));
+
+                                // Find the manifestCipherPass
+                                if (! strEq(strLstGet(filePathSplitLst, 2), STRDEF("backup.history")) &&
+                                    ! strEndsWithZ(relativeFile, BACKUP_MANIFEST_FILE) &&
+                                    ! strEndsWithZ(relativeFile, BACKUP_MANIFEST_FILE".copy"))
+                                {
+                                   const Manifest *manifest = manifestLoadFile(
```
dwsteele, line 117: Looks like the indentation here is one space short.

src/command/repo/get.c
```diff
@@ -46,13 +49,81 @@ storageGetProcess(IoWrite *destination)
...
+                        if (strEq(strLstGet(filePathSplitLst, 0), STORAGE_PATH_ARCHIVE_STR))
+                        {
+                            // Find the archiveCipherPass
+                            if (! strEndsWithZ(relativeFile, INFO_ARCHIVE_FILE) &&
+                                ! strEndsWithZ(relativeFile, INFO_ARCHIVE_FILE".copy"))
+                            {
+                                InfoArchive *info = infoArchiveLoadFile(
+                                    storageRepo(), strNewFmt(STORAGE_PATH_ARCHIVE "/%s/%s", strPtr(stanza), INFO_ARCHIVE_FILE),
+                                    repoCipherType, cipherPass);
+                                cipherPass = strDup(infoArchiveCipherPass(info));
```
dwsteele, line 96: I don't think this strDup() is needed. All this code is in one mem context so nothing should get freed.

src/command/repo/get.c
```diff
@@ -46,13 +49,81 @@ storageGetProcess(IoWrite *destination)
...
+                        if (strEq(strLstGet(filePathSplitLst, 0), STORAGE_PATH_BACKUP_STR))
+                        {
+                            if (! strEndsWithZ(relativeFile, INFO_BACKUP_FILE) &&
+                                ! strEndsWithZ(relativeFile, INFO_BACKUP_FILE".copy"))
+                            {
+                                // Find the backupCipherPass
+                                InfoBackup *info = infoBackupLoadFile(
+                                    storageRepo(), strNewFmt(STORAGE_PATH_BACKUP "/%s/%s", strPtr(stanza), INFO_BACKUP_FILE),
+                                    repoCipherType, cipherPass);
+                                cipherPass = strDup(infoBackupCipherPass(info));
```
dwsteele, line 110: Remove strDup().

src/command/repo/get.c
```diff
@@ -46,13 +49,81 @@ storageGetProcess(IoWrite *destination)
...
+
+                                // Find the manifestCipherPass
+                                if (! strEq(strLstGet(filePathSplitLst, 2), STRDEF("backup.history")) &&
+                                    ! strEndsWithZ(relativeFile, BACKUP_MANIFEST_FILE) &&
+                                    ! strEndsWithZ(relativeFile, BACKUP_MANIFEST_FILE".copy"))
+                                {
+                                   const Manifest *manifest = manifestLoadFile(
+                                        storageRepo(), strNewFmt(STORAGE_PATH_BACKUP "/%s/%s/%s", strPtr(stanza), 
+                                        strPtr(strLstGet(filePathSplitLst, 2)), BACKUP_MANIFEST_FILE), repoCipherType, cipherPass);
+                                    cipherPass = strDup(manifestCipherSubPass(manifest));
```
dwsteele, line 120: Remove strDup().

src/command/repo/get.c
pgstef: Isn't it the case with https://github.com/pgbackrest/pgbackrest/blob/master/src/storage/storage.c#L703 already ?

src/command/repo/get.c
pgstef: With the current code, `cfgOptRepoCipherPass` will be the default if no `cfgOptCipherPass` is given.
That match the repo-put possibilities also (and the previous tests written).

It would be easy to modify to:
```
if (archive)..
else if (backup)..
else AssertError
```

But would it really be useful since we can use the default RepoCipherPass if no specific CipherPass is provided?
In the future, if something is added, we can still modify this. I just assumed that since archive/backup.info are encrypted using the RepoCipherPass, if we add an extra level, it would use the same idea.

src/command/repo/get.c
pgstef: fixed, thanks

src/command/repo/get.c
pgstef: fixed, thanks

src/command/repo/get.c
pgstef: fixed, thanks

src/command/repo/get.c
pgstef: fixed, thanks

src/command/repo/get.c
pgstef: fixed, thanks for the advice !

src/command/repo/get.c
pgstef: fixed, thanks

src/command/repo/get.c
pgstef: fixed, thanks

src/command/repo/get.c
dwsteele: > Isn't it the case with https://github.com/pgbackrest/pgbackrest/blob/master/src/storage/storage.c#L703 already?

Yes, but that's an assert error which a user should never see. The assert in storagePath() is the last resort if user-provided data is not properly validated.

src/command/repo/get.c
dwsteele: It's hard to say what we might do in the future so I'd rather not try to plan ahead. What I'm worried about is if the user puts some file/dir at the top level of the repo and we try to decrypt it with the master key. That's not likely to work and there should not be any files there anyway. We should be throwing an error, something like `unable to determine encryption key for '/path'`. The error type can be FormatError.

src/command/repo/get.c
pgstef: Ok. Let's throw a FormatError also then ?

```
if (!strBeginsWith(relativeFile, cfgOptionStr(cfgOptRepoPath)))
{
    THROW_FMT(
        FormatError, "absolute path '%s' is not in base path '%s'", strPtr(relativeFile),
        strPtr(cfgOptionStr(cfgOptRepoPath)));
}
```

src/command/repo/get.c
pgstef: Ok.

Added the else if as mentioned above and:

```
else
{
    // Nothing should be stored at the top level of the repo
    THROW_FMT(FormatError, "unable to determine encryption key for '%s'", strPtr(relativeFile));
}
```

(I'll have to modify the tests once you think the get.c code is ready ^^)

src/command/repo/get.c
dwsteele: Actually, let's use `OptionInvalidValueError` for both these errors. Since this is invalid user input I think that's more appropriate.

src/command/repo/get.c
dwsteele: Looks good, just use `OptionInvalidValueError` instead of `FormatError`.

src/command/repo/get.c
```diff
@@ -46,13 +49,96 @@ storageGetProcess(IoWrite *destination)
...
             {
+                /*******************************************************************************************************************
+                Repository encryption design:
+                REPO / (repoCipherPass)
+                     / archive   / (repoCipherPass)
+                     / archive   / stanza / (archiveCipherPass)
+                     / backup    / (repoCipherPass)
+                     / backup    / stanza / (backupCipherPass)
+                     / backup    / stanza / set / (manifestCipherPass)
+                     / backup    / stanza / backup.history / (manifestCipherPass)
```
dwsteele, line 60: Isn't the backup.history dir encrypted with backupCipherPass?

src/command/repo/get.c
pgstef: Seems better indeed. I've updated the 3rd Error regarding stanza not found which is the same type of error.

src/command/repo/get.c
pgstef: Done

src/command/repo/get.c
pgstef: Indeed, wrong comment you're right (as always).
That's what is done in the code, apply backupCipherPass.
I'll update the comment and look at coverage tests.

Thanks !

dwsteele: Haha, well, I've been doing this for a while. I saw the code was correct so the comment was just a typo.

pgstef: Modified, thanks ;-)

src/command/repo/get.c
dwsteele: I think what we need is to move this validation code up to line 33 so it happens before storageNewRead(). So, something like (not tested!):
```
    if (strLstSize(cfgCommandParam()) == 1)
    {
        file = strLstGet(cfgCommandParam(), 0);

        if (strBeginsWith(relativeFile, FSLASH_STR))
        {
            if (!strBeginsWith(relativeFile, cfgOptionStr(cfgOptRepoPath)))
            {
                THROW_FMT(
                    OptionInvalidValueError, "absolute path '%s' is not in base path '%s'", strPtr(relativeFile),
                    strPtr(cfgOptionStr(cfgOptRepoPath)));
            }

            file = strSub(file, strSize(cfgOptionStr(cfgOptRepoPath)) + 1);
        }
    }
    else
        THROW(ParamRequiredError, "source file required");
```
Then we don't need the "relativeFile" var anymore. This validation needs to be done regardless of encryption so it makes sense to move it to the beginning.

dwsteele: So, yeah, we were getting an assert before, but repo-get has been an internal command used only for testing. If we want it to be public-facing we need to make it cleaner.

pgstef: Indeed, good catch.

I've modified the code accordingly and added extra verifications:
* nothing in repo root directory
* only archive/backup subdirectories allowed

Still the possibility to get unencrypted files and files encrypted with custom `--cipher-pass`.

The tests have been modified accordingly and coverage reports `INFO: tested modules have full coverage`.

pgstef: Hm, github mystery. I pushed a new commit appeared in the fork, not here :(

dwsteele: I also didn't get an email notification of your last comment. Maybe wait a bit and see if github catches up?

dwsteele: Doesn't matter either way. I can see the commits on your repo so I'll pull those down to do my final review.

src/command/repo/get.c
```diff
@@ -49,10 +68,78 @@ storageGetProcess(IoWrite *destination)
...
+                // -----------------------------------------------------------------------------------------------------------------
                 if (cipherPass == NULL)
+                {
                     cipherPass = cfgOptionStr(cfgOptRepoCipherPass);
 
+                    // Encrypted files could be stored directly at the root of the repo if needed
+                    // We should then at least be able to determine the archive or backup directory
+                    StringList *filePathSplitLst = strLstNewSplit(file, FSLASH_STR);
+
+                    if (strLstSize(filePathSplitLst) > 1)
```
dwsteele, line 89: I feel like this should be `strLstSize(filePathSplitLst) > 1`. The minimum path that makes sense is archive|backup/<stanza>/<file>, right?

src/command/repo/get.c
```diff
@@ -49,10 +68,78 @@ storageGetProcess(IoWrite *destination)
...
+                                    const Manifest *manifest = manifestLoadFile(
+                                        storageRepo(), strNewFmt(STORAGE_PATH_BACKUP "/%s/%s/%s", strPtr(stanza),
+                                        strPtr(strLstGet(filePathSplitLst, 2)), BACKUP_MANIFEST_FILE), repoCipherType, cipherPass);
+                                    cipherPass = manifestCipherSubPass(manifest);
+                                }
+                            }
+                        }
+                        // Only archive and backup directories are allowed
+                        else
+                            THROW_FMT(OptionInvalidValueError, "unable to determine encryption key for '%s'", strPtr(file));
```
dwsteele, line 136: I don't like that we have the same error message two times. There should be a way to combine them.  Perhaps set cipherPass = NULL if we can't find it and base the error message on that?

src/command/repo/get.c
pgstef: Isn't it already the case ? Or do you mean > 2 ?

After the repo-path is removed from the file path, indeed, 1st argument should be archive or backup. The next one should be the stanza name. If we look at what are inside the stanza directories, it should be `strLstSize(filePathSplitLst) > 2` then...

src/command/repo/get.c
pgstef: That seems reasonable indeed. I pushed a small modification to do so.

pgstef:
Thanks a lot for your review. I'll check and adjust the tests accordingly.

src/command/repo/get.c
dwsteele: Whoops! Yes, that's what I meant.

src/command/repo/get.c
dwsteele: Looks good. That's pretty much what I was thinking.

src/command/repo/get.c
pgstef: Ok, I'll update it, thanks

dwsteele: Excellent. This pretty much looks ready to me once this is done.

pgstef: I just have to figure out where the error comes from in the mock tests:

```
2020-07-16 09:13:07.016 P00  ERROR: [032]: /home/travis/test/bin/none/pgbackrest --config=/home/travis/test/test-0/db-primary/pgbackrest.conf --stanza=db --log-level-console=off --log-level-stderr=error --no-repo1-azure-verify-tls --repo1-azure-host=172.17.0.2  --ignore-missing repo-get /backup/db/backup.info terminated unexpectedly [032]: ERROR: [032]: unable to determine encryption key for 'ackup/db/backup.info'
```

It looks like the repo-path would only be '/'. I'll have a look at how the `file = strSub(file, strSize(cfgOptionStr(cfgOptRepoPath)) + 1);` reacts when repo-path ends with/without /

dwsteele: In other places we handle this as a special case and that is probably needed here as well.

dwsteele: Yeah, that looks right to me. I'll have a closer look later and commit if everything looks OK.

pgstef: I think there are still issues with MockStanzaTest. It tries to `repo-get /archive/db/archive.info.old`. 
We defined only `.info` and `.info.copy` are authorized to use the repoCipherPass. Otherwise, it will try to get the archiveCipherPass.
Should we allow files on the same directory level as `.info` to use the repoCipherPass ? Or should we adjust the tests ?

dwsteele: I actually have a PR in to get rid of these tests, but for now I think it would be OK to copy this file raw. Let me have a look at it and see where it would be best to adjust the integration test.

pgstef: Great, thanks!

dwsteele: Just pushed a commit up that will hopefully fix it. This was arguably a bug in the integration tests which was just exposed by the stricter rules in repo-get.

pgstef: Well, that was a good catch, it fixed the tests :-)

dwsteele:
OK, this looks read to commit to me. I made a few more comment updates and committed the integration test fix separately.

@pgstef Look good to you?

pgstef:
If it doesn't need more explanation in the docs, it looks good to me :-)

dwsteele:
Hmm, good point, but not sure I want to tackle that right now. The command remains internal for now so we can do a review of the docs when we make it public.

The general plan is to let you integrate repo-ls and repo-get into check_pgbackrest after the upcoming release (planned for Monday). If everything looks good we'll probably make them public in the August/September release.

Issue #1092, opened 2020-07-07
[Feature request] Some kind of "pre-expire" callback hook for the "backup" command


kmoppel:
Main idea - would be great if one could specify some script that is called just before the "auto-expire" of the regular "backup" command. Here in our use we'd want to make sure that all "to be expired" backups are copied off-site.

Even nicer of course would be built in support for such "remote archiving" (together with needed WAL) to S3 or normal FS mounts.

pgstef:
Hi,

I think the pre/post-backups hooks are something that will be added in the future (see https://github.com/pgbackrest/pgbackrest/projects/2#card-8957002).

In the meantime, should we add an extra option to disable "automatic" purge after the backup ? This would let users dry-run expire in external scripts before running the real purge themselves ?

kmoppel:
Ah nice, good to see it written down already :)
@pgstef sure, this "disable auto-purge" flag would make life a already easier. Currently one needs to archive off-site somewhere between the nightly runs but on failures / slowness there's still a chance that it's wiped before finished.

dwsteele:
In the meantime, you can run your backup with maxed out retention options to prevent expiration from happening. For example if you want to prevent full/diff backup expiration you could do something like:
```
pgbackrest --repo1-retention-full=9999999 --repo1-retention-diff=9999999 ...
```
Expire will still run but it won't have anything to do.

bradnicholson:
Is success of the post-backup expiration tied to success of the backup command?  I _think_ it is, and if the backup succeeds and the expiration fails, the backup fails.  We use the high expiration workaround and I seem to recall having backup failures in such cases (although I don''t recalls the exact specifics - the failure might have been from our caller not catching that case properly).

Also, the expire command performance degrades as the number of backups grows.  It's especially noticeable with s3 type storage.  While not a big deal, it is burning compute for no real reason when doing manual expirations.

pgstef:
The main logic is there https://github.com/pgbackrest/pgbackrest/blob/master/src/main.c#L116. 

I would have to try to confirm but I'm not sure the exit code will report an error:
```
$ pgbackrest backup  --stanza=my_stanza --type=full
...
P00   INFO: backup command end: completed successfully (8425ms)
P00   INFO: expire command begin 2.27
P00   INFO: expire full backup 20200709-160925F
P00   INFO: remove expired backup 20200709-160925F
[...]

PR #1098, opened 2020-07-09
Add expire-auto option to the backup command

pgstef:
The idea comes from #1092. 

In some cases, it might be interesting to not execute the expire command automatically. 
Indeed, we could set retention-options to maximum to avoid the purge but it seems a bit odd and would force to have different options for backup / expire commands. I'm afraid it could lead to some confusion there. That's why I tried to create this option.

Many thanks in advance for your ideas,
Kind regards


doc/xml/reference.xml
```diff
@@ -744,6 +744,15 @@
                         <example>5G</example>
                     </config-key>
 
+                    <!-- CONFIG - BACKUP SECTION - EXPIRE-AUTO -->
+                    <config-key id="expire-auto" name="Expire Auto">
+                        <summary>Automatically run the <cmd>expire</cmd> command after a successful backup.</summary>
+
+                        <text>The setting is enabled by default. Don't forget to manually run the <cmd>expire</cmd> command if you disable it.</text>
```
cmwshang, line 751: Instead of 
`Don't forget to manually run the <cmd>expire</cmd> command if you disable it.` 

I suggest: 
`Use caution when disabling this option; doing so will result in retaining all backups and archives indefinitely which could cause your repository to run out of space.`

dwsteele:
> I do not see any tests, however, for setting the option to "n" so I would request that one be added.

There are no unit tests for `main.c` so this would need to be an integration test. I'm not exactly sure what that would look like. I'm not too interested in adding it to the mock/all tests since we are hoping to get rid of those. It could probably be added somewhere in real/all, though. Thoughts?

cmwshang:
> There are no unit tests for main.c so this would need to be an integration test. I'm not exactly sure what that would look like. I'm not too interested in adding it to the mock/all tests since we are hoping to get rid of those. It could probably be added somewhere in real/all, though. Thoughts?

If we'd rather not update MockExpireTest.pm then in the RealAllTest.pm we could probably get a count of the number of backup labels in the repo and then ensure that the retention settings would expire one of the backups then run the backup with those options but also the expire-auto option set then confirm that the number of backup labels in the repo are the prior count +1.

doc/xml/reference.xml
pgstef: Hi,
Thanks for the review and the suggestion. It has been applied.

pgstef:
To be honest, I've not played with RealAllTest yet. But the idea to count backups seems good to me.
Currently, simple test builds are failing and I'm not sure why. I've updated the branch to the current master state without any change there.

pgstef:
Fixed the tests on my laptop with `git clean -f -d -X`

```
2020-07-10 10:23:35.012 P00   INFO: P1-T1/1 - vm=u12, module=real, test=all, run=1, pg-version=8.3 (98.62s)
2020-07-10 10:23:35.196 P00   INFO: TESTS COMPLETED SUCCESSFULLY (100s)
```

I don't know why the CI doesn't work though. Any idea ?

dwsteele:
Not sure what was going on with the CI builds. I merged new commits from master and pushed and now it seems OK. Your commit worked fine when I pushed it to my repo so I think something was weird with the PR. Github mysteries.

I cancelled the arm64 test because the queue for arm64 is really long right now. This happens from time to time and because it is still beta they don't say why. Lack of resources, mostly likely.

pgstef:
Great, thanks @dwsteele !
I'm trying to work on RealAllTest based on @cmwshang idea. I'm tinking to use `BackupInfo->list()` to get the backups list and compare the numbers.

cmwshang:
@pgstef looks good! I know running the FULL backups are a bit expensive (time-wise) in the tests but I think readability and considering the ReallAllTest is currently the only way to test this feature, I think it is appropriate. But I'll defer to so @dwsteele on that :-)

pgstef:
@cmwshang : indeed, running all the tests take some time :-( 
I discovered some missing parts in test/expect/real-all-001.log for co7 tests. 
I think directly add the `--no-expire-auto` rather than setting it globally will be easier and won't impact other tests. I've done the modification (not pushed yet). Just running the co7 tests again to confirm :-)

dwsteele:
> I think directly add the --no-expire-auto

Yes -- it's generally better to pass a setting like this directly to the command rather than set and unset it.

> I discovered some missing parts in test/expect/real-all-001.log for co7 tests.

The easiest way to update the co7 tests is just to use `--expect`, e.g.
```
test.pl --expect --module=real
```
That will run just the tests that generate the expect log and update the log as well. Then you just need to review before commit.

As for the tests, I think we should just go with what you have for now. We're working on identifying real/all tests that are already covered in unit tests and I think we should also look at running tests that are not version specific on just one version of PG -- probably the one we already run expect tests for.

> indeed, running all the tests take some time

It kind of sounds like you are running all the tests locally before pushing but we rarely do that. If you link your github account to Travis CI you can run CI yourself before pushing up to the PR. I usually just run a few tests to see if anything obviously breaks and then let CI work while I do something else.

pgstef:
> As for the tests, I think we should just go with what you have for now. We're working on identifying real/all tests that are already covered in unit tests and I think we should also look at running tests that are not version specific on just one version of PG -- probably the one we already run expect tests for.

The latest commit defines the `expire-auto` only for the backup command and not globally, as suggested.
If builds are OK, I think this PR could be ready for a final review ?

> It kind of sounds like you are running all the tests locally before pushing but we rarely do that. If you link your github account to Travis CI you can run CI yourself before pushing up to the PR. I usually just run a few tests to see if anything obviously breaks and then let CI work while I do something else.

Indeed, I run the tests locally with the test vm. I'll have to find out how travis-ci works then :-) (linked my github account, could relaunch the tests for this PR but I'll have to find how to run it before "push" or before submitting a PR then ^^)

Thanks for the advice !

pgstef:
The new comment looks indeed way better. +1 for me :-) 
Thank you both !

PR #1119, opened 2020-07-24
Suppress errors when closing local/remote processes.

dwsteele:
The main part of the fix is to ignore errors when closing local/remote processes that have finished their work. This is how it was supposed to be working but the code in `common/exit.c` was not quite right (it didn't include a `CATCH()` clause). Fix this logic and move it down to `protocol/helper.c` so it applies no matter where the connection is closed. A warning is still included to indicate that something is not right.

Also address the reason the error was happening. Local processes were not being shut down after parallel processing completed and the amount of time between them finishing and being shutdown was long enough that the local process has timed out and exited.


test/src/module/protocol/protocolTest.c
```diff
@@ -195,6 +195,27 @@ testRun(void)
...
+    // *****************************************************************************************************************************
+    if (testBegin("protocolHelperClientFree()"))
+    {
+        TEST_TITLE("free with errors output as warnings");
+
+        // Create and free a mem context to give us an error to use
+        MemContext *memContext = memContextNew("test");
+        memContextFree(memContext);
+
+        // Create bogus client and exec witht the freed memcontext to generate errors
```
cmwshang, line 207: Typo `witht`

dwsteele: Fixed.

PR #1120, opened 2020-07-24
Fix issue with = character in file or database names.

dwsteele:
The manifest uses the = character as the key/value separator so = characters in the key cause parsing errors and lead to an error or segfault.

Since the value must be valid JSON we can keep checking the value on the right side of the = and stop building the key when the value is valid. It's a bit hackish but it does seem to do the job without breaking the manifest format.

Unsurprisingly this makes parsing about 50% slower but it's still more than fast enough. Parsing 10 million key/values takes about 6.5s for the old code and 10s for the new code. Since the value is used as JSON downstream we can reclaim most of this time by just passing the JSON value rather than making the callback reparse it. We'll save that for another commit, though.


dwsteele:
> An alternative could have been just to put out an error message that prompts the user to a small tutorial on how to fix errors in the manifest.

Well, as you saw it's not such an easy thing to do (esp if on s3/azure storage). In 2.28 we have a `repo-get` command that will automatically figure out the encryption keys for you but it is not public yet. And there is no equivalent `repo-put` yet.

> Just a thought: Are equal-signs the only characters that pose a threat?

We believe so. The Perl code also had problems with UTF8 but we fixed that in the C migration. The = character is a bigger deal because it is part of the format.

dwsteele:
Config parse does not use this code -- it still uses the old Ini object that loads the ini all at one. If we decide to use this function for config we'll need to add a flag to tell it whether to expect JSON or not.

True that the other info files don't need this code but I'd rather it just ran all the time.

PR #1123, opened 2020-07-27
Improve memory usage of unlogged relation detection in manifest build.

dwsteele:
This loop was using a lot of memory without freeing it at intervals like we normally do. Rewrite to use char arrays when possible to reduce memory that needs to be allocated and freed.

The manifest data structures still use a fair amount of memory, ~200MB per million files.


dwsteele:
Not a lot of time, but enough that I didn't want to spend it. This code runs even if there are no unlogged tables so I want to keep it fast enough that we won't hear complaints about making it detect unlogged tables first.

Also, when I get into code with performance problems I just tend to optimize the heck out of it while I'm there.

PR #1125, opened 2020-07-28
Preserve Variant parsed from JSON in iniLoad().

dwsteele:
The fix for = characters in info files (039d3144) adds JSON validation but discards the resulting Variant which means the JSON is being parsed twice. This nearly doubles the time to load a manifest since a lot of complex JSON is involved.

Time to load a million file manifest:
Before 039d3144: 7.8s
039d3144: 15.5s
This patch: 7.5s

To fix this regression return the Variant in the callback so the caller does not have to parse it again. The new code appears slightly more efficient overall, perhaps because working with Variants is faster than working with JSON.


src/info/info.c
```diff
@@ -165,13 +165,14 @@ typedef struct InfoLoadData
 } InfoLoadData;
 
 static void
-infoLoadCallback(void *data, const String *section, const String *key, const String *value)
+infoLoadCallback(void *data, const String *section, const String *key, const String *value, const Variant *valueVar)
```
cmwshang, line 168: I don't understand why we need the `String *value` - it is only used in `INFO_CHECKSUM_KEY_VALUE(loadData->checksumActual, key, value);` so why not pass `varStr(valueVar)` instead and remove `String *value` altogether??

src/info/infoPg.c
```diff
@@ -121,13 +121,13 @@ infoPgLoadCallback(void *data, const String *section, const String *key, const S
     if (strEq(section, INFO_SECTION_DB_STR))
     {
         if (strEq(key, INFO_KEY_DB_ID_STR))
-            loadData->currentId = jsonToUInt(value);
+            loadData->currentId = varUIntForce(value);
     }
     // Process db:history section
     else if (strEq(section, INFO_SECTION_DB_HISTORY_STR))
     {
         // Load JSON loadData into a KeyValue
```
cmwshang, line 129: Comment should probably remove the word JSON since we're not loading the JSON to a kv anymore.

src/info/manifest.c
```diff
@@ -1649,11 +1656,11 @@ manifestLoadCallback(void *callbackData, const String *section, const String *ke
     else if (strEq(section, MANIFEST_SECTION_BACKUP_DB_STR))
     {
         if (strEq(key, MANIFEST_KEY_DB_ID_STR))
-            manifest->data.pgId = jsonToUInt(value);
+            manifest->data.pgId = varUIntForce(value);
         else if (strEq(key, MANIFEST_KEY_DB_SYSTEM_ID_STR))
-            manifest->data.pgSystemId = jsonToUInt64(value);
+            manifest->data.pgSystemId = varUInt64(value);
```
cmwshang, line 1661: Why does `manifest->data.pgId = varUIntForce(value);` (among others) require Force but `manifest->data.pgSystemId = varUInt64(value);` does not require `varUInt64Force(value); `?

src/info/info.c
dwsteele: varStr() won't work because it might not be a string -- also it needs to be a string in JSON format. So, if we don't return the string we need to re-encode it to JSON to calculate the checksum.

Also, if we decide to use this function for loading configs it will need to be able to return a normal string.

src/info/infoPg.c
dwsteele: Done.

src/info/manifest.c
dwsteele: Since numbers are not typed in JSON we always convert them to uint64. So if you want uint you need to force it.

PR #1130, opened 2020-08-04
Proactively close file descriptors after forking async process.

dwsteele:
PostgreSQL may be using most of the available file descriptors when it executes the the archive-get/archive-push commands (especially archive-get). This can lead to problems depending on how many file descriptors are needed for parallelism in the async process.

Proactively free file descriptors between 3 and 1023 to help ensure there are enough available for reasonable values of process-max, i.e. <= 300.

### DIFF
--- a/src/command/archive/common.c
+++ b/src/command/archive/common.c
@@ -249,6 +249,12 @@ archiveAsyncExec(ArchiveMode archiveMode, const StringList *commandExec)
         // Detach from parent process
         forkDetach();
 
+        // Close any open file descriptors above the standard three (stdin, stdout, stderr). Don't check the return value since we
+        // don't know which file descriptors are actually open (might be none). It's possible that there are open files >= 1024 but
+        // there is no easy way to detect that and this should give us enough descriptors to do our work.
+        for (int fd = 3; fd < 1024; fd++)
+            close(fd);
+
         // Execute the binary.  This statement will not return if it is successful.
         THROW_ON_SYS_ERROR_FMT(
             execvp(strZ(strLstGet(commandExec, 0)), (char ** const)strLstPtr(commandExec)) == -1, ExecuteError,
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -55,6 +55,15 @@
 
                         <p>Improve memory usage of unlogged relation detection in manifest build.</p>
                     </release-item>
+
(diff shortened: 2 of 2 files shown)


src/command/archive/common.c
```diff
@@ -249,6 +249,12 @@ archiveAsyncExec(ArchiveMode archiveMode, const StringList *commandExec)
         // Detach from parent process
         forkDetach();
 
+        // Close any open file descriptors above the standard three (stdin, stdout, stderr). Don't check the return value since we
+        // don't know which file descriptors are actually open (might be none). It's possible that there are open files >= 1024 but
+        // there is no easy way to detect that and this should give us enough descriptors to do our work.
+        for (int fd = 3; fd < 1024; fd++)
+            close(fd);
```
cmwshang, line 256: I'm a little confused. If we inherit file descriptors from Postgresql then are we in danger here of closing FDs actually in use? Also I don't see any tests for this. And:

"Not checking the return value when closing the file may lead to silent loss of data. This can especially be observed with NFS and with disk quota.

A successful close does not guarantee that the data has been successfully saved to disk, as the kernel defers writes. It is not common for a file system to flush the buffers when the stream is closed. If you need to be sure that the data is physically stored use fsync(2). (It will depend on the disk hardware at this point.) "

dwsteele: > If we inherit file descriptors from Postgresql then are we in danger here of closing FDs actually in use?

Yeah, we might close FDs in use but *we* are not using them, Postgres is. The FD will still be open in Postgres.

> Not checking the return value when closing the file may lead to silent loss of data.

Postgres is responsible for syncing and closing these files.

> Also I don't see any tests for this

We are testing it in `command/archive-get` and `command/archive-push` at least to the extent that it is not breaking anything. We don't have a good test to make sure that the code does not disappear -- it's not clear to me how we would do that since it is happening two forks down. Even if the code disappeared it wouldn't exactly represent a bug on our part. We are just trying to compensate for PostgreSQL not closing file handles before exec'ing pgbackrest.

sfrost: Note that, technically, there's two independent sets of file descriptors after the fork() happens, which is why it doesn't make any different to PG that we close *our* copy of the file descriptors that PG had open when we were fork()'d from PG.

PR #1131, opened 2020-08-04
Improve TLS error reporting.

dwsteele:
Before 9f2d647b TLS errors included additional details in at least some cases. After 9f2d647b a connection to an HTTP server threw `TLS error [1]` instead of `unable to negotiate TLS connection: [336031996] unknown protocol`.

Bring back the detailed messages to make debugging TLS errors easier. Since the error routine is now generic the `unable to negotiate TLS connection` context is still lost so the error looks like `TLS error [1:336031996] unknown protocol`.


dwsteele:
Had to change the type of the `errorTlsDetail` param to keep `ERR_reason_error_string()` happy on 32 bit without a bunch of casts.

PR #1088, opened 2020-06-29
Sample flow chart for IoReadInternal().

dwsteele:
[...]

PR #1132, opened 2020-08-05
Add IoClient and IoSession interfaces.

dwsteele:
These interfaces allow the HttpClient and HttpSession objects to work with protocols other than TLS, .e.g. plain sockets. This is necessary to allow standard HTTP -- right now we only allow HTTPS, i.e. HTTP over TLS.

For now only TlsClient and TlsSession have been converted to the new interfaces. SocketClient and SocketSession will also need to be converted but first sckSessionReadyRead() and sckSessionReadyWrite() need to be moved into the IoRead and IoWrite interfaces, since they are not a good fit for IoSession.


src/common/io/tls/client.c
```diff
@@ -351,7 +307,92 @@ tlsClientOpen(TlsClient *this)
...
+        FUNCTION_LOG_PARAM(SOCKET_CLIENT, socket);
+        FUNCTION_LOG_PARAM(TIME_MSEC, timeout);
+        FUNCTION_LOG_PARAM(BOOL, verifyPeer);
+        FUNCTION_LOG_PARAM(STRING, caFile);
+        FUNCTION_LOG_PARAM(STRING, caPath);
+    FUNCTION_LOG_END();
+
+    ASSERT(socket != NULL);
+
+    IoClient *this = NULL;
```
sfrost, line 334: wouldn't you normally declare this later, since you don't use it until later..?  Seems like it's here just because it was being used earlier before.

dwsteele: It needs to be here because it must be outside the MEM_CONTEXT_NEW block and this is last chance to do that.

dwsteele:
> Some kind of high-level discussion of ioSession's and httpClient and tlsSession, etc.

Cynthia and I are working on that -- slowly to be sure, but it is on our radar. PR #1088 is the very beginning of that effort.

src/common/io/tls/session.c
```diff
@@ -272,7 +293,44 @@ tlsSessionEof(THIS_VOID)
 }
 
 /**********************************************************************************************************************************/
-TlsSession *
+static IoRead *tlsSessionIoRead(THIS_VOID)
```
cmwshang, line 296: Carriage return needed before tlsSessionIoRead

src/common/io/tls/session.c
```diff
@@ -272,7 +293,44 @@ tlsSessionEof(THIS_VOID)
...
+        FUNCTION_TEST_PARAM(TLS_SESSION, this);
+    FUNCTION_TEST_END();
+
+    ASSERT(this != NULL);
+
+    FUNCTION_TEST_RETURN(this->read);
+}
+
+/**********************************************************************************************************************************/
+static IoWrite *tlsSessionIoWrite(THIS_VOID)
```
cmwshang, line 310: Carriage return needed before tlsSessionIoWrite

src/common/io/tls/session.c
```diff
@@ -60,7 +80,7 @@ tlsSessionClose(TlsSession *this, bool shutdown)
     if (this->session != NULL)
     {
         // Shutdown on request
-        if (shutdown)
+        if (this->shutdownOnClose)
             SSL_shutdown(this->session);
 
         // Free the socket session
```
cmwshang, line 86: My comment is for lines 91 and 92 - here the calling order is:
```
memContextCallbackClear(this->memContext);
tlsSessionFreeResource(this);
```
but in storageReadPosixClose it is called in the opposite order:
```
storageReadPosixFreeResource(this);
memContextCallbackClear(this->memContext);
```
Are they both correct? If so, why does order not matter?

src/common/io/session.c
```diff
@@ -0,0 +1,98 @@
...
+
+    ASSERT(this != NULL);
+
+    this->interface->close(this->driver);
+
+    FUNCTION_LOG_RETURN_VOID();
+}
+
+/**********************************************************************************************************************************/
+IoRead *ioSessionIoRead(IoSession *this)
```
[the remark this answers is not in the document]
dwsteele: Fixed.

src/common/io/session.c
```diff
@@ -0,0 +1,98 @@
...
+        FUNCTION_TEST_PARAM(IO_SESSION, this);
+    FUNCTION_TEST_END();
+
+    ASSERT(this != NULL);
+
+    FUNCTION_TEST_RETURN(this->interface->ioRead(this->driver));
+}
+
+/**********************************************************************************************************************************/
+IoWrite *ioSessionIoWrite(IoSession *this)
```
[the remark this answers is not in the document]
dwsteele: Fixed.

src/common/io/session.c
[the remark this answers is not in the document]
dwsteele: Fixed.

src/common/io/tls/session.c
dwsteele: Fixed.

src/common/io/tls/session.c
dwsteele: Hmm, I'd say it is correct here and wrong in storageReadPosixClose(). Clearing the callback is so the callback won't be called if there is an error in storageReadPosixFreeResource(), since this is the same function will be called on error.

Corrected in storageReadPosixClose() here 0b349c0f2f54d1.

PR #1133, opened 2020-08-06
Move file descriptor read/write ready into IoRead/IoWrite.

dwsteele:
Move sckSessionReadyRead()/Write() into the IoRead/IoWrite interfaces. This is a more logical place for them and the alternative would be to add them to the IoSession interface, which does not seem like a good idea.

This is mostly a refactor, but a big change is the select() logic in fdRead.c has been replaced by ioReadReady().  This was duplicated code that was being used by our protocol but not TLS. Since we have not had any problems with requiring poll() in the field this seems like a good time to remove our dependence on select().

Also, IoFdWrite now requires a timeout so update where required, mostly in the tests.


dwsteele:
> but I certainly appreciate that's a ... rather larger change.

It's not huge, but definitely seems better to tackle at another time. Since I didn't modify any of the poll() code (just moved it) I didn't want to get into it.

Issue #900, opened 2019-12-30
optional wal archiving


sean0101n:
I'm suggesting a new feature (I don't believe there's currently an option for this, or a work-around) that can toggle on/off wal archiving for backups.

Our use case simply requires a consistent state to be reached during a restore, which would be met when simply restoring from a specific backup set.  WAL generated between/after finished backups is not necessarily needed, and simply adds on to s3 costs as we're continuously archiving to s3.  I'd assume it would also cut down on cpu load on the primary during high traffic.

This might be an unusual use-case, but I thought I'd throw this suggestion out there.  Let me know your thoughts.

Also, thanks for all the hard work on this project!

-Sean

hlakshmi:
👍 for the feature as we are also having a similar use case.

dwsteele:
We have thought about this before but it is a bit more complicated that it looks.  The archive-get and backup commands don't currently communicate but for this feature they would need to do so somehow.  There are a number of ways we could go about it but it would require some work.

In general giving up PITR capability is a pretty big negative so while we are not against the feature it is not currently on our roadmap.  We would be willing to look at a contribution, though!

sean0101n:
@dwsteele Maybe backup and archive-push commands don't need to communicate to each other?  Could the archive-push command execute a quick pg_is_in_backup(), and if it returns true, push WAL to archive?  I'm most likely over-simplifying it here, but just an idea.

dwsteele:
`pg_is_in_backup()` does not work with non-exclusive backup, which we use for PostgreSQL >= 9.6.

sean0101n:
Just had another idea.. could the archive-push check for an existing backup lock file, and decide to archive if it exists?

dwsteele:
Yeah, that could work and it is fairly cheap.

sean0101n:
I simply tried out adding a test to the archive_command like this:
`archive_command = 'test -f /tmp/pgbackrest/db2-backup.lock || exit 0 && pgbackrest --stanza=db2 archive-push %p'`
This seems to work fine, and the backup restored as expected.  Maybe it doesn't need to be implemented in the code.  A side note in the documentation or something could let people know that adding a check for the lock file in the archive command will avoid accumulating potentially unnecessary WAL, but will forfeit PITR functionality.

dwsteele:
That is very clever!

I'm not sure we want to endorse it in the documentation (yet) but it is a perfect proof of concept.

[...]

PR #1134, opened 2020-08-07
Delay backup remote connection close until after archive check.

fvannee:
Only close the connection to the remote after verifying that the WAL files are received.
This is necessary if the archive_command on pg servers is conditional. Then we need to
keep the backup lock until we know all WAL files have been sent.

Without this, the work-around as given in https://github.com/pgbackrest/pgbackrest/issues/900 sometimes fails, because the backup lock file is released before the last WAL segment is sent to the backup repo.

I know this isn't a full solution to proper conditional archiving, which would require a lot more changes. However, at least this makes the solution work. Would this be ok to merge?

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -67,6 +67,15 @@
 
                     <release-item>
                         <release-item-contributor-list>
+                            <release-item-contributor id="floris.van.nee"/>
+                            <release-item-reviewer id="david.steele"/>
+                        </release-item-contributor-list>
+
+                        <p>Delay backup remote connection close until after archive check.</p>
+                    </release-item>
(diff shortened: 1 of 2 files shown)


fvannee:
Thanks for the changes @dwsteele . I've put them in this branch. The changes make sense to me.

Strangely enough the 'Allow edits by maintainers' checkbox has been ticked from the beginning of the PR. I'm not sure what's still blocking access to the forked branch for you.

dwsteele:
> I'm not sure what's still blocking access to the forked branch for you.

Hmmm, weird, might be on my side somehow. I'm not going to worry about it for now, though.

PR #1135, opened 2020-08-10
Implement IoClient/IoSession interfaces for SocketClient/SocketSession.

dwsteele:
Following up on 111d33c1 implement the new interfaces for socket client/session. Now HTTP objects can be used over TLS or plain sockets.

This required adding ioSessionFd() and ioSessionRole() to provide the functionality of sckSessionFd() and sckSessionType(). sckClientHost() and sckClientPort don't make sense in a generic interface so the were replaced with ioSessionName().


src/common/io/socket/session.c
```diff
@@ -84,19 +190,18 @@ sckSessionNew(SocketSessionType type, int fd, const String *host, unsigned int p
...
         };
 
-        memContextCallbackSet(this->memContext, sckSessionFreeResource, this);
         strFree(name);
+
+        // Open read/write io
+        ioReadOpen(driver->read);
+        ioWriteOpen(driver->write);
+
+        // Ensure file descriptor is close
```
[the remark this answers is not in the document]
dwsteele: Fixed.

src/common/io/session.c
```diff
@@ -92,9 +106,25 @@ ioSessionIoWrite(IoSession *this)
...
+
 /**********************************************************************************************************************************/
 String *
 ioSessionToLog(const IoSession *this)
 {
-    return strNewFmt("{type: %s, driver: %s}", strZ(*this->interface->type), strZ(this->interface->toLog(this->driver)));
+    return strNewFmt(
+        "{type: %s, role: %s, driver: %s}", strZ(*this->interface->type),
+        this->interface->role(this->driver) == ioSessionRoleClient ? "client" : "server",
+        strZ(this->interface->toLog(this->driver)));
```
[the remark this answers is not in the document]
dwsteele: This will be shown in the driver if it is valid, but sure. More info is generally better.

src/common/io/session.c
```diff
@@ -92,9 +106,25 @@ ioSessionIoWrite(IoSession *this)
...
+}
+
 /**********************************************************************************************************************************/
 String *
 ioSessionToLog(const IoSession *this)
 {
-    return strNewFmt("{type: %s, driver: %s}", strZ(*this->interface->type), strZ(this->interface->toLog(this->driver)));
+    return strNewFmt(
+        "{type: %s, role: %s, driver: %s}", strZ(*this->interface->type),
+        this->interface->role(this->driver) == ioSessionRoleClient ? "client" : "server",
```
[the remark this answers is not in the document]
dwsteele: Just didn't think of it. Updated.

src/common/io/socket/client.c
```diff
@@ -25,63 +31,47 @@ static SocketClientStat sckClientStatLocal;
...
     unsigned int port;                                              // Port to connect to host on
+    String *name;                                                   // Socket name (host:port)
     TimeMSec timeout;                                               // Timeout for any i/o operation (connect, read, etc.)
-};
-
-OBJECT_DEFINE_MOVE(SOCKET_CLIENT);
+} SocketClient;
 
-OBJECT_DEFINE_GET(Host, const, SOCKET_CLIENT, const String *, host);
-OBJECT_DEFINE_GET(Port, const, SOCKET_CLIENT, unsigned int, port);
```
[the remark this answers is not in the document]
dwsteele: They won't be accessible anymore since sckClientNew() returns IoSession and it doesn't make sense for IoSession to expose something so socket-specific. They were just used for logging and were replaced with ioClientName().

src/common/io/session.c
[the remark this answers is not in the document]
dwsteele: Actually, I think it's better just to leave this logged in the `driver: {...}` part. ioSessionFd() requires a non-const pointer so we need to either duplicate the logic or `UNCONSTIFY()` it. Neither is very attractive since the info is already there.

src/common/io/tls/client.c
```diff
@@ -288,48 +287,64 @@ tlsClientOpen(THIS_VOID)
         if (verifyResult != X509_V_OK)                                                                              // {vm_covered}
         {
             THROW_FMT(                                                                                              // {vm_covered}
-                CryptoError, "unable to verify certificate presented by '%s:%u': [%ld] %s",                         // {vm_covered}
-                strZ(sckClientHost(this->socketClient)), sckClientPort(this->socketClient), verifyResult,           // {vm_covered}
-                X509_verify_cert_error_string(verifyResult));                                                       // {vm_covered}
+                CryptoError, "unable to verify certificate presented by '%s': [%ld] %s",                            // {vm_covered}
+                strZ(ioClientName(this->ioClient)), verifyResult, X509_verify_cert_error_string(verifyResult));     // {vm_covered}
```
sfrost, line 291: This went from sckClientHost to ioClientName...  but those are two different things, aren't they?  hostname vs. "name" (which seems to include the port...?)

src/common/io/tls/client.c
```diff
@@ -288,48 +287,64 @@ tlsClientOpen(THIS_VOID)
...
         }
 
         // Verify that the hostname appears in the certificate
         X509 *certificate = SSL_get_peer_certificate(session);                                                      // {vm_covered}
-        bool nameResult = tlsClientHostVerify(sckClientHost(this->socketClient), certificate);                      // {vm_covered}
+        bool nameResult = tlsClientHostVerify(this->host, certificate);                                             // {vm_covered}
```
sfrost, line 296: This seems like what above should have been doing...?  sckClientHost -> this->host...

src/common/io/tls/client.c
```diff
@@ -288,48 +287,64 @@ tlsClientOpen(THIS_VOID)
...
+{
+    THIS(TlsClient);
+
+    FUNCTION_TEST_BEGIN();
+        FUNCTION_TEST_PARAM(TLS_CLIENT, this);
+    FUNCTION_TEST_END();
+
+    ASSERT(this != NULL);
+
+    FUNCTION_TEST_RETURN(ioClientName(this->ioClient));
```
sfrost, line 323: Not sure I entirely get what the purpose of the "name" construct is...   If it's for identifying information, I would think you'd want to add a "TLS: " or something to it instead of just passing it through as-is from the underlying layer...

src/common/io/tls/client.c
```diff
@@ -340,7 +355,8 @@ tlsClientNew(SocketClient *socket, TimeMSec timeout, bool verifyPeer, const Stri
         *driver = (TlsClient)
         {
             .memContext = MEM_CONTEXT_NEW(),
-            .socketClient = sckClientMove(socket, MEM_CONTEXT_NEW()),
+            .ioClient = ioClientMove(ioClient, MEM_CONTEXT_NEW()),
+            .host = strDup(host),
```
sfrost, line 359: This seems a bit grotty, do we actually need two copies of it...?  What if one is changed?  Which is canonical..?

src/common/io/tls/client.c
dwsteele: It's because we don't know what ioClient() implements -- does it even have a hostname? The output of this error is the same currently because SocketClient combines the home and port as it was before. But in the future this might not be the case.

src/common/io/tls/client.c
dwsteele: this->host is new because SocketClient does not expose it anymore.

src/common/io/tls/client.c
dwsteele: We already have the type that gets output in logging as `tls` so no need to repeat it. The name is just some identifier for the ioClient. When the ioClient is a socket this will be `host:port`.

src/common/io/tls/client.c
dwsteele: Well, this is the host we are going to use to verify TLS -- there's no reason that has to be the same host we pass to SocketClient, though that's the case now.

src/common/io/tls/client.c
sfrost: Yeah, makes sense.

src/common/io/tls/client.c
sfrost: ok, makes sense.

src/common/io/tls/client.c
sfrost: We should figure out a way to make this cleaner because we need to make absolutely sure that the host we're using for verification is the exact value of what was configured in the pgbackrest.conf.

PR #1136, opened 2020-08-10
WIP: Add archives gap detection in the info command

pgstef:
NOTE: THIS PR IS A WORK IN PROGRESS AND NOT READY FOR DETAILED REVIEW.

With a new `archive-gap-detection` in the `info` command, it is possible to create a list of expected archives:
- start/stop wal for each backup;
- min/max wal found in the repo;
- exclude archives expired by the archive-retention policy.

The function that generate this expected list will parse .history file for the timeline of the "stop" wal specified. Based on that, we're able to follow all the timeline switches and check that particular timeline.

While it's not (yet) easy to get the wal-segment-size directly from the repo, I used `PG_WAL_SEGMENT_SIZE_DEFAULT` until we can put that to archive.info for example. 

Then, it's pretty easy (even if it comes with performance penalty for the info command) to list the archives in the repo and compare with the expected list.

The `archive-gap-detection` option is disabled by default of course.

I tried to run as much test cases as I could imagine and this POC seems to work pretty correctly.

Before putting efforts to the tests part, quick review and advice would be very much appreciated :-)


src/command/info/info.c
```diff
@@ -350,6 +358,271 @@ backupList(VariantList *backupSection, InfoBackup *info, const String *backupLab
...
+
+    // Get .history file if timelines don't match
+    uint32_t startTimeline = (uint32_t)strtol(strZ(strSubN(archiveStart, 0, 8)), NULL, 16);
+    uint32_t stopTimeline = (uint32_t)strtol(strZ(strSubN(archiveStop, 0, 8)), NULL, 16);
+
+    // Store history content: WAL segments and target timeline
+    StringList *timelineSwitchWal = strLstNew();
+    KeyValue *targetTimelines = kvNew();
+
+    if(stopTimeline > startTimeline)
```
cmwshang, line 391: If I'm understanding correctly, the problem that is trying to be solved is determining if all the WAL exists in the repo to get from a backup through the latest timeline. If that is the case, passing the archive wal start/stop for a specific backup to this function will never satisfy the stopTimeline > startTimeline condition since a single backup will never cross timelines - a set will, but not a backup within a set. 

As an example, if you have docker and can run the pgbackrest docs though the pitr section:
`/<YOUR DIR>/doc/doc.pl --out=html --include=user-guide --require=/pitr --var=encrypt=n --no-cache`
then 
`docker exec -it doc-pg-primary bash`
you can see that the `info` command might produce:

```
   db (current)
        wal archive min/max (12-1): 000000010000000000000001/000000030000000000000007

        full backup: 20200810-171426F
            timestamp start/stop: 2020-08-10 17:14:26 / 2020-08-10 17:14:40
            wal start/stop: 000000010000000000000002 / 000000010000000000000002
            database size: 23.4MB, backup size: 23.4MB
            repository size: 2.8MB, repository backup size: 2.8MB

        diff backup: 20200810-171426F_20200810-171442D
            timestamp start/stop: 2020-08-10 17:14:42 / 2020-08-10 17:14:44
            wal start/stop: 000000010000000000000003 / 000000010000000000000003
            database size: 23.4MB, backup size: 8.3KB
            repository size: 2.8MB, repository backup size: 426B
            backup reference list: 20200810-171426F

        diff backup: 20200810-171426F_20200810-171445D
            timestamp start/stop: 2020-08-10 17:14:45 / 2020-08-10 17:14:47
            wal start/stop: 000000010000000000000004 / 000000010000000000000004
            database size: 23.4MB, backup size: 8.3KB
            repository size: 2.8MB, repository backup size: 425B
            backup reference list: 20200810-171426F

        incr backup: 20200810-171426F_20200810-171459I
            timestamp start/stop: 2020-08-10 17:14:59 / 2020-08-10 17:15:03
            wal start/stop: 000000020000000000000006 / 000000020000000000000006
            database size: 23.4MB, backup size: 2.2MB
            repository size: 2.8MB, repository backup size: 232.5KB
            backup reference list: 20200810-171426F, 20200810-171426F_20200810-171445D
```

(you can use --out=json to get the kv structured output). So a single backup archive `wal start/stop` will never cross a timeline. But 20200810-171426F_20200810-171459I relies on prior backups (backup reference list) so the set does cross timelines - but that is not the information that is being passed to the `generateNeededArchivesList()` function.

It's a little hard to see what is the expected outcome of the patch without tests so my apologies in advance if I'm misunderstanding your intent.

pgstef: Well, indeed, for the backups it's more unlikely that there will be a timeline switch. 
But this function is also used (a bit further, line 586) to check archives consistency for (in your example) `wal archive min/max (12-1): 000000010000000000000001/000000030000000000000007` where a timeline switch occurred :-)

A good test is to retention-archive-type to diff, so archives between the 1st full and 1st diff will be expired. That's why we need to check globally between min (and change the min according the archive retention policy) and max wal but also between each backup start/stop location.

I'll look to add some tests if it's easier to review then ;-)

cmwshang: Yes, I think providing tests would be helpful. Also, take care on your historyExp as it will not only match the `2	0/7000000` in 
`2	0/7000000	before 2000-01-01 00:00:00+00`
but also 
```
2 01111/7000000
0111/012345766985640856045968409
```
etc

dwsteele: I don't think we need full test coverage here, yet, but a few tests would definitely help us review the code.

src/command/info/info.c
```diff
@@ -350,6 +358,271 @@ backupList(VariantList *backupSection, InfoBackup *info, const String *backupLab
...
+            LOG_WARN_FMT(
+                "min wal archive ('%s') found in the repo is older than the first backup start WAL location ('%s').\n"
+                "HINT: this might indicate a configuration error in the retention policy.",
+                strZ(archiveStart), strZ(firstStartWal));
+
+        // If repo-retention-archive is set, exclude potentially expired archives
+        if (cfgOptionTest(cfgOptRepoRetentionArchive))
+        {
+            archiveStart = strLstGet(backupStartWalsList,
+                strLstSize(backupStartWalsList) - cfgOptionUInt(cfgOptRepoRetentionArchive));
```
cmwshang, line 582: I'm not clear on the backupStartWalList - if we take the example I gave above, then by the time we get here, backupStartWalsList would contain 000000010000000000000002 (FULL start), 000000010000000000000003 and 000000010000000000000004 (start from the 2 DIFFs). If archive-retention-type=DIFF and archive-retention=4 then this code would error (3-4 = -1). Even if archive-retention were 2 instead, the archiveStart would be backupStartWalsList[1] = 000000010000000000000003 which would only mean the neededArchiveList would start at 000000010000000000000002 instead of 000000010000000000000001.

Also, this does not take into account time-based backup retention or adhoc backup expiration. 

So what is the real goal here? Are we trying to inform users that there are backups that have gaps in the WAL between the backup wal start/stop, meaning that backup is invalid? Or that there are gaps between the backups so that they can't perform PITR? 

I know you are writing some test, so maybe it will be clearer to me once there are some examples of the expected results.

pgstef: With archive-retention=4, we'd have a WARNing saying we keep more archives than backups.

My idea is to inform if there are missing WAL archives between start/stop backup location but also between backups themselves. But I'd also like to take care correctly of the retention policy, to avoid reporting missing archives where they're not missing since they've been expired. I should verify that part to also take care of the time-based retention.

You're right, it should be easier to see with some tests. I'll try to submit them asap.

cmwshang:
One thing that might be an issue is collecting all of the WAL rather than ranges, but I'll let David weigh in on that.

dwsteele:
Agreed, we have users with hundreds of millions of WAL -- printing out missing WAL needs to be in ranges, not an exhaustive list. If they are missing a big chunk we could run out of memory trying to build the info output.

pgstef:
Hi,

I've added some tests (not pushed yet) while improving some of your previous remarks (.history regex,...).
Last point on my "to do" list is to have a look on how to support the time-based retention.

Regarding the missing archives list display, is that something for text and json output ? It would have been nice to have the accurate information in the json imho but I understand that we have to avoid OOM :-)

What kind of ranges would you imagine ? 1st - last missing, 1st - last of each continuous missing archives, 10 ou 20 first missing (could be a specific option) ?

pgstef:
Hm, in fact it supported already the time-based retention since the "aggressive" expire with `retention-archive-type` works the same way for `retention-full-type` count or time (regarding the tests I tried, not looked at the code though).

This implementation should detect:
* missing archives between start/stop wal location of each backup;
* missing archives in the complete list between min / max archives found in the repo (min is changed according to the archive retention specific policy).

Repo encryption is already covered in the tests. Not all the branches are covered yet `c module command/info/info is not fully covered (231/240 branches)` but I hope this will help for the review :-)

I'm waiting your advice regarding the "output" to change that part ;-)

cmwshang:
One thing we need to consider is the history file format. According to this: http://www.interdb.jp/pg/pgsql10.html there was a change in the format:

```
The timeline history file format is changed in version 9.3. Formats of versions 9.3 or later and earlier both are shown below but not in detail.
Later version 9.3:

timelineId	LSN	"reason"

Until version 9.2:

timelineId	WAL_segment	"reason"
```

However, I have been looking at 9.1 version of postgres code and am having trouble confirming the format. I am not sure if the "WAL_Segment" being referred to includes the timeline (e.g. 000000010000000000000009) or does not include it.

Since we support 9.2 versions and below, this will be something we will need to address.

test/src/module/command/infoTest.c
```diff
@@ -30,6 +30,7 @@ testRun(void)
     {
         StringList *argList = strLstNew();
         strLstAdd(argList, strNewFmt("--repo-path=%s/", strZ(repoPath)));
+        strLstAddZ(argList, "--repo1-retention-full=1");  // avoid warning
```
cmwshang, line 33: This change on line 33 was not necessary - not sure why it was added and would recommend reverting it.

levkk:
👍 Would be great if we could run this as the archiving is taking place and automatically upload missing segments. A race condition is possible where Postgres archives the segment before we can "heal" the archive though.

test/src/module/command/infoTest.c
```diff
@@ -1308,6 +2359,7 @@ testRun(void)
     {
         StringList *argList = strLstNew();
         strLstAdd(argList, strNewFmt("--repo-path=%s", strZ(repoPath)));
+        strLstAddZ(argList, "--repo1-retention-full=1");  // avoid warning
```
cmwshang, line 2362: Again, this change on line 2362 was not necessary - not sure why it was added and would recommend reverting it.

test/src/module/command/infoTest.c
pgstef: That's because of the `CFGOPT_REPO_RETENTION_FULL_TYPE` in `Data.pm`. I need to know the retention policy in the info command to generate the expected archives list.

Without that line in the tests, we get a warning in the log file.

Maybe there's another way to handle that information without allowing it in the `Data.pm` file ?

test/src/module/command/infoTest.c
pgstef: Same as above comment :-)

pgstef:
> One thing we need to consider is the history file format. According to this: http://www.interdb.jp/pg/pgsql10.html there was a change in the format:
> 
> ```
> The timeline history file format is changed in version 9.3. Formats of versions 9.3 or later and earlier both are shown below but not in detail.
> Later version 9.3:
> 
> timelineId	LSN	"reason"
> 
> Until version 9.2:
> 
> timelineId	WAL_segment	"reason"
> ```
> 
> However, I have been looking at 9.1 version of postgres code and am having trouble confirming the format. I am not sure if the "WAL_Segment" being referred to includes the timeline (e.g. 000000010000000000000009) or does not include it.
> 
> Since we support 9.2 versions and below, this will be something we will need to address.

I tried this morning to play the same test cases on PG 9.2. The `00000003.history` file content gives:

```
1	000000010000000000000009	before 2000-01-01 01:00:00+01
2	00000002000000000000000C	no recovery target specified
```

I guess I'll have to change the history file parsing part depending on the PG version and add some tests for that

test/src/module/command/infoTest.c
cmwshang: Yes there is another way. You didn't add any tests in this section (nor the other one) so somewhere in the tests you added, you're missing clearing the logs with harnessLogResult(). So once you figure out where that is, you should remove this additional `strLstAdd(arglist, "--repo1-retention-full=1");`

pgstef: Oh, I indeed didn't thought about clearing the logs between the tests. I'll have a look at that.

I was also wondering if there were any other way to access the retention policy options without changing the `CFGOPT_REPO_RETENTION_FULL_TYPE`,.. in `Data.pm`. Which would avoid the message in the logs..

pgstef: Hi @cmwshang,
Sorry it took me so long to check this.
The `harnessLogResult` is called correctly in the tests I added.

But as I was saying, in info.c, I need to get the `cfgOptRepoRetention*` info. So, I've modified `Data.pm` to make it available for the `info` command. 

The side effect is that if you run the info command without a proper `repo-retention-full` config, you get the warning message in the log. That's why I added the `strLstAdd(arglist, "--repo1-retention-full=1");` to the other tests, not because of not clearing logs.

That's why I was wondering if there were any other way to access the retention policy options without changing the `CFGOPT_REPO_RETENTION_FULL_TYPE`,.. in `Data.pm`.

Kind regards,

dwsteele: This warning is raised by the logic here: https://github.com/pgbackrest/pgbackrest/blob/master/src/config/load.c#L215.

So you just need an exception for the info command.

pgstef: Indeed, adding the exception and removing the unneeded test line seems to work as expected (commit 
24a1dd8).

Thanks for the tip!

dwsteele (changes requested):
I have now had a chance to go through this commit in detail. We are definitely interested in this functionality but there are a few issues with the current implementation that we'll need to address.

- Only the last history file is checked, so I'm guessing you are only concerned about the most recent timeline. I don't think that's enough, though -- at the end of the day we are concerned about backups so we need to be sure each backup can be recovered. The last timeline in the repo may skip the middle timelines (i.e. 1 -> 6) which does not tell us much about whether a backup on timeline 5 can be recovered. It's pretty common to get an extra timeline in the repo when somebody accidentally promotes a replica so the last timeline may be useless for determining if the backups will recover.

- The code builds two *full* lists of the archive files, which could consume a huge amount of memory. I think we need to follow the same strategy as in #1167, i.e. build up ranges of WAL files in the repo. It's certainly possible to have a pathological case where every other file is missing, but that's pretty unlikely. Once those lists are created, you can read the history files to be sure the timelines can be followed.

- Missing WAL should also be reported in ranges for the same reason.

- There is significant overlapping functionality with verify #1167.

I think verify is closer to the implementation we want so we should focus on getting that committed and see how we can use the same functionality for info. Note that the verify functionality is partial and we'll be committing as an internal command a little at a time.

It's been pretty helpful for me to review two different implementation with a similar goal. I think I now have a better idea overall of what this code needs to look like to work in both places, so I don't feel there is any wasted effort here.

pgstef:
> I have now had a chance to go through this commit in detail. We are definitely interested in this functionality but there are a few issues with the current implementation that we'll need to address.
> 
> I think verify is closer to the implementation we want so we should focus on getting that committed and see how we can use the same functionality for info. Note that the verify functionality is partial and we'll be committing as an internal command a little at a time.
> 
> It's been pretty helpful for me to review two different implementation with a similar goal. I think I now have a better idea overall of what this code needs to look like to work in both places, so I don't feel there is any wasted effort here.

Hi @dwsteele,

Thanks for your feedback. If I understand it correctly, I should wait until `verify` command (or at least the archives part) is committed before continue working on this patch. I'll have a closer look at the `verify` patch then.

Kind regards,

dwsteele:
> I'll have a closer look at the verify patch then.

Sounds like a good plan. I'm going to do my final review today and work on the memory management. Then it should be ready for commit.

dwsteele:
@pgstef Now that verify is officially released, I would like to close this in favor of a solution that leverages the verify code.

Not exactly sure what that looks like, but I think verify should have JSON output like info. Or verify should save some JSON in the repo that the next info run could see.

Thoughts?

pgstef:
> @pgstef Now that verify is officially released, I would like to close this in favor of a solution that leverages the verify code.

+1. Info should stay a "fast"-replying command.

> Not exactly sure what that looks like, but I think verify should have JSON output like info. Or verify should save some JSON in the repo that the next info run could see.
> 
> Thoughts?

If verify could store its last run in the repo, and info command pick it up, that would definitely speed up the monitoring checks. Updating the archive.info might be an easy possibility?

The monitoring check would also have to make sure the last verify run isn't too old and be able to trigger it if needed. So the JSON output would be handy from an automation point-of-view anyway.

If I had to chose between the 2 possibilities, I'd go for `verify --output=json` first. But having both would be awesome.

pgstef:
Hi,
Should we close this PR (given this implementation won't be kept) and open a project card to keep track of this topic?

dwsteele:
Makes sense to me. Did you want to do that?

pgstef:
Done [there](https://github.com/pgbackrest/pgbackrest/projects/2#card-84703818). Don't hesitate to adjust it if needed ;-)

dwsteele:
Looks good to me.

Issue #721, opened 2019-05-06
S3 AWS credentials auto resolution from IAM server role


aleszeleny:
Hi,
I've found post (see bellow) suggesting there were intentions for 2018 to use IAM host credentials for pgBackrest to authenticate to S3 instead of creating configuration include file with S3 credentials somewhat like:

`curl --silent  http://169.254.169.254/latest/meta-data/identity-credentials/ec2/security-credentials/ec2-instance/ | jq -r '"repo1-s3-key-secret="+.AccessKeyId, "repo1-s3-key="+.SecretAccessKey, "repo1-s3-token="+.Token,"#LastUpdated: "+.LastUpdated, "#Expiration: "+.Expiration'
`
What generally works for backup & restore, can be included in the beginning of backup/restore script, but i doubt how to exchange S3 credentials to make sure archive-push/archive-get commands to avoid transient issues while old file was removed and before new one is created. Therefore I'd like to kindly ask, whether and if yes when build-in AWS credential usage is expected to be available, or how to handle include configuration file update while avoiding transient issues for archive-push/archive-get commands.

Thanks Ales

@farrellit We are planning to provide this functionality in the fall.

Unfortunately, the available Perl APIs are a bit of a mess and seem to think you'd like to retrieve a file from S3 all at once.  For a highly-parallel process having the chance that each process could use multiple gigabytes of memory is not OK.  Also, none of them are available as packages, so that means rolling our own or requiring CPAN, which we won't do.

_Originally posted by @dwsteele in https://github.com/pgbackrest/pgbackrest/issues/518#issuecomment-417725677_

dwsteele:
Yes, we are still planning to implement this functionality.

The delay is because we made the decision to migrate the entire code base to C and so adding new features means adding them in two places and writing two sets of tests.  You can imagine how we might not find that very productive.

The C migration is planned to be completed in October but we are contemplating adding an interface between Perl and the C storage layer to enable us to add features like this earlier.  I'm doing some of the prelim work for that right now, but we haven't made a final decision yet.

So the answer to your question is sometime between this summer and the end of the year depending on which development path we take.

aleszeleny:
That makes sense, thanks for sharing this information. Thanks for all the work in pgBackrest!

dwsteele:
We have implemented this feature in PR #1137. Please have a look if you can.

PR #1137, opened 2020-08-10
Automatically retrieve temporary S3 credentials on AWS instances.

dwsteele:
Automatically retrieve temporary credentials for S3 when the AWS instance is associated with an IAM role. Credentials are automatically updated when they are <= 5 minutes from expiring.

Basic configuration is to set `repo1-s3-key-type=auto`. `repo1-s3-role` can be used to set a specific role, otherwise it will be retrieved automatically. We welcome discussion of these option names/values.

Most of the changed files deal with configuration -- the main code changes are in `src/storage/s3/storage.c`.


dwsteele:
@Xof, you have expressed interest in this feature in the past. We'd appreciate it if you could have a look, doing some testing, comment on the functionality, etc.

src/storage/s3/storage.c
```diff
@@ -273,6 +315,99 @@ storageS3RequestAsync(StorageS3 *this, const String *verb, const String *uri, St
...
+                httpRequestError(request, response);
+
+            // Free old credentials
+            strFree(this->accessKey);
+            strFree(this->secretAccessKey);
+            strFree(this->securityToken);
+
+            // Get credentials from the JSON response
+            KeyValue *credential = jsonToKv(strNewBuf(httpResponseContent(response)));
+            (void)credential;
```
cmwshang, line 380: Casting to void is often used to suppress warnings for unused variables but I don't see why this is done here in (void)credential; - there are no conditionally compiled code sections with the credential variable so can you explain why this is needed here?

dwsteele: Looks I put that in so I could get it to compile before writing the rest of the section. Removed.

youattd:
Built and tested on Amazon Linux v1 and v2.  
Using CloudTrail you can see the AssumeRole API events happen about every hour (default).
In the AssumeRole event body, the new temp key has an expiration time about 6 (or 8?) hours from the event timestamp.

aleszeleny:
Hello,

thanks a lot for the amazing work done!

Based on https://github.com/pgbackrest/pgbackrest/issues/721 I've tried to run a backup having:
`repo1-s3-key-type=auto`
and commented out
```
#repo1-s3-key=
#repo1-s3-key-secret=
```

Stanza creation proceeded, `info` command for other stanzas having some backups in place worked as well.

The backup is running (testing on ~1TB pg cluster, so it'll take some time to finish (dev server only 2 parallel processes).

I doubt the credential updates "<= 5 minutes from expiring", there are many cases (like `vacuumdb` on an active instance running 4 parallel jobs causing pgBackRest (async push, 4 parallel processes on 16vCPU machine) archiving thousands (30k...) of WAL files (default size 16MB each) takes 20-40 minutes to complete (server was not overloaded, so adding more parallel push processes is an option). From Credential timing perspective, it depends whether pgBackrest use somewhat cashed credentials and if all WAL files are sent using single HTTP stream to S3 (therefore once authentication is done, the credentials might expire with no issue) or if it opens a new session to S3 for every WAL file using the cached credentials ( and it'll be an issue unless retrieval fo new credential will be triggered).

I'll update you the backup results, it'll take some more time to prepare an environment where to restore from the testing backup.

Once again, thanks for that change!
Ales

dwsteele:
@aleszeleny Yes -- we don't reuse credentials between executions of pgBackRest. To do so we'd need to cache them somewhere and for complexity/security reasons it seems better just to request them again. So renewal is just for longer running session that have credentials that will expire before the work is done.

JBromage:
Hi, 
I've also tested this against a 5TB db ( what are limits if they aren't to be pushed after all ) - in all it took 15 hours, 34 minutes, but it had been throttled back as I wanted it to potentially fail as it had been doing after 6 hours......I just used the parameter repo1-s3-key-type=temp, still using an instance profile IAM , so normally the temp token would expire after 6 hours.... and without any intervention, its completed without any errors / having to resume the backup. 
This is absolutely what I needed, and thank you for all your effort on this. If you need any details / help with documentation for this, I've got all the logfiles etc so can provide sample outputs and sample config. Thank you again, this is brilliant!

aleszeleny:
Hello, my testing backup succeeded as well:
```2020-08-25 04:58:57.597 P00   INFO: full backup size = 2405.4GB
2020-08-25 04:58:57.597 P00   INFO: execute non-exclusive pg_stop_backup() and wait for all WAL segments to archive
2020-08-25 04:58:58.098 P00   INFO: backup stop archive = 00000002000058790000003B, lsn = 5879/3B97FB28
2020-08-25 04:58:58.262 P00   INFO: check archive for segment(s) 000000020000587900000014:00000002000058790000003B
2020-08-25 04:58:59.950 P00   INFO: new backup label = 20200824-123648F
2020-08-25 04:59:00.970 P00   INFO: backup command end: completed successfully (58852039ms)
```
Thanks for the fantastic feature!

dwsteele:
@JBromage I have added the user guide documentation if you would like to have a look: https://github.com/pgbackrest/pgbackrest/pull/1137/files#diff-a62458990193106959efca32e11bc8ef

I'd rather not provide specific AWS-side configuration in this case since I don't think we need to cover that level of detail in the pgBackRest user guide.

JBromage:
> @JBromage I have added the user guide documentation if you would like to have a look: https://github.com/pgbackrest/pgbackrest/pull/1137/files#diff-a62458990193106959efca32e11bc8ef
> 
@dwsteele - yep, that all looks good to me  - -thank you again for this

dwsteele:
And committed! Thank you to everyone for your assistance reviewing and testing this patch.

PR #1151, opened 2020-08-18
Asynchronous list/remove for S3/Azure storage.

dwsteele:
Improve the performance of list/delete operations by using async requests.

It's questionable whether this will have any impact on Azure deletes since they are sent one at a time with little work done in between, but it doesn't hurt to try.

### DIFF
--- a/src/storage/azure/storage.c
+++ b/src/storage/azure/storage.c
@@ -337,8 +337,6 @@ storageAzureListInternal(
 
     MEM_CONTEXT_TEMP_BEGIN()
     {
-        const String *marker = NULL;
-
         // Build the base prefix by stripping off the initial /
         const String *basePrefix;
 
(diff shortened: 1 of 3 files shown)


src/storage/azure/storage.c
```diff
@@ -363,35 +361,61 @@ storageAzureListInternal(
...
-                if (marker != NULL)
-                    httpQueryAdd(query, AZURE_QUERY_MARKER_STR, marker);
+                // Get sync response on first request or async response on continuation
+                HttpResponse *response = NULL;
 
-                // Add the delimiter to not recurse
-                if (!recurse)
-                    httpQueryAdd(query, AZURE_QUERY_DELIMITER_STR, FSLASH_STR);
+                if (request == NULL)
+                    response = storageAzureRequestP(this, HTTP_VERB_GET_STR, .query = query);
```
cmwshang, line 394: I'm a little confused why line 394 would not be wrapped in storageAzureResponseP. If line 397 is 
`response = storageAzureResponseP(request); `
then response should be coming from a call to a response function. By line 397, `request` was set by line 415:
`request = storageAzureRequestAsyncP(this, HTTP_VERB_GET_STR, .query = query);`

then why wouldn't line 394 be:
`response = storageAzureResponseP(storageAzureRequestP(this, HTTP_VERB_GET_STR, .query = query));`

I mean is `response` a response or is it a request?  That's the part that is a bit confusing - what really is the variable `response`?

src/storage/s3/storage.c
```diff
@@ -391,32 +389,59 @@ storageS3ListInternal(
...
             {
-                HttpQuery *query = httpQueryNewP();
+                // Get sync response on first request or async response on continuation
+                HttpResponse *response = NULL;
 
-                // Add continuation token from the prior loop if any
-                if (continuationToken != NULL)
-                    httpQueryAdd(query, S3_QUERY_CONTINUATION_TOKEN_STR, continuationToken);
+                if (request == NULL)
+                    response = storageS3RequestP(this, HTTP_VERB_GET_STR, FSLASH_STR, query);
```
cmwshang, line 419: Same question I had in the Azure storage.c: is `response` a response or a request?

src/storage/azure/storage.c
dwsteele: `storageAzureResponseP()` is used to get a response from a request started asynchronously by `storageAzureRequestAsyncP()`. storageAzureRequestP() does an sync request and returns the response immediately.

If the request is NULL it means we don't have any date yet, so we need to do a sync request to get some. Doing an async request here does us not good since we don't have anything else we can do while waiting for the data.

> what really is the variable response

It's defined as HttpResponse on line 391.

Is the confusion here over the naming of the functions, perhaps? We could work on that, but as part of another commit.

src/storage/s3/storage.c
dwsteele: It's the same, HttpResponse.

src/storage/azure/storage.c
dwsteele: Fixed some typos in the comment above, perhaps you should re-read.

cmwshang: It's definitely a naming thing. I think at a minimum, could you add a couple comments: 
at line 394: 
`get an immediate response from a synchronous request`
and at line 397:
`get a response from a request started asynchronously`

src/storage/s3/storage.c
cmwshang: same request for comments as in the Azure.c

src/storage/azure/storage.c
dwsteele: I kind of thought that's what this comment said:
```
// Get sync response on first request or async response on continuation
```
Do you want to expand on that comment, move it, or just say it again in the if-else?

dwsteele: How about:
```
                // If there is no outstanding async request then get the response immediately from a sync request
                HttpResponse *response = NULL;

                if (request == NULL)
                {
                    response = storageS3RequestP(this, HTTP_VERB_GET_STR, FSLASH_STR, query);
                }
                // Else get the response from the outstanding sync request
                else
                {
                    response = storageS3ResponseP(request);

                    httpRequestFree(request);
                    request = NULL;
                }
```

dwsteele: Or maybe it is clearer to invert it:
```
                // If there is an outstanding async request then wait for the response
                HttpResponse *response = NULL;

                if (request != NULL)
                {
                    response = storageS3ResponseP(request);

                    httpRequestFree(request);
                    request = NULL;
                }
                // Else get the response immediately from a sync request
                else
                {
                    response = storageS3RequestP(this, HTTP_VERB_GET_STR, FSLASH_STR, query);
                }
```

cmwshang: If you locate 
`// If there is no outstanding async request then get the response immediately from a sync request`
right above 
`if (request == NULL)`
then that would help. Proximity of the comment is important here. The naming of the functions is the issue and I know I should have caught it before but that was the PR I had a bit of trouble following. As you said, we can make adjustments to the function names in another commit. And I KNOW naming things is HARD! :-)

dwsteele: OK, done that way.

src/storage/s3/storage.c
dwsteele: Same as above.

PR #1160, opened 2020-08-25
Add archive-mode option to disable archiving on restore.

dwsteele:


src/command/restore/restore.c
```diff
@@ -1329,6 +1334,26 @@ restoreRecoveryOption(unsigned int pgVersion)
...
+        const String *archiveMode = cfgOptionStr(cfgOptArchiveMode);
+
+        if (!strEq(archiveMode, ARCHIVE_MODE_PRESERVE_STR))
+        {
+            if (pgVersion < PG_VERSION_12)
+            {
+                THROW_FMT(
+                    OptionInvalidError,
+                    "option '" CFGOPT_ARCHIVE_MODE "' is not supported on " PG_NAME " < " PG_VERSION_12_STR "\n"
+                        "HINT: archive_mode should be set manually in postgresql.conf.");
```
sfrost, line 1347: I would have this explicitly say "archive_mode should be set to off manually in postgresql.conf" or something along those lines, since that's what the user asked to try and do.

test/src/module/command/helpTest.c
```diff
@@ -124,6 +124,8 @@ testRun(void)
             "\n"
             "Command Options:\n"
             "\n"
+            "  --archive-mode                   disable archiving on restored cluster\n"
```
sfrost, line 127: Seems like this would be better as "preserve or disable archiving" or something along those lines..  seems a bit odd to describe it only as "disable"

src/command/restore/restore.c
dwsteele: Sounds good, done that way.

test/src/module/command/helpTest.c
dwsteele: Preserve is the default -- so if you are using it you are probably disabling. But I'm fine with the new language as well.

PR #1165, opened 2020-09-03
Add hint about starting the stanza when WAL segment not found.

machack666:
Pull in approved wording changes for WAL hints.


dwsteele:
Amended this commit at 9fd31913a8cb to update more tests.

PR #1171, opened 2020-09-11
Improve working directory error message.

dwsteele:
Improve the wording of the error message and add a hint to make it clearer what is wrong and how the user can fix it.

Also change the assert to a regular error since this is not an internal error.


pgstef (approved):
Hi,

The comment appears to be consistent with those in `cmdArchivePush()` in **push.c**.
The hint should help in most of miss-configuration cases.

It looks good to me.

PR #1175, opened 2020-09-16
Error with hints when backup user cannot read pg_settings.

dwsteele:
This condition used to give a not-very-clear error which we have been intending to improve. But in the meantime the changes in fbff299 resulted in a segfault for this condition instead because the data_directory was assumed to be non-NULL.

Fix this by explicitly throwing an error with hints when any row in pg_settings cannot be selected.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -19,6 +19,8 @@
                         <release-item-contributor-list>
                             <release-item-ideator id="mohamed.insaf.k"/>
                             <release-item-contributor id="david.steele"/>
+                            <release-item-reviewer id="stefan.fercot"/>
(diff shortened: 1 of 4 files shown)


pgstef (approved):
The test seems pretty simple.
Even if system-side users may be different than "postgres" and the confuse a little bit the users, I like the HINTs. I would probably prefer something like `is the backup running as superuser?` but it might add more confusion then. 

So, this looks good to me as it is.

Kind regards,

cmwshang (approved):

dwsteele:
> I would probably prefer something like is the backup running as superuser?

Yeah, I went back and forth on this but the vast majority of clusters use the postgres user. We can certainly tweak the language but I think for this release fixing the segfault is most important.

PR #1181, opened 2020-09-21
Update yum repository path for CentOS/RHEL user guide.

crunchyheath:
This change allows the running of the ./doc.pl script when testing a pgbackrest build based on PostgreSQL 13.  Without this change the version of the pgdg-redhat-repo that is installed does not container the pgdg-13 repo at all.  I have built and run this on centos-7 and with it I am able to fully run the ./doc.pl script to complete the regression.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -76,6 +76,19 @@
                     </release-item>
                 </release-development-list>
             </release-core-list>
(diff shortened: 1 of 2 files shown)


dwsteele (approved):
Looks, good. Thanks!

PR #1191, opened 2020-10-06
Improve performance of large file lists in backup/restore commands.

dwsteele:
lstRemoveIdx(list, 0) resulted in the entire list being moved down to the first position which could take a long time for big lists. This is a common pattern in backup/restore when processing file queues.

Instead simply move the list pointer up when first item is removed. Then on insert check if there is space at the beginning when there is no longer space at the end and do the move then. This way if a list is built and then drained without any new inserts then no move is required.


src/common/type/list.c
```diff
@@ -294,9 +295,19 @@ lstInsert(List *this, unsigned int listIdx, const void *item)
...
+    // Else if there is space before the beginning of the list then move the list down
+    else if (
+        (this->list != this->listAlloc) &&
+        (this->listSize + ((uintptr_t)(this->list - this->listAlloc) / this->itemSize) == this->listSizeMax))
+    {
+        memmove(this->listAlloc, this->list, this->listSize * this->itemSize);
+        this->list = this->listAlloc;
+    }
 
     // If not inserting at the end then move items down to make space
```
cmwshang, line 312: Could you move this comment  `// If not inserting at the end then move items down to make space` to right before the `if (listIdx != lstSize(this))` statement?

cmwshang (commented):
Looks good. One minor request to relocate a comment...

src/common/type/list.c
dwsteele: Sure. Also added a comment.

cmwshang: Awesome.

cmwshang (approved):

PR #1192, opened 2020-10-08
Optimize small reads in IoRead.

dwsteele:
Instead of using memmove() to manage the internal output buffer for every small read, track the current buffer position and only move data when the small read cannot be satisfied and more data is needed.


sfrost (approved):
This generally looks good but I am left wondering if there might be some way to eliminate all of the memmove's that we have...  It's just not a terribly efficient thing to do, in the general sense, and this doesn't end up eliminating all of them even though it does look like it reduces the number of them quite a bit which is good.

dwsteele:
I'm not sure it is possible to eliminate them entirely. Sometimes not all the data required for a read will be in the buffer so we'll need to move it down to make space (or create another buffer).

But this should get rid of the vast majority of them, and even when a memmove() is needed this reduces the likelihood of the regions overlapping which should improve performance and reduce cache pollution.

src/common/io/read.c
```diff
@@ -191,18 +192,18 @@ ioRead(IoRead *this, Buffer *buffer)
...
     {
-        // Determine how much data should be copied
-        size_t size = bufUsed(this->output) > bufRemains(buffer) ? bufRemains(buffer) : bufUsed(this->output);
+        // Internal output buffer remains taking into account the position
+        size_t outputRemains = bufUsed(this->output) - this->outputPos;
 
-        // Copy data to the user buffer
-        bufCatSub(buffer, this->output, 0, size);
+        // Determine how much data should be copied
+        size_t size = outputRemains > bufRemains(buffer) ? bufRemains(buffer) : outputRemains;
```
cmwshang, line 202: can you rename outputRemains inside the IF to outputRemainsInternal or something? it would help clarify this section of code

cmwshang (changes requested):
just a minor change request

src/common/io/read.c
dwsteele: Done, and also done in ioReadLineParam().

cmwshang (approved):

PR #1194, opened 2020-10-09
Add pack type.

dwsteele:
This patch had three parts. They are presented together to help provide context but will be committed separately.

1) ioReadSmall() function in read.c. This function is optimized for small binary reads (very similar to ioReadLineParam().

2) The pack type in pack.c. See pack.h for details.

3) A pack implementation in storage/remote/protocol.c and storage/remote/storage.c. that greatly reduces the amount of data sent over the protocol layer when transmitting info lists.


src/common/type/pack.h
```diff
@@ -0,0 +1,500 @@
...
+PackWrite *write = pckWriteNew(buffer);
+pckWriteU64P(write, 77);
+pckWriteBoolP(write, false, .defaultNull = true);
+pckWriteI32P(write, -1, .defaultNull = true, .defaultValue = -1);
+pckWriteStringP(write, STRDEF("sample"));
+pckWriteEndP();
+
+A string representation of this pack is `1:uint64:77,4:str:sample`. The boolean was not stored because we requested NULL on default
+and the default was false (because it was not set to anything else). The int32 field was also not stored but we set an explicit
+default value. Note that there is a gap in the ID stream, which represents the NULL values.
```
cmwshang, line 37: I'm still not clear on why the Int32 value was not stored. Are you trying to say that not only do we not store NULL but we also do not store values that are the same as the default value? So if we had not had a default value for the int32 then if we'd set the `pckWriteI32P(write, 0, .defaultNull = true);` then this would mean the same thing?

src/common/type/pack.h
```diff
@@ -0,0 +1,500 @@
...
+
+A pack is an object by default. Objects can store fields, objects, or arrays. Objects and arrays will be referred to collectively as
+containers. Fields contain data to be stored, e.g. integers, strings, etc.
+
+Here is a simple example of a pack:
+
+PackWrite *write = pckWriteNew(buffer);
+pckWriteU64P(write, 77);
+pckWriteBoolP(write, false, .defaultNull = true);
+pckWriteI32P(write, -1, .defaultNull = true, .defaultValue = -1);
```
cmwshang, line 31: Can you clarify what defaultNull is vs defaultValue? It seems to me defaultNull means the field can be Nullable which in the case of the integer example would be "it can be zero"? But that doesn't really make sense. Maybe a definition for defaultNull and defaultValue would help.

src/common/type/pack.h
```diff
@@ -0,0 +1,500 @@
...
+A string representation of this pack is `1:uint64:77,4:str:sample`. The boolean was not stored because we requested NULL on default
+and the default was false (because it was not set to anything else). The int32 field was also not stored but we set an explicit
+default value. Note that there is a gap in the ID stream, which represents the NULL values.
+
+This pack can be read with:
+
+PackRead *read = pckReadNew(buffer);
+pckReadU64P(read);
+pckReadBoolP(read, .defaultNull = true);
+pckReadI32P(read, .defaultNull = true, .defaultValue = -1);
```
cmwshang, line 44: We're reading so why do we have to specify the defaultValue again?  I see in line 48 you state "If we don't care about the NULLs" then we don't need to specify defaultNull and then defaultValue also because - what - defaulValue is treated as NULL?

src/common/type/pack.h
```diff
@@ -0,0 +1,500 @@
...
+NULLs are not stored in a pack and are therefore not typed. A NULL is essentially just a gap in the field IDs. Fields that are
+frequently NULL are best stored at the end of an object.
+
+A pack is an object by default. Objects can store fields, objects, or arrays. Objects and arrays will be referred to collectively as
+containers. Fields contain data to be stored, e.g. integers, strings, etc.
+
+Here is a simple example of a pack:
+
+PackWrite *write = pckWriteNew(buffer);
+pckWriteU64P(write, 77);
```
cmwshang, line 29: U64P, BoolP, I32P - these all make me think we're reading pointers. I strongly recommend we drop the "P" on the end since we're reading a "type". And, how do we read a pointer anyway?

cmwshang (commented):
Partial review - discussion needed on some fundamental questions before continuing.

src/common/type/pack.c
```diff
@@ -0,0 +1,1513 @@
+/***********************************************************************************************************************************
+Pack Type
+
+Each pack field begins with a one byte tag. The four high order bits of the tag contain the field type (PackType). The four lower
+order bits vary by type.
+
+Integer types (packTypeData[type].valueMultiBit) when the value is >= -1 and <= 1:
+  3 more value indicator set to 0
```
cmwshang, line 8: Recommend using a dash and being consistent in using "bit" in the explanation. e.g. 
`3 - more value indicator bit set to 0`

cmwshang (commented):
Interim comments.

src/common/type/pack.c
```diff
@@ -0,0 +1,1513 @@
...
+}
+
+/***********************************************************************************************************************************
+Read bytes from the buffer
+
+IMPORTANT NOTE: To avoid having dyamically created return buffers the current buffer position (this->bufferPos) is stored in the
+object. Therefore this function should not be used as a parameter in other function calls since the value of this->bufferPos will
+change.
+***********************************************************************************************************************************/
+static size_t pckReadBuffer(PackRead *this, size_t size)
```
cmwshang, line 237: CRLF needed before pckReadBuffer

src/common/type/pack.c
```diff
@@ -0,0 +1,1513 @@
...
+} PackTagStack;
+
+struct PackRead
+{
+    MemContext *memContext;                                         // Mem context
+    IoRead *read;                                                   // Read pack from
+    Buffer *buffer;                                                 // Buffer to contain read data
+    const uint8_t *bufferPtr;                                       // Pointer to buffer
+    size_t bufferPos;                                               // Position in the buffer
+    size_t bufferMax;                                               // Maximum position of buffer
```
cmwshang, line 147: What do you mean by `Maximum position of buffer`? Is this the position of the last used byte in the buffer (line 262 makes me think that)? Or total size? The way you're using it in line 246 makes me think it is more like bufferLimit? And `of buffer` - do you mean `of the data in the buffer`?

src/common/type/pack.c
```diff
@@ -0,0 +1,1513 @@
...
+            bufUsedZero(this->buffer);
+
+            // Limit the buffer for the next read so we don't read past the end of the pack
+            bufLimitSet(this->buffer, size < bufSizeAlloc(this->buffer) ? size : bufSizeAlloc(this->buffer));
+
+            // Read bytes
+            ioReadSmall(this->read, this->buffer);
+            this->bufferPos = 0;
+            this->bufferMax = bufUsed(this->buffer);
+            remaining = this->bufferMax - this->bufferPos;
```
cmwshang, line 263: `remaining` is always going to be equal to bufferMax since bufferPos is always reset to 0. Why is that?

cmwshang (commented):
partial review

src/common/type/pack.c
```diff
@@ -0,0 +1,1513 @@
...
+
+Each pack field begins with a one byte tag. The four high order bits of the tag contain the field type (PackType). The four lower
+order bits vary by type.
+
+Integer types (packTypeData[type].valueMultiBit) when the value is >= -1 and <= 1:
+  3 more value indicator set to 0
+  2 value low-order bit
+  1 more ID delta indicator
+  0 ID delta low order bit
+
```
cmwshang, line 12: Given the "Structure:" representation below this (from the tests), can you clarify the following:

high bits:      low bits:
3   2   1   0 | 3   2   1   0
-----------------------------  // array item 2, type u64, value 1
1   0   1   1   0   1   0   0 = B4 => B = pckTypeU64 (11), 4 =  "value of low order bit" which is 1 so what would it be if it were -1? 
-----------------------------  // id 10, type u64, value 1
1   0   1   1   0   1   1   0 = B601 => pckTypeU64 (11), 6 = "value low-order bit" + "more ID delta indicator" then, 01 according to the description "the tag will be followed by a base-128 encoded integer with the higher order ID delta bits. The ID delta represents the delta from the ID of the previous field.", the previous field ID was 7 and this field is 10 (8, 9, are missing/null) so the delta from the previous field is 3 so why is the value only 01? Since the "more ID delta indicator" seems to mean nothing here then why is this field not also represented as B4 as is the array one above?    

Structure:
           "1:u64:488"
             ", 2:u64:1911246845"
             ", 7:u64:18446744073709551615"
            **", 10:u64:1"**
            ", 11:u64:77"
            ", 12:u32:127"
            ", 13:i64:-1"
            ", 14:i32:-1"
            ", 15:bool:true"
            ", 20:bool:false"
            ", 28:obj:"
            "{"
                   "1:bool:true"
                   ", 2:bool:false"
            "}"
            ", 37:array:"
            "["
                  "1:u64:0"
                **", 2:u64:1"**
                ", 3:u64:2"
                ", 4:u64:3"
            "]"

src/common/type/pack.c
```diff
@@ -0,0 +1,1513 @@
...
+0-1 ID delta low order bit
+
+String and binary types (packTypeData[type].valueSingleBit):
+  3 value bit
+  2 more ID delta indicator
+0-1 ID delta low order bits
+
+Array and object types:
+  4 more ID delta indicator
+0-3 ID delta low order bit
```
cmwshang, line 25: I thought we were referring to 4 bits, not 5, so why are we referencing "4" and then "0-3"?

cmwshang (commented):
Partial review

src/common/type/pack.h
dwsteele: Clarified in 919dc9668c9.

src/common/type/pack.h
dwsteele: Same as above.

src/common/type/pack.h
dwsteele: Clarified in 021a830254e62.

src/common/type/pack.c
dwsteele: Updated in 919dc9668c93.

src/common/type/pack.c
dwsteele: Fixed in f50252aa78e9ef4.

src/common/type/pack.h
```diff
@@ -0,0 +1,500 @@
...
+    pckReadI64(this, (PckReadInt64Param){VAR_PARAM_INIT, __VA_ARGS__})
+
+int64_t pckReadI64(PackRead *this, PckReadInt64Param param);
+
+// Read object begin/end
+#define pckReadObjBeginP(this, ...)                                                                                                \
+    pckReadObjBegin(this, (PackIdParam){VAR_PARAM_INIT, __VA_ARGS__})
+
+void pckReadObjBegin(PackRead *this, PackIdParam param);
+
```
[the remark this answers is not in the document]
dwsteele: Removed in 84128037a1ae5.

src/common/type/pack.h
```diff
@@ -0,0 +1,500 @@
...
+    pckWriteObjBegin(this, (PackIdParam){VAR_PARAM_INIT, __VA_ARGS__})
+
+PackWrite *pckWriteObjBegin(PackWrite *this, PackIdParam param);
+
+#define pckWriteObjEndP(this)                                                                                                      \
+    pckWriteObjEnd(this)
+
+PackWrite *pckWriteObjEnd(PackWrite *this);
+
+// Read pointer. Use with extreme caution. Pointers cannot be sent to another host -- they must only be used locally.
```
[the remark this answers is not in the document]
dwsteele: Yes, fixed in 84128037a1ae5.

src/common/type/pack.h
```diff
@@ -0,0 +1,500 @@
...
+
+void pckReadObjBegin(PackRead *this, PackIdParam param);
+
+
+#define pckReadObjEndP(this)                                                                                                       \
+    pckReadObjEnd(this)
+
+void pckReadObjEnd(PackRead *this);
+
+// Read pointer. See pckWritePtrP() for cautions.
```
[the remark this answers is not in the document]
dwsteele: Done in 84128037a1ae5.

src/common/type/pack.c
```diff
@@ -0,0 +1,1513 @@
...
+        .name = STRDEF("obj"),
+    },
+    {
+        .type = pckTypePtr,
+        .valueMultiBit = true,
+        .name = STRDEF("ptr"),
+    },
+    {
+        .type = pckTypeStr,
+        .valueSingleBit = true,
```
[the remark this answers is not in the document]
dwsteele: Strings tend to be longish in terms of bits so all we're storing here is if the string length is 0 or > 0. Clarified in 6eff7981d8.

src/common/type/pack.c
dwsteele: Fixed in 84128037a1a.

src/common/type/pack.c
dwsteele: Yeah, this is confusing. It is set from bufUsed() so I think it makes sense to just call it bufUsed. Also updated the comment. Done in 84128037a1a.

We can't just used bufUsed because there are two different buffers we might be looking at depending on how the object was constructed, i.e. a buffer passed from the user or an internal buffer used with IoRead.

src/common/type/pack.c
dwsteele: Good point --  probably copied that from above. Fixed in 84128037a1a.

src/common/type/pack.c
dwsteele: Hopefully clarifications in f50252aa78e9e help some here. Also, you are planning to add some examples.

src/common/type/pack.h
dwsteele: We have convention to suffix parameterized function macros with P. If there's confusion with the `FUNCTION_LOG_PARAM_P()` macro (and others like it) I vote we rename those macro, .e.g `FUNCTION_LOG_PARAM_PTR()`, which would follow our general convention of using `ptr` in function names when specifically dealing with pointers.

cmwshang: ok

src/common/type/pack.c
```diff
@@ -0,0 +1,1571 @@
...
+Some tags and data may be skipped based on the value of the id parameter.
+***********************************************************************************************************************************/
+static uint64_t
+pckReadTag(PackRead *this, unsigned int *id, PackType type, bool peek)
+{
+    FUNCTION_TEST_BEGIN();
+        FUNCTION_TEST_PARAM(PACK_READ, this);
+        FUNCTION_TEST_PARAM_P(UINT, id);
+        FUNCTION_TEST_PARAM(ENUM, type);
+        FUNCTION_TEST_PARAM(BOOL, peek);
```
cmwshang, line 476: Can you give me an idea as to what the purpose of peek really is? It's not very clear except that it can be true or false when the type is unknown but otherwise must be false.

src/common/type/pack.c
```diff
@@ -0,0 +1,1571 @@
...
+    FUNCTION_TEST_BEGIN();
+        FUNCTION_TEST_PARAM(PACK_READ, this);
+        FUNCTION_TEST_PARAM(UINT, param.id);
+        FUNCTION_TEST_PARAM(BOOL, param.defaultNull);
+        FUNCTION_TEST_PARAM(INT, param.defaultValue);
+    FUNCTION_TEST_END();
+
+    ASSERT(this != NULL);
+
+    if (pckReadDefaultNull(this, param.defaultNull, &param.id))
```
cmwshang, line 755: param.id is not always set so in those cases we're always going to look at the object that we consider the next to be read, right? But I guess everything about the pack is you must read and write in order so say I have all int32s then if I think I'm reading ID 4 with a default of 5 but I didn't specify the ID so I'm actually reading ID 3 because the last thing I read was ID 2, then I could be checking the defaultNull for the wrong ID. But that is the chance we take, right?

src/common/type/pack.h
```diff
@@ -0,0 +1,505 @@
...
+
+Fields in a pack are identified by IDs. A field ID is stored as a delta from the previous ID, which is very efficient, but means
+that reading from the middle is generally not practical. The size of the gap between field IDs is important -- a gap of 1 never
+incurs extra cost, but depending on the field type larger gaps may require additional bytes to store the field ID delta.
+
+NULLs are not stored in a pack and are therefore not typed. A NULL is essentially just a gap in the field IDs. Fields that are
+frequently NULL are best stored at the end of an object. When using .defaultNull in write functions a NULL will be written (by
+making a gap in the IDs) if the value matches the default. When using .defaultNull in read functions the default will be returned
+when the field is NULL (i.e. missing). The standard default is the C default for that type (e.g. bool = false, int = 0) but can be
+changed with the .defaultValue parameter.
```
cmwshang, line 24: What do you think about clarifying the last, especially with respect to the WriteStr:
```
The standard default is the C default for that type (e.g. bool = false, int = 0) but 
can be changed with the .defaultValue parameter. For example, 
pckWriteBoolP(write, false) will write a 0 with an ID into the pack, but 
pckWriteBoolP(write, false, .defaultNull = true) will not write to the pack,
it will simply skip the ID. However, pckWriteStrP(packWrite, NULL) is not valid,
instead .defaultNull=true must also be passed.
```

src/common/type/pack.h
```diff
@@ -0,0 +1,505 @@
...
+
+PackRead *read = pckReadNew(buffer);
+pckReadU64P(read);
+pckReadBoolP(read, .defaultNull = true);
+pckReadI32P(read, .defaultNull = true, .defaultValue = -1);
+pckReadStringP(read);
+pckReadEndP();
+
+Note that defaults are not stored in the pack so any defaults that were applied when writing (by setting .defaultNull and
+optionally .defaultValue) must be applied again when reading (by setting .defaultNull and optionally .defaultValue).
```
cmwshang, line 52: What do you think about some clarification:
```
Note that defaults are not stored in the pack so any defaults that were applied when writing 
(by setting .defaultNull and optionally .defaultValue) must be applied again when reading
(by setting .defaultNull and optionally .defaultValue), otherwise the value returned will be 
the .defaultValue specified when reading (e.g. pckReadI32P(read, .defaultNull = true,
.defaultValue = 200); will return 200 even though the original 
pckWriteI32P(write, -1, .defaultNull = true, .defaultValue = -1); specified -1). Similarly, if
a default was set when the value was written, but not passed when the value is read and the
value read is the default, then an error will occur because the field will not exist.
```

src/common/type/pack.h
```diff
@@ -0,0 +1,505 @@
...
+
+If we don't care about the NULLs, another way to read is:
+
+PackRead *read = pckReadNew(buffer);
+pckReadU64P(read);
+pckReadStringP(read, .id = 4);
+pckReadEndP();
+
+This skips the NULLs but in practice it is better to read all the fields just as they were written so constant IDs are not required.
+By default each read and write advances the ID by one.
```
cmwshang, line 62: What do you think about clarifying this last statement:
```
By default each read and write advances the ID by one, meaning when specifying an ID
to read/write if the ID was already read/written or skipped, an error will occur because it
cannot be read/written again.
```

cmwshang (commented):
partial review with some questions/comments

src/common/type/pack.c
dwsteele: Peek allows you to look at the next tag without updating the internal state of the pack. So, for instance, you can look for gaps that indicate NULLs without advancing the field id. If the field id advances, then a function like pckReadNull() would fail because the current field id in the pack would be greater than the field id requested, because field ids for NULLs don't exist. It is also used to find the end of containers.

src/common/type/pack.c
dwsteele: Correct -- if you are going to be skipping fields you really need an enum that specifies the field ids to be used on both sides. There will be cases like this.

However, our current use case (the one in this patch) does not create any gaps so constant IDs are not needed. But the serializer/deserializer need to be coded with the same logic.

src/common/type/pack.h
dwsteele: That looks fine to me.

src/common/type/pack.h
dwsteele: That looks good, but no need to put ; after functions when writing them in comments.

src/common/type/pack.h
dwsteele: That's a bit awkward. How about:
```
By default each read/write advances the ID by one. If the ID is specified it must be unique and increasing. 
An error will occur if an ID is read/written and then the same ID is read/written again.
```

src/common/type/pack.c
```diff
@@ -0,0 +1,1571 @@
...
+
+    // Read the tag byte
+    pckReadBuffer(this, 1);
+    unsigned int tag = this->bufferPtr[this->bufferPos];
+    this->bufferPos++;
+
+    // If the current container is complete (e.g. object)
+    if (tag == 0)
+    {
+        this->tagNextId = 0xFFFFFFFF;
```
cmwshang, line 389: 0xFFFFFFFF is used 4 times. I'd like to see a constant for it since it would be very easy to type 0xFFFFFFF accidentally.

cmwshang (commented):
partial review

src/common/type/pack.c
```diff
@@ -0,0 +1,1571 @@
...
+        *id = this->tagStackTop->idLast + 1;
+    }
+    // Else check that the id has been incremented
+    else if (*id <= this->tagStackTop->idLast)
+        THROW_FMT(FormatError, "field %u was already read", *id);
+
+    // Search for the requested id
+    do
+    {
+        // Get the next tag if it has been read yet
```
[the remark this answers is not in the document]
dwsteele: That's what I meant, will update.

src/common/type/pack.h
```diff
@@ -0,0 +1,505 @@
...
+Read Functions
+***********************************************************************************************************************************/
+typedef struct PackIdParam
+{
+    VAR_PARAM_HEADER;
+    unsigned int id;
+} PackIdParam;
+
+// Read next field
+bool pckReadNext(PackRead *this);
```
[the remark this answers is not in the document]
dwsteele: pckReadNext() is intended for external use (see hrnPackToStr()) to do generic reads and also to read arrays (see tests).

I can certainly add to the comments here.

src/common/type/pack.c
```diff
@@ -0,0 +1,1571 @@
...
+
+    if (!pckWriteDefaultNull(this, param.defaultNull, value == param.defaultValue))
+        pckWriteTag(this, pckTypeU32, param.id, value);
+
+    FUNCTION_TEST_RETURN(this);
+}
+
+/**********************************************************************************************************************************/
+PackWrite *
+pckWriteU64(PackWrite *this, uint64_t value, PckWriteUInt64Param param)
```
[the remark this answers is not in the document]
dwsteele: There no way to know sure sure that defaultValue is set because it could be set to zero.

But, we could automatically set defaultNull when defaultValue is non 0/NULL/false. So you could set either defaultNull or defaultValue and it would work. Would that be better or would it make things even more confusing?

src/common/type/pack.c
```diff
@@ -0,0 +1,1571 @@
...
+            while (sizeExpected != 0)
+            {
+                size_t sizeRead = pckReadBuffer(this, sizeExpected);
+                sizeExpected -= sizeRead;
+                this->bufferPos += sizeRead;
+            }
+        }
+
+        // Increment the last id to the id just read
+        this->tagStackTop->idLast = this->tagNextId;
```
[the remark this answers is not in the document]
dwsteele: It does if the thing you want to peek at is beyond other tags. So, it doesn't advance the field tag to the thing you are peeking at.

src/common/type/pack.c
dwsteele: Agreed, and constant here would be good.

test/src/module/common/typePackTest.c
```diff
@@ -0,0 +1,315 @@
...
+        TEST_ERROR(pckReadArrayEndP(packRead), FormatError, "not in array");
+        TEST_RESULT_BOOL(pckReadBoolP(packRead, .defaultNull = true), true, "read true");
+        TEST_RESULT_BOOL(pckReadBoolP(packRead), false, "read false");
+        TEST_ERROR(pckReadBoolP(packRead), FormatError, "field 3 does not exist");
+        TEST_RESULT_BOOL(pckReadNullP(packRead), true, "field 3 is null");
+        TEST_RESULT_BOOL(pckReadNullP(packRead, .id = 4), true, "field 4 is null");
+        TEST_RESULT_BOOL(pckReadBoolP(packRead, .defaultNull = true), false, "read default false");
+        TEST_RESULT_VOID(pckReadObjEndP(packRead), "read object end");
+
+        TEST_ERROR(pckReadArrayEndP(packRead), FormatError, "not in array");
```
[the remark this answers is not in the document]
dwsteele: pckReadNext() skips NULLs so this will not work the way you expected. Basically, pckReadNext() is going to read the next thing that *exists*.

src/common/type/pack.c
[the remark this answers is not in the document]
dwsteele: Or we have a separate functions for defaults, e.g. `pckReadU64DefaultP()`.

test/src/module/common/typePackTest.c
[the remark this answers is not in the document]
cmwshang: Hmmm, well that does not explain the behavior since 29 through 36 do not exist so it should have been "skipping" and still doesn't explain the error when reading the non-existent 31.

src/common/type/pack.h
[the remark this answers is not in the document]
cmwshang: If it has limitations, they should be called out in the interface (pack.h) comments so that it is clear to a user how/when to use/not to use this function.

src/common/io/read.h
```diff
@@ -29,6 +29,9 @@ bool ioReadOpen(IoRead *this);
 // Read data from IO and process filters
 size_t ioRead(IoRead *this, Buffer *buffer);
 
+// Same as ioRead() but optimized for small reads
```
cmwshang, line 32: What constitutes a "small read"? As a user, it might be nice to know it is optimized, but I'd, more importantly, want to know how/when to use ioReadSmall vs isRead.

src/common/type/pack.c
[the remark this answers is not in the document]
cmwshang: defaultNull = true means 0/NULL/false, right? So if I say defaultNull = true, why would I need defaultValue? I think it would be much clearer to only set detaultValue if we do not want to store the default when it is something other than that that is classified as defaultNull. I personally don't like defaultValue, but if we're going to use it, I think it would be much clearer to have each of these parameters have a clear and separate purpose (even if we have to set defaultNull behind the scenes).

src/common/type/pack.c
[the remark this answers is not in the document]
cmwshang: I'll fix

src/common/type/pack.h
cmwshang: We're missing the "skipping" part. How about: 
```
By default each read/write advances the field ID by one. If an ID is specified it must be unique and increasing, because it will 
advance the field ID beyond the value specified. An error will occur if an ID is attempted to be read/written but the field ID has 
advanced beyond it.
```

src/common/type/pack.c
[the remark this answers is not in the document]
cmwshang: commit 6e24fbba2aa1b27b04f2c9df41730472e263c6db

src/common/type/pack.h
cmwshang: commit 6e24fbba2aa1b27b04f2c9df41730472e263c6db

test/src/module/common/typePackTest.c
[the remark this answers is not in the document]
dwsteele: So, the main thing is the code doesn't pay attention to the type of the next field (which is an array) and just tries to keep reading integers. The first one works OK because all there are all NULLs up to the array. So the defaultNull works as expected.

But the next pckReadNext() reads the first u64 of the array, which happens to be zero so it looks like the NULL did before. But then the next read is to a u64 that is 1 (the second element of the array) so the code dies.

So the test code here is not right, but there's also no error so it's not at all clear what has gone wrong.  So I've added an error that if the next type is a container and you don't read that container it will error. In other words, you can't just "read past" a container because it's not clear if you want the data in the container or the data after the container.

src/common/type/pack.h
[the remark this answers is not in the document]
dwsteele: Agreed, will do once I can work on the branch again.

dwsteele:
Sure, I'm OK with updating defaultNull behind the scenes.

src/common/type/pack.h
dwsteele: OK.

src/common/io/read.h
dwsteele: That's hard to say for sure -- really depends on the situation. We have lots of "small" reads right now when only a small amount of data is available.

So I guess it's when your intention is make many repetitive reads that are smaller than ioBufferSize().

src/storage/remote/protocol.c
```diff
@@ -92,47 +93,120 @@ storageRemoteFilterGroup(IoFilterGroup *filterGroup, const Variant *filterList)
...
+// Fields that do no change from one call to the next are omitted to save bandwidth.
 static void
-storageRemoteInfoWrite(ProtocolServer *server, const StorageInfo *info)
+storageRemoteInfoWrite(StorageRemoteProtocolInfoListCallbackData *data, const StorageInfo *info)
 {
     FUNCTION_TEST_BEGIN();
-        FUNCTION_TEST_PARAM(PROTOCOL_SERVER, server);
+        FUNCTION_TEST_PARAM_P(VOID, data);
         FUNCTION_TEST_PARAM(STORAGE_INFO, info);
     FUNCTION_TEST_END();
```
cmwshang, line 118: Two things:
1) Can *data or *info be NULL? If not, maybe need ASSERTs?
2) Why is FUNCTION_TEST_PARAM(STORAGE_INFO, info); not with an _P:  FUNCTION_TEST_PARAM_P(STORAGE_INFO, info);

src/storage/remote/protocol.c
```diff
@@ -92,47 +93,120 @@ storageRemoteFilterGroup(IoFilterGroup *filterGroup, const Variant *filterList)
...
 
 static void
-storageRemoteProtocolInfoListCallback(void *server, const StorageInfo *info)
+storageRemoteProtocolInfoListCallback(void *dataVoid, const StorageInfo *info)
 {
     FUNCTION_TEST_BEGIN();
-        FUNCTION_LOG_PARAM(PROTOCOL_SERVER, server);
+        FUNCTION_LOG_PARAM_P(VOID, dataVoid);
         FUNCTION_LOG_PARAM(STORAGE_INFO, info);
     FUNCTION_TEST_END();
```
cmwshang, line 202: Same comment:
1) Can *data or *info be NULL? If not, maybe need ASSERTs?
2) Why is FUNCTION_TEST_PARAM(STORAGE_INFO, info); not with an _P:  FUNCTION_TEST_PARAM_P(STORAGE_INFO, info);

src/storage/remote/storage.c
```diff
@@ -141,30 +194,37 @@ storageRemoteInfoList(
...
+
+                storageRemoteInfoParse(&parseData, &info);
                 callback(callbackData, &info);
 
                 // Reset the memory context occasionally so we don't use too much memory or slow down processing
                 MEM_CONTEXT_TEMP_RESET(1000);
 
-                // Read the next item
-                name = protocolClientReadLine(this->client);
+                pckReadObjEndP(read);
```
cmwshang, line 217: Why is pckReadObjEndP(read);  after  MEM_CONTEXT_TEMP_RESET(1000); instead of before?

cmwshang (commented):
Review complete

src/common/io/read.h
cmwshang: Maybe a comment stating just that would be good.

test/src/module/common/typePackTest.c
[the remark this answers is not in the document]
cmwshang: Fix in 4f1302fd9c67de86eea4b1ed5a76217759eccbc0 looks good.

src/common/type/pack.h
cmwshang: Clarified in commit 0ea578426aab7a582cb8e1bf2b4cfda5205b0d94

src/common/io/read.h
cmwshang: Clarified in commit 0ea578426aab7a582cb8e1bf2b4cfda5205b0d94

src/common/type/pack.c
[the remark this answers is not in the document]
cmwshang: > Sure, I'm OK with updating defaultNull behind the scenes.

After a brief discussion 12/7/2020, defaultNull may become defaultWrite and treating it as writing the default into the pack when set to true.

src/common/type/pack.h
cmwshang: This comment (and maybe others) may need to be modified if it is decided that defaultWrite instead of defaultNull is the chosen parameter to use.

src/storage/remote/protocol.c
```diff
@@ -92,47 +93,120 @@ storageRemoteFilterGroup(IoFilterGroup *filterGroup, const Variant *filterList)
...
+    uid_t userIdLast;                                               // userId from last call
+    gid_t groupIdLast;                                              // groupId from last call
+    String *user;                                                   // user from last call
+    String *group;                                                  // group from last call
+} StorageRemoteProtocolInfoListCallbackData;
+
 // Helper to write storage info into the protocol. This function is not called unless the info exists so no need to write exists or
 // check for level == storageInfoLevelExists.
+//
+// Fields that do no change from one call to the next are omitted to save bandwidth.
```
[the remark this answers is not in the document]
cmwshang: Fixed in 198918b9af844cbb66c4c042357f28efabfc6c39

src/common/type/pack.c
```diff
@@ -0,0 +1,1571 @@
...
+/**********************************************************************************************************************************/
+PackWrite *
+pckWriteBin(PackWrite *this, const Buffer *value, PckWriteBinParam param)
+{
+    FUNCTION_TEST_BEGIN();
+        FUNCTION_TEST_PARAM(PACK_WRITE, this);
+        FUNCTION_TEST_PARAM(BUFFER, value);
+        FUNCTION_TEST_PARAM(UINT, param.id);
+        FUNCTION_TEST_PARAM(BOOL, param.defaultNull);
+    FUNCTION_TEST_END();
```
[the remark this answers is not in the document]
cmwshang: Only assert on THIS is needed. Added in commit 198918b9af844cbb66c4c042357f28efabfc6c39

src/common/type/pack.c
```diff
@@ -0,0 +1,1571 @@
...
+/***********************************************************************************************************************************
+Read next field tag
+***********************************************************************************************************************************/
+static bool
+pckReadTagNext(PackRead *this)
+{
+    FUNCTION_TEST_BEGIN();
+        FUNCTION_TEST_PARAM(PACK_READ, this);
+    FUNCTION_TEST_END();
+
```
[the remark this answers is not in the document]
cmwshang: Added in commit 198918b9af844cbb66c4c042357f28efabfc6c39

src/storage/remote/storage.c
dwsteele: Because if it was before the value just read would be freed. We need it for the next iteration of the loop.

src/storage/remote/protocol.c
dwsteele: 1. Asserts added.
2. Because STORAGE_INFO is a defined logging type (see storage/info.h) that happens to be a pointer but we don't need to know that here. void * is not a define logging type so we need to specify _P to let the macro know it is a pointer.

src/storage/remote/protocol.c
dwsteele: 1. Asserts added.
2. Because STORAGE_INFO is a defined logging type (see storage/info.h) that happens to be a pointer but we don't need to know that here. void * is not a define logging type so we need to specify _P to let the macro know it is a pointer.

src/common/type/pack.c
```diff
@@ -0,0 +1,1571 @@
...
+    FUNCTION_TEST_RETURN(this);
+}
+
+/**********************************************************************************************************************************/
+String *
+pckWriteToLog(const PackWrite *this)
+{
+    return strNewFmt(
+        "{depth: %u, idLast: %u}", this->tagStackTop == NULL ? 0 : lstSize(this->tagStack),
```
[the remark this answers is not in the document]
dwsteele: Hmmm, not quite sure what you mean here. `this->tagStack` can never be NULL since it is initialized in the constructor.

src/common/type/pack.h
dwsteele: Updated in 522ae2eb3dfe332.

src/common/type/pack.h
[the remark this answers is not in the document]
dwsteele: Added in fa5bff95a5d5.

src/common/type/pack.c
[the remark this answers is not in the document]
dwsteele: Done in 522ae2eb3dfe.

src/common/type/pack.c
dwsteele: Fixed in fa5bff95a5d51.

src/common/type/pack.c
```diff
@@ -0,0 +1,1571 @@
...
+
+    ASSERT(this != NULL);
+
+    uint64_t result = 0;
+    uint8_t byte;
+
+    for (unsigned int bufferIdx = 0; bufferIdx < PACK_UINT64_SIZE_MAX; bufferIdx++)
+    {
+        pckReadBuffer(this, 1);
+        byte = this->bufferPtr[this->bufferPos];
```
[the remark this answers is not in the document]
dwsteele: Add in fa5bff95a5d51.

src/common/type/pack.c
```diff
@@ -0,0 +1,1571 @@
...
+    }
+    MEM_CONTEXT_END();
+
+    FUNCTION_TEST_RETURN(this);
+}
+
+/***********************************************************************************************************************************
+Write to io or buffer
+***********************************************************************************************************************************/
+static void pckWriteBuffer(PackWrite *this, const Buffer *buffer)
```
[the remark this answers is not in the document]
dwsteele: bufferPos was removed, but can't seem to find the commit where I did it. Or I'm not understanding what you mean here.

src/common/type/pack.c
```diff
@@ -0,0 +1,1571 @@
...
+
+    // Reset tagNextId to keep reading
+    this->tagNextId = 0;
+
+    FUNCTION_TEST_RETURN_VOID();
+}
+
+/**********************************************************************************************************************************/
+Buffer *
+pckReadBin(PackRead *this, PckReadBinParam param)
```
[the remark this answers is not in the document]
dwsteele: Added many more comments in fa5bff95a5d51.

src/common/type/pack.c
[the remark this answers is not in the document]
cmwshang: Right but is there any possibility that this->tagStackTop == NULL but this->tagStack has a size? (or visa versa). I mean, why not always just write out lstSize(this->tagStack) - it's always going to be >= 0 and then also write out tagStackTop as a separate thing...

src/common/type/pack.c
```diff
@@ -0,0 +1,1611 @@
...
+
+    // Search for the requested id
+    do
+    {
+        // Get the next tag if it has not been read yet
+        if (this->tagNextId == 0)
+            pckReadTagNext(this);
+
+        // Return if the id does not exist
+        if (*id < this->tagNextId)
```
cmwshang, line 507: I'm a little confused here - before, if the ID did not exist then we'd error if they weren't just trying to "peek" at it. So now that we're breaking out of the loop if it doesn't exist, then what are we returning with this->tagNextValue?

cmwshang (commented):
Just 2 open comments - everything else looks good.

src/common/type/pack.c
dwsteele: This is because we no longer error if you try to read something and it does not exist (when defaultNull = false before). Now we just always default it so there's no way to get an error.

this->tagNextValue will be the value of the next existing tag -- if we get to this condition it means the thing we are looking for is before it and therefore NULL.

src/common/type/pack.c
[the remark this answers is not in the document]
dwsteele: Works for me. Done in ef50938f967df6699.

PR #1196, opened 2020-10-12
Expire history files.

pgstef:
Remove history files that are prior to minimum WAL timeline after expiration

card link: https://github.com/pgbackrest/pgbackrest/projects/2#card-46191249

### DIFF
--- a/src/command/expire/expire.c
+++ b/src/command/expire/expire.c
@@ -734,7 +734,7 @@ removeExpiredArchive(InfoBackup *infoBackup, bool timeBasedFullRetention)
                                     storageListP(
                                         storageRepo(),
                                         strNewFmt(STORAGE_REPO_ARCHIVE "/%s", strZ(archiveId)),
-                                        .expression = STRDEF("^[0-F]{8}.history$")),
+                                        .expression = STRDEF(HISTORY_FILES_REGEXP)),
                                     sortOrderAsc);
 
                             for (unsigned int historyFileIdx = 0; historyFileIdx < strLstSize(historyFilesList); historyFileIdx++)
--- a/src/command/expire/expire.c
+++ b/src/command/expire/expire.c
@@ -727,6 +727,37 @@ removeExpiredArchive(InfoBackup *infoBackup, bool timeBasedFullRetention)
                             // Log if there is more to log
                             else
                                 logExpire(&archiveExpire, archiveId);
+
+                            // Based on the timeline of the backupArchiveStart, look for history files to expire
+                            StringList *historyFilesList =
+                                strLstSort(
+                                    storageListP(
+                                        storageRepo(),
+                                        strNewFmt(STORAGE_REPO_ARCHIVE "/%s", strZ(archiveId)),
+                                        .expression = STRDEF("^[0-F]{8}.history$")),
+                                    sortOrderAsc);
+
+                            for (unsigned int historyFileIdx = 0; historyFileIdx < strLstSize(historyFilesList); historyFileIdx++)
+                            {
+                                String *historyFile = strLstGet(historyFilesList, historyFileIdx);
+
+                                // Extract timeline to compare
+                                if (strCmp(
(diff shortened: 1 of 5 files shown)


src/command/expire/expire.c
```diff
@@ -727,6 +727,37 @@ removeExpiredArchive(InfoBackup *infoBackup, bool timeBasedFullRetention)
...
                             else
                                 logExpire(&archiveExpire, archiveId);
+
+                            // Based on the timeline of the backupArchiveStart, look for history files to expire
+                            StringList *historyFilesList =
+                                strLstSort(
+                                    storageListP(
+                                        storageRepo(),
+                                        strNewFmt(STORAGE_REPO_ARCHIVE "/%s", strZ(archiveId)),
+                                        .expression = STRDEF("^[0-F]{8}.history$")),
```
dwsteele, line 737: This seems like a useful constant that should be added to `src/command/archive/common.h`.

dwsteele (changes requested):
This looks pretty straightforward.

One thing -- I wonder if we should keep the history file for the oldest timeline in the archive dir (when it exists, e.g. there won't be a history file for timeline 1). I know this is not useful for recovery but I think it might be useful for forensics.

So, in your test we'd keep `00000002.history`. That's means you'd need to bump up the timeline in your test to 3 and create a history file for both timelines and make sure only 2 is removed.

What do you think?

src/command/expire/expire.c
pgstef: Good point. Moved it (for the next commit)

pgstef:
If `backup-archive-start` would have been `0000000400000000000000XX`, you would keep `00000003.history` and remove `00000002.history` then ?

I don't really think this would be handy. If we decide to remove, we should remove it all. Or it will lead to some confusion for users I think. We could also imagine keeping it in some `archive.history` or `timelines.history` directory like the `backup.history` perhaps. 

My vote would be to remove it all but I can change to "keep -1" if you find it better ;-)

dwsteele:
I meant If backup-archive-start was 0000000400000000000000XX you would keep *00000004.history* and remove 00000003.history (and below).

dwsteele:
Which actually seems to be what you are doing, so my bad...

dwsteele:
But to make this clearer can you add a `00000003.history` file in your test and check that it does *not* get removed?

pgstef:
Okay. I'll move "20181119-152900F_20181119-152500I" to timeline4 and test if 3.history still exists ;-)

dwsteele:
We don't need to move 20181119-152900F_20181119-152500I to timeline 4. Just create 3.history and make sure it still exists after expiration. So, the code is fine, I just want to add a test to be sure 3.history is preserved.

pgstef:
Test added :-)

I'm not sure why the coverage isn't complete though:
```
P00  ERROR: [125]: c module command/expire/expire is not fully covered (209/210 branches)
...
749 | [+ -][+ +] | if (!cfgOptionValid(cfgOptDryRun) \|\| !cfgOptionBool(cfgOptDryRun))
```

Thanks for your review!

dwsteele:
> I'm not sure why the coverage isn't complete though

This is because you don't have a test where the dry-run option is not valid, i.e. backup. So you just need to repeat the test at 1157 using `cfgCmdBackup` instead of `cfgCmdExpire`.

No need to recreate the 2.history file for this test -- I think it is fine for it to do nothing. I'd still check that 3.history exists, though.

pgstef:
Fixed :-)
Thanks !

dwsteele (approved):
Update some naming, comments, and formatting. Also fixed the string constant that was declared but not defined.

Added detail logging to the tests. We are planning to make detail the default level for testing so this saves some changes later.

This looks ready to me if you agree with the changes.

pgstef:
It looks better indeed, thanks !

Just a small remark. In `expireTest.c`, `harnessLogLevelSet(logLevelDetail);` is defined twice without reset. 
I've removed the extra line and pushed the clean-up.

dwsteele:
> harnessLogLevelSet(logLevelDetail); is defined twice without reset.

Good catch, thanks!

Issue #1199, opened 2020-10-17
INFO command is empty


uspen:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version: 2.30-1.rhel7

2. PostgreSQL version: 9.6

3. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each: CentOS 7

4. Did you install pgBackRest from source or from a package? rpm from pgdg-common

5. Please attach the following as applicable:
    - `pgbackrest.conf` file(s)
    - `postgresql.conf` settings applicable to pgBackRest (`archive_command`, `archive_mode`, `listen_addresses`, `max_wal_senders`, `wal_level`, `port`)
    - errors in the postgresql log file before or during the time you experienced the issue
    - log file in `/var/log/pgbackrest` for the commands run (e.g. `/var/log/pgbackrest/mystanza_backup.log`)

7. Describe the issue:

If pgbackrest backup dir is empty, INFO command empty too

$ pgbackrest --config-path=/etc/pgbackrest --stanza=fake --output=json info
[]

$ touch /data/pgbackrest/backup/foo

$ pgbackrest --config-path=/etc/pgbackrest --stanza=fake --output=json info
[{"backup":[],"db":[],"name":"fake","status":{"code":1,"lock":{"backup":{"held":false}},"message":"missing stanza path"}}]
[pgbackrest-info.txt](https://github.com/pgbackrest/pgbackrest/files/5395323/pgbackrest-info.txt)

pgstef:
Hi,

This seems normal. Have you run `pgbackrest --stanza=fake stanza-create` ? This will initiate the repo, manifests, sub-directories,...

uspen:
no, if stanza not exist, INFO must be:
[{"backup":[],"db":[],"name":"fake","status":{"code":1,"lock":{"backup":{"held":false}},"message":"missing stanza path"}}]

uspen:
But if I make other stanza or any file, fake stanza has a good answer (missing stanza path)

pgstef:
Blank json output is actually `No stanzas exist in the repository.`: https://github.com/pgbackrest/pgbackrest/blob/master/src/command/info/info.c#L850.

```
$ pgbackrest info --stanza=my_stanza
No stanzas exist in the repository.
$ pgbackrest info --stanza=my_stanza --output=json
[]
```

[...]

PR #1208, opened 2020-10-21
Improve info command output when a stanza is specified but missing.

pgstef:
Idea based on issue https://github.com/pgbackrest/pgbackrest/issues/1199.

Output would give:

```
$ pgbackrest_build info --output=json
[]

$ pgbackrest_build info --output=json --stanza=foo
[{"backup":[],"db":[],"name":"foo","status":{"code":1,"lock":{"backup":{"held":false}},"message":"missing stanza path"}}]

$ pgbackrest_build info 
No stanzas exist in the repository.

$ pgbackrest_build info --stanza=foo
stanza: foo
    status: error (missing stanza path)
```

It's a pretty minor code change and I think it doesn't break the actual json format used... This can be a better output than the empty one.


test/src/module/command/infoTest.c
```diff
@@ -42,6 +42,37 @@ testRun(void)
...
+                    "\"status\":{"
+                        "\"code\":1,"
+                        "\"lock\":{\"backup\":{\"held\":false}},"
+                        "\"message\":\"missing stanza path\""
+                        "}"
+                "}"
+            "]",
+            "json - empty repo, stanza option specified");
+
+
```
cmwshang, line 66: Extra empty line not needed.

cmwshang (commented):
Overall the change looks good. I am just concerned that anyone out there relying on the empty format for their tests or monitoring might not be happy. Has anyone else complained? We might want to solicit a bit more feedback...

test/src/module/command/infoTest.c
pgstef: removed, thanks

pgstef:
Currently I think that's the only one who reported an issue. I would find a bit strange to rely on "empty" for monitoring purpose though.

This new behaviour seems more understandable at first sight:
- asking for a specific stanza, not there -> error
- asking for the complete info, nothing there -> empty

From a monitoring point of view, I find it better to rely on error codes rather than on empty return (which could mean anything).
Regarding tests, if the question is "does my stanza exist? if not, create it", I think it's better to use an error code rather then an empty output. Although, the text output isn't empty.

Solicit feedback might make sense but how can we actually do that ?

cmwshang:
> From a monitoring point of view, I find it better to rely on error codes rather than on empty return (which could mean anything).

Good point.

cmwshang (approved):

dwsteele (approved):

PR #1210, opened 2020-10-21
Allow [, #, and space as the first character in database names.

dwsteele:
iniLoad() was trimming lines which meant that a leading space would not pass checksum validation when a manifest was reloaded. Remove the trims since files we write should never contain extraneous spaces. This further diverges the format for the functions that read conf files (e.g. pgbackrest.conf) and those that read info (e.g. manifest) files.

While we are at it also allow [ and # as initial characters. # was reserved for comments but we never put comments into info files. [ denotes a section but we can get around this by never allowing arrays as values in info files, so if a line ends in ] it must be a section. This is currently the case but enforce it by adding an assert to info/info.c.


pgstef:
Looks good to me.
I've made a few tests with this branch, trying to mess around with the .conf (comments, spaces,..), and db names.

Maybe we could add some "visual" surrounding in the backup set info output: `database list:  test (16407), #test (16430)..`. It's not currently very clear that the dbname starts with a space :-)

But that's just a cosmetic detail. Otherwise this patch seems fine.

dwsteele:
> Maybe we could add some "visual" surrounding in the backup set info output

Hmm, maybe, but I'm not sure how we'd detect all possible cases. There are plenty of unicode characters that would be hard to detect as well. If people put weird characters in their database names I'm not sure how much we need to cater to them. Other than making work, of course.

pgstef (approved):

pgstef:
Indeed, and add extra characters to surround the names would not be very clever. 
Anyway, the PR is good for me. I just forgot to approve :-(

src/common/ini.c
```diff
@@ -358,27 +358,23 @@ iniLoad(
 
             do
```
cmwshang, line 359: Not sure where to leave this comment - but I am glad to see that the reading of pgbackrest.conf (and thus iniParse) is not affected since I have seen people put comments in the conf file. However, I might suggest in the ini.h file that it is made clear that iniLoad should never be used for the conf files - only iniParse.

cmwshang (commented):
I tried to break it with different DB, File and Dir names, but wasn't able to, so good job! However, I'm not a fan of allowing # because that limits our ability (or anyone else's) to put in comments and I did not see it as an ask to do so.

src/common/ini.c
dwsteele: Yeah, these are separate code paths and now they really need to stay that way. Added some comments to help clarify what to use for what.

I added support for # and [ since I was in there and it was clear that they would also not work as expected. Trying to save us an issue down the road.

cmwshang (approved):

Issue #1251, opened 2020-12-03
Allow to set the config directory via autoconf


mbanck:
The default config file seems to be hardcoded to `/etc/{pgbackrest/,}pgbackrest.conf` (not sure whether the `pgbackrest` directory is needed, I think both work?); however, for sites that install pgbackrest locally the usually used `postres` user will not have the permission to create `/etc/pgbackrest` or `/etc/pgbackrest.conf`. So it would be nice if one could set that via e.g. `--sysconfdir=/foo/bar/pgbackrest`. Well, one can set that, but it's not being used by pgBackRest. Another option would be to optionally check `$PREFIX/etc/[pgbackrest]/pgbackrest.conf` as configuration file location.

As an aside, `--log-level-console=debug` is not helpful at all about where/how pgBackRest tries to find/read its configuration file, all I see is
```
2020-12-03 13:52:23.633 P00  DEBUG:     config/load::cfgLoad: => void
```

cmwshang:
/etc/pgbackrest/pgbackrest.conf is checked first and if it does not exist then /etc/pgbackrest.conf is used. This was done to provide a smooth transition from 1.x to 2.x. For more custom configurations see the Command page https://pgbackrest.org/command.html an excerpt is below but there are also options for overriding per PG or repo host (see Configuration page https://pgbackrest.org/configuration.html (example: https://pgbackrest.org/configuration.html#section-repository/option-repo-host-config))

**4.2.7 Config Option (--config)**
pgBackRest configuration file.
Use this option to specify a different configuration file than the default.
```
default: /etc/pgbackrest/pgbackrest.conf
example: --config=/conf/pgbackrest/pgbackrest.conf
```

**4.2.8 Config Include Path Option (--config-include-path)**
Path to additional pgBackRest configuration files.
Configuration files existing in the specified location with extension .conf will be concatenated with the pgBackRest configuration file, resulting in one configuration file.
```
default: /etc/pgbackrest/conf.d
example: --config-include-path=/conf/pgbackrest/conf.d

```
**4.2.9 Config Path Option (--config-path)**
Base path of pgBackRest configuration files.
This setting is used to override the default base path setting for the --config and --config-include-path options unless they are explicitly set on the command-line.

For example, passing only --config-path=/conf/pgbackrest results in the --config default being set to /conf/pgbackrest/pgbackrest.conf and the --config-include-path default being set to /conf/pgbackrest/conf.d.
```
default: /etc/pgbackrest
[...]

PR #1255, opened 2020-12-07
Allow custom config-path default with ./configure --with-configdir.

mschout:
This adds --with-confdir=DIR option to configure, which can be used to
override the deafult configuration directory of /etc/pgbackrest

Probably in the future it would be better to just leverage ${sysconfdir}
which is based on prefix, but since previously the config dir was hard
coded to /etc/pgbacrest, we retain that default value by not relying on
sysconfdir for now.

refs #1251


dwsteele:
This looks reasonable. There are two things to consider.

1) The code will still look at `/etc/pgbackrest.conf` if the config file does not exist in the specified location. Changing the code to ignore this file means also writing new test cases, which I'm not very excited about for a sandbox feature.
2) You have manually modified an auto-generated file so the changes actually need to be made in `build/lib/pgBackRestBuild/Config/BuildDefine.pm`. But it'd be easier for me to make that change since I know exactly how.

@mbanck Does that look reasonable to you, esp #1?

mbanck:
I've tested the PR now and it works fine for me, thanks!

dwsteele:
OK, I finally got to updating code generation so defaults can be based on C constants. That makes this commit a lot smaller.

Can you `git pull https://github.com/pgbackrest/pgbackrest.git autoconf-configdir` with the updates? Github won't let me push this directly because there are updates in master to our github workflows.

mschout:
Thanks!  I'll take a look at this over the weekend.

mschout:
Looking at this now....

mschout:
I think this one is ready to merge. Let me know if it needs anything else.

dwsteele:
Looks good, thank you!

PR #1275, opened 2021-01-13
Update info command to allow for multiple repositories.

cmwshang:
For multiple repositories, the info command has been enhanced to search over all repositories when a backup set is not requested. If a backup set is requested, the --repo option is required if more than one repository is configured. 

The json output from the command will now include information about each repo, including the status of the stanza on the repo. The text output has been updated to include the repo key to indicate where each backup is located. 

A new status code of 'mixed' has been introduced for the overall stanza status and overall cipher type when they differ across repos. For the text output, when mixed is displayed, the status/cipher type will include additional lines indicating the status/cipher type for each repo.


src/command/info/info.c
```diff
@@ -140,10 +287,11 @@ archiveDbList(const String *stanza, const InfoPgData *pgData, VariantList *archi
     String *archiveStart = NULL;
     String *archiveStop = NULL;
     Variant *archiveInfo = varNewKv(kvNew());
+    const Storage *storageRepo = storageRepoIdx(repoIdx);
```
dwsteele, line 290: Not sure we need to create these storageRepo variables. I think using storageRepoIdx(repoIdx) in place is clearer and there aren't so many storage calls that it will make any difference performance-wise.

src/command/info/info.c
```diff
@@ -770,34 +1160,116 @@ infoRender(void)
...
+        // If the repo was specified then set index to the array location and max to loop only once
+        if (cfgOptionTest(cfgOptRepo))
+        {
+            repoIdxStart = cfgOptionGroupIdxDefault(cfgOptGrpRepo);
+            repoIdxMax = repoIdxStart + 1;
+        }
+
+        for (unsigned int repoIdx = repoIdxStart; repoIdx < repoIdxMax; repoIdx++)
+        {
+            // Get the repo storage in case it is remote and encryption settings need to be pulled down (performed here for testing)
```
dwsteele, line 1195: I'm not sure what "performed here for testing" means.

src/command/info/info.c
```diff
@@ -370,388 +591,557 @@ backupList(VariantList *backupSection, InfoBackup *info, const String *backupLab
...
-        stanzaStatus(
-            INFO_STANZA_STATUS_CODE_MISSING_STANZA_PATH, INFO_STANZA_STATUS_MESSAGE_MISSING_STANZA_PATH_STR, false, stanzaInfo);
-        varLstAdd(result, stanzaInfo);
+        char timeBufferStart[20];
+        char timeBufferStop[20];
+        time_t timeStart = (time_t)varUInt64(kvGet(timestampInfo, KEY_START_VAR));
+        time_t timeStop = (time_t)varUInt64(kvGet(timestampInfo, KEY_STOP_VAR));
+
+        strftime(timeBufferStart, sizeof(timeBufferStart), "%Y-%m-%d %H:%M:%S", localtime(&timeStart));
+        strftime(timeBufferStop, sizeof(timeBufferStop), "%Y-%m-%d %H:%M:%S", localtime(&timeStop));
```
dwsteele, line 786: Perhaps a helper function would make this less repetitive?

src/command/info/info.c
```diff
@@ -81,30 +85,121 @@ STRING_STATIC(INFO_STANZA_STATUS_MESSAGE_MISSING_STANZA_PATH_STR,   "missing sta
...
+
+#define FUNCTION_LOG_INFO_STANZA_REPO_TYPE                                                                                         \
+    InfoStanzaRepo *
+#define FUNCTION_LOG_INFO_STANZA_REPO_FORMAT(value, buffer, bufferSize)                                                            \
+    objToLog(value, "InfoStanzaRepo", buffer, bufferSize)
+
+// Group all databases with the same system-id and version together regardless of db-id or repo
+typedef struct DbGroup
+{
+    uint64_t systemId;
```
dwsteele, line 133: Comments for these struct members?

dwsteele (changes requested):
Any other plans for the user-guide other than the update to the info command description?

I was thinking that when we add new repos in the S3 and Azure sections maybe we could do an info and limit it to just that repo.

dwsteele (changes requested):
Review complete.

src/command/info/info.c
cmwshang: fixed 8af20b6e1cacc2b4b68cbe5a4b7ab891ffa48768

src/command/info/info.c
cmwshang: fixed 8af20b6e1cacc2b4b68cbe5a4b7ab891ffa48768

dwsteele (approved):

PR #1285, opened 2021-01-27
Update expire command log messages with repo prefix.

cmwshang:
In preparation for multi-repo support, a repo tag is added in this commit to the expire command log and error messages. This change also affects the expect logs and the user-guide. The format of the tag is "repoX:" where X is the repo key used in the configuration. Until multi-repo support has been completed, this tag will always be "repo1:".


dwsteele (approved):
Looks good, just removed one added linefeed.

PR #1289, opened 2021-02-03
Enhance expire command multi-repo support.

cmwshang:
The expire command has been enhanced to expire backups and archives from all configured repositories by default. In addition, it will accept the --repo option to expire backups and archives only from the specified repository. Using the --repo options the --set option can also be refined further to the specified repo. If --set is provided but the --repo option has not, then all repositories will be searched and retention settings will be applied on each whether the backup set has been found or not.


dwsteele (approved):
Looks good. Made minor changes to repo min/max logic and some formatting/comments.

PR #1300, opened 2021-02-16
Enhance restore command multi-repo support.

cmwshang:
The restore command automatically defaults to selecting the latest backup from the single repository. With multiple repositories configured, the restore command will now default to selecting the latest backup from the first repository where backups exist. The order in which the repositories are checked is dictated by the pgbackrest.conf. 

To select from a specific repository, the --repo option can be passed (e.g. --repo=1). The --set option can be passed if a backup other than the latest is desired.


src/command/restore/restore.c
```diff
@@ -247,52 +327,80 @@ restoreBackupSet(InfoBackup *infoBackup)
...
+                        break;
+                    else
                     {
-                        LOG_WARN_FMT(
-                            "unable to find backup set with stop time less than '%s', latest backup set will be used",
-                            strZ(cfgOptionStr(cfgOptTarget)));
-                    }
-                }
-            }
+                        InfoBackupData latestBackup = infoBackupData(infoBackup, infoBackupDataTotal(infoBackup) - 1);
```
dwsteele, line 344: We get this data twice, here and line 356. Perhaps we should just move this up to 316? I'm not worried about the performance hit but de-duping the code might be nice.

dwsteele (changes requested):
Looks good except for the one suggestion, which is totally up to you.

Perhaps we should wait to merge this until archive-get is done and we can update the real tests?

cmwshang:
> Perhaps we should wait to merge this until archive-get is done and we can update the real tests?

Agreed. Let's get everything ready and then take a hard look at the Real Tests.

src/command/restore/restore.c
cmwshang: Fixed f20b7b7d8031a0615a645f57aec223454c7fbc19

dwsteele (approved):
As discussed, we'll look at the real tests in a separate commit.

PR #1301, opened 2021-02-16
Fix option warnings breaking async archive-get/archive-push.

dwsteele:
Option warnings will cause the async process to fail because a warning is logged but stdout is closed so the process aborts.

This bug has existed for quite some time, but it was made worse by abb8ebe because now the async role can have different valid options than the default role. Previously at least a warning would be emitted before the async process died.

Fix this by only allowing warnings for the default role. Warnings were already suppressed for local and remote roles so the logic already exists.

### DIFF
--- a/test/src/module/config/parseTest.c
+++ b/test/src/module/config/parseTest.c
@@ -776,8 +776,9 @@ testRun(void)
             configParse(storageTest, strLstSize(argList), strLstPtr(argList), false), OptionInvalidValueError,
             "'/path1/path2//' cannot contain // for 'pg1-path' option");
 
-        // Local and remove commands should not modify log levels during parsing
         // -------------------------------------------------------------------------------------------------------------------------
+        TEST_TITLE("only reset log levels for default role");
+
         argList = strLstNew();
         strLstAdd(argList, strNew("pgbackrest"));
         hrnCfgArgRawZ(argList, cfgOptPg, "2");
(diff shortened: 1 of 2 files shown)


test/src/module/config/parseTest.c
```diff
@@ -816,6 +816,21 @@ testRun(void)
         TEST_RESULT_STR_Z(cfgCommandRoleStr(cfgCmdRoleRemote), "remote", "    remote role name");
         TEST_RESULT_INT(logLevelStdOut, logLevelError, "console logging is error");
         TEST_RESULT_INT(logLevelStdErr, logLevelError, "stderr logging is error");
+
+        argList = strLstNew();
+        strLstAdd(argList, strNew("pgbackrest"));
+        hrnCfgArgRawZ(argList, cfgOptPgPath, "/path/to");
+        strLstAdd(argList, strNew("--stanza=db"));
+        strLstAdd(argList, strNew("--log-level-stderr=info"));
```
cmwshang, line 824: I tried this same test with archive-get because I did not see a test where the default proves it is reset. In any event, it does reset the stderr to a warning vs the requested info - and although I can't find it - I assume there must be somewhere in the config that we restore the logging to the levels requested after we're done parsing the config?

cmwshang (commented):
Review complete with a question.

test/src/module/config/parseTest.c
cmwshang: Per discussion with David, the settings for log-levels requested by the user, will get reapplied after the parsing, by calling cfgLoadLogSetting(). And the tests that prove the logging is set to warning then reset are partly here but mostly in the load tests.

cmwshang (approved):

PR #1308, opened 2021-02-22
Enhance expire to continue processing when a repo fails.

cmwshang:
When the expire command fails on a single repo it will continue on to the next one rather than stopping.

Errors are logged to the log file rather than thrown. If, after processing all repos, one or more errors occurred, then it will throw a single error to indicate there were errors and the log file should be inspected.


src/command/expire/expire.c
```diff
@@ -900,68 +899,103 @@ cmdExpire(void)
...
+            CATCH_ANY()
+            {
+                LOG_ERROR_FMT(
+                    errorTypeCode(errorType()), "repo%u: %s", cfgOptionGroupIdxToKey(cfgOptGrpRepo, repoIdx), errorMessage());
+                repoError = true;
+                errorTotal++;
+            }
+            TRY_END();
+
+            if (!repoError)
```
dwsteele, line 968: I can't figure out why expire is broken into two try blocks. Why not just run everything in the first try block?

src/command/expire/expire.c
```diff
@@ -900,68 +899,103 @@ cmdExpire(void)
                 THROW_FMT(OptionInvalidValueError, "'%s' is not a valid backup label format", strZ(adhocBackupLabel));
         }
 
+        // Track any errors that may occur
+        unsigned int errorTotal = 0;
```
dwsteele, line 903: Seems like a simple boolean would do the job.

dwsteele (changes requested):

src/command/expire/expire.c
cmwshang: Fixed https://github.com/pgbackrest/pgbackrest/pull/1308/commits/54e3bf08da92f1c157dc6d3f79f80439e4f35265

src/command/expire/expire.c
cmwshang: Kept as count and added it to final THROW. Commit https://github.com/pgbackrest/pgbackrest/pull/1308/commits/54e3bf08da92f1c157dc6d3f79f80439e4f35265

dwsteele (approved):

PR #1313, opened 2021-02-26
Dockerfile - To run everywhere

mohsinhijazee:
As many of the libraries are dynamically linked, it might not be possible to run a binary on system where there's no distribution's package manager is bundling everything together. In such scenarios, this docker file would be helpful and would make this utility accessible on a wider range of systems.

I tried to statically link the whole binary and use the docker `scratch` image to reduce the size but compilation failed. Moving the binary alone into an empty image also fails because there are dependencies on external libraries.

To build, from root of the git repo:
```
docker build -t pgbackrest .
```
And to run:
```
docker run -it --rm pgbackrest
```
Or you can make an alias like so:

```
alias pgbackrest='docker run -it --rm pgrest'

# Now it just works as if a normal binary:
pgbackrest version
```
Hope this is useful. Looking forward to feedback/adjustments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM ubuntu:18.04
+
+RUN    apt-get update && \
+       apt-get upgrade -y && \
+       apt-get install -y rsync git devscripts build-essential valgrind lcov autoconf \
+       autoconf-archive libssl-dev zlib1g-dev libxml2-dev libpq-dev pkg-config \
+       libxml-checker-perl libyaml-perl libdbd-pg-perl liblz4-dev liblz4-tool \
+       zstd libzstd-dev bzip2 libbz2-dev
+
+WORKDIR /pgbackrest
+
+ADD . /pgbackrest
+
+RUN cd /pgbackrest/src && \
+       ./configure && \
+       make && \
+       mv /pgbackrest/src/pgbackrest /usr/bin/pgbackrest && \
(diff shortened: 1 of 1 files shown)


dwsteele:
Hi! This looks pretty good but we don't include any packages or containers in this project. If we have one then we'll get requests for me and we simply don't have time to keep up with that. You are welcome to upload this to Dockerhub yourself, of course.

My main comment is that it would be better to have one container to build and another to deploy. That makes the deployed container leaner and meaner.

mohsinhijazee:
Hey, thank you for your response. I think the docker would be useful and I can provide maintenance if you want help in that.
The other point that you mentioned I think is very important - separating the build and runtime dependencies. I guess the best way would be to copy all the dynamic libraries along side the binary into a new image. I will see if I can get that far.

dwsteele:
> I think the docker would be useful and I can provide maintenance if you want help in that.

Understood, but we are not going to commit this to the project. We have no way to automatically test it and we are not expert enough in Docker to properly review it.

You can host this container yourself on Github and Dockerhub. There are already several others out there.

> I guess the best way would be to copy all the dynamic libraries along side the binary into a new image.

I'd say you'd just want install the packages that are needed in the deployed container, e.g. `apt-get install libpq5 liblz4 ...`.

mohsinhijazee:
Okay, thank you. I'll do it that way. Closing.

PR #1315, opened 2021-02-26
Repo option is optional when performing a backup.

cmwshang:
When performing a backup with multiple repositories configured, the highest priority repository will be chosen (i.e. repo1) if the --repo option is not specified.


dwsteele (approved):
Looks good!

PR #1316, opened 2021-02-26
Add write fault-tolerance to archive-push command.

dwsteele:
The archive-push command will continue to push even after it gets a write error on one or more repos. The idea is to archive to as many repos as possible even we still need to throw an error to PostgreSQL to prevent it from removing the WAL file.

Also refactor the warnings to work like archive-get, i.e. they are returned as an array so they can be logged separately.


test/src/module/command/archivePushTest.c
```diff
@@ -851,7 +876,7 @@ testRun(void)
             "P00   INFO: push 1 WAL file(s) to archive: 000000010000000100000002\n"
             "P01   WARN: WAL file '000000010000000100000002' already exists in the repo1 archive with the same checksum\n"
             "            HINT: this is valid in some recovery scenarios but may also indicate a problem.\n"
-            "            WAL file '000000010000000100000002' already exists in the repo3 archive with the same checksum\n"
+            "P01   WARN: WAL file '000000010000000100000002' already exists in the repo3 archive with the same checksum\n"
             "            HINT: this is valid in some recovery scenarios but may also indicate a problem.\n"
             "P01 DETAIL: pushed WAL file '000000010000000100000002' to the archive");
```
cmwshang, line 881: This isn't new, but I was curious as to why this is a DETAIL level
`"P01 DETAIL: pushed WAL file '000000010000000100000002' to the archive"`
but on line 575 it is an INFO level:
`"P00   INFO: pushed WAL file '000000010000000100000002' to the archive"`

Is it because the Async process already spits out 
`P00   INFO: push 1 WAL file(s) to archive: 000000010000000100000002`

test/src/module/command/archivePushTest.c
```diff
@@ -521,11 +521,36 @@ testRun(void)
...
+            storageTest, strNewFmt("repo3/archive/test/11-1/0000000100000001/000000010000000100000002-%s", walBuffer2Sha1),
+            .errorOnMissing = true);
+
+        HRN_STORAGE_MODE(storageTest, "repo2/archive/test/11-1/0000000100000001", .mode = 0500);
+
+        TEST_ERROR(
+            cmdArchivePush(), CommandError,
+            "archive-push command encountered error(s):\n"
+            "repo2: [FileOpenError] unable to open file '/home/vagrant/test/test-0/repo2/archive/test/11-1/0000000100000001"
+                "/000000010000000100000002-755defa48a0a0872767b6dea49bdd3b64902f147' for write: [13] Permission denied");
```
cmwshang, line 539: So, at this point, even though one repo succeeded, we don't indicate that because there was a hard error on the other repo?

cmwshang (commented):

src/command/archive/push/push.c
```diff
@@ -377,13 +377,13 @@ cmdArchivePush(void)
                 ArchivePushCheckResult archiveInfo = archivePushCheck(cfgOptionTest(cfgOptPgPath));
```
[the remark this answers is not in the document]
dwsteele: Hmm, good point. I'll put that in as a card and we can decide what to do about it on Monday.

Hardening that part of archive-get was really difficult so maybe we are out of time for that in this release.

test/src/module/command/archivePushTest.c
dwsteele: It is detail level in the async log to cut down on logging. It is info level in the foreground log because it's a lot of logging anyway, and info level is usually disabled for the foreground.

test/src/module/command/archivePushTest.c
dwsteele: Yes -- any repos not in this list worked, but I think it add clutter to report that.

src/command/archive/push/push.c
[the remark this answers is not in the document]
cmwshang: Per our discussion, this PR will be committed and a note will be made in the card on the project board to address the need to update archivePushCheck().

cmwshang (approved):

src/command/archive/push/push.c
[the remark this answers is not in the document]
dwsteele: Note has been made.

PR #1318, opened 2021-03-01
Add HttpUrl object.

dwsteele:
Parse a URL into component parts.


src/common/io/http/url.c
```diff
@@ -0,0 +1,213 @@
...
+    "(:[1-9][0-9]{0,4}){0,1}"                                       // Optional port
+    "(\\/[^?\\/]*)*$");                                             // Optional path
+
+/***********************************************************************************************************************************
+Object type
+***********************************************************************************************************************************/
+struct HttpUrl
+{
+    HttpUrlPub pub;
+    MemContext *memContext;                                         // Mem context
```
cmwshang, line 30: Is the MemContext below the pub for alignment? I just found it strange that MemContext was not the first thing in the struct as it is everywhere else, and the way it's constructed in lines 75-82, so I was curious as to why.

cmwshang (commented):
Review complete - with a question/possible change

src/common/io/http/url.c
dwsteele: It needs to be first so we know where to find it, i.e. when casting from HttpUrl to HttpUrlPub the pub elements need to be at the beginning of the private struct.

It's the same for String, the MemContext comes after the STRING_COMMON define.

cmwshang (approved):

PR #1328, opened 2021-03-10
Log archive copy during backup.

dwsteele:
Copying can be a fairly expensive operation so it makes sense to log it.


src/command/backup/backup.c
```diff
@@ -1773,6 +1773,9 @@ backupArchiveCheckCopy(Manifest *manifest, unsigned int walSegmentSize, const St
 
                     if (cfgOptionBool(cfgOptArchiveCopy))
                     {
+                        // Copy can be a pretty expensive operation so log it
+                        LOG_DETAIL_FMT("copy segment %s to backup", strZ(walSegment));
```
cmwshang, line 1777: I was wondering if it wouldn't be more efficient to log one time and also filter the FOR loop on the cfgOptArchiveCopy. We've gotten complaints before about too much logging, so maybe something like:

```
StringList *walSegmentList = pgLsnRangeToWalSegmentList(
    manifestData(manifest)->pgVersion, timeline, lsnStart, lsnStop, walSegmentSize);

if (cfgOptionBool(cfgOptArchiveCopy) && !strLstEmpty(walSegmentList))
{
    // Copy can be a pretty expensive operation so log it
    LOG_DETAIL_FMT("copy to backup segments (%s)", strZ(strLstJoin(walSegmentList, ", ")));

    for (unsigned int walSegmentIdx = 0; walSegmentIdx < strLstSize(walSegmentList); walSegmentIdx++)
    {
        MEM_CONTEXT_TEMP_BEGIN()
        {
            const String *walSegment = strLstGet(walSegmentList, walSegmentIdx);

            // Find the actual wal segment file in the archive
            const String *archiveFile = walSegmentFind(
                storageRepo(), archiveId, walSegment,  cfgOptionUInt64(cfgOptArchiveTimeout));

            // Get compression type of the WAL segment and backup
            CompressType archiveCompressType = compressTypeFromName(archiveFile);

            ...
            
       }
        MEM_CONTEXT_TEMP_END();

        // A keep-alive is required here for the remote holding the backup lock
        protocolKeepAlive();
    }
}
```

cmwshang (commented):
Review complete with one suggestion.

src/command/backup/backup.c
dwsteele: We already log what segments we are checking/copying at line 1745. The idea here is to provide some status so if it takes three hours to do the copy the user gets regular updates that something is still happening.

dwsteele: BTW, there is some precedent for this in async archive-get/push where we log the range at info level then each file at detail level.

cmwshang: > We already log what segments we are checking/copying at line 1745. The idea here is to provide some status so if it takes three hours to do the copy the user gets regular updates that something is still happening.

Ah, ok. But couldn't you stil wrap the FOR loop in the `if (cfgOptionBool(cfgOptArchiveCopy))`?

cmwshang (approved):

pgstef (approved):
Good idea, approved :-)

PR #1332, opened 2021-03-12
Improve protocol handlers.

dwsteele:
Make protocol handlers have one function per command. This allows the logic of finding the handler to be in ProtocolServer, isolates each command to a function, and removes the need to test the "not found" condition for each handler.


src/storage/remote/protocol.c
```diff
@@ -342,16 +497,25 @@ storageRemoteProtocolBlockSize(const String *message)
...
-        MEM_CONTEXT_BEGIN(memContextTop())
+        // If the context is not already allocated it means this function is being used locally so allocate it
+        if (storageRemoteProtocolLocal.memContext == NULL)
         {
-            MEM_CONTEXT_NEW_BEGIN("StorageRemoteFileReadLocal")
+            MEM_CONTEXT_BEGIN(memContextTop())
             {
-                storageRemoteProtocolLocal.memContext = memContextCurrent();
-                storageRemoteProtocolLocal.blockRegExp = regExpNew(BLOCK_REG_EXP_STR);
+                MEM_CONTEXT_NEW_BEGIN("StorageRemoteProtocol")
```
cmwshang, line 507: So in this incarnation of the "StorageRemoteProtocol" we will not have a driver. So in the storageRemoteProtocolLocal has been set up here and then in line 161, the storageRemoteProtocolLocal.driver will be null - but the memContext has already been set up here, so why would we overwrite it in line 167?

test/src/module/config/protocolTest.c
```diff
@@ -37,8 +37,13 @@ testRun(void)
                 harnessCfgLoad(cfgCmdArchiveGet, argList);
 
                 ProtocolServer *server = protocolServerNew(strNew("test"), strNew("config"), read, write);
-                protocolServerHandlerAdd(server, configProtocol);
-                protocolServerProcess(server, NULL);
+
+                static const ProtocolServerHandler commandHandler[] =
+                {
+                    {.command = PROTOCOL_COMMAND_CONFIG_OPTION, .handler = configOptionProtocol},
```
cmwshang, line 43: I'm not sure I'm enamored with the rename of configProtocol to configOptionProtocol, but I guess it will have to do - but can you please rename it in line 21 here since we no longer have configProtocol()?

cmwshang (changes requested):
Review complete with one change request and several comments/questions.

src/protocol/server.c
```diff
@@ -146,14 +133,21 @@ protocolServerProcess(ProtocolServer *this, const VariantList *retryInterval)
...
-                    ProtocolServerProcessHandler handler = *(ProtocolServerProcessHandler *)lstGet(this->handlerList, handlerIdx);
+                    if (strEqZ(command, handlerList[handlerIdx].command))
+                    {
+                        handler = handlerList[handlerIdx].handler;
+                        break;
+                    }
+                }
 
+                // If handler was found then process
+                if (handler != NULL)
```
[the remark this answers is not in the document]
dwsteele: Yes. If no handler is found then it is likely one of the built-in commands (noop, exit). But there will still be an error if the command is invalid.

src/storage/remote/protocol.c
```diff
@@ -138,197 +139,351 @@ storageRemoteProtocolInfoListCallback(void *server, const StorageInfo *info)
...
-    const Storage *storage = protocolStorageTypeEnum(
-        cfgOptionStr(cfgOptRemoteType)) == protocolStorageTypeRepo ? storageRepoWrite() : storagePgWrite();
-    StorageInterface interface = storageInterface(storage);
-    void *driver = storageDriver(storage);
+    MEM_CONTEXT_TEMP_BEGIN()
+    {
+        // Get storage based on remote type
+        const Storage *storage =
+            protocolStorageTypeEnum(cfgOptionStr(cfgOptRemoteType)) == protocolStorageTypeRepo ?
+                storageRepoWrite() : storagePgWrite();
```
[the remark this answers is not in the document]
dwsteele: The remote always has --repo and --pg set so it can just use the default. Each remote addresses one and only one pg/repo storage. The logic for this is in protocol/helper.c.

src/storage/remote/protocol.c
dwsteele: There are two ways this function can be called -- in the remote *or* the local/main process. So the function can't depend on the context having been created for the driver. Really we are just trying to cache this regexp so it does not need to be created for each block.

I admit that this is a bit convoluted, but since this function will go away when the protocol is binary I'm not very motivated to make it cleaner or pick a better place to store the context.

test/src/module/config/protocolTest.c
dwsteele: Well, this `configProtocol -> configOptionProtocol` was to make it match all the other protocol functions, so I think that is fine. Fixed the test title, though, in 127aaf348a57.

What I did not notice is that we also have `configOptionProtocol` which is very confusing. I have renamed this to `configOptionRemote` in 127aaf348a57. Better?

src/command/local/local.c
```diff
@@ -17,6 +17,18 @@ Local Command
...
+Command handlers
+***********************************************************************************************************************************/
+static const ProtocolServerHandler commandLocalHandlerList[] =
+{
+    {.command = PROTOCOL_COMMAND_ARCHIVE_GET_FILE, .handler = archiveGetFileProtocol},
+    {.command = PROTOCOL_COMMAND_ARCHIVE_PUSH_FILE, .handler = archivePushFileProtocol},
+    {.command = PROTOCOL_COMMAND_BACKUP_FILE, .handler = backupFileProtocol},
+    {.command = PROTOCOL_COMMAND_RESTORE_FILE, .handler = restoreFileProtocol},
+    {.command = PROTOCOL_COMMAND_VERIFY_FILE, .handler = verifyFileProtocol},
+};
```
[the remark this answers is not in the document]
dwsteele: I've added some basic directions in protocol/server.h in 127aaf348a57.

As for local/remote, that really depends on what you are trying to do. The ProtocolServer has no notion of local/remote so that's not the right place to put it. local.h/remote.h could have more info on what is local vs remote, but I don't want to do that as part of this patch. We add new commands so seldom that anything I write will probably be wrong by the time it might be used.

src/protocol/server.c
```diff
@@ -125,11 +108,15 @@ protocolServerError(ProtocolServer *this, int code, const String *message, const
...
-protocolServerProcess(ProtocolServer *this, const VariantList *retryInterval)
+protocolServerProcess(
+    ProtocolServer *this, const VariantList *retryInterval, const ProtocolServerHandler *const handlerList,
+    const unsigned int handlerListSize)
 {
     FUNCTION_LOG_BEGIN(logLevelDebug);
         FUNCTION_LOG_PARAM(PROTOCOL_SERVER, this);
         FUNCTION_LOG_PARAM(VARIANT_LIST, retryInterval);
+        FUNCTION_LOG_PARAM_P(VOID, handlerList);
+        FUNCTION_LOG_PARAM(UINT, handlerListSize);
```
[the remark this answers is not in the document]
dwsteele: Yes, definitely need asserts here, added in 127aaf348a57. I can't find `protocolServerProcess(server, NULL)` -- you may be looking at the wrong branch.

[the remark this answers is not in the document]
cmwshang: Line 182 of module/performance/storageTest.c has protocolServerProcess(server, NULL); - this is what is confusing me.

test/src/module/config/protocolTest.c
cmwshang: Exactly why I didn't like the name, so the rename looks good except   line 21 of config/protocolTest.c   
`if (testBegin("configOptionProtocol() and configProtocolOption()"))`

src/storage/remote/protocol.c
cmwshang: I need to think on this. Take a look at my other 2 replies and I'll revisit tomorrow.

test/src/module/config/protocolTest.c
dwsteele: Already fixed in cf663563c2160461 but I didn't link it before. I noticed these after fixed the others.

src/protocol/server.c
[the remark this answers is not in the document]
dwsteele: Not sure how I missed that -- must have been looking in the wrong branch! Anyway, it wasn't causing problems because that test was not being run in CI. Looks like an oversight when we moved to Github Actions. Not too big a deal since it does not provide coverage but nice if it works when we need it for performance testing. So, fixed that separately in integration 0eb5b22ffa02b. The test was fixed in 96b0afcfa0df3dfa1.

src/storage/remote/protocol.c
cmwshang: Why would the function `storageRemoteFeatureProtocol` at line 142 not look like this, then (I added checking the memory context):
```
void
storageRemoteFeatureProtocol(const VariantList *paramList, ProtocolServer *server)
{
    FUNCTION_LOG_BEGIN(logLevelDebug);
        FUNCTION_LOG_PARAM(VARIANT_LIST, paramList);
        FUNCTION_LOG_PARAM(PROTOCOL_SERVER, server);
    FUNCTION_LOG_END();

    ASSERT(paramList == NULL);
    ASSERT(server != NULL);

    MEM_CONTEXT_TEMP_BEGIN()
    {
        // Get storage based on remote type
        const Storage *storage =
            protocolStorageTypeEnum(cfgOptionStr(cfgOptRemoteType)) == protocolStorageTypeRepo ?
                storageRepoWrite() : storagePgWrite();

        // Store local variables in the server context
        if (storageRemoteProtocolLocal.driver == NULL)
        {
            // If the memoryContext has already been alocatted
            if (storageRemoteProtocolLocal.memContext != NULL)
            {
                MEM_CONTEXT_BEGIN(storageRemoteProtocolLocal.memContext)
                {
                    storageRemoteProtocolLocal.driver = storageDriver(storage);
                }
                MEM_CONTEXT_END();
            }
            else
            {
                MEM_CONTEXT_PRIOR_BEGIN()
                {
                    MEM_CONTEXT_NEW_BEGIN("StorageRemoteProtocol")
                    {
                        storageRemoteProtocolLocal.memContext = memContextCurrent();
                        storageRemoteProtocolLocal.driver = storageDriver(storage);
                    }
                    MEM_CONTEXT_NEW_END();
                }
                MEM_CONTEXT_PRIOR_END();
            }
        }

        // Return storage features
        protocolServerWriteLine(server, jsonFromStr(storagePathP(storage, NULL)));
        protocolServerWriteLine(server, jsonFromUInt64(storageInterface(storage).feature));

        protocolServerResponse(server, NULL);
    }
    MEM_CONTEXT_TEMP_END();

    FUNCTION_TEST_RETURN_VOID();
}
```

dwsteele: Mostly because that's more to undo when `storageRemoteProtocolBlockSize()` is removed, which should be in the next few weeks. There might be a simpler way to deal with this, though.

dwsteele: 534bb04c96bf40894f2 is a simplified version that works fine if we need to leave it in longer than I expect, but removes the interaction with the main context.

cmwshang (approved):

PR #1341, opened 2021-03-22
Exclude content-length from S3 signed headers.

dwsteele:
The `content-length` header was being signed since it was the only header that didn't need to be and it seemed simpler just to sign it as well. However, some proxies munge this header causing authentication failure.

### DIFF
--- a/src/storage/s3/storage.c
+++ b/src/storage/s3/storage.c
@@ -34,9 +34,11 @@ Defaults
 /***********************************************************************************************************************************
 S3 HTTP headers
 ***********************************************************************************************************************************/
-STRING_STATIC(S3_HEADER_CONTENT_SHA256_STR,                         "x-amz-content-sha256");
-STRING_STATIC(S3_HEADER_DATE_STR,                                   "x-amz-date");
-STRING_STATIC(S3_HEADER_TOKEN_STR,                                  "x-amz-security-token");
+#define S3_HEADER_PREFIX                                            "x-amz-"
+    STRING_STATIC(S3_HEADER_PREFIX_STR,                             S3_HEADER_PREFIX);
+STRING_STATIC(S3_HEADER_CONTENT_SHA256_STR,                         S3_HEADER_PREFIX "content-sha256");
+STRING_STATIC(S3_HEADER_DATE_STR,                                   S3_HEADER_PREFIX "date");
+STRING_STATIC(S3_HEADER_TOKEN_STR,                                  S3_HEADER_PREFIX "security-token");
 
 /***********************************************************************************************************************************
 S3 query tokens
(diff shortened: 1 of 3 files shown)


test/src/module/storage/s3Test.c
```diff
@@ -57,14 +57,13 @@ testRequest(IoWrite *write, Storage *s3, const char *verb, const char *path, Tes
     {
         strCatFmt(
             request,
-                "authorization:AWS4-HMAC-SHA256 Credential=%s/\?\?\?\?\?\?\?\?/us-east-1/s3/aws4_request,"
-                    "SignedHeaders=content-length",
-                param.accessKey == NULL ? strZ(driver->accessKey) : param.accessKey);
+            "authorization:AWS4-HMAC-SHA256 Credential=%s/\?\?\?\?\?\?\?\?/us-east-1/s3/aws4_request,SignedHeaders=",
+            param.accessKey == NULL ? strZ(driver->accessKey) : param.accessKey);
```
cmwshang, line 62: It's a bit unclear to me that we are proving we do not sign the Content-Length:0 header, as was the ask, in this test file. Maybe you could point me to where we test this?

src/storage/s3/storage.c
```diff
@@ -206,8 +208,9 @@ storageS3Auth(
             const String *headerKey = strLstGet(headerList, headerIdx);
             const String *headerKeyLower = strLower(strDup(headerKey));
 
-            // Skip the authorization header -- if it exists this is a retry
-            if (strEq(headerKeyLower, HTTP_HEADER_AUTHORIZATION_STR))
+            // Skip headers that do not require signing
+            if (!strEq(headerKeyLower, HTTP_HEADER_HOST_STR) && !strEq(headerKeyLower, HTTP_HEADER_CONTENT_MD5_STR) &&
+                !strBeginsWith(headerKeyLower, S3_HEADER_PREFIX_STR))
```
cmwshang, line 213: I'm a little concerned that the code change, which does more than the ask, may result in issues that we are unaware of. Maybe you could point me to where the S3 documentation indicates that only headers that begin with "x-amz-" or are "host" or  "content-md5" are the only things that need to be signed?

cmwshang (commented):

test/src/module/storage/s3Test.c
dwsteele: content-length was in the signed headers test list before, but now it is not. What's unclear?

src/storage/s3/storage.c
dwsteele: It's here: https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-header-based-auth.html. Search for `SignedHeaders`.  content-md5 is not in the list but that's not the only place where there docs are inaccurate.

Since content-length is the only thing we don't need to sign and we are not likely to add new headers, I would be OK with flipping this and just excluding content-length and authorization from signing.

dwsteele: Done in cf8f611ac6dbd4c.

test/src/module/storage/s3Test.c
cmwshang: I know, but I was looking for when you call storageS3Auth with HttpHeader *httpHeader that includes content-length - but since you must be getting coverage on the new code in that function, it must be somewhere, so OK.

src/storage/s3/storage.c
cmwshang: Super, thanks.

cmwshang (approved):

test/src/module/storage/s3Test.c
dwsteele: Oh, I see. Every request contains content-length so it definitely gets tested. If you look at s3Test line 75 you can see that the header is tested unconditionally. We just removed it from the signed headers list.

PR #1343, opened 2021-03-24
Improve archive-push command fault tolerance.

dwsteele:
3b8f0ef7ae8a missed some cases that could cause archive-push to fail:

1) While checking archive info.
2) While checking to see if a WAL segment already exists.

These cases are now handled so archive-push can succeed on any valid repos.


cmwshang (approved):
Fixed minor formatting - see commit baf3631e2e71b59dd7ddd8e2d2c78926dc03cecf

Issue #1242, opened 2020-11-25
Clarify default compress-level


mbanck:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version:

2.30

[...]

7. Describe the issue:

I am trying to compare backup sizes between pgBackRest and manual base backups. pgBackRest has the `--compress-level`  option which goes from 0 to 9, however, it is unclear which compression level is chosen if the option is not mentioned (and neither the `pgbackrest help backup` help nor the command reference mention a default).
It could be that there is a different default based on the `compress-type` option (i.e. zstd has different defaults/compress-level scales than bz2 or so), but then, it would be useful to document them, either along `compress-level` or `compress-type`.
I tried to dig around a bit in the source, but couldn't find it defined anywhere off-hand.

As an aside, doing a backup with `compress-type=zst` and `compress-level=9` results in a 5% larger backup size than a manual `zstd -10` (while `zstd` allows levels as high as 19)

mbanck:
I've now found the place in the code, it's in `src/common/compress/helper.c`:
```
        .type = STRDEF(BZ2_EXT),
        .levelDefault = 9,
        .type = STRDEF(GZ_EXT),
        .levelDefault = 6,
        .type = STRDEF(LZ4_EXT),
        .levelDefault = 1,
        .type = STRDEF(ZST_EXT),
        .levelDefault = 3,
```
Would be nice to have those documented.

dwsteele:
> there is a different default based on the compress-type option. 

Exactly -- and since there are multiple possible programmatic defaults the documentation doesn't have a fixed default to pick up. However, these are very unlikely to change so we can just document them statically.

> As an aside, doing a backup with compress-type=zst and compress-level=9 results in a 5% larger backup size than a manual zstd -10

That's a bigger difference than I had thought. We do need to update the lz4/zstd ranges to reflect reality. Since they are version-specific and programmatic (e.g. `ZSTD_maxCLevel(void)`) they don't fit very well with how we constrain option ranges.

If you want to expand the range for testing you can update it here: https://github.com/pgbackrest/pgbackrest/blob/release/2.30/src/config/define.auto.c#L852.

PR #1346, opened 2021-03-25
Add compress-level defaults per compress-type value.

cmwshang:
REFERENCE: https://github.com/pgbackrest/pgbackrest/issues/1242


doc/xml/reference.xml
```diff
@@ -155,7 +155,15 @@
                     <config-key id="compress-level" name="Compress Level">
                         <summary>File compression level.</summary>
 
-                        <text>Sets the level to be used for file compression when <setting>compress-type</setting> does not equal <id>none</id> or <setting>compress=y</setting> (deprecated).</text>
+                        <text>Sets the level to be used for file compression when <setting>compress-type</setting> does not equal <id>none</id> or <setting>compress=y</setting> (deprecated).
+
+                        The following are the defaults levels based on the type when compress-level is not specified:
+                        <ul>
```
dwsteele, line 161: I think this should be something like:
```
based on <setting>compress-type</setting> when <setting>compress-level</setting> is not specified
```

dwsteele (changes requested):

doc/xml/reference.xml
cmwshang: Fixed 18b612b10f4d65affa901171bfa468546955f46e

dwsteele (approved):

Issue #871, opened 2019-11-05
NFS : doc should give officlal recommandations


Krysztophe:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version: 2.17
2. PostgreSQL version: 12
3. Operating system/version : Linux
4. Did you install pgBackRest from source or from a package? package
7. Describe the issue:

For people not so fluent with NFS, is it possible to add recommendations on how to use an NFS share as a repo in a secure way and be sure not to lose some wals? In my understanding, the recommendations from the PG12 doc (https://www.postgresql.org/docs/current/creating-cluster.html#NFS : 'sync' on the server, 'hard' on the client) is probably enough, but I may be wrong.

dwsteele:
The Postgres recommendations are also fine for pgBackRest.  We'll add this recommendation with a link to the Postgres docs.

cmwshang:
PR created for this issue.

PR #1347, opened 2021-03-25
Add note about required NFS settings being the same as PostgreSQL.

cmwshang:
REFERENCE: https://github.com/pgbackrest/pgbackrest/issues/871


dwsteele (approved):
Did a bit of editing in cd8a3e5ed. If that's OK then we are ready to commit.

PR #1356, opened 2021-04-06
Fix issue when checking links for large numbers of tablespaces.

dwsteele:
The manifestLinkCheck() was pretty inefficient so large numbers of links caused it to use a lot of memory and eventually crash.

This is a more efficient implementation which runs O(nlogn) and uses far less memory.

Checking for duplicate file links has been added, which represents a change in behavior that we'll need to consider carefully.


src/info/manifest.c
```diff
@@ -381,6 +381,178 @@ manifestNewInternal(void)
...
+                ASSERT(linkIdx != LIST_NOT_FOUND);
+
+                // Check the link destination path to be sure it is not a subpath of a prior link destination path. Skip file
+                // links since they are allowed to be in the same path with each other and in the parent path of a linked
+                // destination path.
+                for (unsigned int priorLinkIdx = linkIdx - 1; priorLinkIdx < linkIdx; priorLinkIdx--)
+                {
+                    const ManifestLinkCheckItem *const priorLink = lstGet(linkCheck->linkList, priorLinkIdx);
+
+                    if (strBeginsWith(path, priorLink->path))
```
cmwshang, line 511: This could still be a file link at this point since `->file`  isn't checked until after this (my understanding is `->file = NULL` means it is a path link) - so wouldn't that be a problem if there are links to different files in the same path? I'm not sure I saw a test for this.

test/src/module/info/manifestTest.c
```diff
@@ -530,6 +530,29 @@ testRun(void)
         varLstAdd(tablespace, varNewStrZ("tblspc1"));  // tablespaceName
         varLstAdd(tablespaceList, varNewVarLst(tablespace));
 
+        // Add more tablespaces to test bug
+        // for (unsigned int tsIdx = 3; tsIdx <= 4001; tsIdx++)
```
cmwshang, line 534: Should we comment that this is removed but left here for a reason?

cmwshang (commented):

test/src/module/info/manifestTest.c
dwsteele: I'm going to remove it completely. Perhaps we could add it to the performance test at some point but I don't think it is worth the time right now.

src/info/manifest.c
```diff
@@ -381,6 +381,178 @@ manifestNewInternal(void)
...
+
+            MEM_CONTEXT_BEGIN(lstMemContext(linkCheck->linkList))
+            {
+                path = strNewFmt("%s/", strZ(manifestTargetPath(this, target1)));
+
+                // Get base bath
+                if (linkCheck->basePath == NULL)
+                {
+                    linkCheck->basePath = strNewFmt(
+                        "%s/", strZ(manifestTargetPath(this, manifestTargetFind(this, MANIFEST_TARGET_PGDATA_STR))));
```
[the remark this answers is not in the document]
dwsteele: The storage module will never return a path with a slash at the end. And it should not matter to the algorithm any, but I also added a few final slashes in the tests to be sure.

src/info/manifest.c
dwsteele: There is a test for it at ManifestTest.c line 1981. But the comment certainly doesn't match the code. I think this may be tricky to test due to ordering. I'll need to look into it further, but my brain does not want to cooperate today.

test/src/module/info/manifestTest.c
dwsteele: Removed in ef221c6e23f57e22.

src/info/manifest.c
dwsteele: Added the logic to skip file links in ef221c6e23f57e22. Luckily this already had tests.

Also add checks when the path link is the *parent* of a subsequent link. Depending on the link ordering this can happen.

src/info/manifest.c
```diff
@@ -381,6 +381,209 @@ manifestNewInternal(void)
...
+
+                // Find the path in the sorted list
+                unsigned int linkIdx = lstFindIdx(linkCheck->linkList, &path);
+                ASSERT(linkIdx != LIST_NOT_FOUND);
+
+                // Check path links against other links (file links have already been checked)
+                if (target1->file == NULL)
+                {
+                    // Check the link destination path to be sure it is not a subpath of a prior link destination path
+                    for (unsigned int priorLinkIdx = linkIdx - 1; priorLinkIdx < linkIdx; priorLinkIdx--)
```
cmwshang, line 508: I'm a little confused on how the test expression `priorLinkIdx < linkIdx` ever evaluates to false. If every `priorLink->file != NULL` we continue until when? Shouldn't this be `(int)priorLinkIdx >= 0`?

cmwshang (commented):
Review complete - one issue found

src/info/manifest.c
dwsteele: We can make it `(int)priorLinkIdx >= 0` if you like, but this works as well because `(unsigned)0 - 1 == UINT_MAX`. I was hoping it would be more readable...

cmwshang: we use the `(int)priorLinkIdx >= 0` construct in other places in the code so I'd rather be consistent...

dwsteele: Sure thing, updated in d8e9d3db6ec.

cmwshang (approved):

PR #1358, opened 2021-04-08
Add StringId type.

dwsteele:
It is often useful to represent identifiers as strings when they cannot easily be represented as an enum/integer, e.g. because they are distributed among a number of unrelated modules or need to be passed to remote processes. Strings are also more helpful in debugging since they can be recognized without cross-referencing the source. However, strings are awkward to work with in C since they cannot be directly used in switch statements leading to less efficient if-else structures.

A StringId encodes a short string into an integer so it can be used in switch statements but may also be readily converted back into a string for debugging purposes. StringIds may also be suitable for matching user input providing the strings are short enough.

This patch includes a sample of StringId usage by converting protocol commands to StringIds. There are many other possible use cases. To list a few:

1) All "types" in storage, filters. IO , etc. These types are primarily for identification and debugging so they fit well with this model.

2) MemContext names would work well as StringIds since these are entirely for debugging.

3) Option values could be represented as StringIds which would mean we could remove the functions that convert strings to enums, e.g. `CipherType`.

4) There are a number of places where enums need to be converted back to strings for logging/debugging purposes. An example is `protocolParallelJobToConstZ`. If `ProtocolParallelJobState` were defined as:
```
typedef enum
{
    protocolParallelJobStatePending = STRID5("pending", ...),
    protocolParallelJobStateRunning = STRID5("running", ...),
    protocolParallelJobStateDone = STRID5("done", ...),
} ProtocolParallelJobState;
```
then `protocolParallelJobToConstZ()` could be replaced with `strIdToZ()`.

This also applies to many enums that we don't covert to strings for logging, such as `CipherMode`.


src/common/type/stringId.h
```diff
@@ -0,0 +1,123 @@
...
+String *strIdToStr(const StringId strId);
+
+// Convert StringId to zero-terminated string. See strIdToZN() for buffer sizing and return value.
+size_t strIdToZ(const StringId strId, char *const buffer);
+
+/***********************************************************************************************************************************
+Generate constant StringIds
+
+To generate a constant StringId call strIdGenerate() in any debug build. It will throw an error with the generated StringId macro
+in the error message.
```
cmwshang, line 92: I'm a little confused about the strIdGenerate. 

`strIdGenerate("mytest0123abc")` produces `STRID5("mytest0123abc", 0x107dde02932d32d4)`
but then I run `strIdToStr(0x107dde02932d32d4)` and I get `mytest!123ab+`

Since `mytest0123abc` contains a 0 I was thinking it would use STRID6, how do I get it to not replace the `0` with a `!` ?

src/common/type/stringId.h
```diff
@@ -0,0 +1,123 @@
...
+Macros to define constant StringIds. ALWAYS use strIdGenerate() to create these macros. The parameters in the macros are not
+verified against each other so the str parameter is included only for documentation purposes.
+***********************************************************************************************************************************/
+#define STRID5(str, strId)                                          strId
+#define STRID6(str, strId)                                          strId
+
+/***********************************************************************************************************************************
+Functions
+***********************************************************************************************************************************/
+// Convert N chars to StringId, If the string is longer than the allowable length for the selected encoding then the StringID will
```
cmwshang, line 58: In this statement, I took `allowable length` to be the STRID5_MAX, or STRID6_MAX, but a string that is 11 characters `strIdGenerate("mytest0123a")` - is equal to, not longer than, the allowable length of 11 yet when I convert it back, the 11th character becomes a + sign - so is there something I'm missing?

cmwshang (commented):
Partial review.

src/common/type/stringId.h
dwsteele: It should have used STRID6(). There was problem with the boundary checking, so one extra character worked, but more than one extra did not work. Fixed in 196c2b047a267.

src/common/type/stringId.h
dwsteele: The max includes the + indicator when present. I clarified the comment a bit in aabc411388a3ca.

src/common/type/stringId.c
```diff
@@ -0,0 +1,375 @@
...
+
+            // Char 12
+            buffer[11] = map[strId & 0x1F];
+            ASSERT(strId >> 5 == 0);
+
+            // If prefix flag is set then append +
+            if (prefix)
+            {
+                buffer[12] = '+';
+                FUNCTION_TEST_RETURN(13);
```
cmwshang, line 265: Shouldn't we be using the STRID5_MAX constant instead of 13?

src/common/type/stringId.c
```diff
@@ -0,0 +1,375 @@
...
+            }
+
+            FUNCTION_TEST_RETURN(12);
```
cmwshang, line 268: So this is STRID5_MAX - 1?

src/common/type/stringId.c
```diff
@@ -0,0 +1,375 @@
...
+            // Macro to decode all but the last character
+            #define STR5ID_TO_ZN_IDX(idx)                                                                                          \
+                buffer[idx] = map[strId & 0x1F];                                                                                   \
+                strId >>= 5;                                                                                                       \
+                                                                                                                                   \
+                if (strId == 0)                                                                                                    \
+                    FUNCTION_TEST_RETURN(idx + 1)
+
+            // Char 1-11
+            STR5ID_TO_ZN_IDX(0);
```
cmwshang, line 245: Can't we do this in a loop and use the STRID5_MAX like 
```
for (unsigned int idx = 0, idx < STRID5_MAX - 3; idx++)
    STR5ID_TO_ZN_IDX(idx);
```

src/common/type/stringId.c
```diff
@@ -0,0 +1,375 @@
...
+            // Macro to decode all but the last character
+            #define STR6ID_TO_ZN_IDX(idx)                                                                                          \
+                buffer[idx] = map[strId & 0x3F];                                                                                   \
+                strId >>= 6;                                                                                                       \
+                                                                                                                                   \
+                if (strId == 0)                                                                                                    \
+                    FUNCTION_TEST_RETURN(idx + 1)
+
+            // Char 1-9
+            STR6ID_TO_ZN_IDX(0);
```
cmwshang, line 288: Can't we do this in a loop and use the STRID6_MAX like 
```
for (unsigned int idx = 0, idx < STRID6_MAX - 3; idx++)
    STR6ID_TO_ZN_IDX(idx);
```

src/common/type/stringId.c
```diff
@@ -0,0 +1,375 @@
...
+
+            // Char 10
+            buffer[9] = map[strId & 0x3F];
+            ASSERT(strId >> 6 == 0);
+
+            // If prefix flag is set then append +
+            if (prefix)
+            {
+                buffer[10] = '+';
+                FUNCTION_TEST_RETURN(11);
```
cmwshang, line 306: Again, maybe we should be using STRID6_MAX?

src/common/type/stringId.c
```diff
@@ -0,0 +1,375 @@
...
+            }
+
+            FUNCTION_TEST_RETURN(10);
```
cmwshang, line 309: Again, maybe we should be using STRID6_MAX - 1?

src/common/type/stringId.h
```diff
@@ -0,0 +1,123 @@
...
+StringId typedef to make them more recognizable in the code
+***********************************************************************************************************************************/
+typedef uint64_t StringId;
+
+/***********************************************************************************************************************************
+Number of bits to use for encoding. The number of bits affects the character set that can be encoded.
+***********************************************************************************************************************************/
+typedef enum
+{
+    stringIdBit5 = 0,                                               // 5-bit encoding for a-z, 1-4, and - characters
```
cmwshang, line 44: Will this be used for options? What about reset-pg8-host-cmd - how will it be able to be used with an 8 and also how will we be able to distinguish reset-pg8-host when 13 characters are max (or if it must use STRID6, that would be even worse, no?)

cmwshang (changes requested):
Partial review 2.

src/common/type/stringId.c
dwsteele: Yes, but this is much faster.

src/common/type/stringId.c
dwsteele: Yes, but this is much faster.

src/common/type/stringId.h
dwsteele: There is no plan to use StringIds for option names.

src/common/type/stringId.h
```diff
@@ -0,0 +1,123 @@
...
+
+It is often useful to represent identifiers as strings when they cannot easily be represented as an enum/integer, e.g. because they
+are distributed among a number of unrelated modules or need to be passed to remote processes. Strings are also more helpful in
+debugging since they can be recognized without cross-referencing the source. However, strings are awkward to work with in C since
+they cannot be directly used in switch statements leading to less efficient if-else structures.
+
+A StringId encodes a short string into an integer so it can be used in switch statements but may also be readily converted back into
+a string for debugging purposes. StringIds may also be suitable for matching user input providing the strings are short enough.
+
+See strIdGenerate() for information on StringId constants.
```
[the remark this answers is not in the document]
dwsteele: I'd rather not split the documentation for strIdGenerate() as it will be much harder to keep up to date. Can we put this text down in that section?

src/common/type/stringId.c
```diff
@@ -0,0 +1,375 @@
...
+                 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
+                 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
+                 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
+                 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
+                 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
+                 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
+                 0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,
+            };
+
+#ifdef DEBUG
```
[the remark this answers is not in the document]
dwsteele: There are plenty of places in the prod code where this is true, so not sure why this place is special. These checks are slow and currently not needed outside of dev/test.

src/common/type/stringId.c
```diff
@@ -0,0 +1,375 @@
...
```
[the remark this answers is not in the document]
dwsteele: Same as above.

src/common/type/stringId.c
dwsteele: Yes, but then one wonders whether we should do STRID5_MAX - 2, STRID5_MAX - 3, etc.

It's a constant, but it cannot ever be changed without breaking everything. The only reason it exists is to help the caller properly size a buffer so not sure it make sense to propagate it through the code.

src/common/type/stringId.h
```diff
@@ -0,0 +1,123 @@
...
+#define COMMON_TYPE_STRINGID_H
+
+#include <stddef.h>
+#include <stdint.h>
+#include <string.h>
+
+#include "common/type/string.h"
+
+/***********************************************************************************************************************************
+Maximum number of characters in the string representation of a StringId, including the + prefix indicator when present. This is a
```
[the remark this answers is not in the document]
dwsteele: Well, the + sign is a suffix that indicates the *string* is a prefix. So it's a *prefix indicator* as the comment says.

But your wording also seems fine to me, so updated like that in 3104b8dfbac1ec97dd56.

src/common/type/stringId.c
```diff
@@ -0,0 +1,375 @@
...
+
+/***********************************************************************************************************************************
+Constants used to extract information from the header
+***********************************************************************************************************************************/
+#define STRING_ID_BIT_MASK                                          3
+#define STRING_ID_HEADER_SIZE                                       4
+#define STRING_ID_PREFIX                                            4
+
+/**********************************************************************************************************************************/
+StringId strIdFromZN(const StringIdBit bit, const char *const buffer, const size_t size)
```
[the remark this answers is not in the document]
dwsteele: Fixed in 3104b8dfbac1ec97dd56.

src/common/type/stringId.c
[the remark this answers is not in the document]
cmwshang: That does not explain why commenting out the code causes the test `TEST_ERROR(strIdGenerate("Watcha"), FormatError, "STRID6(\"Watcha\", 0x480d407c1)"); `to fail and return the following - I won't get a chance to look at it until much later today but maybe you have an explanation:

  ```
EXPECTED FormatError: STRID6("Watcha", 0x480d407c1)

     BUT GOT FormatError: STRID5("Watcha", 0x281d0200)
```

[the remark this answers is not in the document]
Comment (dwsteele) [reply src/common/type/stringId.c:57]: `strIdGenerate()` is only available in DEBUG mode so it requires all the DEBUG code to be active in order to work.

File: src/protocol/server.c [hunk 15]
```diff
@@ -131,15 +131,15 @@ protocolServerProcess(
             {
                 // Read command
                 KeyValue *commandKv = jsonToKv(ioReadLine(protocolServerIoRead(this)));
-                const String *command = varStr(kvGet(commandKv, VARSTR(PROTOCOL_KEY_COMMAND_STR)));
+                const StringId command = strIdFromStr(stringIdBit5, varStr(kvGet(commandKv, VARSTR(PROTOCOL_KEY_COMMAND_STR))));
```
cmwshang, line 134: Do we specify anywhere that all PROTOCOL_COMMAND_* must be STRID5? What happens if it is stringIdBit6 format?

test/src/module/command/backupTest.c
```diff
@@ -1396,7 +1396,7 @@ testRun(void)
         // -------------------------------------------------------------------------------------------------------------------------
         TEST_TITLE("report job error");
 
-        ProtocolParallelJob *job = protocolParallelJobNew(VARSTRDEF("key"), protocolCommandNew(STRDEF("command")));
+        ProtocolParallelJob *job = protocolParallelJobNew(VARSTRDEF("key"), protocolCommandNew(strIdFromZ(stringIdBit6, "X")));
```
cmwshang, line 1399: It doesn't actually matter what you set here, e.g. `strIdFromZ(stringIdBit5, "bp-f")` instead of `strIdFromZ(stringIdBit6, "X")` would still fail because the error is set via the next line `protocolParallelJobErrorSet`, right?

test/src/module/protocol/protocolTest.c
```diff
@@ -434,7 +434,7 @@ testRun(void)
 
         MEM_CONTEXT_TEMP_BEGIN()
         {
-            TEST_ASSIGN(command, protocolCommandNew(strNew("command1")), "create command");
+            TEST_ASSIGN(command, protocolCommandNew(strIdFromZ(stringIdBit6, "cmd1")), "create command");
```
cmwshang, line 437: Given src/protocol/server.c line 134, I thought only stringIdBit5 was valid for protocol commands - how is it you can use stringIdBit6?

cmwshang (commented):
Review complete with some final questions.

src/common/type/stringId.h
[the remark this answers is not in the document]
cmwshang: Well, this is more than just documentation about strIdGenerate - it's the whole processes of generating, assigning and retrieving, which I thought was more appropriate in this summary area. But if you feel strongly that it should go there, just add it but please don't removed the other example you have.

src/protocol/server.c
dwsteele: There's no requirement that they be STRID5, it is just slightly more efficient that way since they are smaller.

src/common/type/stringId.c
cmwshang: Discussed with David and agreed it was ok to leave the hardcoded values.

src/protocol/server.c
dwsteele: Oops, yeah, for the time being they do need to be STRID5 -- but that's a temporary condition until the binary protocol is in place.

dwsteele: 6-bit will work when we move the binary protocol but I see no reason why we should. I added comments to specify that these StringIds should be 5-bit in d8902130e0.

test/src/module/command/backupTest.c
dwsteele: You are right, it does not matter. But let's just make these 5-bit to avoid confusion. Done in d8902130e0.

test/src/module/protocol/protocolTest.c
dwsteele: Originally I had not added any numbers to 5-bit so I had made these 6-bit. It works because they are value 5-bit strings and they get converted to a string in the protocol layer. Clearer just to make them 5-bit so I did that in d8902130e0.

src/common/type/stringId.h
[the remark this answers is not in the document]
dwsteele: Updated as discussed in 19b4be5c5faa35.

cmwshang (approved):

Issue #1363, opened 2021-04-12
configure on FreeBSD does not handle configuration file location


fluca1978:
pgbackrest version 2.33 (release) on FreeBSD 12.2-release.
I was expecting that having binaries and libraries under `/usr/local` now by default, even configuration should be placed on `/usr/local/etc` instead of `/etc`.
However it seems to me this is not.

Example:

```
 % uname -a
FreeBSD miguel 12.2-RELEASE FreeBSD 12.2-RELEASE r366954 GENERIC  amd64

 % ./configure && gmake && sudo gmake install
...

% ls -sh /usr/local/etc/pgbackrest.conf
4 /usr/local/etc/pgbackrest.conf
```

With everything in place, the following does not work:

```
% sudo -u postgres pgbackrest --stanza miguel info 
stanza: miguel
    status: error (missing stanza path)
```

while it works if I copy/move the configuration file to the default location:

```
 % sudo mv /usr/local/etc/pgbackrest.conf /etc 

 % sudo -u postgres pgbackrest --stanza miguel info 
stanza: miguel
    status: ok
    cipher: none

    db (current)
        wal archive min/max (12): 00000001000000C200000021/00000001000000C200000029
...
```

Interestingly, a link does not work either:

```
 % sudo ln -s /usr/local/etc/pgbackrest.conf /etc/pgbackrest

 % sudo -u postgres pgbackrest --stanza miguel info   
ERROR: [041]: unable to open file '/etc/pgbackrest/pgbackrest.conf' for read: [20] Not a directory

 % ls -l /etc/pgbackrest 
lrwxr-xr-x  1 root  wheel  30 Apr 12 12:37 /etc/pgbackrest -> /usr/local/etc/pgbackrest.conf
```

Configuring the software with `--with-configdir` works as expected:

```
% ./configure --with-configdir=/usr/local/etc && gmake && sudo gmake install
...
% ls /usr/local/etc/pgbackrest.conf
/usr/local/etc/pgbackrest.conf

 % ls -l /etc/pgbackrest.conf       
ls: /etc/pgbackrest.conf: No such file or directory

 % sudo -u postgres pgbackrest --stanza miguel info                          
stanza: miguel
    status: ok
    cipher: none

    db (current)
        wal archive min/max (12): 00000001000000C200000021/00000001000000C200000029

...
```

[...]

PR #1364, opened 2021-04-12
Use ac_default_prefix as default path for configuration on FreeBSD.

fluca1978:
See #1363.

Modifies the `configure.ac` to regenerate `configure` to get the correct path for configuration path.

Tests the OS type, and in the case of FreeBSD sets a variable
to use when defining the default configuration file path.
This makes /usr/local/etc the path for FreeBSD, keeping /etc for Linux
and Darwin.
Tested on FreeBSD 12.2 and Fedora 33.

Since I've no experience in configuring and using autotools, this requires a review because I'm sure there is some AC macro to do this in a single pass.

### DIFF
--- a/src/build/configure.ac
+++ b/src/build/configure.ac
@@ -18,6 +18,10 @@ AC_PROG_CC_C99
 AC_CANONICAL_HOST
 AC_SUBST(CFLAGS, "${CFLAGS} -std=c99")
 
+# default path for configuration file
+default_configdir="/etc"
+
+
 case $host_os in
     darwin*)
         AC_SUBST(CPPFLAGS, "${CPPFLAGS} -D_DARWIN_C_SOURCE")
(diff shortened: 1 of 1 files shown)


fluca1978:
The failed tests denote how much do I know about autotools.
For the records: I generated the programs by doing

```
cd src/build && autoconf && cd .. && cp build/configure . &&  ./configure
```

before the usual `make` steps.

However, I realized this PR does not make sense: I thought the usage of `/usr/local/bin` as default installation target has something related to Unix, but instead it used as the default target for both Unix and Linux platforms. Therefore, I suspect the only sane way of telling where the configuration should be stored is by using the `--with-configdir` option when compiling.

dwsteele:
Yeah, autoconf is a real pain. It might be possible to just set the best default when FreeBSD is detected rather than try to use the `ac_default_prefix` macro. Hopefully somebody will pick up the proposed FreeBSD package and we'll get some input.

mschout:
I'm interested in making pgbackrest better on freebsd, but I think getting accepted into the ports tree, plus `--with-configdir` is enough probably.  The way the proposed FreeBSD port works, everything gets installed in the place where FreeBSD users would expect.

I really have no idea why the ports committers have not picked it up.  All I can suggest is commenting on the freebsd bug report where the port submission is pending.

mschout:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=251687

PR #1373, opened 2021-04-19
Do not write files atomically or sync paths during backup copy.

dwsteele:
There is no need to write the file atomically (e.g. via a temp file on Posix) because checksums are tested on resume after a failed backup. The path does not need be synced for each file because all paths are synced at the end of the backup.

This functionality was not lost during the migration -- it never existed in the Perl code, though these settings are used in restore. See commit where backupFile() was migrated to C: https://github.com/pgbackrest/pgbackrest/commit/59f135340d4f76c522c8b24ecb23d56b57b2b0f8#diff-531b06ff58546d53a3931a9dbcee3e11967b23dda804c8e99c2e44cea37792fa

So, any reason why this seems like a bad idea?

Project card: https://github.com/pgbackrest/pgbackrest/projects/2#card-41800801


sfrost:
I can't think of any reason we need to use a temp file or sync paths during the backup when we sync them all at the end before claiming the backup has been successful...

One interesting question though- do we do anything to make sure that all the files used for a resumed backup are sync'd out to disk?  Consider a case where fsync() just hangs for some reason- the first backup dies because of a timeout and everything is killed, a subsequent backup checks the checksum of the files that exist and decides that those files are "ok" because the checksum matches but those files never got a confirmed and successful fsync.  Seems like we should explicitly fsync those files even if the checksums match.

dwsteele:
> Seems like we should explicitly fsync those files even if the checksums match.

Checksums are not stored in the manifest until the file is successfully fsync'd. And only files with checksums in the manifest can be resumed. So that seems safe enough.

sfrost:
Ah, right, agreed, that works.  Sounds good to me.

pgstef:
> I can't think of any reason we need to use a temp file or sync paths during the backup when we sync them all at the end before claiming the backup has been successful...

+1

> > Seems like we should explicitly fsync those files even if the checksums match.
> 
> Checksums are not stored in the manifest until the file is successfully fsync'd. And only files with checksums in the manifest can be resumed. So that seems safe enough.

Sounds good to me too.

I gave a try to the feature with my usual test cases and didn't found anything special. The backup seems a little bit faster though:
```bash
# 2.33
$ pgbackrest version
pgBackRest 2.33
$ time pgbackrest backup --type=full --stanza=my_stanza
real	0m34.369s
user	0m23.195s
sys	0m1.282s
$ pgbackrest --stanza=my_stanza info
stanza: my_stanza
    status: ok
    cipher: none

    db (current)
        wal archive min/max (13): 0000000200000000000000E1/0000000200000000000000E1

        full backup: 20210422-110548F
            timestamp start/stop: 2021-04-22 11:05:48 / 2021-04-22 11:06:20
            wal start/stop: 0000000200000000000000E1 / 0000000200000000000000E1
            database size: 1.5GB, database backup size: 1.5GB
            repo1: backup set size: 88.8MB, backup size: 88.8MB

$ time pgbackrest backup --type=full --stanza=my_stanza
real    2m7.987s
user    1m50.489s
sys 0m3.437s
$ pgbackrest --stanza=my_stanza info
stanza: my_stanza
    status: ok
    cipher: none

    db (current)
        wal archive min/max (13): 00000003000000040000001B/00000003000000040000001B

        full backup: 20210422-114151F
            timestamp start/stop: 2021-04-22 11:41:51 / 2021-04-22 11:43:57
            wal start/stop: 00000003000000040000001B / 00000003000000040000001B
            database size: 7.3GB, database backup size: 7.3GB
            repo1: backup set size: 412MB, backup size: 412MB

# 3.34dev
$ pgbackrest version
pgBackRest 2.34dev
$ time pgbackrest backup --type=full --stanza=my_stanza
real	0m32.090s
user	0m22.735s
sys	0m1.223s
$ pgbackrest --stanza=my_stanza info
stanza: my_stanza
    status: ok
    cipher: none

    db (current)
        wal archive min/max (13): 0000000200000000000000E3/0000000200000000000000E3

        full backup: 20210422-111400F
            timestamp start/stop: 2021-04-22 11:14:00 / 2021-04-22 11:14:31
            wal start/stop: 0000000200000000000000E3 / 0000000200000000000000E3
            database size: 1.5GB, database backup size: 1.5GB
            repo1: backup set size: 88.8MB, backup size: 88.8MB

$ time pgbackrest backup --type=full --stanza=my_stanza
real    2m1.677s
user    1m49.051s
sys 0m3.174s
$ pgbackrest --stanza=my_stanza info
stanza: my_stanza
    status: ok
    cipher: none

    db (current)
        wal archive min/max (13): 000000030000000400000019/000000030000000400000019

        full backup: 20210422-113604F
            timestamp start/stop: 2021-04-22 11:36:04 / 2021-04-22 11:38:04
            wal start/stop: 000000030000000400000019 / 000000030000000400000019
            database size: 7.3GB, database backup size: 7.3GB
            repo1: backup set size: 412MB, backup size: 412MB
```

I didn't found any simple/fast use-case to deeper verify the impact of this change but the assumptions seems safe enough for me.

dwsteele:
> The backup seems a little bit faster though

Those time differences are what I've seen as well. Modest, but measurable.

I would expect this to have a more dramatic impact on NFS mounts or other network file systems where skipping the two additional ops per file (mv, sync path) will likely be more noticeable.

> I didn't found any simple/fast use-case to deeper verify the impact of this change but the assumptions seems safe enough for me.

Yeah, this is an incredibly hard thing to test empirically because you'd need a reliable way to *not* have files/paths sync'd properly. There are ways to do this by "breaking" the kernel but I think that is further than we want to go here.

So what we need is plenty of eyeballs on it to make sure it looks safe. Ultimately the restore will fail on checksums fail if the data was not written correctly but we'd rather avoid that.

If you are done reviewing, please approve the PR.

pgstef (approved):

src/command/backup/file.c
```diff
@@ -227,8 +227,11 @@ backupFile(
                         storageReadIo(read)), cipherBlockNew(cipherModeEncrypt, cipherType, BUFSTR(cipherPass), NULL));
             }
 
-            // Setup the repo file for write
-            StorageWrite *write = storageNewWriteP(storageRepoWrite(), repoPathFile, .compressible = compressible);
+            // Setup the repo file for write. There is no need to write the file atomically (e.g. via a temp file on Posix) because
+            // checksums are tested on resume after a failed backup. The path does not need be synced for each file because all
```
cmwshang, line 231: `The path does not need be synced` --> add "to": `The path does not need to be synced`

cmwshang (commented):
So are there no documentation changes? Not likely for the intent of this PR but I ask because you stated `because checksums are tested on resume after a failed backup.` and I see the following regarding the delta option and it makes me think we don't use the checksums by default...

4.2.11 Delta Option (--delta)
Restore or backup using checksums.
During a restore, by default the PostgreSQL data and tablespace directories are expected to be present but empty. This option performs a delta restore using checksums.

During a backup, this option will use checksums instead of the timestamps to determine if files will be copied.

dwsteele:
> So are there no documentation changes?

No, this is not user facing -- and it depends on which driver you are using. For instance on objects stores these setting have no effect.

> I see the following regarding the delta option and it makes me think we don't use the checksums by default...

Backup does not use checksums to determine which files will be copied by default (--delta), but it always generates checksums for the backup. And restore always verifies those checksums (Whether --delta is enabled or not).

src/command/backup/file.c
dwsteele: Fixed in 5b4e95ed14b72d.

cmwshang (approved):

sfrost (approved):

PR #1374, opened 2021-04-20
Free no longer needed remotes so they do not timeout during restore.

dwsteele:
The remote is no longer needed in the main process after the manifest is loaded. If the restore is enough the connection will timeout and WARN at the end of the restore. This is harmless for the restore but distracting for the user.


cmwshang (commented):
There are a couple tests in MockArchiveStop that result in this WARN (see mock-archive-stop-002.log) - would they need to be updated? Or do we still expect those to appear.

dwsteele:
> There are a couple tests in MockArchiveStop that result in this WARN (see mock-archive-stop-002.log) 

I can't find any warning that seem to fit the description (line number?). If you mean:
```
[UnknownError] remote-0 process on 'bogus' terminated unexpectedly [255]: ssh: Could not resolve hostname bogus: Name or service not known
```
then yes those need to stay. These are errors on connect due to a bogus hostname, not errors on disconnect handled by this PR.

cmwshang (approved):

PR #1381, opened 2021-04-23
Add cfgOptionDisplay()/cfgOptionIdxDisplay().

dwsteele:
Centralize the formatting of the configuration value for display to the user or passing on a command line.

For the new functions, if the value was set by the user via the command line, config, etc., then that exact value will be displayed. This makes it easier for the user to recognize the value and saves having to format it into something reasonable, especially for time and size option types.

Note that cfgOptTypeHash and cfgOptTypeList option types are not supported by these functions, but they are generally not displayed to the user as a whole.

This also fixes a bug in `config/load.c` where time values where not being formatted correctly in an error message.


src/command/archive/push/push.c
```diff
@@ -390,7 +390,7 @@ cmdArchivePush(void)
             {
                 THROW_FMT(
                     ArchiveTimeoutError, "unable to push WAL file '%s' to the archive asynchronously after %s second(s)",
-                    strZ(archiveFile), strZ(strNewDbl((double)cfgOptionInt64(cfgOptArchiveTimeout) / MSEC_PER_SEC)));
+                    strZ(archiveFile), strZ(cfgOptionDisplay(cfgOptArchiveTimeout)));
```
cmwshang, line 393: There is an inconsistency with archive-timeout and it is not in this file but in the archive/common.c. When I set the archive-time=0.1 in the pgbackrest.conf file, then run the check command I get:
P00  ERROR: [082]: WAL segment 000000020000000000000005 was not archived before the **100ms** timeout

But if I run `pgbackrest help archive-push archive-timeout` it states **current: 0.1** 
And if I change the archivePushTest, line 698 to `strLstAddZ(argList, "--archive-timeout=0.1");` I get:
ArchiveTimeoutError: unable to push WAL file 'bogus' to the archive asynchronously after 0.1 second(s)

Since I entered 0.1, I expected I would always see 0.1 second(s) as a display - not 100ms sometimes and 0.1 second(s) other times.

test/src/module/config/parseTest.c
```diff
@@ -1335,6 +1335,7 @@ testRun(void)
         TEST_RESULT_STR_Z(cfgOptionIdxStr(cfgOptPgSocketPath, 0), "/path/to/socket", "    pg1-socket-path is set");
         TEST_RESULT_INT(cfgOptionIdxSource(cfgOptPgSocketPath, 0), cfgSourceConfig, "    pg1-socket-path is config param");
         TEST_RESULT_BOOL(cfgOptionBool(cfgOptOnline), false, "    online not is set");
+        TEST_RESULT_STR_Z(cfgOptionDisplay(cfgOptOnline), "false", "    online display is false");
```
cmwshang, line 1338: I feel like we are trying to do a bit too much in this PR and not actually accomplishing the requirement: "If the value was set by the user via the command line, config, etc., then that exact value will be displayed." 
"true" and "false" are never used as values - only "y" and "n". The timeouts are an issue and even there they are inconsistent (see previous comment). I think we should rethink the requirements and what is trying to be accomplished. We might just want to solve the timeout/size problem as that appears to be what motivated this PR. Having a general Display function that is only useful sometimes may not be the answer.

cmwshang (changes requested):

src/command/archive/push/push.c
dwsteele: That's true, but that mismatch was not introduced by this commit. I didn't want to change any messages (except load.c which refers to options) in this commit because it then it is harder to evaluate the change.

test/src/module/config/parseTest.c
dwsteele: I did this because we have messages that use true/false the output booleans, e.g. backup.c lines 434 and 455. I didn't think this commit should try to go through and update existing messages to y/n, though I agree this makes more sense.

Agreed that this is all a bit messy but the goal here is to centralize some logic, not make all the error messages consistent.

src/command/archive/push/push.c
dwsteele: And this is a place where cfgOptionDisplayVar() could be used. The difference between these two places is in one the code is aware that it is working with an option, and in the other place it is not. Again, that hasn't changed for this commit but it would be good to have a general purpose solution.

cmwshang: If it is not appropriate for this commit, then when will it get done? The goal was to display the exact value displayed by the user.

test/src/module/config/parseTest.c
cmwshang: I don't feel it is simplifying the option display - I think it is now just adding more ways to display it, which can get confusing on which to use.

dwsteele: We definitely need to make boolean messages more consistent, but as discussed I have created a card for it: https://github.com/pgbackrest/pgbackrest/projects/2#card-59948715

src/command/archive/push/push.c
dwsteele: As discussed, added a card to make these more consistent. https://github.com/pgbackrest/pgbackrest/projects/2#card-59949268

cmwshang (approved):
Approving with caveat that https://github.com/pgbackrest/pgbackrest/projects/2#card-59949268 and https://github.com/pgbackrest/pgbackrest/projects/2#card-59948715 will be addressed at some point.

PR #1384, opened 2021-04-27
Update CipherType/CipherMode to StringId.

dwsteele:
As in 6cc521b6b, this allows option values and enums to be easily mapped together.


src/command/info/info.c
```diff
@@ -640,7 +640,7 @@ stanzaInfoList(List *stanzaRepoList, const String *backupLabel, unsigned int rep
         VariantList *repoSection = varLstNew();
 
         int stanzaStatusCode = -1;
-        unsigned int stanzaCipherType = 0;
+        uint64_t stanzaCipherType = 0;
```
cmwshang, line 643: I'm debating whether this should be `= cipherTypeNone` instead of `= 0`. We should never get to this function if there are no repositories configured so it's probably mute, but the intent was this would be the value of None which meant if we ever got to line 748 without going through the FOR loop at line 649 (which we won't), then this would be the right type. I know the value is overloaded and this was not the best way to go about this, but...

cmwshang (commented):
Review partial - part 1

src/command/info/info.c
dwsteele: No objections so done in c0d8f7eb649c4.

test/src/common/harnessStorage.c
```diff
@@ -293,15 +293,15 @@ hrnStoragePutLog(const Storage *storage, const char *file, const Buffer *buffer,
...
             param.cipherPass = TEST_CIPHER_PASS;
 
         if (param.compressType != compressTypeNone)
             strCatZ(log, "/");
 
-        strCatFmt(log, "enc[%s,%s]", strZ(cipherTypeName(param.cipherType)), param.cipherPass);
+        strCatFmt(log, "enc[%s,%s]", strZ(strIdToStr(param.cipherType)), param.cipherPass);
     }
 
     // Add a space if compression/encryption defined
```
cmwshang, line 307: On the line below this (line 308 - I'm unable to click on it) it has:
```
if (param.compressType != compressTypeNone || param.cipherType != cipherTypeNone)
```

Shouldn't it be checking param.cipherType !=0 as well? 
```
if (param.compressType != compressTypeNone || (param.cipherType != 0 && param.cipherType != cipherTypeNone)
```

cmwshang (commented):
Review complete with one possible issue.

test/src/common/harnessStorage.c
dwsteele: Yes, it should. Fixed in f3a74498cc0.

cmwshang (approved):

PR #1388, opened 2021-05-03
Convert option values in commands to StringId.

dwsteele:
Convert most of the remaining options that benefit from being StringIds. Since all the command modules can include config.h directly it makes sense to auto-generate these values instead of manually creating an enum for each one.

For the time being StringIds are not being auto-generated because the StringId code does not exist in Perl. However, the *_Z zero-terminated constants for each allowed option value are now auto-generated.


build/lib/pgBackRestBuild/Config/Build.pm
```diff
@@ -32,6 +32,7 @@ use constant BLDLCL_CONSTANT_OPTION_GROUP                           => '02-const
 use constant BLDLCL_CONSTANT_OPTION_GROUP_TOTAL                     => 'CFG_OPTION_GROUP_TOTAL';
 use constant BLDLCL_CONSTANT_OPTION                                 => '03-constantOption';
 use constant BLDLCL_CONSTANT_OPTION_TOTAL                           => 'CFG_OPTION_TOTAL';
+use constant BLDLCL_CONSTANT_OPTION_VALUE                           => '04-constantOption';
```
cmwshang, line 35: Might be nice if the value was '04-constantOptionValue' instead of '04-constantOption' - it's just a bit too similar to '03-constantOption' - but not really a big deal.

cmwshang (commented):
Review complete with one comment.

build/lib/pgBackRestBuild/Config/Build.pm
dwsteele: It should have been `04-constantOptionValue` but I forgot to rename it. Fixed in 059f6b4b23a.

PR #1389, opened 2021-05-06
Add repo-retention-history option to expire backup history.

pgstef:
This adds a new `repo-retention-history` option (in days) to clean the **backup.history** directory.

Without any default value, all the history is kept.
In minimal configuration (0 days), unexpired backups are always kept in history.

When full backup manifests expires, all differential/incremental manifests associated will expire too.

I've tried to keep the comments (and tests) relevant so we could more easily discuss this feature. I'm just not confident on the line-breaks in the tests though.

Coverage tests seems running fine on my laptop:

```bash
$ pgbackrest/test/test.pl --gen-only
$ pgbackrest/test/test.pl --vm-build --vm=f32
$ pgbackrest/test/test.pl --vm-out --dev --module=command --test=expire --coverage-only --vm=f32
2021-05-06 11:45:34.025 P00   INFO: P1-T1/1 - vm=f32, module=command, test=expire (13.88s)
2021-05-06 11:45:34.752 P00   INFO: tested modules have full coverage
2021-05-06 11:45:34.753 P00   INFO: writing C coverage report
2021-05-06 11:45:34.955 P00   INFO: TESTS COMPLETED SUCCESSFULLY (18s)
```


src/command/expire/expire.c
```diff
@@ -968,7 +1050,8 @@ cmdExpire(void)
 
                 // Remove all files on disk that are now expired
                 removeExpiredBackup(infoBackup, adhocBackupLabel, repoIdx);
-                removeExpiredArchive(infoBackup, timeBasedFullRetention, repoIdx);
+                removeExpiredArchive(infoBackup, timeBasedFullRetention, repoIdx);
```
dwsteele, line 1053: A bunch of trailing spaces were added here, please remove. It's better if your editor does this automatically.

src/command/expire/expire.c
```diff
@@ -863,6 +863,88 @@ removeExpiredBackup(InfoBackup *infoBackup, const String *adhocBackupLabel, unsi
...
+        FUNCTION_LOG_PARAM(UINT, repoIdx);
+    FUNCTION_LOG_END();
+
+    ASSERT(infoBackup != NULL);
+
+    MEM_CONTEXT_TEMP_BEGIN()
+    {
+        if (cfgOptionIdxTest(cfgOptRepoRetentionHistory, repoIdx))
+        {
+            // Get all the current backups in backup.info - these will not be expired.
```
dwsteele, line 884: Remove periods from single sentence comments -- check all comments below.

src/command/expire/expire.c
```diff
@@ -863,6 +863,88 @@ removeExpiredBackup(InfoBackup *infoBackup, const String *adhocBackupLabel, unsi
...
+
+                for (unsigned int historyIdx = 0; historyIdx < strLstSize(historyList); historyIdx++)
+                {
+                    const String *historyBackupFile = strLstGet(historyList, historyIdx);
+                    const String *historyBackupLabel = strLstGet(strLstNewSplitZ(historyBackupFile, "."), 0);
+
+                    // Keep the history manifests for unexpired backups and compare dates, extracted from backup labels since the 
+                    // retention is set in days.
+                    if (!strLstExists(currentBackupList, historyBackupLabel) &&
+                        strCmp(strSubN(historyBackupLabel, 0, 8), strSubN(minBackupLabel, 0, 8)) < 0)
```
dwsteele, line 920: I can't see why we need these strSubN() calls. This should work just as well:
```
if (!strLstExists(currentBackupList, historyBackupLabel) && strCmp(historyBackupLabel, minBackupLabel) < 0)
```

src/command/expire/expire.c
```diff
@@ -863,6 +863,88 @@ removeExpiredBackup(InfoBackup *infoBackup, const String *adhocBackupLabel, unsi
...
+            // Get all the current backups in backup.info - these will not be expired.
+            StringList *currentBackupList = strLstSort(infoBackupDataLabelList(infoBackup, NULL), sortOrderDesc);
+
+            // If the full backup history manifests are expired, we should expire diff and incr too. So, format the oldest full
+            // backup label to retain.
+            const time_t minTimestamp = time(NULL) - (time_t)(cfgOptionIdxUInt(cfgOptRepoRetentionHistory, repoIdx) * SEC_PER_DAY);
+            char buffer[16];
+            THROW_ON_SYS_ERROR(strftime(
+                buffer, sizeof(buffer), "%Y%m%d-%H%M%S", localtime(&minTimestamp)) == 0, AssertError, "unable to format time");
+            String *minBackupLabel = strNewFmt("%sF", buffer);
```
dwsteele, line 893: There is already a function in `command/backup/backup.c` to do this formatting, `backupLabelFormat()`. I think it would be best to move that function into `command/backup/common.c` and reuse it. You may also need to move some tests to get coverage.

src/command/expire/expire.c
```diff
@@ -863,6 +863,88 @@ removeExpiredBackup(InfoBackup *infoBackup, const String *adhocBackupLabel, unsi
...
+
+            // Get all the history years.
+            const StringList *historyYearList = strLstSort(storageListP(
```
dwsteele, line 896: This should be formatted as:
```
            const StringList *historyYearList = strLstSort(
                storageListP(
                    storageRepo(), STRDEF(STORAGE_REPO_BACKUP "/" BACKUP_PATH_HISTORY), .expression = STRDEF("^2[0-9]{3}$")),
                sortOrderDesc);
```

src/command/expire/expire.c
```diff
@@ -863,6 +863,88 @@ removeExpiredBackup(InfoBackup *infoBackup, const String *adhocBackupLabel, unsi
...
+                storageRepo(), STRDEF(STORAGE_REPO_BACKUP "/" BACKUP_PATH_HISTORY), .expression = STRDEF("^2[0-9]{3}$")),
+                sortOrderDesc);
+            
+            for (unsigned int historyYearIdx = 0; historyYearIdx < strLstSize(historyYearList); historyYearIdx++)
+            {
+                // Get all the backup history manifests.
+                const String *historyYear = strLstGet(historyYearList, historyYearIdx);
+                const StringList *historyList = strLstSort(storageListP(
```
dwsteele, line 904: This should be formatted as:
```
                const StringList *historyList = strLstSort(
                    storageListP(
                        storageRepo(), strNewFmt(STORAGE_REPO_BACKUP "/" BACKUP_PATH_HISTORY "/%s", strZ(historyYear)),
                        .expression = strNewFmt(
                            "%s\\.manifest\\.%s$",
                            strZ(backupRegExpP(.full = true, .differential = true, .incremental = true, .noAnchorEnd = true)),
                            strZ(compressTypeStr(compressTypeGz)))),
                    sortOrderDesc);
```

src/command/expire/expire.c
```diff
@@ -863,6 +863,88 @@ removeExpiredBackup(InfoBackup *infoBackup, const String *adhocBackupLabel, unsi
...
+                {
+                    const String *historyBackupFile = strLstGet(historyList, historyIdx);
+                    const String *historyBackupLabel = strLstGet(strLstNewSplitZ(historyBackupFile, "."), 0);
+
+                    // Keep the history manifests for unexpired backups and compare dates, extracted from backup labels since the 
+                    // retention is set in days.
+                    if (!strLstExists(currentBackupList, historyBackupLabel) &&
+                        strCmp(strSubN(historyBackupLabel, 0, 8), strSubN(minBackupLabel, 0, 8)) < 0)
+                    {
+                        LOG_INFO_FMT(
```
dwsteele, line 922: Formatting:
```
                        LOG_INFO_FMT(
                            "repo%u: remove expired history backup manifest %s", cfgOptionGroupIdxToKey(cfgOptGrpRepo, repoIdx),
                            strZ(historyBackupFile));
```

src/command/expire/expire.c
```diff
@@ -863,6 +863,88 @@ removeExpiredBackup(InfoBackup *infoBackup, const String *adhocBackupLabel, unsi
...
+                                storageRepoIdxWrite(repoIdx),
+                                strNewFmt(
+                                    STORAGE_REPO_BACKUP "/" BACKUP_PATH_HISTORY "/%s/%s",
+                                    strZ(historyYear), strZ(historyBackupFile)));
+                        }
+                    }
+                }
+
+                // Now that we removed the history manifests, we could remove the year directory if empty. However, it seems better
+                // to leave it there (and empty) to avoid the extra callback to the storage.
```
dwsteele, line 939: I think we should remove the year directory once it is empty. No extra call should be required if you track the number of files removed in the loop above.

But we should never remove the `backup.history` directory itself.

dwsteele (changes requested):
Please also rebase against master. Since `.github/workflows/test.yaml` has been updated I can't push to your branch.

src/command/expire/expire.c
pgstef: Oh, indeed, sorry.
(Just added trim_trailing_white_space_on_save and ensure_newline_at_eof_on_save to my sublimeText3 config)
Fixed in 6d2b9f5.

src/command/expire/expire.c
pgstef: Done in 001035e.

src/command/expire/expire.c
dwsteele: I'm surprised that Sublime Text is not honoring the project `.editorconfig` file. Maybe a setting?

pgstef: I've tried to add a specific plugins for that but it doesn't seem to load the `.editorconfig`. I'll have to investigate further on this later but at least the trim option is there now ;-)

src/command/expire/expire.c
pgstef: Fixed in 620979d

src/command/expire/expire.c
pgstef: Fixed in 620979d

src/command/expire/expire.c
pgstef: Fixed in 620979d

src/command/expire/expire.c
pgstef: I first wanted to focus on full backups to expire diff and incr too but that's enough with `minBackupLabel` indeed.
Removing strSubN in dafca49

src/command/expire/expire.c
pgstef: Agreed. I first wanted to try that but wondered if it should be part of this patch or a new one

src/command/expire/expire.c
pgstef: I thought about it too but that's if we **only** have `x.manifest.x` files in it (matching the storage expression). Else, the `storagePathRemoveP` function would ERROR returning `[39] Directory not empty`.

And I'm not really optimistic to use `.recurse = true` which would remove all the files there in case user would put some files there manually...

So:
- or we leave the directory there (even if empty)
- or we add an extra callback to really check if empty
- or we remove with `.recurse = true`
- or we let the ERROR happen and recommend the users to clean it manually

I was in favor of (1) :-) What's your thoughts about this?

src/command/expire/expire.c
cmwshang: Since this PR is what is driving the need to have `backupLabelFormat()` more available, I'd suggest the move to `common.c` be part of this PR.

src/command/expire/expire.c
cmwshang: @dwsteele Correct me if I'm wrong, but won't we create the directory each time we want to store a manifest in the history directory and the directory does not exist? If so, then if the directory is empty after expiring the files, we should remove it since it will be automatically created the next time if it is needed.

pgstef: The directory is created if missing (given a quick test I just ran). So, d799a7a adds the directory removal using `.recurse = true`. I've not updated the tests yet, just waiting your advice on this commit before doing so.

I list the *manifest* files in the year directory, count the number of removal and remove the directory if removed all the *manifest* found.

My concern is that users would add other files in that directory manually. The only way to make sure that the directory is empty is to trigger another look on disk without filtering any file expression. Perhaps a side note in the doc like **"When all history manifests files are expired, the year directory will be expired too."** (so, users know they probably shouldn't put anything else in those directories...)

src/command/expire/expire.c
pgstef: Is there any precedence in the include files? I'm struggling with an awkward error.

Moving `backupLabelFormat()` to `common.c`, I also have to include `infoBackup.h` to support **BackupType**. But then, I get a compilation error:
```
In file included from ./command/backup/common.h:11:0,
                 from ./info/manifest.h:13,
                 from info/manifest.c:17:
./info/infoBackup.h:126:1: error: unknown type name ‘Manifest’
 void infoBackupDataAdd(const InfoBackup *this, const Manifest *manifest);
 ^
cc1: warning: unrecognized command line option "-Wno-implicit-fallthrough" [enabled by default]
make: *** [.build/info/manifest.o] Error 1
```

Would it rather make sense to move the `BackupType` definition to `common.h`?

src/command/expire/expire.c
dwsteele: I have little sympathy for users placing their own files in the repository. It's certainly not a good reason for us to not clean up after ourselves.

In archive expiration, we simply remove the old directories regardless of the contents. The same is true for backups, where we remove the directory without checking for user-added files. I'm in favor of the same behavior here, i.e. use `recurse`.

This would be easier if we did it the way archive expiration does. If the entire year is less than the year of the minimum backup to retain, then just remove the dir (see https://github.com/pgbackrest/pgbackrest/blob/release/2.33/src/command/expire/expire.c#L656). Otherwise find and remove individual files.

Adding a note about this to the option reference is also fine, of course.

src/command/expire/expire.c
dwsteele: > Would it rather make sense to move the BackupType definition to common.h?

No, that will definitely cause a lot of breakage. The issue was the ordering of the header files in manifest.h, see dc0e6645cd5c4.

Pull from master and you should be good to go. Please don't rebase as we'll lose the PR history.

src/command/expire/expire.c
pgstef: Hm, you're right, that'd be totally better. And the behavior would be consistent with other expires.
I've then changed the year directory removal and added the tests in 9886532.
Thanks for your inputs!

src/command/expire/expire.c
pgstef: fetch upstream + merge upstream/master solved it indeed. Thanks !
I'll look at the tests and push afterwards then :-)

src/command/expire/expire.c
dwsteele: That looks good!

src/command/expire/expire.c
pgstef: Just pushed 520da96.
`--vm=u20 --c-only` tests are running fine on my laptop. I'll watch the CI to check if more changes are needed in the tests.
Thanks again for your help!

dwsteele: Yes, this looks pretty good, but also be sure to use this function to simply formatting in the tests.

dwsteele: @cmwshang It's a bit hard to migrate the tests here since this function is called through the parent function in the tests. Perhaps this is the time to combine the tests? It would make moving functions around less painful.

pgstef: Tests updated in d70c052. Including `common.h` was enough apparently.

dwsteele:
A bit of cleanup in fe7a90c640d5ed3ea153235f7159c4412e8d6204 and 860edfcdcd71d76bcb069c095404b6f0d64d9c2b.

The big change is to the tests in 6840463497ae9fb2f416bf9533b0989bce6eb485. The idea is to bring the tests into line with how we'd like to see them written going forward. You were copying the existing tests in this module, which was completely OK, but we'd like new tests to be written using the new macros. We are working on documentation to help with this. And don't feel bad, Cynthia and I have been doing the same!

Also, in general, there is no need to create a bunch of variables with calculations in the tests. Just figure out what the constant should be (in this case the backup labels) and use that. It makes the tests quite a bit more readable.

pgstef:
Hi,

Thanks for the explanation.

> Also, in general, there is no need to create a bunch of variables with calculations in the tests. Just figure out what the constant should be (in this case the backup labels) and use that. It makes the tests quite a bit more readable.

I used to read variable names to understand the purpose of those constants but ofc you're right, it can also be described in comments. Comments are more important / readable indeed.

> The big change is to the tests in 6840463. The idea is to bring the tests into line with how we'd like to see them written going forward. You were copying the existing tests in this module, which was completely OK, but we'd like new tests to be written using the new macros. We are working on documentation to help with this. And don't feel bad, Cynthia and I have been doing the same!

Indeed I reused the old macros since I already knew them and also for consistency in that test file. It's nice to see a working example with the new macros, they seem easier to use indeed. I'll re-use that example next time I write tests :-)

I follow the contributing update PR carefully indeed. Let me know if there's something I can help with on that topic ;-)

Do you plan to rewrite the current test files from old to new macros some days ? I could perhaps help with that (at least with the command modules for example)...

dwsteele (approved):

src/command/expire/expire.c
```diff
@@ -863,6 +863,100 @@ removeExpiredBackup(InfoBackup *infoBackup, const String *adhocBackupLabel, unsi
...
+            // If the full backup history manifests are expired, we should expire diff and incr too. So, format the oldest full
+            // backup label to retain.
+            const time_t minTimestamp = time(NULL) - (time_t)(cfgOptionIdxUInt(cfgOptRepoRetentionHistory, repoIdx) * SEC_PER_DAY);
+            String *minBackupLabel = backupLabelFormat(backupTypeFull, NULL, minTimestamp);
+
+            // Get all history years
+            const StringList *historyYearList = strLstSort(
+                storageListP(
+                    storageRepo(), STRDEF(STORAGE_REPO_BACKUP "/" BACKUP_PATH_HISTORY), .expression = STRDEF("^2[0-9]{3}$")),
+                sortOrderDesc);
```
cmwshang, line 895: It would be more efficient if the year list was sorted Ascending. For example, if a user would like to keep the last 2 years of history then we can stop looking when we get to the current-year - 2 years. 

Also, the historyList is only needed when we are going to look at the year that satisfies the request, so no need to get a list of files for each year that we are simply going to remove the directory. 

See https://github.com/cmwshang/pgbackrest/commit/a5cd80d36d5adfbaa242031cd1defa8e32564228 for changes and updates to tests to incorporate if you agree with the modifications. 

@dwsteele and @pgstef - I kept the currentBackupList and historyList Descending since it is a pattern we use when expiring files to expire the dependents first in the event something happens during file removal, we still have the Full - although I don't see how that would be helpful since the history files are useless if the backup is removed and by removing a dependent first, we would leave the history incorrect if an error occurs during the removal of a dependent.

cmwshang (commented):
Looks good but I have a couple of suggestions - please see my review comment.

src/command/expire/expire.c
dwsteele: > t would be more efficient if the year list was sorted Ascending. For example, if a user would like to keep the last 2 years of history then we can stop looking when we get to the current-year - 2 years.

That looks good to me.

> Also, the historyList is only needed when we are going to look at the year that satisfies the request, so no need to get a list of files for each year that we are simply going to remove the directory.

Very good point -- saving IO is always a worthwhile.

> although I don't see how that would be helpful since the history files are useless if the backup is removed

For the purposes of history, each manifest is valuable by itself. For instance, if you wanted to do trend analysis you don't need every manifest, maybe just one per month or one per week max. For these purposes full, diff, incr does matter because you can still measure total database size and WAL generated per day, probably the most important metrics. Anyway, if there are gaps because of failed deletes they would be outside the retention period anyway so it doesn't seem very important.

dwsteele: Changed does to doesn't above.

pgstef: The suggested change looks good to me too. Thanks @cmwshang  :-)

cmwshang (approved):

PR #1395, opened 2021-05-12
Add automatic GCS authentication for GCE instances.

dwsteele:
When running on a GCE instance the authentication token can be pulled directly from the instance metadata. This is configured with repo-gcs-key-type=auto.

In a separate commit (26fefa6a), moved the code that parses the token response into a separate function, storageGcsAuthToken(), since it is now needed by two key types. This drastically improves the readability of the main commit.

NOTE: this PR is best viewed without whitespace changes since there were indentation changes.


src/storage/gcs/storage.c
```diff
@@ -98,6 +98,61 @@ struct StorageGcs
...
+    time_t timeExpire;
+} StorageGcsAuthTokenResult;
+
+static StorageGcsAuthTokenResult
+storageGcsAuthToken(HttpRequest *request, time_t timeBegin)
+{
+    FUNCTION_TEST_BEGIN();
+        FUNCTION_TEST_PARAM(HTTP_REQUEST, request);
+        FUNCTION_TEST_PARAM(TIME, timeBegin);
+    FUNCTION_TEST_END();
```
wieck: Some indentation issues here

src/storage/gcs/storage.c
```diff
@@ -98,6 +98,61 @@ struct StorageGcs
     HttpClient *authClient;                                         // Client to service auth requests
 };
 
+/***********************************************************************************************************************************
+Parse HTTP JSON response containing an authentication token and expiration
+
+Note that the function is intended to run directly in the caller's mem context and results will be placed in the caller's prior mem
+context.
```
wieck: Isn't that somehow backwards? Shouldn't the caller decide which memory context the result ends up in?

And if this is the intended behavior, then it might be good to indicate that somehow in the function name, like naming it storageGcsAuthTokenIntoPrior() (don't particularly like that name but it is just an example).

src/storage/gcs/storage.c
```diff
@@ -213,35 +261,39 @@ storageGcsAuthToken(StorageGcs *this, time_t timeBegin)
...
-        }
+Based on the documentation at https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances#applications
+***********************************************************************************************************************************/
+static StorageGcsAuthTokenResult
+storageGcsAuthAuto(StorageGcs *this, time_t timeBegin)
+{
+    FUNCTION_TEST_BEGIN();
+        FUNCTION_TEST_PARAM(STORAGE_GCS, this);
+        FUNCTION_TEST_PARAM(TIME, timeBegin);
+    FUNCTION_TEST_END();
```
wieck, line 285: Same indentation issue

src/build/config/config.yaml
```diff
@@ -1536,7 +1536,11 @@ option:
     group: repo
     secure: true
     command: repo-type
-    depend: repo-gcs-bucket
+    depend:
+      option: repo-gcs-key-type
+      list:
+        - service
```
wieck, line 1542: Why does auto not appear in these lists?

src/storage/gcs/storage.c
dwsteele: Not sure what you mean. Is it that you don't expect FUNCTION_TEST_PARAM() to be indented? That's conventions we used in these debug preambles to make them a bit more readable.

src/storage/gcs/storage.c
dwsteele: Yeah, we could pass in the mem context, but moving to prior is a pattern we use pretty regularly. If this function were extern'd we probably make the caller do their own move, but since it is module level I think it is fine with the documentation. Thoughts, @cmwshang?

src/storage/gcs/storage.c
dwsteele: Same comment as above.

src/build/config/config.yaml
dwsteele: What this is saying is that `repo-gcs-key` is only valid (i.e. settable) when `key-type` is `service` or `token`, since only those types require the user to supply a key. When the key type is auto we get the key programmatically so letting the user set it would be misleading.

src/storage/gcs/storage.c
wieck: Oh ... I missed that it was BEGIN, PARAM, PARAM, END ... I read all of them as the same macro. Never mind.

src/build/config/config.yaml
wieck: Thanks.

wieck (approved):

src/storage/gcs/storage.c
cmwshang: > Yeah, we could pass in the mem context, but moving to prior is a pattern we use pretty regularly. If this function were extern'd we probably make the caller do their own move, but since it is module level I think it is fine with the documentation. Thoughts, @cmwshang?

If it were not a static module-level function (i.e. extern'd), I would agree with Jan.

PR #1396, opened 2021-05-13
Replace getopt_long() with custom implementation.

dwsteele:
getopt_long() requires an exhaustive list of all possible options that may be found on the command line. Because of the way options are indexed (e.g. repo1-4, pg1-8) optionList[] has 827 entries and we have kept it small by curtailing the maximum indexes very severely. Another issue is that getopt_long() scans the array sequentially so parsing gets slower as the index maximums increase.

Replace getopt_long() with a custom implementation that behaves the same but allows options to be parsed with a function instead of using optionList[]. This commit leaves the list in place in order to focus on the getopt_long() replacement, but cfgParseOption() could be replaced with a more efficient implementation that removes the need for optionList[].

This implementation also fixes an issue where invalid options are misreported in the error message if they only have one dash, e.g. -config.

Tests were added at 0825428fef00, 2b8d2daca1, 34dd6636b, and 384f2470 to check that previously untested getopt_long() behavior doesn't change.

NOTE: processing options before commands/parameters would be a more natural flow, but this way makes a much better diff. We can reverse it before commit or reverse it in a follow-up commit.

REFERENCE SOURCE: https://sourceware.org/git/?p=glibc.git;a=blob;f=posix/getopt.c;h=958111bad0f4f379847637fe6a20cee56b049597;hb=HEAD


cmwshang (approved):
Great job - I tried my best to break this but couldn't. Code and comments look good as well.

dwsteele:
> Great job - I tried my best to break this but couldn't.

Haha, from you that's really saying something.

wieck (approved):
Looks all god to me and helps to get familiar with the overall structure.

dwsteele:
Thank you @cmwshang @wieck!

PR #1398, opened 2021-05-14
WIP: Simplify catalog version handling.

dwsteele:


dwsteele:
Close since we decided this is not the direction we'd like to go.

PR #1400, opened 2021-05-14
Update contributing documentation.

cmwshang:
Given the extent of the changes, it may be easier to view the CONTRIBUTING.md and README.md markdown versions in this branch https://github.com/cmwshang/pgbackrest/tree/cs-contributing-ci.


doc/xml/contributing.xml
```diff
@@ -118,191 +131,250 @@
...
+
+// Declare getters and setters inline for the publicly visible variables.
+// Only setters require "Set" appended to the name.
+__attribute__((always_inline)) static inline unsigned int
+myObjMyData(const MyObj *const this)
+{
+    return THIS_PUB(MyObj)->myData;    // Use the built-in THIS_PUB macro
+}
+
+// TYPE and FROMAT macros for function logging
```
pgstef, line 188: **TYPE and FORMAT** ?

CONTRIBUTING.md
```diff
@@ -31,11 +39,171 @@ pgbackrest-dev => Clone pgBackRest repository
...
+
+// Declare getters and setters inline for the publicly visible variables.
+// Only setters require "Set" appended to the name.
+__attribute__((always_inline)) static inline unsigned int
+myObjMyData(const MyObj *const this)
+{
+    return THIS_PUB(MyObj)->myData;    // Use the built-in THIS_PUB macro
+}
+
+// TYPE and FROMAT macros for function logging
```
pgstef, line 86: **TYPE and FORMAT** ?

CONTRIBUTING.md
```diff
@@ -164,189 +331,275 @@ pgbackrest/test/test.pl --vm=u18 --dev --module=mock --test=archive --run=2
...
+// Run the test
+TEST_RESULT_STR_Z(
+    infoRender(), "No stanzas exist in the repository.\n", "text output - no stanzas");
+```
 
-**Section 2**
+#### Storing a file
 
-To better explain this section, `CFGOPT_ONLINE` will be used as an example:
+Sometimes it is desirable to store or manipulate files before or during a test and then confirm the contents. The [harnessStorage.h](https://github.com/pgbackrest/pgbackrest/blob/master/test/src/common/harnessStorage.h) file contains macros (e.g. `HRN_STORAGE_PUT` and `TEST_STORAGE_GET`) for doing this. In addition, `HRN_INFO_PUT` is convenient for writing out info files (archive.info, backup.info, backup.manifest) since it will automatically add header and checksum information.
```
pgstef, line 388: Does this mean that `HRN_STORAGE_PUT` should be used everywhere instead of `TEST_RESULT_VOID(storagePutP(...` or is there some caveats where both should/could be used?

pgstef (commented):
The `.md` being easier to read, I've left the comments on this rather than on `.xml`.
Not being an English native speaker, I've only found one small typo.

That said, I like this modification. IMHO, the coding example about memory contexts is great. I better understood the layers involved thanks to this example.

The guide says `--coverage-only - selects a vm to use (although best when overridden with --vm=f32)`. I've noticed this too on my laptop. I have to build the f32 vm since it's not built by default.

Shouldn't we change the test script to build the f32 vm by default and use that one for coverage tests then?

dwsteele:
> Shouldn't we change the test script to build the f32 vm by default and use that one for coverage tests then?

f32 is obsolete and will be removed soon. We'll be going to u20 for general coverage testing, since the gcc/lcov versions work well together. I think what we need here is to change `--coverage-only` so it does not automatically select a VM.

CONTRIBUTING.md
cmwshang: good catch - thanks! Fixed in 269b9f13e715b061cc3d006c62e0b32525af3f3f

CONTRIBUTING.md
cmwshang: We're trying to encourage everyone to use the newer macros - the old ways still work, but we'd rather not detail the history of the progression to the macros - rather just outline the way it should be done. Maybe we can work on a project to update the tests to be more consistent throughout...

cmwshang:
> > Shouldn't we change the test script to build the f32 vm by default and use that one for coverage tests then?
> 
> f32 is obsolete and will be removed soon. We'll be going to u20 for general coverage testing, since the gcc/lcov versions work well together. I think what we need here is to change `--coverage-only` so it does not automatically select a VM.

Updated in commit 269b9f13e715b061cc3d006c62e0b32525af3f3f : Removed references to --coverage-only since it is now the default for unit tests. Provided more details in the Code Coverage section.

pgstef (approved):

dwsteele (changes requested):

doc/xml/contributing.xml
```diff
@@ -321,87 +393,324 @@
...
+                <code-block>
+  set:
+    type: string
+    command:
+      backup:
+        depend:
+          option: stanza
+        required: false
+      restore:
+        default: latest
```
dwsteele, line 596: The default command role has been renamed to main in PR #1410. So if we commit that first we can add command role here, or leave it in the config.yaml documentation.

CONTRIBUTING.md
```diff
@@ -2,7 +2,15 @@
...
+
+Code fixes or new features can be submitted via pull requests. Ideas for new features and improvements to existing functionality or documentation can be [submitted as issues](https://github.com/pgbackrest/pgbackrest/issues). You may want to check the [Project Boards](https://github.com/pgbackrest/pgbackrest/projects) to see if your suggestion has already been submitted.
+
+Bug reports should be [submitted as issues](https://github.com/pgbackrest/pgbackrest/issues). Please provide as much information as possible to aid in determining the cause of the problem.
+
+You will always receive credit in the [release notes](http://www.pgbackrest.org/release.html) for your contributions.
+
+Coding standards are defined in [CODING.md](https://github.com/pgbackrest/pgbackrest/blob/master/CODING.md) and some important coding details and an example are provided in the [Coding](#coding) section below. At a minimum, unit tests must be written and run and the documentation generated before submitting a Pull Request; see the [Testing](#testing) section below for details.
```
pgstef, line 13: Perhaps another idea here. Would it be interesting to outline how to work in PR's ? Like "the reviewer will solve the comment once fixed,..."
I know it since you guided me on that path but perhaps it could be nice to write it here too ?

cmwshang: > Perhaps another idea here. Would it be interesting to outline how to work in PR's ? Like "the reviewer will solve the comment once fixed,..."
> I know it since you guided me on that path but perhaps it could be nice to write it here too ?

Agreed, and there is a card to address this in addition to a few other things https://github.com/pgbackrest/pgbackrest/projects/2#card-54174338 - but we are not sure if that will make it into this PR or be worked on in a separate PR.

CONTRIBUTING.md
```diff
@@ -164,189 +329,273 @@ pgbackrest/test/test.pl --vm=u18 --dev --module=mock --test=archive --run=2
...
+- `online` - the name of the option
 
-- `TEST_RESULT_STR` - Test the actual value of the string returned by the function.
+- `type` - the type of the option. Valid values for types are: `boolean`, `hash`, `integer`, `list`, `path`, `size`, `string`, and `time`
 
-- `TEST_RESULT_UINT` / `TEST_RESULT_INT` - Test for an unsigned integer / integer.
 
-- `TEST_RESULT_BOOL` - Test a boolean return value.
+- `negate` - being a command-line only boolean option, this rule would automatically default to false so it must be defined if the option is negatable. Ask yourself if negation makes sense, for example, would a --dry-run option make sense as --no-dry-run? If the answer is no, then this rule can be omitted as it would automatically default to false. Any boolean option that cannot be negatable, must be a command-line only and not a configuration file option as all configuration boolean options must be negatable.
```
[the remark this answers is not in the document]
cmwshang: Fixed, commit bce4ce3e3fb92323d56fd86f66d8caca0fd16df4

doc/xml/contributing.xml
```diff
@@ -118,191 +131,250 @@
...
-                </execute>
-            </execute-list>
+            <p>Logging is also used for providing information to the user via the <code>LOG_*()</code> macros, such as <code>LOG_INFO("some informational message")</code> and <code>LOG_WARN_FMT("no prior backup exists, %s backup has been changed to full", strZ(cfgOptionDisplay(cfgOptType)))</code> and also via <code>THROW_*()</code> macros for throwing an error. See <link url="{[github-url-src-common]}/log.h">log.h</link> and <link url="{[github-url-src-common]}/error.h">error.h</link> for more details and the <link section="/coding/coding-example">Coding Example</link> below.</p>
         </section>
 
-        <section id="with-docker" depend="/environment">
-            <title>With Docker</title>
-
-            <p>Build a container to run tests.  The vm must be pre-configured but a variety are available.  The vm names are all three character abbreviations, e.g. <id>u18</id> for <proper>Ubuntu 18.04</proper>.</p>
+        <section id="coding-example">
```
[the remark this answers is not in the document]
cmwshang: Fixed, commit 4f0913f0db4c68c68f31bc44b01dc8a32d2c9ead

CONTRIBUTING.md
dwsteele: Agreed, let's keep that for a separate PR.

dwsteele (approved):
This looks good to me now with the updates in c1e933199f2111d304e.

CONTRIBUTING.md
pgstef: Seen the working example in `repo-retention-history` PR. It would perhaps be interesting to update the tests to be consistent but stating that we're now only adding/using new macros is fine :-)

PR #1413, opened 2021-05-21
Add PostgreSQL 14 support.

dwsteele:
There are no code changes from PostgreSQL 13 so simply add the new version.

Add CATALOG_VERSION_NO_MAX to allow the catalog version to "float" during the PostgreSQL beta/rc period so new pgBackRest versions are not required when the catalog version changes.

Update the integration tests to handle new PostgreSQL startup messages added in https://github.com/postgres/postgres/commit/df9384492b89aac370ab9d12eb89375aeb38a1d4.

### DIFF
--- a/test/lib/pgBackRestTest/Common/ContainerTest.pm
+++ b/test/lib/pgBackRestTest/Common/ContainerTest.pm
@@ -482,7 +482,7 @@ sub containerBuild
             {
                 $strScript .=
                     "    echo 'deb http://apt.postgresql.org/pub/repos/apt/ " .
-                    $$oVm{$strOS}{&VM_OS_REPO} . '-pgdg main' .
+                    $$oVm{$strOS}{&VM_OS_REPO} . '-pgdg main' . ($strOS eq VM_U18 ? ' 14' : '') .
                         "' >> /etc/apt/sources.list.d/pgdg.list && \\\n" .
                     "    wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - && \\\n" .
                     "    apt-get update && \\\n" .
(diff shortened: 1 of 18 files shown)


test/lib/pgBackRestTest/Common/ContainerTest.pm
```diff
@@ -482,7 +482,7 @@ sub containerBuild
             {
                 $strScript .=
                     "    echo 'deb http://apt.postgresql.org/pub/repos/apt/ " .
-                    $$oVm{$strOS}{&VM_OS_REPO} . '-pgdg main' .
+                    $$oVm{$strOS}{&VM_OS_REPO} . '-pgdg main' . ($strOS eq VM_U18 ? ' 14' : '') .
```
cmwshang, line 485: Is this something that will have to change every time now for u18 when a new PG version is added? We don't have pg version 14 yet so I'm not 100% clear on what this is doing.

cmwshang (commented):
Review complete but a couple questions...

test/lib/pgBackRestTest/Env/HostEnvTest.pm
```diff
@@ -293,6 +294,7 @@ sub dbControlVersion
         &PG_VERSION_11 => 1100,
         &PG_VERSION_12 => 1201,
         &PG_VERSION_13 => 1300,
+        &PG_VERSION_14 => 1300,
```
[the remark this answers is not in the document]
dwsteele: The control version did not change for PG14, which is part of the reason we are doing the song and dance with the catalog version.

test/lib/pgBackRestTest/Common/ContainerTest.pm
dwsteele: This adds the pre-release repo for PG14. Once it is prod it will move into main and we can remove the extra `14` repo. This happens each year (see history for details).

This would be better as a setting in VmTest.pm, but so it is for now.

test/lib/pgBackRestTest/Env/HostEnvTest.pm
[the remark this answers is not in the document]
cmwshang: Oh, boy - yeah, ok, you did mention this.

test/lib/pgBackRestTest/Common/ContainerTest.pm
cmwshang: ok

cmwshang (approved):

PR #1420, opened 2021-06-04
Update command/expire test to use standard patterns.

cmwshang:


test/src/module/command/expireTest.c
```diff
@@ -134,7 +129,7 @@ testRun(void)
         "\"backup-archive-stop\":\"000000010000000000000011\",\"backup-info-repo-size\":2369186,"
         "\"backup-info-repo-size-delta\":346,\"backup-info-size\":20162900,\"backup-info-size-delta\":8428,"
         "\"backup-prior\":\"20181119-152900F\",\"backup-reference\":[\"20181119-152900F\"],"
-        "\"backup-timestamp-start\":%" PRIu64 ",\"backup-timestamp-stop\":%" PRIu64 ",\"backup-type\":\"diff\","
+        "\"backup-timestamp-start\":%" PRIu64 ",\"backup-timestamp-stop\":%" PRIu64 ",\"backup-type\":\"incr\","
```
dwsteele, line 132: This should be diff.

dwsteele (changes requested):

test/src/module/command/expireTest.c
cmwshang: Yep. Made the change in the wrong spot. It isn't used in the tests - just nice to have the label match the type. Fixed in two places in https://github.com/pgbackrest/pgbackrest/pull/1420/commits/334c502225450dc7215f27369d89bd1feb840eeb

dwsteele (approved):
Added some calls to TEST_STORAGE_EXISTS() in 7e380ecf5db54d7.

Also changed the return value of archiveExpectList() in 67928632519ab6 since this saves a lot of calls to strZ(). I'll revert that if you don't like it.

cmwshang:
> Added some calls to TEST_STORAGE_EXISTS() in [7e380ec](https://github.com/pgbackrest/pgbackrest/commit/7e380ecf5db54d7fae8a456434941e9a2d99a3dd).
> 
> Also changed the return value of archiveExpectList() in [6792863](https://github.com/pgbackrest/pgbackrest/commit/67928632519ab645d60a00c46f1f9aac8b59f84c) since this saves a lot of calls to strZ(). I'll revert that if you don't like it.

Both of these look good, thanks.

PR #1421, opened 2021-06-08
Add support for more Pack types.

dwsteele:
Since the pack type was stored in 4 bits, only 15 values were allowed (0 was reserved).

Allow virtually unlimited types by storing type info in a base-128 encoded integer following the tag when the type bits in the tag are set to 0xF.

Also separate the type IDs used in the pack (PackTypeMap) from those presented to the user (PackType). The prior PackType enum exposed implementation details to the user, e.g. pckTypeUnknown.


cmwshang (approved):
Looks good - but I did fix a comment in commit e0097329e6e31aba105c4126c4c9b44f8a98e52d

PR #1425, opened 2021-06-10
WIP: POC to see if this fixes regression/doc generation issues on rhel based systems

crunchyheath:
Issue:  Currently if you try to run the doc generation/regression target on a CentOS 8.4 base OS with the latest docker-ce, the "systemctl stop" commands all hang and never recover.  After some investigation, it appears that the commands are hanging because systemd is incorrectly trying to kill the wrong process ID when terminating the service.  The reason for this pocess ID mismatch is still unknown.  However, it appears to be related to the following Red Hat issue: 
https://bugzilla.redhat.com/show_bug.cgi?id=1875056

To make sure that the problem isn't solely caused by the 2.34 release, I ran a sanity check by building against the previous 2.33 release on the same system. The result was that it hangs as well, which makes it appear to not be related to anything that has changed in the backrest code for 2.34, but instead seems to be an issue with the docker/OS combination that is now available on these machines.

IMPORTANT NOTE: This change I included below will definitely break the current Debian systems and is only intended as a Proof of Concept.  In its current form, it solves the issue of broken documentation generation on a RHEL-based system.  As this is strictly just a proof of concept, the path to the pg_ctl binary is hardcoded to use PG10.

With this change, documentation generation succeeds on RHEL-based systems, but we also noticed an added bonus-- RHEL builds the docs significantly faster using pg_ctl commands than it did using the previous systemctl method.  However, there is one caveat: the final "/upgrade-stanza" stage fails because of these hardcoded values, which is expected.  I am also aware that this sort of implementation will change the final documents to tell the end user to use the pg_ctl commands instead of running postgres as a service, which is not desirable. As this is a Work in Progress and definitely not a final fix for this issue, please let me know if you need any further testing or changes to this branch to make this into something we can successfully and non-invasively incorporate into the code base.

### DIFF
--- a/doc/xml/user-guide.xml
+++ b/doc/xml/user-guide.xml
@@ -220,25 +220,25 @@
         <variable key="pg-cluster-create-upgrade" if="{[os-type-is-debian]}">pg_createcluster {[pg-version-upgrade]} {[postgres-cluster-demo]}</variable>
 
         <variable key="pg-cluster-start" if="{[os-type-is-debian]}">pg_ctlcluster {[pg-version]} {[postgres-cluster-demo]} start</variable>
-        <variable key="pg-cluster-start" if="{[os-type-is-rhel]}">systemctl start postgresql-{[pg-version]}.service</variable>
+	<variable key="pg-cluster-start" if="{[os-type-is-rhel]}">/usr/pgsql-10/bin/pg_ctl start -D /var/lib/pgsql/10/data</variable>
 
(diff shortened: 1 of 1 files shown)


dwsteele:
I've been looking through this and I don't see any clean way to display `systemctl` while running `pg_ctl`. Part of the issue is that we show the output from these commands in some cases and expect regular expressions to match. So the user would see one command displayed but the output from another.

We could switch to using `pg_ctl` but that strikes me as a bad move since that is not how users interact with postgres on RHEL.

crunchyheath:
This does not seem like a possibility and we have figured out a workaround for this issue.  I will close this and update it if we find a more permanent workaround for the issue.

PR #1426, opened 2021-06-11
Update command/stanza test to use standard patterns.

cmwshang:
No core code changes; only changes for stanza-create, stanza-upgrade and stanza-delete commands unit tests.


dwsteele (approved):
Added a few test titles at efc5b29a51ced. Looks like more could be done there if you want to have a look.

PR #1438, opened 2021-06-24
Detect errors in S3 multi-part upload finalize.

dwsteele:
Multi-part upload may fail despite returning an HTTP success code. Check for the ETag field in the result and if not present consider the upload to have failed. This will trigger a retry at the local job level. Worst case, an error will be returned to postgres and it will try again later.


test/src/module/storage/s3Test.c
```diff
@@ -686,11 +686,69 @@ testRun(void)
...
+                        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
+                        "<CompleteMultipartUpload>"
+                        "<Part><PartNumber>1</PartNumber><ETag>WxRt1</ETag></Part>"
+                        "<Part><PartNumber>2</PartNumber><ETag>WxRt2</ETag></Part>"
+                        "</CompleteMultipartUpload>\n");
+                testResponseP(
+                    service,
+                    .content =
+                        "<?xml version=\"1.0\" encoding=\"UTF-8\"?>"
+                        "<Error><Code>AccessDenied</Code><Message>Access Denied</Message></Error>");
```
cmwshang, line 730: If I change line 730 and 750 to be 
`"<CompleteMultipartUploadResult></CompleteMultipartUploadResult>");`

it is the absence of an Etag element, so I expected

cmwshang (commented):
question

test/src/module/storage/s3Test.c
cmwshang: Ah, I changed the content-length from 110 to 101 and it worked.

cmwshang (approved):

PR #1452, opened 2021-07-07
Fix detection of circular symlinks.

dwsteele:
Links were followed before they were checked for validity so a circular link would send the manifest build into endless recursion leading to a crash. Fix by moving the recursion after the link check.

Note that this issue has existed since the C migration and was not introduced by the refactor in eba013b4.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -14,6 +14,21 @@
     <release-list>
         <release date="XXXX-XX-XX" version="2.35dev" title="UNDER DEVELOPMENT">
             <release-core-list>
+                <release-bug-list>
+                    <release-item>
+                        <github-issue id="1447"/>
+                        <github-pull-request id="1452"/>
+
+                        <release-item-contributor-list>
+                            <release-item-ideator id="rohit.raveendran"/>
+                            <release-item-contributor id="david.steele"/>
+                            <release-item-reviewer id="stefan.fercot"/>
+                        </release-item-contributor-list>
+
+                        <p>Fix detection of circular symlinks.</p>
+                    </release-item>
+                </release-bug-list>
+
                 <release-improvement-list>
                     <release-item>
                         <github-issue id="1445"/>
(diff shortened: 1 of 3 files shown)


pgstef:
hm, I'm just curious. How did you reproduce the segfault ?

I tried with:

```
$ ls -l /var/lib/pgsql/13/data/pg_wal
lrwxrwxrwx. 1 postgres postgres 21 Jul  8 08:58 /var/lib/pgsql/13/data/pg_wal -> /var/lib/pgsql/13/wal
$ ls -l /var/lib/pgsql/13/data/pg_wal/pg_wal
lrwxrwxrwx. 1 postgres postgres 21 Jul  8 08:59 /var/lib/pgsql/13/data/pg_wal/pg_wal -> /var/lib/pgsql/13/wal
```

But then, the backup command tells me `ERROR: [041]: unable to get info for path/file '...': [40] Too many levels of symbolic links`.

Using this branch, I get `ERROR: [070]: link 'pg_wal/pg_wal' (/var/lib/pgsql/13/wal) destination is the same directory as link 'pg_wal' (/var/lib/pgsql/13/wal)`. This error is way more explicit for users / debug. So, +1 ! :-)

dwsteele:
> How did you reproduce the segfault ?

Without the fix the new test case throws `assertion 'memContextCurrentStackIdx < MEM_CONTEXT_STACK_MAX - 1' failed` on a debug build. This would be a segfault in a production build.

But the error you get may be based on your distro, memory, debug builds, etc. The point of this patch is just to stop the runaway recursion.

pgstef:
Tested on Ubuntu:

```
P00   INFO: backup command begin 2.34:...
Segmentation fault (core dumped)
```
```
P00   INFO: backup command begin 2.35dev: ...
ERROR: [070]: link 'pg_wal/pg_wal' destination '/var/lib/postgresql/13/main/pg_wal' is in PGDATA
P00   INFO: backup command end: aborted with exception [070]
```

pgstef (approved):

PR #1455, opened 2021-07-08
Change level of backup/restore copied file logging to detail.

dwsteele:
This will make the info level less noisy and more usable.


pgstef (approved):
Tried it. The INFO log is really less noisy now and IMHO much more usable by default for console log level.

I'm still a bit concerned with log-level-file not being detail so perhaps we should emphases on this in the release note so it's up to the users to change their config accordingly if they don't want to lose the progress status ?

dwsteele:
> so perhaps we should emphases on this in the release note

Done in 31355aab0dd and had Cynthia review.

I'd really like to increase the file log level but that could be a *really* nasty surprise for uses with async logging at info level. Also, even for backup/restore there were already a *lot* more messages at detail level that could easily increase log sizes by several times. It just doesn't seem safe.

PR #1458, opened 2021-07-12
WIP: Check that pgBackRest the same consistent view of the filesystem as PostgreSQL

dwsteele:
In order to make a good copy of PGDATA we need to have the same view of the filesystem as PostgreSQL. This includes any pages in the file system cache.

This code repeatedly writes a small file to PGDATA in Postgres using COPY then immediately checks the contents of the file in the pgBackRest process running on the same host. The idea is to catch any issues with sidecars, novel filesystems, etc. that might cause an inconsistent backup.

The same check can be done on pg_control, but because of the relatively long delay between pg_control being updated and control being returned to pgBackRest it is not nearly as good a test.

See dbSyncCheckHelper() https://github.com/pgbackrest/pgbackrest/compare/integration...dev-fs-consistent?expand=1#diff-63e96f16bd384bbf5f5b4fee87ce218f81dc89ab2f854ba853ffdf9736f77a50R470 for the important code. The rest is just support.

This patch is not ready for a full review. I just want to see if the basic idea makes sense or could be improved.


src/db/db.c
```diff
@@ -452,6 +461,113 @@ dbBackupStop(Db *this)
...
+        char fileActual[DB_SYNC_TOTAL][DB_SYNC_SIZE];
+        ssize_t fileActualSize[DB_SYNC_TOTAL];
+
+        for (unsigned int fileIdx = 0; fileIdx < DB_SYNC_TOTAL; fileIdx++)
+        {
+            // Write data into the file using a copy on the pg server
+            CHECK(pgClientQuery(pgClient, fileCmd[fileIdx]) == NULL);
+
+            // Stat the file
+            THROW_ON_SYS_ERROR_FMT(stat(file, &fileStat[fileIdx]) == -1, FileReadError, "unable to stat '%s'", file);
```
wieck, line 507: Since we are opening the file just below, should we move this check after the open(2) and use fstat(2)?

src/db/db.c
```diff
@@ -452,6 +461,113 @@ dbBackupStop(Db *this)
...
+
+            // Read the file
+            int fd = open(file, O_RDONLY, 0);
+            THROW_ON_SYS_ERROR_FMT(fd == -1, FileOpenError, "unable to open '%s'", file);
+
+            fileActualSize[fileIdx] = read(fd, fileActual[fileIdx], DB_SYNC_SIZE);
```
wieck, line 513: Reading exactly the size of the data we intended to write, without whatever end of line marker COPY produced, misses the possibility that there could be added junk at the end of the file.

src/db/db.c
```diff
@@ -452,6 +461,113 @@ dbBackupStop(Db *this)
...
+            THROW_ON_SYS_ERROR_FMT(fileActualSize[fileIdx] == -1, FileReadError, "unable to read '%s'", file);
+
+            THROW_ON_SYS_ERROR_FMT(close(fd) == -1, FileCloseError, "unable to close '%s'", file);
+        }
+
+        // Check results
+        for (unsigned int fileIdx = 0; fileIdx < DB_SYNC_TOTAL; fileIdx++)
+        {
+            CHECK(fileStat[fileIdx].st_size == DB_SYNC_SIZE + 1);
```
wieck, line 522: This assumes that the end of line is exactly one byte and it doesn't check that it is that EOL character instead of some other junk. I think a better way would be to make this read buffer larger by 4 bytes, attempt the read in that size, compare that the number of bytes actually read matches st_size from fstat(2), then right-strip every '\r' and '\n' character and strcmp(3) the result.

The reason for allocating 4 more bytes is that a read(2) does not NUL terminate the string, so we need to first make sure that read(2) in the length of DB_SYNC_SIZE+4 did return less bytes than that and NUL terminate the data at the offset, read(2) returned. We expect an end-of-line sequence, which may be LF or CRLF, but handle the case that there could be some garbage after that.

We may not expect to run on anything but Unix and I am not aware of any filesystem or setting that could cause the end-of-line to be CRLF. But that could change tomorrow.

wieck (commented):
I think dbSyncCheckHelper() could be improved. Comments inline.

dwsteele:
Closing this in favor of #1581. The method in this PR is more rigorous but requires additional permissions and writes files into PGDATA. It is possible that this method could be revived for the `check` command.

PR #1460, opened 2021-07-13
Loop while waiting for checkpoint LSN to reach replay LSN.

pgstef:
Based on [this card](https://github.com/pgbackrest/pgbackrest/projects/2#card-64554287).

After standby replayed the target lsn, loop and wait that checkpoint writes it too.

Edit: tests seems to be running fine now
```
$ pgbackrest/test/test.pl --vm-out --dev --module=db --coverage-only --vm=f32
2021-07-13 09:33:22.276 P00   INFO: P1-T1/1 - vm=f32, module=db, test=db (6.41s)
2021-07-13 09:33:22.983 P00   INFO: tested modules have full coverage
2021-07-13 09:33:22.983 P00   INFO: writing C coverage report
2021-07-13 09:33:23.103 P00   INFO: TESTS COMPLETED SUCCESSFULLY (11s)
```


dwsteele (changes requested):
The code looks correct to me but I think this is more functionality than we need. I can see you worked from the loop above, but replay may take minutes or even hours to complete, so marking progress along the way is pretty important.

In this case I think `pg_control` is trailing replay lsn by milliseconds and every once in a while somebody will be unlucky enough to miss the update. It seems sufficient to retry for just one timeout interval (which is probably way longer than needed) and not worry about progress.

It's not a question of whether the code works (it does!) but I would prefer to keep the code and tests as simple as possible for a given use case.

On the bright side, you now know how to use the pq shim!

pgstef:
> It's not a question of whether the code works (it does!) but I would prefer to keep the code and tests as simple as possible for a given use case.

Point taken. Attempt to simplify in c120e5e.

> On the bright side, you now know how to use the pq shim!

That one was tricky, but I guess I made it :-) 

New run after c120e5e:
```
$ pgbackrest/test/test.pl --vm-out --dev --module=db --coverage-only --vm=f32
2021-07-14 12:45:04.365 P00   INFO: P1-T1/1 - vm=f32, module=db, test=db (5.47s)
2021-07-14 12:45:06.837 P00   INFO: tested modules have full coverage
2021-07-14 12:45:06.838 P00   INFO: writing C coverage report
2021-07-14 12:45:06.962 P00   INFO: TESTS COMPLETED SUCCESSFULLY (9s)
```

src/db/db.c
```diff
@@ -568,29 +568,37 @@ dbReplayWait(Db *this, const String *targetLsn, TimeMSec timeout)
...
-            // Execute query
-            VariantList *row = dbQueryRow(this, query);
+            do
+            {
+                // Build the query
+                const String *query = strNewFmt(
+                    "select checkpoint_%s::text,\n"
+                    "       (checkpoint_%s > '%s')::bool as targetReached\n"
+                    "  from pg_catalog.pg_control_checkpoint() as checkpointLsn",
+                    lsnName, lsnName, strZ(targetLsn));
```
dwsteele, line 585: The new query is the same as the old query, just the select fields have been swapped (and `as  checkpointLsn` is now in the wrong place). This also causes churn in the macro. Can we restore the original query and reduce the churn?

dwsteele (changes requested):

src/db/db.c
pgstef: I preferred to adjust the query so it is consistent with the replay query. 
I guess I can go back to previous query if you'd prefer...

pgstef: Should be solved with 6b4b253.

dwsteele: If we wanted to make the queries look more like each other (which I'm not sure is needed) we'd want to do that in a separate patch. Here it just looks like noise.

Update looks good.

pgstef: > If we wanted to make the queries look more like each other (which I'm not sure is needed) we'd want to do that in a separate patch. Here it just looks like noise.

Right, understood :-)

> Update looks good.

Cool, thanks :-)

dwsteele (approved):
Approved with a few minor updates.

PR #1474, opened 2021-07-26
Add check for test path inside repo path.

dwsteele:


turnstep:
This works well for me, and gives a good error message.

.cirrus.yml
```diff
@@ -18,7 +18,7 @@ freebsd_12_task:
   install_script: pkg install -y bash git postgresql-libpqxx pkgconf libxml2 gmake perl5 libyaml p5-YAML-LibYAML rsync
 
   script:
-    - perl ${CIRRUS_WORKING_DIR}/test/test.pl --no-gen --make-cmd=gmake --vm=none --vm-max=2 --no-coverage --no-valgrind --module=command --test=backup
+    - cd .. && perl ${CIRRUS_WORKING_DIR}/test/test.pl --no-gen --make-cmd=gmake --vm=none --vm-max=2 --no-coverage --no-valgrind --module=command --test=backup
```
turnstep, line 21: Worth doing a `cd -` at the end?

turnstep, line 21: `--vm=none` would seem to make `--vm-max` a no-op?

turnstep (approved):

.cirrus.yml
dwsteele: Doesn't seem like it is needed.

dwsteele: --vm-max will still run things in parallel even when --vm=none. The option names have gotten a bit creaky -- they date back to when all this was done with vagrant VMs.

Issue #1476, opened 2021-07-26
Use a hostname other than "pg1" in the user guide


turnstep:
Doc improvement idea: use a different hostname other than "pg1" which leads to confusion with the args that require a literal "pg1" not related to the hostname. In the user guide:

 `A new host named pg1 is created to contain the demo cluster and run backrest examples.`

dwsteele:
Suggestions?

pgstef:
Is it really called `pg1`? I've got the feeling that is appears as `pg-primary` ?
Perhaps that mention is some left-over and we should adjust to `pg-primary` too ?

dwsteele:
Good catch, @pgstef. I have added PR #1479 to fix this and one other host name issue.

I would like to rename the hosts from `pg-primary` and `pg-standby` at some point because it would be good to add some failover examples, but this addresses the immediate issue.

pgstef:
> Good catch, @pgstef. I have added PR #1479 to fix this and one other host name issue.

Seen and approved ;-)

> I would like to rename the hosts from `pg-primary` and `pg-standby` at some point because it would be good to add some failover examples, but this addresses the immediate issue.

In such cases, I like to use the "node" term (like vagrant does for example). Why not something like `pg-node1`, `pg-node2`, `repo-node1`,... ? Anyway, that's another day problem :-D

dwsteele:
So, anyone interested in doing the rename to `pg-node1` or `pg-host1`? I personally favor `pg-host1` since `host` is the terminology we use throughout the documentation.

PR #1479, opened 2021-07-29
Fix incorrect host names in user guide.

dwsteele:


pgstef (approved):
Looks good to me to fix #1476.

doc/xml/user-guide.xml
```diff
@@ -866,7 +866,7 @@
     <section id="installation">
         <title>Installation</title>
 
-        <p>A new host named <host>pg1</host> is created to contain the demo cluster and run <backrest/> examples.</p>
+        <p>A new host named <host>{[host-pg1]}</host> is created to contain the demo cluster and run <backrest/> examples.</p>
```
turnstep, line 869: This is an improvement, but I'd like to see us not use `pg1` at all, to make it clear it is absolutely unrelated to things like `pg1-port`. Maybe `postgres-node1` or similar?

pgstef: Hm, afaik, we're not using `pg1` -> `<variable key="host-pg1">pg-primary</variable>`
That's why `pg-primary` is displayed in the user-guide...

dwsteele: Right, `pg-*` is just the name of the internal variable, not what is displayed to the user (`pg-primary`, `pg-standby`), except for that one typo. Renaming the variables would cause a lot of unneeded churn, in my opinion.

dwsteele: @turnstep I think whether or not we rename these variables should be discussed in #1476. This PR is pretty much just a straight bug fix and I'd like to get it into this release.

PR #1480, opened 2021-08-02
Only pass selected repo options to the remote.

dwsteele:
Options for other repos can cause conflicts and should never be used. Each remote can address exactly one repo or pg cluster.


test/src/module/protocol/protocolTest.c
```diff
@@ -346,6 +346,7 @@ testRun(void)
         strLstAddZ(argList, "--repo1-host-config-include-path=/path/include");
         strLstAddZ(argList, "--repo1-host-config-path=/path/config");
         strLstAddZ(argList, "--repo1-host-user=repo-host-user");
+        hrnCfgArgKeyRawZ(argList, cfgOptRepoType, 2, "posix");      // Will not be passed to the remote
```
pgguru, line 349: This test is sufficient to prove what you're doing, but a little too subtle in explanation.  So per some offline discussion, this is related to verifying the following:

- if we pass repo2 options to repo1, we will end up with an error related to the conflicting local repo types.
- the patch is supposed to suppress repo2 options going to repo1
- ergo when we add this option in the test, it will not be passed to repo1 if things are working as expected ("the comment")

I think we need to be more explicit in *why* we are adding this, specifically that this particular line serves to verify that repo-specific option filters will prevent this from being passed to the wrong remote, so eliminates what would have been an error condition if the filter was not in place.

(Probably could be more concise than that... :-))

dwsteele: Added a better comment in 5bcd86aba0006. Look good?

dwsteele: Revised in ec13f245697c083.

pgguru (approved):

PR #1488, opened 2021-08-11
Rearrange backup documentation in user guide.

dwsteele:
Remove the "Automatic Stop Option" section since it only applies to PostgreSQL <= 9.6, which will soon be EOL. Since we no longer build the user guide for PostgreSQL < 10 this section was no longer being tested. The stop-auto option is still documented in the reference.

Move the "Fast Start Option" to "Quick Start - Perform Backup". This is a commonly-used option so it makes sense to mention it earlier. This also makes the backups run more quickly. In the worst case, backups in "Quick Start - Perform Backup" could take minutes to start

Move the "Archive Timeout" section to "Quick Start - Perform Backup" since it is the last section in "Backup".

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -194,6 +194,17 @@
                     </release-item>
 
                     <release-item>
+                        <github-pull-request id="1488"/>
+
+                        <release-item-contributor-list>
+                            <release-item-contributor id="david.steele"/>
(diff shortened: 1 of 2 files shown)


dwsteele:
My hope is that it is more likely to be seen there. Otherwise it is hard to argue even mentioning it outside of the reference.

PR #1489, opened 2021-08-11
Log backup file total and restore size/file total.

dwsteele:
The backup size was a bit off because it did not include any files (e.g. backup_label, WAL files) that were added to the manifest after the main copy. To fix this move the log message to the very end of the backup.

Add size/file total log message to restore since it did not exist before.


cmwshang (approved):
I was a little confused because sometimes the detail messages don't all say "backup file" - some say "copy segment", "hardlink", or "wrote" etc, but all the numbers match up when I take those into account, so looks good.

PR #1508, opened 2021-09-09
Report backup file validation errors in backup.info.

pgstef:
Hi,

This is a proposal to report the page checksum error flag from backup manifest to `backup.info`, and show an additional text message in the `info` command.

`infoBackupLoadCallback` is responsible to ensure the key is returned even if not existing previously.
`info` will then have a new key to report the boolean and the error list separately. The error list will only be returned when requesting info for a specific backup set.
Imho, the key name change won't have any impact given it's only shown in json output and until PR #1501, json output didn't support `--set`.

```bash
$ pgbackrest info
...
        full backup: 20210909-092339F
            timestamp start/stop: 2021-09-09 09:23:39 / 2021-09-09 09:23:45
            wal start/stop: 000000010000000000000019 / 000000010000000000000019
            database size: 180.3MB, database backup size: 180.3MB
            repo1: backup set size: 11.9MB, backup size: 11.9MB

        incr backup: 20210909-092339F_20210909-092539I
            timestamp start/stop: 2021-09-09 09:25:39 / 2021-09-09 09:25:41
            wal start/stop: 00000001000000000000001B / 00000001000000000000001B
            database size: 180.3MB, database backup size: 547.3KB
            repo1: backup set size: 11.9MB, backup size: 48.8KB
            backup reference list: 20210909-092339F
            page checksum error(s) detected

$ pgbackrest info --stanza=xxx --set=20210909-092339F_20210909-092539I
...
        incr backup: 20210909-092339F_20210909-092539I
            timestamp start/stop: 2021-09-09 09:25:39 / 2021-09-09 09:25:41
            wal start/stop: 00000001000000000000001B / 00000001000000000000001B
            database size: 180.3MB, database backup size: 547.3KB
            repo1: backup set size: 11.9MB, backup size: 48.8KB
            backup reference list: 20210909-092339F
            database list: postgres (13319), test-checksums (16406)
            page checksum error: base/16406/16407
```

Coverage test seems good too:
```
2021-09-09 08:47:23.203 P00   INFO: P1-T1/1 - vm=f33, module=info, test=info-backup (6.07s)
2021-09-09 08:47:24.558 P00   INFO: tested modules have full coverage
2021-09-09 08:47:24.559 P00   INFO: writing C coverage report
2021-09-09 08:47:24.684 P00   INFO: TESTS COMPLETED SUCCESSFULLY (11s)
...
[...]

PR #1519, opened 2021-09-28
Rename page checksum error to error list in info text output.

dwsteele:
"file error" makes it clear that other errors may be reported. If checksum-page is true in the manifest but no checksum-page-error list is provided then the error is in alignment, i.e. the file size is not a multiple of the page size, with allowances made for a valid-looking partial page at the end of the file.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -29,7 +29,7 @@
                     </release-item>
 
                     <release-item>
-                        <github-pull-request id="1434"/>
+                        <github-pull-request id="1519"/>
 
                         <release-item-contributor-list>
                             <release-item-contributor id="david.steele"/>
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -29,6 +29,17 @@
                     </release-item>
 
                     <release-item>
+                        <github-pull-request id="1434"/>
+
+                        <release-item-contributor-list>
+                            <release-item-contributor id="david.steele"/>
+                            <release-item-reviewer id="stefan.fercot"/>
+                        </release-item-contributor-list>
+
+                        <p>Rename page checksum error to error list in info text output.</p>
+                    </release-item>
+
+                    <release-item>
                         <github-issue id="1434"/>
 
                         <release-item-contributor-list>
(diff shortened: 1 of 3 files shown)


pgstef (commented):
`file-error` makes it clear that there's a problem indeed. But then, the question for the user is where to find the extra info.

Should we add more examples in the docs to help people investigate (they might not have the backup logs anymore) or could we change the display to reflect "invalid page checksum"/"page misalignment in file" like in the backup log ?

dwsteele:
> could we change the display to reflect "invalid page checksum"/"page misalignment in file" like in the backup log ?

I don't think this would scale very well. I'm working on adding more errors -- in the end there could be 5-10. I think we would need some kind of `--detail` option since there could be 100+k errors per file.

I'm starting to remember why we didn't put this data into the JSON output. We just haven't finalized what we want to do here enough to commit to a format.

pgstef:
> I don't think this would scale very well. I'm working on adding more errors -- in the end there could be 5-10. I think we would need some kind of `--detail` option since there could be 100+k errors per file.
> 
> I'm starting to remember why we didn't put this data into the JSON output. We just haven't finalized what we want to do here enough to commit to a format.

Given all the errors would be in the backup manifest (I guess), it would be nice to reflect those errors in the `--set` text output imho. Running info globally or with json format gives you the overview and then you could investigate with a more detailed view a specific backup set. 

That could be a feature/PR on its own, so this one looks fine to me ;-)

pgstef (approved):

dwsteele:
I started looking at #1508 and realized it was already making some of the changes in this commit. I think it would be better to name this field `error-list` and the new field added by #1508 `error`. So we'll have a global error flag that is always present and a list that is present when `--set` is specified.

Make sense?

pgstef:
> I started looking at #1508 and realized it was already making some of the changes in this commit. I think it would be better to name this field `error-list` and the new field added by #1508 `error`. So we'll have a global error flag that is always present and a list that is present when `--set` is specified.
> 
> Make sense?

Totally make sense. The global report is just a flag indicating "oops, something went wrong, please investigate". And then, we have the details when `--set` is specified. we might even limit some deeper prints for the text output, given we'd be supposed to investigate manually...

So +1 with the latest change :-)

PR #1520, opened 2021-09-30
Allow link-map to create new links.

dwsteele:
Currently link-map only allows links that exist in the backup manifest to be remapped to a new destination.

This commit allows link-map to create a new link as long as a valid path/file from the backup is referenced.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -29,6 +29,20 @@
                     </release-item>
 
                     <release-item>
+                        <github-issue id="1507"/>
+                        <github-pull-request id="1518"/>
+
+                        <release-item-contributor-list>
+                            <release-item-ideator id="tom.swartz"/>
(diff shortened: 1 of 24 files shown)


pgstef (approved):
Hi,

I've just tried this PR branch and the re-map works great.
All my usual tests passed successfully and I didn't noticed anything "obvious" code-side. 

So, +1, LGTM :-)

With `master`, when the restore target (pg1-path) doesn't exist, the restore command creates it.
With this branch, I get `ERROR: [041]: unable to get info for missing path/file '...': [2] No such file or directory`.

Isn't that case supposed to be checked by the CI tests?

dtseiler:
This works great for me. I took backup, then set `link-map` configuration and restored with symlink as desired. Then backed that up again and restored again.

dtseiler (approved):

cbandy (approved):

PR #1521, opened 2021-10-01
Report original error and retries on local job failure.

dwsteele:
The local process will retry jobs (e.g. backup file) but after a certain number of failures gives up. Previously, the last error was reported but generally the first error is far more valuable. The last error is likely to be a cascade failure such as the protocol being out of sync.

Report the first error (and stack trace) and append the retry errors to the first error without stack trace information. It is likely that the formatting could be improved. Any comments on that are welcome.

This commit builds on 136d309dd4 and ae40ed6ec9c.


pgstef (commented):
Just tried this branch. My usual tests passed successfully.

I'm just pondering about the error message. Example:
```
ERROR: [055]: raised from local-2 protocol: unable to open missing file '.../pg_data/global/pg_control.gz' for read
       [FileMissingError] on retry after 0ms: unable to open missing file '.../pg_data/global/pg_control.gz' for read
       [FileMissingError] on retry after 15000ms: unable to open missing file '.../pg_data/global/pg_control.gz' for read
```

It allows to identify clearly that there was a retry. But the error message is still the same, so not really interesting. Couldn't we only print error if that's not the same ? I'm probably over-thinking this...

dwsteele:
> Couldn't we only print error if that's not the same?

Maybe, but it does complicate the logic a bit and we'd need a consistent message to use. I suppose we wouldn't need the error code either -- but that means there would be four different ways to format the retry: same error/same message, different error/same message, same error/different message, different error/different message.

Seem a bit complicated?

pgstef:
> Maybe, but it does complicate the logic a bit and we'd need a consistent message to use. I suppose we wouldn't need the error code either -- but that means there would be four different ways to format the retry: same error/same message, different error/same message, same error/different message, different error/different message.
> 
> Seem a bit complicated?

I'm not sure duplicating the error type would be an issue. I'm kinda feeling that only the message could be really long.
Perhaps only avoiding to print the message if it has already been printed before would be good enough..

```
ERROR: [055]: raised from local-2 protocol: unable to open missing file '.../pg_data/global/pg_control.gz' for read
       [FileMissingError] on retry after 0ms
       [FileMissingError] on retry after 15000ms
```

I find this a little bit easier to read, but it probably doesn't worth the effort...

dwsteele:
OK, how about cb7bd4a12c80fc7e?

pgstef (approved):
> OK, how about [cb7bd4a](https://github.com/pgbackrest/pgbackrest/commit/cb7bd4a12c80fc7ec95675668dae301cf3546d3a)?

Looks great to me !

Issue #1537, opened 2021-10-12
The use of a wrapper for pgbackrest.bin leads to the impossibility of a restoration without intervention


collectionneurfou:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version:
2.25 

2. PostgreSQL version:
Not relevant.

3. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each:
RHEL 7.9

4. Did you install pgBackRest from source or from a package?
source.

5. Please attach the following as applicable:
    - `pgbackrest.conf` file(s)
    - `postgresql.conf` settings applicable to pgBackRest (`archive_command`, `archive_mode`, `listen_addresses`, `max_wal_senders`, `wal_level`, `port`)
    - errors in the postgresql log file before or during the time you experienced the issue
    - log file in `/var/log/pgbackrest` for the commands run (e.g. `/var/log/pgbackrest/mystanza_backup.log`)

N/A
7. Describe the issue:

I'm using pgbackrest with 2 files :
pgbackrest.bin
pgbackrest

pgbackrest is a wrapper which add specific libpq path to use pgbackrest (all version of postgresql installed locally are installed with EDB installer of postgres, which is static version, required for having multiple postgres version in the system).

pgbackrest wrapper is as follow :

export LD_LIBRARY_PATH=/opt/rh/rh-postgresql10/root/usr/lib64${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}
BASEDIR=$(dirname "$0")
${BASEDIR}/pgbackrest.bin $@

When i restore (delta, full) there is an error because the recovery.conf contains the path of "pgbackrest.bin" and NOT the wrapper path. For example 

# Do not edit this file manually!
# It will be overwritten by the ALTER SYSTEM command.

# Recovery settings generated by pgBackRest restore on 2021-09-29 17:13:17
restore_command = '/logiciel/postgres/pgbackrest/bin/pgbackrest.bin --stanza=TEST12B archive-get %f "%p"'

should be : 
# Do not edit this file manually!
# It will be overwritten by the ALTER SYSTEM command.

# Recovery settings generated by pgBackRest restore on 2021-09-29 17:13:17
restore_command = '/logiciel/postgres/pgbackrest/bin/pgbackrest --stanza=TEST12B archive-get %f "%p"'

[...]

PR #1540, opened 2021-10-18
Add --cmd option.

jreidthompson:
Add --cmd option.  Allows users to provide an executable to be used when pgbackrest generates command strings that expect to invoke pgbackrest. These generated commands may be invoked by pgbackrest or written to files by pgbackrest, for example recovery.conf.  Reference issue https://github.com/pgbackrest/pgbackrest/issues/1537

### DIFF
--- a/src/build/help/help.xml
+++ b/src/build/help/help.xml
@@ -145,7 +145,9 @@
                         <summary><backrest/> command.</summary>
 
                         <text>
-                            <p><backrest/> may generate a command string when creating the <code>restore_command</code> during the restore process. The command used to trigger the <backrest/> process will then also be used in this case, unless the <br-option>cmd</br-option> option is provided.</p>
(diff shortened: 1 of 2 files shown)


dwsteele:
I believe his is going to cause unintended consequences because we can't run `execvp()` on a shell script. We'll need to limit this option to just output to `restore_command`.

jreidthompson:
> I believe his is going to cause unintended consequences because we can't run `execvp()` on a shell script. We'll need to limit this option to just output to `restore_command`.

I pushed changes to address this.  Updating to only impact the creation of the restore_command.

dwsteele:
Edits, add cautionary note, remove noise in 01657f71275ffd. Looks good to me otherwise.

Does this look OK @pgstef?

pgstef:
> Edits, add cautionary note, remove noise in [01657f7](https://github.com/pgbackrest/pgbackrest/commit/01657f71275ffde44d1f9329e486e4ce91f8796c). Looks good to me otherwise.
> 
> Does this look OK @pgstef?

Hey,
Yes, looks good to me :thumbsup:

jreidthompson (commented):
looks good for me

PR #1554, opened 2021-11-01
Use cfgOptionStrId() instead of cfgOptionStr() where appropriate.

dwsteele:
The compress-type, repo-type and log-level-* options have allow lists, which means it is more efficient to treat them as StringIds.

For compress-type and log-level-* also update the functions that convert them to enums.


jreidthompson (approved):
I did not see anything amiss

PR #1557, opened 2021-11-03
Display size option default and allowed values with appropriate units.

dwsteele:
Size option default and allowed values were displayed in bytes, which was confusing for the user.

This also lays the groundwork for adding units to time options.

Move option parsing functions into a common module so they can be used from the build module.


jreidthompson (approved):
changes look good to me

Issue #766, opened 2019-07-02
Log shows different incr size vs info command after switchover


mahomedh:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version: 2.15

2. PostgreSQL version: PostgreSQL 11.4 64-bit

3. Operating system/version - Ubuntu 18.04.1 on all three servers

4. Install Source: package (apt from postgresql)

    - `pgbackrest.conf` on backup server
	
			[CustA-Cluster]
			pg1-host=pg1.local
			pg1-path=/var/lib/postgresql/11/main
			pg2-host=pg2.local
			pg2-path=/var/lib/postgresql/11/main
			repo1-path=/CustA

			[global:backup]
			backup-standby=y

			[global]
			repo1-type=s3
			# for compatibility - see pgbackrest issue 645
			repo1-s3-host=pg-backups.mydomain.co.uk
			repo1-s3-endpoint=mydomain.co.uk
			repo1-s3-verify-ssl=n
			repo1-s3-bucket=pg-backups
			repo1-s3-key=xxxxxxxxxxxxxxxxxxx
			repo1-s3-key-secret=yyyyyyyyyyyyyyyyyyyyyyy
			repo1-s3-region=us-east-1
			repo1-retention-full=2
			repo1-retention-diff=2
			repo1-cipher-type=none
			repo1-host-user=postgres
			start-fast=y
			stop-auto=y
			process-max=4
			log-level-file=detail

			[global:archive-push]
			compress-level=3
			
    - `pgbackrest.conf` on db servers
	
			[CustA-Cluster]
			pg1-path=/var/lib/postgresql/11/main
			pg2-path=/var/lib/postgresql/11/main
			repo1-path=/CustA

			[global:backup]
			backup-standby=y

			[global]
			repo1-type=s3
			# for compatibility - see pgbackrest issue 645
			repo1-s3-host=pg-backups.mydomain.co.uk
			repo1-s3-endpoint=mydomain.co.uk
			repo1-s3-verify-ssl=n
			repo1-s3-bucket=pg-backups
			repo1-s3-key=xxxxxxxxxxxxxxxxxxx
			repo1-s3-key-secret=yyyyyyyyyyyyyyyyyyyyyyy
			repo1-s3-region=us-east-1
			repo1-retention-full=2
			repo1-retention-diff=2
			repo1-cipher-type=none
			repo1-host-user=postgres
			start-fast=y
			stop-auto=y
[...]

PR #1562, opened 2021-11-09
Match backup log size with size reported by info command.

jreidthompson:
Properly output to the log the size of files copied during the backup, matching the size of backup returned from the INFO command. In the reference issue, the incremental backup after switchover output to the log the size of all files evaluated rather than only the size of the files copied in the backup. 
Reference https://github.com/pgbackrest/pgbackrest/issues/766


src/command/backup/backup.c
```diff
@@ -1148,15 +1149,15 @@ backupJobResult(
...
+            *sizeProgress += copySize;
 
             // Create log file name
             const String *fileLog = host == NULL ? fileName : strNewFmt("%s:%s", strZ(host), strZ(fileName));
 
             // Format log strings
             const String *const logProgress =
                 strNewFmt(
-                    "%s, %" PRIu64 "%%", strZ(strSizeFormat(copySize)), sizeTotal == 0 ? 100 : sizeCopied * 100 / sizeTotal);
+                    "%s, %" PRIu64 "%%", strZ(strSizeFormat(copySize)), sizeTotal == 0 ? 100 : (*sizeProgress) * 100 / sizeTotal);
```
dwsteele, line 1160: I don't think parens are needed here. We generally add parens only to establish precedence or if the compiler gripes.

src/command/backup/backup.c
```diff
@@ -1182,6 +1183,9 @@ backupJobResult(
             // Else file was copied so update manifest
             else
             {
+                // Increment size of files copied
+                *sizeCopied += copySize;
+
```
dwsteele, line 1188: I think this will also need to be done at 1172 if we want the log output in backup to match what info outputs. That case means that a file existing before a resume is still usable.

test/src/module/command/backupTest.c
```diff
@@ -1325,7 +1325,7 @@ testRun(void)
         ProtocolParallelJob *job = protocolParallelJobNew(VARSTRDEF("key"), protocolCommandNew(strIdFromZ("x")));
         protocolParallelJobErrorSet(job, errorTypeCode(&AssertError), STRDEF("error message"));
 
-        TEST_ERROR(backupJobResult((Manifest *)1, NULL, STRDEF("log"), strLstNew(), job, 0, 0), AssertError, "error message");
+        TEST_ERROR(backupJobResult((Manifest *)1, NULL, STRDEF("log"), strLstNew(), job, 0, 0, 0), AssertError, "error message");
```
dwsteele, line 1328: Let's make this `, NULL, NULL)` since these parameters are pointers. Not that big a deal but may avoid confusion in the future.

dwsteele (changes requested):
Looks pretty good overall, see comments.

src/command/backup/backup.c
jreidthompson: will remove

src/command/backup/backup.c
jreidthompson: will add

test/src/module/command/backupTest.c
jreidthompson: will do

dwsteele (approved):
I'm a little disappointed that this change did not break more tests, but I don't think it is worth adding complex backup tests for a log message.

PR #1566, opened 2021-11-12
httpRequestError: Log verb

df7cb:
The httpRequestError method is logging everything except the actual HTTP
verb, fix that.

### DIFF
--- a/src/common/io/http/request.c
+++ b/src/common/io/http/request.c
@@ -252,7 +252,7 @@ httpRequestError(const HttpRequest *this, HttpResponse *response)
     // Output path/query
     strCatZ(error, ":\n*** Path/Query ***:");
 
-    strCatFmt(error, "\n%s", strZ(httpRequestPath(this)));
+    strCatFmt(error, "\n%s %s", strZ(httpRequestVerb(this)), strZ(httpRequestPath(this)));
 
     if (httpRequestQuery(this) != NULL)
         strCatFmt(error, "?%s", strZ(httpQueryRenderP(httpRequestQuery(this), .redact = true)));


src/common/io/http/request.c
```diff
@@ -243,7 +243,7 @@ httpRequestError(const HttpRequest *this, HttpResponse *response)
     ASSERT(response != NULL);
 
     // Error code
-    String *error = strCatFmt(strNew(), "HTTP request failed with %u", httpResponseCode(response));
+    String *error = strCatFmt(strNew(), "HTTP %s request failed with %u", strZ(httpRequestVerb(this)), httpResponseCode(response));
```
dwsteele, line 246: I think think would be better at line 255, i.e.:
```
strCatFmt(error, "\n%s %s", strZ(httpRequestVerb(this)), strZ(httpRequestPath(this)));
```

dwsteele (changes requested):
This looks sensible but I think it should be moved, as noted.

There are also a number of tests that will need to be updated.

src/common/io/http/request.c
df7cb: I moved the output to that location.

dwsteele: That looks good to me.

dwsteele:
Looks like the test changes are fairly mechanical but let us know if you run into issues. I'm not sure why arm64 is failing here but it doesn't look like the fault of this patch. I'm trying a build from integration to see if something changed in the arm64 env since the last successful build on Friday.

dwsteele:
I just retested he last commit from main and arm64 ran fine: https://cirrus-ci.com/build/6550376959705088. Could be a race condition or just a fluke.

dwsteele:
Committed at c38e2d31709804eb4b9125a15ad84c8fc813f366.

Issue #1430, opened 2021-06-16
Feature request to add S3 server side encryption by specifying a KMS key id as a config


tharinduamila-insta:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version: 2.34

2. PostgreSQL version: 13.3

3. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each: debian

4. Did you install pgBackRest from source or from a package?
from source
5. Please attach the following as applicable:
    - `pgbackrest.conf` file(s)
    - `postgresql.conf` settings applicable to pgBackRest (`archive_command`, `archive_mode`, `listen_addresses`, `max_wal_senders`, `wal_level`, `port`)
    - errors in the postgresql log file before or during the time you experienced the issue
    - log file in `/var/log/pgbackrest` for the commands run (e.g. `/var/log/pgbackrest/mystanza_backup.log`)

7. Describe the issue:

We would like to utilize the capability AWS offers for encrypting objects at server side by specifying a KMS key id in the request. This would allow us to have better control over who would be authorized to decrypt the stored objects and also not have any secrets in the server box itself reducing the risk of exposing that key.  

We will also be willing to contribute for this feature if it's accepted for the project.

dwsteele:
If this is just a header to add to requests it might not be that hard to implement. If you do the implementation we'll update the unit tests, but for now I don't think we have time to take this on.

tharinduamila-insta:
Hi David, you are correct. It just needs several headers to be sent and I am happy to do the implementation. It'll take me a while before I get around to it too.  Thank you for having a look at the request and for suggesting a way forward with this.

PR #1567, opened 2021-11-12
Add support for AWS S3 server-side encryption using KMS.

df7cb:
AWS documentation:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/specifying-kms-enc

Implements #1430.


dwsteele:
I've been trying to decide whether to simply add this header as you have done here, or come up with a more generalized system for adding custom headers, which could then also be used for storage classes, etc.

Something like:
```
repo1-storage-http-header=put=x-amz-server-side-encryption=xxx
```
But that may just be too complicated and error prone.

df7cb:
> ```
> repo1-storage-http-header=put=x-amz-server-side-encryption=xxx
> ```
> 
> But that may just be too complicated and error prone.

That would probably be quite some "what the heck" for users. It's also much harder to document, and explain if someone asks about it.

src/build/config/config.yaml
```diff
@@ -2101,6 +2101,12 @@ option:
       - auto
       - web-id
 
+  repo-s3-kms-key-id:
+    inherit: repo-s3-bucket
+    required: false
+    deprecate:
+      repo-s3-kms-key-id: {}
```
dwsteele, line 2108: This is a new option so no deprecation is needed.

src/storage/s3/storage.c
```diff
@@ -171,6 +175,13 @@ storageS3Auth(
         if (this->securityToken != NULL)
             httpHeaderPut(httpHeader, S3_HEADER_TOKEN_STR, this->securityToken);
 
+        // Set KMS key header if this is a write request
+        if (this->kmsKeyId != NULL && !strCmp(verb, HTTP_VERB_PUT_STR))
+        {
```
dwsteele, line 180: This doesn't look right. I think you mean `strCmp(verb, HTTP_VERB_PUT_STR)` but it seems at least POST would also be required. Sees like this should be implemented in the `StorageWriteS3` object instead where it is more obvious where it should be applied.

dwsteele (changes requested):
See comments, but in general I don't think adding this logic to `storageS3Auth()` is the way to go.

src/build/config/config.yaml
df7cb: TBH this just mirrors all the other repo-s3 options. Looks like a lot of cargo-cult in there.

I'll remove it for this option.

src/storage/s3/storage.c
df7cb: I had originally included POST as well, but that's refused by aws with an "header not applicable here" error.

I don't understand the "you mean strCmp" part, that's exactly what I included there?

During development, I had tried to put the logic into StorageWriteS3, but failed since I couldn't access neither the kmskey option nor the header object there. I'll give it another try, but possibly I'll need help in sorting out the object-oriented C code there.

df7cb: The two sites in s3/write.c that use PUT are these:
```
storageWriteS3PartAsync(StorageWriteS3 *this)
{
...
            this->request = storageS3RequestAsyncP(
                this->storage, HTTP_VERB_PUT_STR, this->interface.name, .query = query, .content = this->partBuffer);
```
and
```
storageWriteS3Close(THIS_VOID)
{
...
            // If a multi-part upload was started we need to finish that way
            if (this->uploadId != NULL)
...
            // Else upload all the data in a single put
            else
                storageS3RequestP(this->storage, HTTP_VERB_PUT_STR, this->interface.name, .content = this->partBuffer);
```
The first one looks easy to modify since it has a "query" member, but the second one is weird as it's starting a new request on close.

To be honest, since this is about encryption, I would prefer to have the header added in the place where it's now, since that seems to be the function that all HTTP requests pass through. If we instead sprinkle the header additions in other places, we might miss one (or some future addition might neglect it). Centralizing that seems safer for security to me, as encryption is optional and AWS S3 does not raise an error if the header is missing.

src/build/config/config.yaml
dwsteele: > Looks like a lot of cargo-cult in there.

These deprecations have purposes and all have tests. The option it looks like you copied from (`repo-s3-region`) was available pre-2.0 before indexes were required. So this deprecation allows `repo-s3-region` and makes it equal to `repo1-s3-region`.

If you identify any dead code in this file we'd be happy to remove it.

src/storage/s3/storage.c
dwsteele: > The first one looks easy to modify since it has a "query" member, 

The first PUT is an UploadPart (https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPart.html) and does not require the key because it was already provided in the POST for CreateMultipartUpload (https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html).

> but the second one is weird as it's starting a new request on close.

That second PUT writes any files that are < 5MB all at once and it requires the key (https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html).

So the verb is not enough to know whether this key is needed or not. Also, I think it should only be added to GET object and not GET list, DELETE, etc.

The easiest thing to do here I think is add a new param to `storageS3RequestP()` by adding a member to the `StorageS3RequestParam` struct:
```
bool sseKms; // ...
```
Then you just need to set it in the `storageS3RequestP()` calls where needed like `storageS3Request(..., sseKms = true)`. Add your header logic to `storageS3RequestAsync()` conditional on param.sseKms being set.

The way I read it you'll need to set sseKms at line 55 in s3/read.c and lines 126 and 251 in write.c, but I may be missing something.

src/build/config/config.yaml
df7cb: Ok. That's really not visible from "repo-s3-region: deprecate: repo-s3-region", sorry for not grasping that.

src/storage/s3/storage.c
dwsteele: Hmmm, you'll also need to add `sseKms` to the `StorageS3RequestAsyncParam` struct and pass it through to `storageS3RequestAsync()` in `storageS3Request()`. You'd don't need async for this feature but that's where the request logic lives.

df7cb: The "may be missing something" part is why I think centralizing that is better.

dwsteele: Agreed, and centralizing it what I'm describing above, but with the ability to turn it on only where needed.

dwsteele: So, something like this: https://github.com/pgbackrest/pgbackrest/commit/1d45b2c25425cf6b145f9a0f577f755d1905a142

This will be a easier if you allow me to update your PR directly. See https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork.

df7cb: It's somewhat hidden in a wall of text on that page, but "allow edits from maintainers" works only for *user-owned* repositories, this fork is owned by the company account unfortunately.

df7cb: Thanks for picking this up.
As @tharinduamila-insta said in https://github.com/pgbackrest/pgbackrest/commit/1d45b2c25425cf6b145f9a0f577f755d1905a142#r60776961 , GET requests don't have the headers, so only the two callers in write.c should remain.
(I unfortunately deleted parts of my test S3 account, and I fail to recreate it, so my ability to test this is somewhat limited at the moment, but AWS is so kind to still spit out "no key allowed here" error messages.)

dwsteele:
> Dropping , .sseKms = true makes it work again.

OK, so looks like we just need the param on one PUT and one POST in `s3/write.c`. I think the prior implementation was only encrypting multi-part uploads, so it was not running but leaving data unencrypted. Honestly, I'm a bit confused about how it was working before, since it looks like the key would have been added to GET requests by `if (this->kmsKeyId != NULL && !strCmp(verb, HTTP_VERB_PUT_STR))`.

> "allow edits from maintainers" works only for user-owned repositories, this fork is owned by the company account unfortunately.

OK, that just means we won't be able to merge the PR using github so the PR will be marked "closed" instead of "merged". I need to add release notes before committing, which are very likely to be in conflict so I always add them at the last moment. In these cases I add the commit to the PR to show that it was merged for anyone who checks the comments.

> I unfortunately deleted parts of my test S3 account, and I fail to recreate it, so my ability to test this is somewhat limited at the moment, but AWS is so kind to still spit out "no key allowed here" error messages.

We'll definitely need to get this tested in a real environment before proceeding. Once that is done the unit tests will ensure we don't have regressions.

tharinduamila-insta:
So I ran this branch on a postgresql db that didn't have any actual data. And used below to get the uploaded file data
`aws s3 ls --recursive --human-readable s3://instaclustr-us-east-1-backups-preprod/1cc4843e-4ad2-4132-9439-a7f921bc1cf2/ | awk '{ print $5 }' | xargs -I {} sh -c 'echo {} && aws s3api head-object --bucket instaclustr-us-east-1-backups-preprod --key {} && echo ||' > head-details.csv`
Then ran this file through below script
`f = open("head-details.csv","r")
lines = f.read()
f.close()

splited = lines.split("||");

count = 0;
for entry in splited:
  data = entry.split('\n', 2);
  object = data[1];
  kmsEncrypted = data[2].find('"ServerSideEncryption": "aws:kms"') != -1;
  if not kmsEncrypted:
    print(object)
  else:
    count += 1;

print('encrypted ' + str(count))
print('total ' + str(len(splited)))`

which gave the output of 

`encrypted 6190
total 6190`

I've attached the data files to be inspected.
Also in these files we have 

`2021-11-24 21:24:48   32.0 MiB 1cc4843e-4ad2-4132-9439-a7f921bc1cf2/a6a6f17a-89be-40a7-8a3f-42961729331e/backup/a6a6f17a-89be-40a7-8a3f-42961729331e/20211124-212445F/pg_data/sample.txt.gz`

Which has large enough size for a multipart upload if I'm not mistaken.

Archiving of wal files were also done and there should be wal archive files as well. Ex:

`2021-11-24 11:31:39   16.1 KiB 1cc4843e-4ad2-4132-9439-a7f921bc1cf2/a6a6f17a-89be-40a7-8a3f-42961729331e/archive/a6a6f17a-89be-40a7-8a3f-42961729331e/13-1/0000000100000000/000000010000000000000002-a78bffd955f477b776dfd729d0a08307243cc5fd.gz`

Is there anything specific that needs to be tested

dwsteele:
Your check to ensure all the files are getting encrypted looks good. The best way to test pgBackRest functionality in general is to run through the Quick Start: https://pgbackrest.org/user-guide.html#quickstart. This will exercise the backup/restore/check commands.

dwsteele:
One question before we get to far on this: why is this feature preferable to simply setting a default encryption key on the bucket? https://docs.aws.amazon.com/AmazonS3/latest/userguide/default-bucket-encryption.html. This seem much safer since it can be done when the bucket is configured and does not need to be added to every pgbackrest config.

tharinduamila-insta:
The advantage to using keys in the call is the ability to use a single bucket shared by different envs/tenants or teams. Which then means we only have to worry about managing one s3 bucket and it's associated object lifecycles and policies. It essentially gives you bit more flexibility on what's encrypted using what keys.

dwsteele:
@tharinduamila-insta How is your testing going? Have you been able to run through the quickstart?

tharinduamila-insta:
@dwsteele I'm sorry I misunderstood your last reply on testing. I thought what I've done was good enough and going through the quickstart was a general approach. 
I've tested below work as expected with the kms key config set

1. Backups 
2. Check Command : verified thewal archive had kms encryption set
3. Restore : verified PostgreSQL would reach running with archive-get set to pgbackrest with kms key set

dwsteele:
> I'm sorry I misunderstood

Or I did. No worries, seems like we are on the same page now!

tharinduamila-insta:
To confirm I'm assuming the testing has been completed now. Like I mentioned above I've gone through all steps to setup pgbackrest and use it to backup and restore a postgresql cluster from s3 with the server side KMS based encryption properties set. And used the info and check commands successfully.

dwsteele:
> To confirm I'm assuming the testing has been completed now.

Your testing looks complete. The patch still needs unit tests, though.

tharinduamila-insta:
Sorry for the long radio silence, we will work on adding the test cases. Any pointers for that would be greatly appreciated.

pgstef:
Hello,

> Sorry for the long radio silence, we will work on adding the test cases. Any pointers for that would be greatly appreciated.

A good starting point might be to read the [contributing](https://github.com/pgbackrest/pgbackrest/blob/main/CONTRIBUTING.md#writing-a-unit-test) guide. There's a part explaining where to find the tests and how to run them locally.

If you look at the GH actions logs, you'll see: `ERROR: [125]: c module storage/s3/storage is not fully covered (485/487 lines, 103/104 branches)`. So the test file you should probably look at is [s3Test.c](https://github.com/pgbackrest/pgbackrest/blob/main/test/src/module/storage/s3Test.c).

Fwiw, I usually find vagrant very handy to run the test locally. Example, once you're logged in the vagrant test box, you could probably generate the coverage report to indicate which part of the test is missing by running:
```bash
$ pgbackrest/test/test.pl --vm=f33 --vm-build
$ pgbackrest/test/test.pl --vm=f33 --coverage-only --module=storage --test=s3
```
(The report will be generated in `test/result/coverage/coverage.html`.)

Hope this helps :-)
Kind Regards

df7cb:
Thanks for the pointers. I got the coverage test running, but after staring that the test file for two hours, I have to admit I have no idea how the requests are built. As this is probably trivial to implement for someone who knows how that testsuite works, I would rather pledge to David's original offer to implement the unit tests once the code is there...

https://github.com/pgbackrest/pgbackrest/issues/1430#issuecomment-875719031

dwsteele:
I'll write the tests. Thanks for having a look, though. Maybe when you see the finished tests it will also make more sense.

dwsteele (approved):
Unit tests have been added and this should be ready to go now.

@tharinduamila-insta How would you like your name to appear in the release notes?

@df7cb You'll need to pull from https://github.com/dwsteele/pgbackrest-dev/tree/dws-kms-ci once I have updated the release notes. I'll let you know when.

dwsteele:
@df7cb Go ahead and pull in https://github.com/dwsteele/pgbackrest-dev/tree/dws-kms-ci now if you can. I can update contributor info later.

df7cb:
I pulled your changes.

df7cb:
Thanks @dwsteele !

PR #1568, opened 2021-11-15
include backuplabel and tablemapspace file size in log output

jreidthompson:
In cases where they are returned from postgres, include backuplabel and tablemapspace file sizes in the backup size value output in the log message .


dwsteele (approved):
Added linefeeds at 334cb050ab3d25cd8deba114c444633f73e11e5c to wrap lines longer than 132 characters.

PR #1571, opened 2021-11-18
Allow y/n arguments for boolean command-line options.

jreidthompson:
Allow arguments of y or n for boolean command line options ala how they are allowed in the config file.
Reference project card:
https://github.com/pgbackrest/pgbackrest/projects/2#card-62440553

parse.c 1417-1429  I removed these lines because I believe that at this point there are no options that cannot accept an argument.

### DIFF
--- a/src/config/parse.c
+++ b/src/config/parse.c
@@ -1423,8 +1423,8 @@ configParse(const Storage *storage, unsigned int argListSize, const char *argLis
                         option.negate = true;
                     else if (!strEqZ(optionArg, "y"))
                         THROW_FMT(
-                                OptionInvalidValueError, "when using argument with option '--%s', argument must be 'y' or 'n'",
-                                strZ(optionName));
+                                OptionInvalidValueError,
+                                "when using argument with boolean option '--%s', argument must be 'y' or 'n'", strZ(optionName));
                 }
 
                 // If the option requires an argument
--- a/src/config/parse.c
+++ b/src/config/parse.c
@@ -1417,19 +1417,15 @@ configParse(const Storage *storage, unsigned int argListSize, const char *argLis
                     // Handle boolean (only y/n allowed as argument)
                     if (parseRuleOption[option.id].type == cfgOptTypeBoolean)
                     {
+                        // Validate argument/set negate when argument present
                         if (optionArg != NULL)
                         {
-                            // Validate argument/set negate when indicated
                             if (strEqZ(optionArg, "n"))
-                            {
                                 option.negate = true;
-                            }
                             else if (!strEqZ(optionArg, "y"))
                             {
                                 THROW_FMT(
-                                    OptionInvalidValueError,
-                                    "when using argument with boolean option '--%s', argument must be 'y' or 'n'",
-                                    strZ(optionName));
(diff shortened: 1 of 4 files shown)


src/config/parse.c
```diff
@@ -1411,6 +1411,22 @@ configParse(const Storage *storage, unsigned int argListSize, const char *argLis
...
+                if (parseRuleOption[option.id].type == cfgOptTypeBoolean && optionArg != NULL)
+                {
+                    // Error on negated (--no-) options that also have an argument
+                    if (strncmp(strZ(optionName), OPTION_PREFIX_NEGATE, sizeof(OPTION_PREFIX_NEGATE) - 1) == 0)
+                        THROW_FMT(OptionInvalidError, "negated options cannot have an argument '--%s'", arg);
+
+                    // Validate argument/set negate when indicated
+                    if (strEqZ(optionArg, "n"))
+                        option.negate = true;
+                    else if (!strEqZ(optionArg, "y"))
```
dwsteele, line 1424: Since this statement breaks onto multiple lines it should be enclosed in braces, see https://github.com/pgbackrest/pgbackrest/blob/main/CODING.md#braces.

src/config/parse.c
```diff
@@ -1411,6 +1411,22 @@ configParse(const Storage *storage, unsigned int argListSize, const char *argLis
                 if (!option.found)
                     THROW_FMT(OptionInvalidError, "invalid option '--%s'", arg);
 
+                // Handle boolean option with argument(only y/n allowed)
```
dwsteele, line 1414: space needed between argument and (.

src/config/parse.c
```diff
@@ -1424,9 +1440,6 @@ configParse(const Storage *storage, unsigned int argListSize, const char *argLis
                         optionArg = strNewZ(argList[++argListIdx]);
                     }
                 }
-                // Else error if an argument was found with the option
```
dwsteele, line 1427: What about something like `--reset-pg1-host=xxx`? Looks like we need a new test for this condition now that the old one is not valid.

src/config/parse.c
```diff
@@ -1411,6 +1411,22 @@ configParse(const Storage *storage, unsigned int argListSize, const char *argLis
                 if (!option.found)
                     THROW_FMT(OptionInvalidError, "invalid option '--%s'", arg);
 
+                // Handle boolean option with argument(only y/n allowed)
+                if (parseRuleOption[option.id].type == cfgOptTypeBoolean && optionArg != NULL)
+                {
+                    // Error on negated (--no-) options that also have an argument
+                    if (strncmp(strZ(optionName), OPTION_PREFIX_NEGATE, sizeof(OPTION_PREFIX_NEGATE) - 1) == 0)
+                        THROW_FMT(OptionInvalidError, "negated options cannot have an argument '--%s'", arg);
```
dwsteele, line 1419: Shouldn't we just check `option.negate` here?

dwsteele (changes requested):

src/config/parse.c
jreidthompson: restored check and added test in commit 473e5597.

dwsteele (approved):
Made some adjustments in dce9dbbff1d. If those look good then this is ready to merge.

PR #1572, opened 2021-11-19
Add SIGTERM and SIGHUP handling to TLS server.

dwsteele:
SIGHUP allows the configuration to be reloaded. Note that the configuration will not be updated in child processes that have already started.

SIGTERM terminates the server process gracefully and sends SIGTERM to all child processes. This also gives the tests an easy way to stop the server.


jreidthompson (approved):
This looks good to me.

PR #1573, opened 2021-11-22
Use infoBackupDataByLabel() to log backup size.

jreidthompson:
Eliminate summing and passing of copied files sizes for logging backup size.   Instead, utilize infoBackupDataByLabel to pull the backup size for the log message.
Reference: https://github.com/pgbackrest/pgbackrest/projects/2#card-72983773


dwsteele (approved):
Converted a single use var into a function param at fea46fa66e508036b4c. Otherwise looks good.

PR #1584, opened 2021-12-07
Check that clusters are alive and correctly configured during a backup.

dwsteele:
Fail the backup if a cluster stops or the standby is promoted. Previously, shutting down the primary would cause an error but it was not detected until the end of the backup. Now the error will happen sooner and a promotion on the standby will also cause an error.


pgstef (approved):
Hi,

I ran my usual test suite against this branch and didn't found any regression. I tried to play a little bit with it and easily encountered the error `ERROR: [058]: standby is no longer is recovery` by promoting the standby during the backup. So, that's another nice safety-net I think.

Didn't noticed anything special about the code itself. So, this PR looks good to me :+1: 

Kind Regards

PR #1586, opened 2021-12-07
Error when restore is unable to find a backup to match the time target.

dwsteele:
This was previously a warning but the warning is easy to miss so a lot of time may be lost restoring and recovering a backup that will not hit the target.

We'll need something in the release notes since this is technically a breaking change:

IMPORTANT NOTE: If the restore command is unable to find a backup that matches a specified time target then an error will be thrown, whereas before a warning was logged.


src/command/restore/restore.c
```diff
@@ -313,18 +311,9 @@ restoreBackupSet(void)
...
-                    {
-                        // If a backup was not yet found then set the latest from this repo as the backup that might be used
-                        RestoreBackupData candidate = restoreBackupData(
-                            latestBackup.backupLabel, repoIdx, infoPgCipherPass(infoBackupPg(infoBackup)));
-
-                        lstAdd(backupCandidateList, &candidate);
-                    }
                 }
                 else
                 {
```
jreidthompson, line 319: line 320 comment, should "(or time provided was not valid)" be removed.  Would the error thrown in getEpoch prevent us from reaching this point due to an invalid date?

jreidthompson (approved):
I noted a question about a comment - other than that looks good to me

src/command/restore/restore.c
dwsteele: Fixed at f6c1549b6b.

hunleyd (approved):

PR #1589, opened 2021-12-08
Fix restore delta link mapping when path/file already exists.

dwsteele:
If a path/file was remapped to a link using either --link-map or --link-all there would be no affect if the path/file already existed. If a link existed it would be properly updated and converting a link to a path/file also worked.

The issue happened during delta cleanup, which failed to check if the existing path/file had been remapped to a link.

Add checks for newly mapped path/file links and remove the old path/file we required.


jreidthompson (approved):
These changes look good to me

PR #1594, opened 2021-12-10
Fix socket leak on connection retries.

dwsteele:
This leak was caused by the file descriptor variable getting clobbered after a long jump. Mark it as volatile to fix.

Testing this is a bit complex because the issue only happens in optimized builds, if at all. Put the test into the performance suite, which is always optimized, until a better idea presents itself.


jreidthompson (approved):
changes look good to me

PR #1597, opened 2021-12-15
Add TLS server documention.

dwsteele:
Add basic documentation and also make the feature visible.

To build the documentation with TLS:
`pgbackrest/doc/doc.pl --out=html --include=user-guide --require=/replication --pre --var=os-type=rhel`


jreidthompson (approved):
These look good to me

dwsteele:
Updated at f06101de77a9 to fix a typo in the commit message.

PR #1607, opened 2021-12-27
Add verify output and verbose options.

jreidthompson:
Add options (none/text) for verify command console output, default to none. Add verbose option for text output, default to no.  Update help and docs for verify command and options
https://github.com/pgbackrest/pgbackrest/projects/2#card-33947719


src/command/verify/verify.c
```diff
@@ -1632,6 +1632,63 @@ verifyProcess(unsigned int *errorTotal)
...
+        String *status = strNew();
+
+        if (*errorTotal == 0)
+            strCatFmt(status,"%s%s",  strZ(strNewZ("All archives and backups passed verification.")), LF_Z);
+        else
+        {
+            strCatFmt(
+                status,
+                "%s%s",
+                strZ(strNewZ("One or more archive or backup contains errors. Invoke verify with verbose option for more details.")),
```
pgstef, line 1659: Hey,
Shouldn't this output `Invoke verify with verbose option for more details.` be removed if the verbose option is already provided? It sounds a bit weird to request it while the user already provided it :confused:

src/command/verify/verify.c
```diff
@@ -1632,6 +1632,63 @@ verifyProcess(unsigned int *errorTotal)
...
+            {
+                strCatFmt(
+                    resultStr, "%s%s%s", strZ(strNewFmt("Stanza: %s", strZ(cfgOptionStr(cfgOptStanza)))), LF_Z,
+                    strZ(strNewFmt("Status: %s", strZ(status))));
+            }
+        }
+
+        MEM_CONTEXT_PRIOR_BEGIN()
+        {
+            result = strDup(resultStr) ;
```
pgstef, line 1683: extra space before `;`?

src/build/help/help.xml
```diff
@@ -2352,6 +2352,26 @@
...
+                        </text>
+
+                        <example>text</example>
+                    </option>
+                    <!-- OPERATION - VERIFY COMMAND - VERBOSE OPTION -->
+                    <option id="verbose" name="Verbose">
+                        <summary>Output verbose result.</summary>
+
+                        <text>
+                            <p>Verbose defaults to false and provides a minimal verify passed/verify failed response. True provides a more detailed response. </p>
```
pgstef, line 2370: extra space before `</p>`?

src/command/verify/verify.c
pgstef: Also wondering, since this is "one or more", perhaps `archive(s) or backup(s)` would make sense? :thinking:

jreidthompson: addressed in 1ab4e8eb20873848a78df60be92e2356098786ec

src/command/verify/verify.c
jreidthompson: addressed in 1ab4e8eb20873848a78df60be92e2356098786ec

src/build/help/help.xml
jreidthompson: addressed in 1ab4e8eb20873848a78df60be92e2356098786ec

jreidthompson:
I have in my notes for text output  'non-verbose, just see the errors'.  This commit implements that, but I'm concerned I may have misinterpreted the intention of that note.

test/src/module/command/verifyTest.c
```diff
@@ -1003,6 +1003,39 @@ testRun(void)
...
+        harnessLogLevelReset();
+
+        // Verify text output, verbose, with verify failures
+        errorTotal = 0;
+        String *result = NULL;
+        result = verifyProcess(&errorTotal);
+        TEST_RESULT_STR_Z(
+            verifyOutputText(result, &errorTotal),
+            "stanza: db\n"
+            "status: error (one or more archive or backup contains errors)\n"
```
pgstef, line 1022: probably the test text should be adjusted here too?

src/command/verify/verify.c
pgstef: Sounds better, thanks :+1:

dwsteele (changes requested):
Before I get too deep into reviewing this, I think we need to resolve what verbose means. My thought was that the errors would always be shown even when verbose is false. This code is only showing that there *were* errors when verbose is false.

I'm not sure we can just reuse the result string as is for this feature. We'll need to filter what is added to the result string based on verbose.

src/command/verify/verify.c
```diff
@@ -1628,6 +1685,62 @@ verifyProcess(unsigned int *errorTotal)
     FUNCTION_LOG_RETURN(STRING, result);
 }
 
+/***********************************************************************************************************************************
+Format the text result
+***********************************************************************************************************************************/
+static String *
+verifyOutputText(const String *verifyresult, const String *errorResult, const unsigned int *errorTotal)
```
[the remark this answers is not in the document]
jreidthompson: Logic moved to verifyProcess().  Added function to dedupe generation of file error type/count string.

dwsteele:
I think it makes more sense to decouple the verbose and output options. It is confusing to me that requesting console output changes what gets logged. Done in 4dc29b79af5f6f but tests and help need to be updated.

PR #1610, opened 2022-01-04
Add backup LSNs to info command output.

pgstef:
Add lsnStart and lsnStop to backup info.
Displaying it in the global part might be overwhelming, so only when requesting the backup details seems enough. This might handy to identify the lsn target to restore.


src/command/info/info.c
```diff
@@ -821,6 +827,19 @@ formatTextBackup(const DbGroup *dbGroup, String *resultStr)
         else
             strCatZ(resultStr, "n/a\n");
 
+        if (kvGet(backupInfo, BACKUP_KEY_LSN_VAR) != NULL)
+        {
+            KeyValue *lsnInfo = varKv(kvGet(backupInfo, BACKUP_KEY_LSN_VAR));
+
+            if (kvGet(lsnInfo, KEY_START_VAR) != NULL && kvGet(lsnInfo, KEY_STOP_VAR) != NULL)
```
dwsteele, line 834: I think this check should be moved up to line 458. Both values should be NULL or non-NULL even in the JSON.

dwsteele (changes requested):
Looks pretty good overall, see comment.

src/command/info/info.c
pgstef: Hm :thinking: Currently, it won't reach json since json format doesn't support `--set` :slightly_smiling_face:

I based the code on the wal start/stop (lines 410 & 821) but I guess if we move the check, the entire lsn section can be skipped directly.

Does 10869a7 sounds like what you had in mind ?

dwsteele: That's the right direction, but I think we do need to add this info to the JSON output. Other things in --set text output are omitted from JSON because they could be very large, but that's not the case here.

In summary, I think this info should be in the JSON and --set text output, but not in the general text output. Make sense?

pgstef: That would make 900d3ac ?

I'd prefer to have both outputs behave accordingly to be honest. I understand it would be hard to change the json key names once we open support for json and `--set` but I'm not sure to see why the lsn info would be more important than (e.g.) the error list.

For now, having it in json and --set text might be enough. We still may move it to json --set for consistency later...

dwsteele: LSN is not more important, but it is much smaller. The reason to exclude error list from general JSON output is because it could be many megabytes of data for one backup. I'd rather include everything *unless* we have a reason to exclude it. In this case, I think LSN is to noisy for general text output, but should be included everywhere else.

dwsteele (approved):
Some updates in 3fd274257560c4, otherwise looks good to me.

pgstef:
> Some updates in [3fd2742](https://github.com/pgbackrest/pgbackrest/commit/3fd274257560c44d0a2590b33d1d1133aa562ab3), otherwise looks good to me.

Thanks!

PR #1611, opened 2022-01-04
Remove support for PostgreSQL 8.3/8.4.

dwsteele:
There is no evidence that users need 8.3/8.4 anymore but it does cost us in terms of development and testing, especially now that we have a number of new backup/restore features planned.

It seems to make sense to remove this support now. If there are users who need to use/migrate from these versions they can use an older version of pgBackRest.


pgstef (approved):
Hi,

Fwiw, I tested those changes against my usual test suite and didn't found any problem so far.

Overall the PR looks good to me. After a quick look I didn't find any other leftover.

+1 for removing PG < 9.0 support :-)

test/src/module/postgres/interfaceTest.c
```diff
@@ -220,18 +219,18 @@ testRun(void)
...
 
         TEST_ERROR(pgWalFromBuffer(result), FormatError, "wal segment size is 33554432 but must be 16777216 for PostgreSQL <= 10");
 
         //--------------------------------------------------------------------------------------------------------------------------
         memset(bufPtr(result), 0, bufSize(result));
-        hrnPgWalToBuffer((PgWal){.version = PG_VERSION_83, .systemId = 0xEAEAEAEA, .size = PG_WAL_SEGMENT_SIZE_DEFAULT}, result);
+        hrnPgWalToBuffer((PgWal){.version = PG_VERSION_90, .systemId = 0xEAEAEAEA, .size = PG_WAL_SEGMENT_SIZE_DEFAULT}, result);
         storagePutP(storageNewWriteP(storageTest, walFile), result);
 
         TEST_ASSIGN(info, pgWalFromFile(walFile, storageTest), "get wal info v8.3");
```
jreidthompson, line 231: line 231 references v8.3

jreidthompson (approved):
looks good to me. noted one reference to v8.3

test/src/module/postgres/interfaceTest.c
dwsteele: Fix in e3975273d30.

PR #1619, opened 2022-01-09
Add read range to all storage drivers.

dwsteele:
The range feature allows reading out an arbitrary chunk of a file and will be important for efficient small file support.

Now that all drivers are required to support ranges remove the storageFeatureLimitRead feature flag that was implemented only by the Posix driver.


jreidthompson (approved):
I don't see anything amiss

PR #1621, opened 2022-01-13
Show backup percent complete in info output.

jreidthompson:
Convert lock file second line to json format.  Have the backup process update the lock file json content with percentage complete as the backup progresses. Currently updates per tenth of a percentage complete.


src/common/lock.c
```diff
@@ -280,3 +287,58 @@ lockRelease(bool failOnNoLock)
 
     FUNCTION_LOG_RETURN(BOOL, result);
 }
+
+/**********************************************************************************************************************************/
+bool
+writeLockPercentageComplete(const String *lockPath, const String *stanza, const String *execId, const double percentageComplete)
```
dwsteele, line 293: All extern'd functions in this module should begin with "lock". So `lockWritePercentageComplete`. Also I think percentage is more verbose than needed. How about using "percent" (here and in variable names)?

src/common/lock.c
```diff
@@ -280,3 +287,58 @@ lockRelease(bool failOnNoLock)
...
+    ASSERT(lockPath != NULL);
+    ASSERT(stanza != NULL);
+    ASSERT(execId != NULL);
+
+    bool status = false;
+
+    // This context should already exist, noop if it doesn't
+    if (lockMemContext != NULL)
+    {
+        MEM_CONTEXT_BEGIN(lockMemContext)
```
dwsteele, line 310: I don't think any of this needs to be in `lockMemContext`. A temp mem context would be better.

src/common/lock.c
```diff
@@ -280,3 +287,58 @@ lockRelease(bool failOnNoLock)
...
+        {
+            // We are only interested in the backup lock file
```
dwsteele, line 312: I don't think most of this is needed. We already have the backup lock file open -- so we should just be able to grab the handle and write to it. No need to check if the file exists. Instead error if the backup lock is not held.

src/common/lock.h
```diff
@@ -37,4 +37,7 @@ bool lockAcquire(
 // Release a lock
 bool lockRelease(bool failOnNoLock);
 
+// Write the backup percentage complete value into the backup lock file
+bool writeLockPercentageComplete( const String *lockPath, const String *stanza, const String *execId, const double percentagComplete);
+
```
dwsteele, line 42: Remove extra space before const.

dwsteele: And remove const before double. This const should only be in the C file -- the caller does not need to know we won't modify this param.

src/common/lock.c
dwsteele: All pointer params should be const.

src/common/lock.c
dwsteele: i.e., `CHECK(AssertError, "backup lock is not held", lockTypeHeld = lockTypeBackup)`.

src/common/lock.c
```diff
@@ -280,3 +287,58 @@ lockRelease(bool failOnNoLock)
...
+                    {
+                        // Build kv for second line json
+                        KeyValue *keyValue = kvNew();
+                        kvPut(keyValue, EXEC_ID_VAR, varNewStr(execId));
+                        const String *const percentageCompleteStr = strNewFmt("%.2lf", percentageComplete);
+                        kvPut(keyValue, PERCENTAGE_COMPLETE_VAR, varNewStr(percentageCompleteStr));
+
+                        // Move to the beginning of the file, and overwrite the contents with the udpated data
+                        lseek(lockFd[lockIdx], 0, SEEK_SET);
+                        ioFdWriteOneStr(lockFd[lockIdx], strNewFmt("%d" LF_Z "%s" LF_Z, getpid(), strZ(jsonFromKv(keyValue))));
```
dwsteele, line 333: Check for error here using THROW_ON_SYS_ERROR. There's an example of how to do this with lseek in the Posix driver.

dwsteele (changes requested):
See comments. Mostly the logic for rewriting the lock file seems more complicated than is required.

src/common/lock.c
jreidthompson: updated in 0f56ed01..1f3fbd2b

src/common/lock.c
jreidthompson: updated in 0f56ed01..1f3fbd2b

src/common/lock.c
jreidthompson: updated in 0f56ed01..1f3fbd2b

src/common/lock.h
jreidthompson: updated in 0f56ed01..1f3fbd2b

src/common/lock.c
jreidthompson: updated in 0f56ed01..1f3fbd2b

dwsteele (changes requested):
This looks pretty good now. I think the next step is to create a struct and two functions to read/write the JSON. Right now there are two places where the JSON is being written and there will be two places where it is read.

dwsteele (changes requested):
I think part of what is making this patch hard is that the lock code is very old and doesn't use the newer IO code that was written since then. There was no reason to update the lock code, but now that we are making things more complex I think there is. Let me look at a bit of refactoring and see if that makes what we need here clearer.

dwsteele (changes requested):
See comments.

src/common/lock.c
```diff
@@ -29,6 +30,8 @@ Constants
 // Indicates a lock that was made by matching exec-id rather than holding an actual lock. This disguishes it from -1, which is a
 // general system error.
 #define LOCK_ON_EXEC_ID                                             -2
+VARIANT_STRDEF_STATIC(EXEC_ID_VAR,                                  "execId");
+VARIANT_STRDEF_STATIC(PERCENT_COMPLETE_VAR,                         "percentComplete");
```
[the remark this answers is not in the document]
jreidthompson: b1d5573 updated

src/command/backup/backup.c
```diff
@@ -1138,7 +1139,8 @@ Log the results of a job and throw errors
 static void
 backupJobResult(
     Manifest *const manifest, const String *const host, const Storage *const storagePg, StringList *const fileRemove,
-    ProtocolParallelJob *const job, const bool bundle, const uint64_t sizeTotal, uint64_t *const sizeProgress)
+    ProtocolParallelJob *const job, const bool bundle, const uint64_t sizeTotal, uint64_t *const sizeProgress,
+    double *const currentPercentComplete)
```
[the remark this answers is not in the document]
jreidthompson: c533677 updated

src/command/backup/backup.c
```diff
@@ -1314,6 +1317,13 @@ backupJobResult(
                         bundleId, bundleOffset);
                 }
             }
+
+            // Update currentPercentComplete and lock file when the change is significant enough
+            if (percentComplete - *currentPercentComplete > .1)
+            {
+                *currentPercentComplete = percentComplete;
+                lockWriteDataP(lockTypeBackup, .percentComplete = &percentComplete);
```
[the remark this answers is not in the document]
jreidthompson: e2a078e

src/command/info/info.c
```diff
@@ -71,6 +71,7 @@ VARIANT_STRDEF_STATIC(STATUS_KEY_CODE_VAR,                          "code");
 VARIANT_STRDEF_STATIC(STATUS_KEY_LOCK_VAR,                          "lock");
 VARIANT_STRDEF_STATIC(STATUS_KEY_LOCK_BACKUP_VAR,                   "backup");
 VARIANT_STRDEF_STATIC(STATUS_KEY_LOCK_BACKUP_HELD_VAR,              "held");
+VARIANT_STRDEF_STATIC(STATUS_KEY_LOCK_BACKUP_PERCENT_COMPLETE_VAR,  "percent-complete");
```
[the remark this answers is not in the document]
jreidthompson: b1d5573 updated

dwsteele (approved):
Approved with minor edits in 81a6f54536a325d.

sumanmichael:
Hello @dwsteele / @jreidthompson

I really appreciate your contribution to this feature, this is very much needed. But It would be great if we could also bring this to JSON output.

I've made some small changes to the `command/info/info.c` to include a percentage completion field as a string in `backupLockKv`. I'd love to get your thoughts on whether you think this approach works. Check out this commit: https://github.com/sumanmichael/pgbackrest/commit/eb0462513bec5f910c004e15819e91eb61a8032e . Let me know your feedback!

dwsteele:
@sumanmichael Please don't respond to closed issues/PRs. Nobody will see it.

I happened to stumble across this while look at #2007. You might want to response there.

PR #1630, opened 2022-01-20
Remove primary flag from manifest.

dwsteele:
This flag was only being used by the backup command after manifestNewBuild() and had no other uses. There was a time when it was important for integration testing but the unit tests now fulfill this role.

Since backup is the only code concerned with the primary flag, move the code into the backup module.

We don't have any cross-version testing but this change was tested manually with the most recent version of pgBackRest to make sure it was tolerant of the missing primary info. When an older version of pgBackRest loads a newer manifest the primary flag will always be set to false, which is fine since it is not used.


pgstef:
I've just run my usual tests against this branch and didn't find any issue, nor seen anything obvious in the code changes.

Aren't you afraid by removing this flag that we'll lose track from where did come those files ? (standby/primary)
Having that info in the manifest seemed interesting at least for info/debug.

But if you think it's best to remove it, this PR looks good to me ;-)

jreidthompson (approved):
Looks good for me

dwsteele:
> Having that info in the manifest seemed interesting at least for info/debug.

Except that we have never used it and it only tells that it copied from the primary *at the time of the backup*. What host was that? We don't know. To know where a file *should* be copied from we only need to look at the regexp.

Being marked as primary in the manifest does not guarantee that the backup code will do the right thing so we still need to have other tests for that.

Finally, this field is in the way of some optimizations that we want to do, so away it goes.

pgstef:
> Finally, this field is in the way of some optimizations that we want to do, so away it goes.

You're indeed right, I didn't saw it like that. Thanks for the explanation :-)

PR #1642, opened 2022-01-28
Add archive-missing-retry option.

dwsteele:
Retry a WAL segment that was previously reported as missing by the archive-get command. This prevents notifications in the spool path from a prior restore from being used and possibly causing a recovery failure if consistency has not been reached.

Disabling this option allows PostgreSQL to more reliably recognize when the end of the WAL in the archive has been reached, which permits it to switch over to streaming from the primary. With retries enabled, a steady stream of WAL being archived will cause PostgreSQL to continue getting WAL from the archive rather than switch to streaming.

When disabling this option it is important to ensure that the spool path for the stanza is empty. The restore command does this automatically if the spool path is configured at restore time. Otherwise, it is up to the user to ensure the spool path is empty.


fdr:
I'm not sure what it means to have a "clean" spool. Am I on the hook to delete the spool sometimes? Is this in reference to multiple restore and multiple clusters on one server?

dwsteele:
> I'm not sure what it means to have a "clean" spool.

I mean empty. I can update the docs to say that if you think it would be clearer.

> Am I on the hook to delete the spool sometimes?

If `spool-path` is configured at restore time, then no. Restore will clean the path.

> Is this in reference to multiple restore and multiple clusters on one server?

This is in reference to multiple restores/recoveries of the same stanza on a single host. Notification files in the spool path from the prior recovery can cause pgBackRest to report a WAL file as missing (with this new option disabled). If PostgreSQL is trying to get to consistency when that happens, it will panic.

pgstef:
Overall the code seems pretty simple and this PR looks good to me.
As far as I understand it, this option only applies to async archive-get right? May I then suggest to add a mention about that in the doc?

src/build/help/help.xml
```diff
@@ -1256,6 +1256,21 @@
                         <example>n</example>
                     </config-key>
 
+                    <!-- ======================================================================================================= -->
+                    <config-key id="archive-missing-retry" name="Retry Missing WAL Segment">
+                        <summary>Retry missing WAL segment</summary>
+
+                        <text>
+                            <p>Retry a WAL segment that was previously reported as missing. This prevents notifications in the spool directly from a prior restore being used and possible causing a recovery failure if consistency has not been reached.</p>
```
pgstef, line 1264: > from a prior restore being used and possible causing a recovery failure
and *possibly* causing ?

pgstef (approved):

dwsteele:
> As far as I understand it, this option only applies to async archive-get right? May I then suggest to add a mention about that in the doc?

Good idea. Done in 5f44b4d91.

src/build/help/help.xml
```diff
@@ -1256,6 +1256,21 @@
...
+
+                            <p>Disabling this option allows <postgres/> to more reliably recognize when the end of the WAL in the archive has been reached, which permits it to switch over to streaming from the primary. With retries enabled, a steady stream of WAL being archived will cause <postgres/> to continue getting WAL from the archive rather than switch to streaming.</p>
+
+                            <p>When disabling this option it is important to ensure that the spool path for the stanza is clean. The <cmd>restore</cmd> command does this automatically if the spool path is configured at restore time. Otherwise, it is up to the user to ensure the spool path is clean.</p>
```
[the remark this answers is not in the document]
dwsteele: Done in 5f44b4d91.

src/build/help/help.xml
dwsteele: Fixed in 5f44b4d91.

pgstef:
> > As far as I understand it, this option only applies to async archive-get right? May I then suggest to add a mention about that in the doc?
> 
> Good idea. Done in [5f44b4d](https://github.com/pgbackrest/pgbackrest/commit/5f44b4d91fac507f4e34e3ed41a376289b738e9f).

> Retry a WAL segment that was previously reported as missing by the <cmd>archive-get</cmd> command

I think the "async" part is important to mention too. Perhaps a simple `When asynchronous mode is enabled, retry...`?

dwsteele:
> I think the "async" part is important to mention too. Perhaps a simple When asynchronous mode is enabled, retry...?

How about this: 8185c9583?

pgstef:
> How about this: [8185c95](https://github.com/pgbackrest/pgbackrest/commit/8185c9583482e49e8d181cd3c0f59b4ad94f726f)?

Fine by me, thanks !

Issue #600, opened 2018-09-27
allow `--type` to filter output of `info` command


hunleyd:
What do you think of allowing `--type` to filter the output of the `info` command? 

For example, `pgbackrest --type=full info` would be the command for 'show me all my full backups' .. instead of having to scroll up/down the output eyeballing it yourself?

Of course, you'd want to be able to use `--type=diff` and `--type=incr` as well and it should also work when `stanza` is specified..

dwsteele:
What do you think @cmwshang?  If it's easy it might be worth doing in Perl, otherwise we can wait for the C port.

cmwshang:
Devil is in the details so it's not as "easy" as you'd think. It's a good requested yet I would strongly suggest we wait until after we have ported it to C.

dwsteele:
OK, agreed.  I've added this to the backlog at https://github.com/pgbackrest/pgbackrest/projects/2#card-13503164

I hope to start the C port of the info command soon, but we can't change the behavior as long as some code paths run through Perl which will probable be the case into December.

dwsteele:
The `info` command has been migrated to C, but this is not a huge priority at the moment. Unless you want to take a whack at it?

dwsteele:
@hunleyd PR #1651 has been submitted for this. You might want to have a look.

PR #1651, opened 2022-02-08
Add backup type filter to info command.

pgstef:
Support `--type` option in info command to display only a specific backup type.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -34,6 +34,18 @@
 
                         <p>Add support for <proper>AWS</proper> <proper>S3</proper> server-side encryption using <proper>KMS</proper>.</p>
                     </release-item>
+
+                    <release-item>
+                        <github-issue id="600"/>
+                        <github-pull-request id="1651"/>
+
+                        <release-item-contributor-list>
+                            <release-item-contributor id="stefan.fercot"/>
+                            <release-item-reviewer id="david.steele"/>
+                        </release-item-contributor-list>
+
+                        <p>Add backup type filter to <cmd>info</cmd> command.</p>
+                    </release-item>
                 </release-feature-list>
 
                 <release-improvement-list>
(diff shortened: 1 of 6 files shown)


dwsteele:
Seems simple enough. Is this in reference to some specific issue or ask? In brief, what's the use case? I mostly see this as being useful for filtering on full backups.

pgstef:
Indeed, I wanted to compare the performance hit by filtering at the code level rather than in an external perl command (or using jq) to fetch only a specific type of backups.

And since it was in the [backlog](https://github.com/pgbackrest/pgbackrest/projects/2#card-13503164) and closed [issue](https://github.com/pgbackrest/pgbackrest/issues/600), I gave it a try.

I also mostly sees it having benefits for full backups but it doesn't seem to cost much to support all types.

Thanks for the review :-)

dwsteele:
I don't think we have any options that are filtering the JSON data yet. Are we sure this is something we want to do?

pgstef:
> I don't think we have any options that are filtering the JSON data yet. Are we sure this is something we want to do?

Can't `--stanza` and `--repo` already been considered as filters ? 
Currently `--set` is on hold but it would be nice to get there at some point.

Why would the backup-type filter be an issue in JSON output ?
From a monitoring point-of-view, it's very common to only look at full backups :man_shrugging:

dwsteele (approved):
Made some help edits in e60a770c28a. If those look good this is ready to commit.

pgstef:
> Made some help edits in [e60a770](https://github.com/pgbackrest/pgbackrest/commit/e60a770c28a4731e1880bfffa68363f0c9e45d44). If those look good this is ready to commit.

Looks great to me. Thanks !

hunleyd:
thanks for tackling this @pgstef ! and thanks for the heads up to come look at it @dwsteele . this looks great to me.

PR #1652, opened 2022-02-08
Remove logic that tried to determine additional file system compression.

dwsteele:
In theory, the additional stat() call after a file has been copied to the repo can determine if additional compression has been applied by the file system. However, it has been a very long time since we tested this in practice. There are currently no unit tests that accurately test this feature, since it requires a compressed file system like ZFS to work, which never seemed worth the extra cost.

It can also add a lot of time to backups if there are a large quantity of small files.

In addition, it stands as a blocker for combining files for small file support since it is no longer possible to get per-file sizes from the viewpoint of the file system. There are several ways this could be reworked but none of them are easy while at the same time maintaining current info functionality.

It doesn't seem worth keeping an untested feature that will only work in some special cases (if it still works) when it is blocking development.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -59,6 +59,18 @@
 
                         <p>Remove support for <proper>PostgreSQL</proper> <id>8.3</id>/<id>8.4</id>.</p>
                     </release-item>
+
+                    <release-item>
+                        <github-pull-request id="1652"/>
+
+                        <release-item-contributor-list>
+                            <release-item-contributor id="david.steele"/>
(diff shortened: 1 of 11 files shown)


jreidthompson (approved):
looks good for me

pgstef (approved):
The compressed/uncompressed size is very nice (to at least have an idea of the compress-level or compress-type impact). Not supporting "storage level" compression size info for performance/development purpose seems acceptable though.

So, the PR looks good to me too :+1:

PR #1655, opened 2022-02-09
Bundle files in the repository during backup.

dwsteele:
Bundle (combine) smaller files during backup to reduce the number of files written to the repository (enable with `--bundle`). Reducing the number of files is a benefit on all file systems, but especially so on object stores such as S3 that have a high file creation cost. Another benefit is that zero-length files are only stored as metadata in the manifest.

Files are batched up to `bundle-size` and then compressed/encrypted individually and stored sequentially in the bundle. The bundle id and offset of each file is stored in the manifest so files can be retrieved randomly without needing to read the entire bundle. Files are ordered by timestamp descending when being assigned to bundles to reduce the amount of random access that needs to be done. The idea is that bundles with older files can be read in their entirety on restore and only bundles with newer files will get fragmented.

Bundles are a custom format with metadata stored in the manifest. `tar` was considered but it is too limited a format, the major issue being that the size of the file must be known in advance and that is very contrary to how pgBackRest works, especially once we introduce page-level incremental backups.

Bundles are stored numbered in the `bundle` directory. *NOTE: should we be creating subdirectories to reduce the total number of files in a dir?* Some files may still end up in `pg_data` if they are added after the backup is complete. `backup_label` is an example. *Maybe it would also be a good idea to store `pg_control` separately?*

Currently, only the backup command works in batches. The restore and verify commands use the offsets to pull individual files out of the bundle. It seems better to finalize how this is going to work before optimizing the other commands. Even as is, this is a major step forward, and all commands function with bundling.

One caveat: `resume` is currently not supported when `bundle` is enabled.

The plan is commit essentially what we have here as an internal feature so people can experiment with it.


Reid-Thompson (approved):
looks good to em

jreidthompson (approved):
looks good to me

PR #1658, opened 2022-02-13
Do not retry fatal errors.

dwsteele:
There is some evidence that retrying fatal errors, especially out of memory errors, may cause lockups. It makes sense to report fatal errors as quickly as possible and bypass retries. This may or not fix the lockup issue but it is worth doing either way.

For now, the only fatal errors will be AssertError and MemoryError.


Reid-Thompson (approved):
looks good for me

PR #1662, opened 2022-02-14
Limit files that can be bundled.

dwsteele:
Limit which files can be added to bundles, which allows resume to work reasonably well. On resume, the bundles are removed and any remaining file is eligible to be to be resumed.

Also reduce the bundle-size default to 20MiB. This is pretty arbitrary, but a smaller default seems better.


pgstef (approved):
This looks good to me. Didn't spot anything obvious in the code.
Resume a bundled backup seemed to work correctly during my tests.

Also, I'm currently collecting a few perf numbers and `--bundle-limit=2MiB --bundle-size=20MiB` gives me the best improvement so far (local backup, more than 1M files - S3 backups in progress). So the new defaults looks good to me too :-)

PR #1667, opened 2022-02-20
Retry on page checksum failure.

dwsteele:
Rather than attempting to filter page checksum failures by LSN, just retry when there is a page checksum failure. If the page has not changed since the last read report it as an error. If the page has changed, then PostgreSQL must be modifying the page so we can ignore the error because a full page write (and possibly updates) will be in the WAL.


sfrost (approved):
Looks good to me.

PR #1669, opened 2022-02-21
Improve validation of zero pages.

dwsteele:
Checking that pd_upper == 0 is not enough since this field may be corrupted. Still use pd_upper as a quick check, but when it is zero proceed to check the rest of the page to ensure it is also all zeroes.


sfrost (approved):
Looks good to me.

PR #1673, opened 2022-02-23
Automatically strip trailing slashes for repo-ls paths.

pgguru:
Trailing slashes in at least some of the backend repo types were preventing `repo-ls` from
displaying any content (presumably backend-specific behavior).  Since the path with the slash should
be equivalent to the path without the slash, just remove it if provided by the user to prevent this
little UI wart.


dwsteele:
So this has exposed another issue -- if an invalid absolute path is passed then the user gets an assertion out of the storage layer. The error is pretty clear, but a user should never see an assertion.

Let me see how easy that is to fix. If it turns out to be a problem we can just commit this as-is for now.

dwsteele:
This affects some other `repo-*` commands as well, so I think it makes sense to centralized the logic. I think it would be better to just commit this as-is and fix the assert error later. I have created a project card for that: https://github.com/pgbackrest/pgbackrest/projects/2#card-78166776.

dwsteele (approved):

PR #1675, opened 2022-02-25
Allow case-insensitive matching of HTTP connection header values.

dwsteele:
The specification allows values for the connection header to be case-insensitive. See https://www.rfc-editor.org/rfc/rfc7230#section-6.1.


jreidthompson (approved):
looks good

dwsteele:
Somewhere in here I managed to kill off the test, so brought that back.

PR #1676, opened 2022-02-26
Improve stop command to honor stanza option.

jreidthompson:
Have the stop command, when force and stanza options are given, terminate only processes holding lock files for the given stanza.  Prior to these changes, termination of all processes holding lock files regardless of stanza occurred.
See
https://github.com/pgbackrest/pgbackrest/projects/2#card-75533135


src/command/control/stop.c
```diff
@@ -53,18 +53,18 @@ cmdStop(void)
             if (cfgOptionBool(cfgOptForce))
             {
                 const String *lockPath = cfgOptionStr(cfgOptLockPath);
+
+                // If no stanza given, operate on all lock files, else restrict to given stanza's lock files
                 StringList *lockPathFileList = strLstSort(
-                    storageListP(storageLocal(), lockPath, .errorOnMissing = true), sortOrderAsc);
+                    storageListP(storageLocal(), lockPath, .errorOnMissing = true, .expression =
+                    cfgOptionStrNull(cfgOptStanza) == NULL ? strNewFmt("-.*\%s$", LOCK_FILE_EXT) : strNewFmt("%s-.*\%s$",
+                    strZ(cfgOptionStrNull(cfgOptStanza)), LOCK_FILE_EXT)), sortOrderAsc);
```
dwsteele, line 61: I'm concerned that using a regexp here is not as precise as we would like. There are basically no limits on stanza names so there is nothing to stop the user from using `.` or some other character that has special meaning for a regexp. I think we might better off generating a prefix and matching each file in the list against the prefix when needed.

dwsteele (changes requested):

src/command/control/stop.c
jreidthompson: altered code to remove regex and match on generated filenames when stanza is provided

dwsteele: Why can't we just check the prefix as I suggested above? So, just check each file that comes back and see if it begins with `[stanza]-` when we are limiting by stanza.

dwsteele: Hmmm, I'm guessing you are concerned about having two stanzas, e.g. `demo` and `demo-2`. In that case prefix matching is not going to work well.

jreidthompson: Thanks for the updates and the review this morning.  The changes make the code much cleaner.

dwsteele (approved):

jreidthompson (commented):
These changes make the code cleaner/clearer.  All look good. Thanks.

PR #1677, opened 2022-02-28
Handle TLS servers that do not close connections gracefully.

dwsteele:
Some TLS server implementations will simply close the socket rather than correctly closing the TLS connection. This causes problems when `connection: close` is specified with no `content-length` or chunked encoding and we are forced to read to EOF. It is hard to know if this is a real EOF or an network error.

In cases where we can parse the content and (hopefully) ensure it is correct, allow the closed socket to serve as EOF. This is not ideal, but the change in 8e1807cd means that currently working servers with this issue will stop working after 2.35 is installed, which seems too risky.

NOTE: a better test will need to be written to be sure this does not regress. We have code coverage, but the tests are too disconnected to ensure this works end to end.


pgguru (approved):
Code review looks fine; is there additional validation that needs to be added for the xml/json incomplete data, or those code paths already existed and this just lets us hit them now instead of dying?

Was there additional checking of `errorSys` that needed to be done in the `SSL_ERROR_SYSCALL` code path before just ignoring, or (based on a little research) does it look like `SSL_ERROR_SYSCALL` is just a catch-all "we don't know what happened", so no useful `errno` would be provided?

sfrost (approved):
Looks good to me.

dwsteele:
@pgguru,

> is there additional validation that needs to be added for the xml/json incomplete data

The validation is done by the caller. It might make sense to build that into the HTTP client but that is for another day.

> so no useful errno would be provided

The errno/message will be appended by THROW_SYS_ERROR_CODE() when it is non-zero. I have never seen a case where that happens, though, so it is getting stomped on my something in openssl, I think.

dwsteele:
Added tests to the tls and http modules to make sure this change has sufficient coverage to prevent a regression.

PR #1688, opened 2022-03-10
Improve error message for invalid repo-azure-key.

jreidthompson:
Indicate that the error relates to the key and hint to ensure that key is validly formed.
See card with associated ticket https://github.com/pgbackrest/pgbackrest/projects/2#card-78855016


src/storage/azure/helper.c
```diff
@@ -43,16 +43,34 @@ storageAzureHelper(const unsigned int repoIdx, const bool write, StoragePathExpr
...
+        CATCH(FormatError)
+        {
+            THROW_FMT(
+                OptionInvalidValueError, "invalid value for '%s' option: %s\n"
+                "HINT: value must be valid base64 when '%s = " CFGOPTVAL_REPO_AZURE_KEY_TYPE_SHARED_Z "'.",
+                cfgOptionIdxName(cfgOptRepoAzureKey, repoIdx), errorMessage(), cfgOptionIdxName(cfgOptRepoAzureKeyType, repoIdx));
+        }
+        TRY_END();
+    }
+
```
dwsteele, line 65: I don't think this is the best place for this error message, because we don't have information about which repo is being configured. It seems best to do this in the helper. See 2787d9b1a1a.

dwsteele: Oops, this is the new code, but you get the idea.

jreidthompson: I do. Added a note to self to think in terms of scaled/non simple operations (i.e. up to 256 repos, etc ) and does this change appropriately provide for that.

Issue #1685, opened 2022-03-09
IMDSv2 support?


nunopi:
Hi,
First of all, I just wanted to say thanks for this great pg backup open-source tool!

I have a question:
I'm working on sending pgbackrest backups to S3 in an AWS environment that has IMDSv2 enabled. ([AWS official documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html))

I was wondering if pgbackrest supports this? 
Is there any specific configuration that has to be done in order to get the credentials from this endpoint?

Thanks!

dwsteele:
> First of all, I just wanted to say thanks for this great pg backup open-source tool!

Thank you!

> I'm working on sending pgbackrest backups to S3 in an AWS environment that has IMDSv2 enabled.

Do you mean *only* IMDSv2? pgBackRest does not support IMDSv2 yet so that's not going to work.

Clearly we need to do this update. I think the question is should we *require* IMDSv2 or allow some sort of soft failback. While IMDSv2 should be available on all AWS instances, I wonder if any third-party implementations have IMDSv1 but not IMDSv2 support.

One thing we could do is add a new `auto2` option that requires IMDSv2 while the current `auto` will use it when available. I'd rather not add that sort of complexity if we can avoid it, though.

Would you be willing to test this feature if we implement it?

nunopi:
Thanks for clarifying! 
Yes, I mean only IMDSv2. The environment I'm working on doesn't allow IMDSv1.
I'd be happy with assisting in testing this feature, please let me know how can I help.
Yesterday I had a quick test to see if I could make it work:
[imdsv2.txt](https://github.com/pgbackrest/pgbackrest/files/8221765/imdsv2.txt)

(Please note that I'm not a C developer and there's probably issues with this patch since I've simply copy pasted functions I found and modified them to retrieve the IMDSv2 token. Also this doesn't handle falling back to IMDSv1)

nunopi:
Yes, it does work in the limited testing I've performed. I did it in an EC2 environment that only allows IMDSv2 auth and was able to interact with the configured S3 bucket.
I've submitted a PR here: https://github.com/pgbackrest/pgbackrest/pull/1691

dwsteele:
OK, I'll close this issue since we have a PR to work on.

How would you like your name to appear in the release notes?

PR #1691, opened 2022-03-10
Add AWS IMDSv2 support.

nunopi:
@dwsteele  as discussed here: https://github.com/pgbackrest/pgbackrest/issues/1685, I'm opening this PR so we can work on adding AWS IMDSv2 support to the S3 repository code. 

As far as I understood, in order to support IMDSv2 authentication it is necessary to change the way that pgbackrest gets credentials for an associated IAM role:
On IMDSv1 it is sufficient to send a GET request to /latest/meta-data/iam/security-credential
On IMDSv2 it is necessary to get a token first from the /latest/api/token endpoint via a PUT request and then add it to the headers in the GET request for the temp credentials to the /latest/meta-data/iam/security-credential endpoint.

See official documentation here: [AWS official documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html)

I've tested this in an EC2 environment that only allows IMDSv2 auth and was able to interact with the configured S3 bucket so in my limited testing this PR does work as intended.
There are a couple of warnings during the build:
```storage/s3/storage.c: In function ‘storageS3AuthAuto’:
storage/s3/storage.c:323:9: warning: passing argument 1 of ‘httpHeaderAdd’ discards ‘const’ qualifier from pointer target type [enabled by default]
         httpHeaderAdd(header, S3_HEADER_IMDSV2_TOKEN_STR, this->token);
         ^
In file included from ./common/io/http/request.h:16:0,
                 from ./storage/s3/storage.intern.h:12,
                 from ./storage/s3/read.h:7,
                 from storage/s3/storage.c:19:
./common/io/http/header.h:27:13: note: expected ‘struct HttpHeader *’ but argument is of type ‘const struct HttpHeader * const’
 HttpHeader *httpHeaderAdd(HttpHeader *this, const String *key, const String *value);
             ^
storage/s3/storage.c: At top level:
cc1: warning: unrecognized command line option "-Wno-implicit-fallthrough" [enabled by default]
```

Please note that I'm not a C developer and there's probably issues with this patch since I've simply copy pasted functions I found and modified them to retrieve the IMDSv2 token. Also this doesn't handle falling back to IMDSv1 as discussed in https://github.com/pgbackrest/pgbackrest/issues/1685


nunopi:
@dwsteele I'd appreciate if you could set Nuno Pires as my name in the release notes. Thanks! :smiley:

dwsteele:
OK, added renewal on expiration and unit tests. Please give it a try.

nunopi:
thanks @dwsteele I can confirm it is working fine in my IMDSv2 environment

dwsteele:
OK, this is (I think) a final version that makes the following improvements:

- Allow failback to IMDSv1 when a token cannot be fetched.
- Move constants to function.
- Don't store token. It will generally expire before the credentials so this complexity is wasted.
- Set TTL based on configured timeouts.

dwsteele:
@nunopi Just to clarify, I think it would be best to test the updates from MAR-09 before we commit this.

nunopi:
@dwsteele I will test your latest changes in the upcoming days and report back here.
Thanks again!

nunopi:
hi @dwsteele I've tested the latest build today and it seems to work fine, thanks!

dwsteele (approved):

PR #1694, opened 2022-03-11
Rename bundle-* options to repo-bundle-*.

dwsteele:
It seems best for these to be repo options so they can be configured per repo, rather than globally.

All clarify usage for `repo-bundle-size` and `repo-bundle-limit`.


jreidthompson (approved):
This looks good to me

pgstef (approved):
Hi,
It sounds good to me too :+1:

PR #1697, opened 2022-03-21
Check for invalid path in `repo-*` commands.

jreidthompson:
Per the card https://github.com/pgbackrest/pgbackrest/projects/2#card-78166776

Check for invalid path in `repo-*` commands. Perform path validation and throw an error when appropriate. Path may not contain '//'. Strip trailing '/' from path. Absolute path must fall under repo path.


dwsteele (approved):
Made a few updates in 0435bd9391. Otherwise looks good.

PR #1699, opened 2022-03-21
Allow hardlink option to be changed after full backup.

dwsteele:
This rule was added because there were not sufficient tests to demonstrate that the `repo-hardlink` option could be changed in a backup set.

Remove the restriction and add/update tests to show that it works.


jreidthompson (approved):
This looks good to me

PR #1706, opened 2022-03-28
Auto select backup for given lsn target

jreidthompson:
For PITR with --type=lsn, attempt to auto select the appropriate backup set based on the --target LSN provided.  Pick the most recent backup where backup-lsn-stop is earlier than the provided LSN.

https://github.com/pgbackrest/pgbackrest/projects/2#card-74364680


pgstef (commented):
Hi,

The code looks good to me at first sight. I also ran my usual test suite without finding any specific regression.

The only side effect I'd be concerned with is when users will try to restore a backup taken with a release prior 2.38 (commit d866dd5c297dc046581134a108032cf9694cae0e I think).

Example: backup taken with 2.36, restored with this branch

```
$ pgbackrest restore --stanza=xxx --type=lsn --target="0/41000028"
ERROR: [075]: unable to find backup set with lsn less than '0/41000028'

$ pgbackrest repo-get backup/c7pg/20220329-122441F/backup.manifest --repo=2 |head
...
backup-lsn-start="0/31000028"
backup-lsn-stop="0/31000100"
```

I'd be ok if we write in the docs (or FAQ) that the user will have to pick the backup set manually in that case but perhaps we might better add a safe-check if backupLsnStop isn't in the global backup info? (and e.g. write a WARN msg to say that auto-select has been disabled and the latest backup set was picked ?)

dwsteele:
> but perhaps we might better add a safe-check if backupLsnStop isn't in the global backup info? (and e.g. write a WARN msg to say that auto-select has been disabled and the latest backup set was picked ?)

Hmmm, but picking the most recent backup set will almost certainly be wrong (as it is now). Maybe we can just add a WARN when we see NULL LSN data so the error makes more sense?

pgstef:
> Hmmm, but picking the most recent backup set will almost certainly be wrong (as it is now). Maybe we can just add a WARN when we see NULL LSN data so the error makes more sense?

Sure, it would be enough I guess. At least the user might have the info right away from the logs and not have to create an issue or look inside the docs ;-)

jreidthompson:
Thanks - I pushed changes related to the above comments

src/build/help/help.xml
```diff
@@ -2173,7 +2173,7 @@
...
+                    <p>When PITR of <br-option>{[dash]}-type=time</br-option> or <br-option>{[dash]}-type=lsn</br-option> is specified, then the target time or target lsn must be specified with the <br-option>{[dash]}-target</br-option> option. If a backup is not specified via the <br-option>{[dash]}-set</br-option> option, then the configured repositories will be checked, in order, for a backup that contains the requested time or lsn. If no matching backup is found, the latest backup from the first repository containing backups will be used for <br-option>{[dash]}-type=time</br-option> while no backup will be selected for <br-option>{[dash]}-type=lsn</br-option> if there is no match. For other types of PITR, e.g. <id>xid</id>, the <br-option>{[dash]}-set</br-option> option must be provided if the target is prior to the latest backup. See <link page="user-guide" section="/pitr">Point-in-Time Recovery</link> for more details and examples.</p>
```
pgstef, line 2176: In the same sentence: `If no matching backup is found` ... `if there is no match`. Probably the second one can be removed?

pgstef:
> Thanks - I pushed changes related to the above comments

The safety added looks good to me. Thanks :-)

pgstef (approved):

dwsteele:
From my testing it looks like the end LSN reported by postgres is a valid LSN target. This makes sense as this is the LSN where the cluster becomes consistent.

I made that change in ae92b6310f5ce58f but there were no test failures. It would be better if the test succeeded with <= and failed with < to help prevent regressions at the boundary.

jreidthompson:
per comment, updated tests to check on boundary.

dwsteele (approved):

jreidthompson:
refactoring and changes look good to me

PR #1710, opened 2022-04-06
Add default for boolean options with unresolved dependencies.

dwsteele:
If a boolean option had an unresolved dependency then the value would be NULL, which meant the dependency would need to be checked in the code to avoid an error. For example, cfgOptionBool(cfgOptOnline) needed to be checked before it was safe to call cfgOptionBool(cfgOptArchiveCheck).

Allow a default for boolean options when they are unresolved to simplify the code. This makes using the options easier and less prone to error. Not all boolean options get a dependency default in this commit, but more may be added in the future.

See also the preparatory commits 5dba0d6e and acc9f3b7.


jreidthompson (approved):
this looks good to me

PR #1712, opened 2022-04-08
Add lockRead*() functions for reading locks from another process.

dwsteele:
Sometimes we need to read a lock from another process. This was done two different ways and in the case of cmdStop() was definitely hacky.

Centralize the logic to make it easier to read the locks for another process. This will also make it easier to add new lock data.


jreidthompson (approved):
looks good to me

PR #1715, opened 2022-04-11
Throw error when unable to read lock process.

dwsteele:
Previously the process id was skipped if it did not exist. Instead, throw an error and handle the errors in downstream code.

This was probably ignored at some point to provide backward-compatibility, but that is no longer required, if it ever was.

### DIFF
--- a/src/command/control/stop.c
+++ b/src/command/control/stop.c
@@ -70,20 +70,17 @@ cmdStop(void)
                     LockReadResult lockRead = lockReadFileP(lockFile, .remove = true);
 
                     // If we cannot open the lock file for any reason then warn and continue to next file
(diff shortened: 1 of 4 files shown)


src/command/control/stop.c
```diff
@@ -70,20 +70,17 @@ cmdStop(void)
                     LockReadResult lockRead = lockReadFileP(lockFile, .remove = true);
 
                     // If we cannot open the lock file for any reason then warn and continue to next file
-                    if (lockRead.status != lockReadStatusValid && lockRead.status != lockReadStatusUnlocked)
+                    if (lockRead.status != lockReadStatusValid)
```
jreidthompson, line 73: should the comment at line 72 be changed from 'open' to 'read' also?

jreidthompson (approved):
looks good to me - one minor question

PR #1718, opened 2022-04-18
Refactor PgClient to return results in Pack format.

dwsteele:
Packs support stronger typing than JSON and are more efficient. For the small result sets that we deal with efficiency is probably not very important, but this removes another place where we are using JSON instead of Pack.

Push checking for result struct (e.g. single row) down into PgClient since it has easy access to this information rather than needing to parse the result set to find out.

Refactor all code downstream that depends on PgClient results.

Also see preparatory commits bc5f6fac, 9751ddc4, and cfd6c7ce.

### DIFF
--- a/src/db/db.c
+++ b/src/db/db.c
@@ -663,7 +663,7 @@ dbReplayWait(Db *const this, const String *const targetLsn, const uint32_t targe
             targetReached = pckReadBoolP(read);
 
             // If the target has not been reached but progress is being made then reset the timer
-            if (!targetReached && replayLsn != NULL && pckReadBoolP(read))
+            if (!targetReached && pckReadBoolP(read, .defaultValue = true))
                 wait = waitNew(timeout);
 
             protocolKeepAlive();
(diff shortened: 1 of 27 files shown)


src/db/db.c
```diff
@@ -100,35 +100,45 @@ dbNew(PgClient *client, ProtocolClient *remoteClient, const Storage *const stora
...
         FUNCTION_LOG_PARAM(DB, this);
+        FUNCTION_LOG_PARAM(STRING_ID, resultType);
         FUNCTION_LOG_PARAM(STRING, query);
     FUNCTION_LOG_END();
 
     ASSERT(this != NULL);
     ASSERT(query != NULL);
 
-    VariantList *result = NULL;
+    Pack *result = NULL;
```
jreidthompson, line 115: should we assert that resultType is not null at line 113

dwsteele: It's not a pointer, but we can at least make sure it is not zero. Done at 5822d3c3e5.

jreidthompson (approved):
Looks good to me

PR #1721, opened 2022-04-22
Improve JSON handling.

dwsteele:
Previously read/writing JSON required parsing/render via a variant, which add many more memory allocations and loops.

Instead allow JSON to be read/written serially to improve performance and simplify the code. This also allows us to get rid of many String and Variant constant which are no longer required.

The goal is to be able to read/write very large (e.g. gigabyte manifest) JSON structures, which would not be practical with the current code.

Note that external JSON (GCS, S3, etc) is still handled using variants. Converting these will require more consideration about key ordering since it cannot be guaranteed as in our own formats.

The use of ASSERT_PARAM() here is experimental. We'll need to decide if we want to keep it.


test/src/module/common/typeJsonTest.c
```diff
@@ -11,303 +11,301 @@ testRun(void)
...
-        TEST_ERROR(jsonToVarLst(STRDEF(" [] ZZZ")), JsonFormatError, "unexpected characters after array at 'ZZZ'");
+        TEST_RESULT_VOID(jsonReadArrayBegin(read), "array begin");
+        TEST_RESULT_BOOL(jsonReadBool(read), true, "bool true");
+        jsonReadConsumeWhiteSpace(read);
+        TEST_ERROR(jsonReadBool(read), JsonFormatError, "missing comma at: true]");
 
-        TEST_RESULT_STRLST_Z(strLstNewVarLst(jsonToVarLst(STRDEF("[\"e1\", \"e2\"]"))), "e1\ne2\n", "json list");
-    }
+        //--------------------------------------------------------------------------------------------------------------------------
+        TEST_TITLE("error on missing comma");
```
jreidthompson, line 39: copy paste error on TEST_TITLE

test/src/module/common/typeJsonTest.c
```diff
@@ -11,303 +11,301 @@ testRun(void)
...
+        TEST_ERROR(
+            jsonReadUInt(jsonReadNew(STRDEF("-9223372036854775807"))), JsonFormatError,
+            "-9223372036854775807 is out of range for uint");
+        TEST_ERROR(
+            jsonReadUInt(jsonReadNew(STRDEF("18446744073709551615"))), JsonFormatError,
+            "18446744073709551615 is out of range for uint");
+        TEST_ERROR(
+            jsonReadUInt64(jsonReadNew(STRDEF("-9223372036854775807"))), JsonFormatError,
+            "-9223372036854775807 is out of range for uint64");
+
```
jreidthompson, line 100: should these test on min/max boundaries -/+ 1 for each type?

jreidthompson (approved):
finished a read through. two comments noted.

Issue #1709, opened 2022-04-05
Question: Is it possible to annotate backup with additional information?


berlin-ab:
I'd like to be able to create a backup and attach metadata about where the backup originated. Is there a recommended approach for doing this?

dwsteele:
If you mean attach metadata like JSON that will be reported back by pgbackrest in the info command, then no. It sounds like an interesting idea.

One thing you could do is use the backup label (which never changes until removed) to associate metadata with the backup, but that would need to be stored elsewhere.

dwsteele:
@pgstef has created PR #1724 for this feature, so closing. @berlin-ab You should have a look and make sure this does what you need.

berlin-ab:
Thanks! Will do.

PR #1724, opened 2022-04-28
Backup key/value annotations.

pgstef:
Based on #1709, here's a proposal to integrate an annotation system.

1) Add an `annotation` option to the backup command

example:
```bash
$ pgbackrest backup --stanza=my_stanza --type=full --annotation=key1=value1 --annotation=key2=value2
$ cat /var/lib/pgbackrest/backup/my_stanza/latest/backup.manifest |grep metadata -A 1
[metadata]
annotation={"key1":"value1","key2":"value2"}
```

Internally it will store the kv into the backup.manifest.

2) Show the annotations in the info command text output for a specific backup set

example:
```bash
$ pgbackrest info --stanza=my_stanza --set=20220428-153040F
...
        full backup: 20220428-153040F
            ...
            database list: postgres (13319)
            annotation(s)
                key1: value1
                key2: value2
```

3) Add an `annotate` command to add/modify/remove annotations

example:
```bash
# remove key1 with --annotation=key1=
# update key2 with --annotation=key2="new value"
# add key3 with --annotation=key3=value3

$ pgbackrest annotate --stanza=my_stanza --set=20220428-153040F --annotation=key1= --annotation=key2="new value" --annotation=key3=value3
2022-04-28 15:35:05.156 P00   INFO: annotate command begin 2.39dev: ...
2022-04-28 15:35:05.157 P00   INFO: backup set '20220428-153040F' to annotate found in repo1
2022-04-28 15:35:05.174 P00   INFO: annotate command end: completed successfully (18ms)

$ pgbackrest info --stanza=my_stanza --set=20220428-153040F
...
        full backup: 20220428-153040F
            ...
            database list: postgres (13319)
            annotation(s)
                key2: new value
                key3: value3

$ cat /var/lib/pgbackrest/backup/my_stanza/latest/backup.manifest |grep metadata -A 1
[metadata]
annotation={"key2":"new value","key3":"value3"}
```

I'm not sure modifying the backup.manifest directly is the best way to go. Using the same approach, we could add extra metadata like "pin" labels to keep backups from the retention policy and so on.

Many thanks in advance for your input and feedback.


pgstef:
Hm, I'll look at the `src/info/manifest.c` conflict :facepalm:

dwsteele:
> Those annotations (created by the annotate command) could be lost if backup.info is lost/broken for any reason...

I think that'll be the least of their worries since the repo will be broken if backup.info can't be read. We do keep two copies of it...

It seems to me that it would be better to split this into two PRs. In this PR, just add the --annotation option to backup and display it in the info. This is such an obviously useful feature that I'd like to get it in as quickly as possible. BTW, why does JSON not get the annotations?

Then we can discuss how to update annotations. I think storing in backup.info is the way to go, but as we think ahead towards read-only repos, maybe we should make this feature friendly with that idea now. Perhaps just drop new files in the backup directory and then combine those into backup.info.

dwsteele:
Thinking some more -- the problem with writing a `backup.annotation` file is that it would require a new format and need an object to support it. I'm actively working on a new info format right now, so I'd rather not get ahead of that.

So, at least for now, writing to backup.info probably makes more sense.

pgstef:
> why does JSON not get the annotations?

Because currently it was only stored in backup.manifest, so only available when --set is provided and json output doesn't support --set currently :innocent:. And also (mainly) because it will be user-defined text which can be very long. So I tend to only display it with `info --set=..`.

> Thinking some more -- the problem with writing a `backup.annotation` file is that it would require a new format and need an object to support it. I'm actively working on a new info format right now, so I'd rather not get ahead of that.
> 
> So, at least for now, writing to backup.info probably makes more sense.

I like the idea of storing the initial annotation inside the manifest (because it was created at that time anyway) and then use the "living" annotations inside backup.info. We could then easily add other metadata into the backup.info too (like pinning a backup to prevent it to be expired,...). I was just afraid to lose those if backup.info was broken but that doesn't seem to be that much concerning after-all and it indeed seems much easier than adding extra files for now.

Regarding splitting the PR, I can remove the annotate command part of this PR if you find it easier to read. But I think the update part is pretty "easy" once I've put the living annotations to backup.info :slightly_smiling_face:

dwsteele:
> So I tend to only display it with info --set=..

Hmmm. It seems to me that annotations are meant to be used in an automated fashion and text does not work for that. Maybe the json output needs an additional option to add annotations.

> Regarding splitting the PR

Just keep working on it all together for now since that will be easier. If we decide to split it later we can do that.

pgstef:
> Hmmm. It seems to me that annotations are meant to be used in an automated fashion and text does not work for that. Maybe the json output needs an additional option to add annotations.

Adding that to my todo list then :-)

dwsteele:
Or just add the annotations to the json and we'll worry about it being too big when somebody screams. It's the user who controls how big these annotations will be.

pgstef:
Ok, now that I added the tests, this PR looks ready for review to me :-)
Many thanks in advance for your time and feedback ;-)

jimmyyih:
This is an amazing feature that I'm very much looking forward to using!

As a request/suggestion (if this PR is the right place to do it), would it make sense to add default "latest" (or make it a usable value) to the `annotate` command's `--set` flag as a way to more quickly annotate the latest backup?  For my personal use case, I plan to store my restore point information into the backup set's metadata as such:
```
incr backup: 20220510-121624F_20220510-121655I
    timestamp start/stop: 2022-05-10 12:16:55 / 2022-05-10 12:16:56
    wal start/stop: 000000010000000000000005 / 000000010000000000000005
    lsn start/stop: 0/14000028 / 0/14000110
    database size: 59.6MB, database backup size: 31.5KB
    repo1: backup set size: 3.5MB, backup size: 2.9KB
    backup reference list: 20220510-121624F
    database list: postgres (13383)
    annotation(s)
        restore_point_20220510122611: 0/180001C0
```

Since I'm only associating new restore points to only the latest backup (to reduce `pgbackrest restore --type name --target restore_point_20220510122611` restore time), it would be convenient to have `--set` value "latest" be default or a usable value.  I also imagine that the `annotate` command would be used slightly more often on the latest backup than the preceding ones... mainly from the perspective of automation.

berlin-ab:
The use-case we have is an automated workflow, as @dwsteele suggested. It would be nice to have JSON also output the annotations.

pgstef:
> The use-case we have is an automated workflow, as @dwsteele suggested. It would be nice to have JSON also output the annotations.

Well, if you look at my previous [comment](https://github.com/pgbackrest/pgbackrest/pull/1724#issuecomment-1113418250) and look at the examples, it is working with `info --output=json`:

```
$ pgbackrest info --stanza=my_stanza --output=json
...,"backup":[{"annotation":{"key1":"value1","key2":"value2"},...
```

It would then be nice if you could try the dev version and see if this does what you need.

Kind Regards

src/command/annotate/annotate.c
```diff
@@ -0,0 +1,91 @@
...
+                    // Update annotations
+                    infoBackupDataAnnotationSet(infoBackup, backupLabel, cfgOptionKv(cfgOptAnnotation));
+
+                    // Write the updated backup info
+                    infoBackupSaveFile(
+                        infoBackup, storageRepoWrite(), INFO_BACKUP_PATH_FILE_STR, repoCipherType,
+                        cfgOptionIdxStrNull(cfgOptRepoCipherPass, repoIdx));
+                }
+            }
+            CATCH(FileMissingError)
```
dwsteele, line 78: We should be handling errors like the similar loop in expire.c line 1003.

src/command/backup/backup.c
```diff
@@ -2253,6 +2253,10 @@ cmdBackup(void)
                     (BackupType)cfgOptionStrId(cfgOptType), manifestData(manifest)->backupLabelPrior, timestampStart));
         }
 
+        // Set annotations in the manifest
+        if (cfgOptionTest(cfgOptAnnotation))
+            manifestAnnotationSet(manifest, cfgOptionKv(cfgOptAnnotation));
+
```
dwsteele, line 2259: There does not seem to be any reason for this to be a separate function. I think annotations should be handled by `manifestBuildComplete()` (like `dbList`).

src/info/infoBackup.c
```diff
@@ -459,6 +468,52 @@ infoBackupDataAdd(const InfoBackup *this, const Manifest *manifest)
...
+        InfoBackupData *infoBackupData = lstFind(this->pub.backup, &backupLabel);
+
+        KeyValue *tmpAnnotationKv = kvNew();
+
+        if (infoBackupData->backupAnnotation != NULL)
+            tmpAnnotationKv = varKv(infoBackupData->backupAnnotation);
+
+        const VariantList *const annotationKeyList = kvKeyList(annotationKv);
+
+        for (unsigned int keyIdx = 0; keyIdx < varLstSize(annotationKeyList); keyIdx++)
```
dwsteele, line 495: I'm not sure why we need to create a copy here? Why not just modify the existing KV? (Yes, you need to remove the const in InfoBackupData.)

src/common/type/keyValue.c
```diff
@@ -331,3 +331,39 @@ kvGetList(const KeyValue *const this, const Variant *const key)
...
+{
+    FUNCTION_TEST_BEGIN();
+        FUNCTION_TEST_PARAM(KEY_VALUE, this);
+        FUNCTION_TEST_PARAM(VARIANT, key);
+    FUNCTION_TEST_END();
+
+    ASSERT(this != NULL);
+    ASSERT(key != NULL);
+
+    MEM_CONTEXT_BEGIN(objMemContext(this))
```
dwsteele, line 347: Not sure why this mem context switch is needed, since we are not allocating anything?

src/common/type/keyValue.c
```diff
@@ -331,3 +331,39 @@ kvGetList(const KeyValue *const this, const Variant *const key)
...
+        if (listIdx != KEY_NOT_FOUND)
+        {
+            // Remove from the list
+            lstRemoveIdx(this->list, listIdx);
+
+            // Remove from the key list
+            for (unsigned int idx = 0; idx < varLstSize(this->pub.keyList); idx++)
+            {
+                if (varEq(key, varLstGet(this->pub.keyList, idx)))
+                    lstRemoveIdx((List *)this->pub.keyList, idx);
```
dwsteele, line 362: We should free the key/value being removed here or we'll leak memory.

dwsteele (changes requested):
Looks pretty good overall. See updates in my commits and review comments.

src/command/annotate/annotate.c
pgstef: This should be fixed in ec62232

src/command/backup/backup.c
pgstef: This should be fixed in cc80268

src/info/infoBackup.c
pgstef: Hm, everything is stored as Variant so I needed to transform into KV to modify it. Perhaps there's indeed an easier way to achieve it, but I didn't found it yet ^^'

src/common/type/keyValue.c
pgstef: Indeed, fixed in 20d3108

src/common/type/keyValue.c
pgstef: Hm, I'm not sure to exactly know how to do that (I thought lstRemoveIdx would handle it) but I tried something in 20d3108

src/command/annotate/annotate.c
dwsteele: That's what I was looking for, thanks!

src/command/backup/backup.c
dwsteele: Looks much simpler now.

src/common/type/keyValue.c
dwsteele: Looks better, but I did some more cleanup in d65796423f1f9d.

pgstef: oh I see the missing varFree now. Thanks for the update, it looks better indeed!

src/info/infoBackup.c
dwsteele: Worked in this in f88abc26925ad670. I misread this a bit -- there was not a copy being made (though there needed to be earlier) but it was leaking a KV. Made general improvements to make the function clearer.

test/src/module/command/infoTest.c
```diff
@@ -652,7 +652,9 @@ testRun(void)
             "\"db-id\":1,\"option-archive-check\":true,\"option-archive-copy\":false,\"option-backup-standby\":false,"
             "\"option-checksum-page\":true,\"option-compress\":true,\"option-hardlink\":false,\"option-online\":true}\n"
             "20181119-152138F_20181119-152155I={"
-            "\"backrest-format\":5,\"backrest-version\":\"2.08dev\",\"backup-archive-start\":\"000000010000000000000003\","
+            "\"backrest-format\":5,\"backrest-version\":\"2.08dev\","
+            "\"backup-annotation\":{\"key1\":null},"
```
dwsteele, line 656: How is it possible to have a NULL key? Don't we remove any keys the user specifies as empty?

pgstef: In theory we shouldn't have null values indeed, but the test is there for the coverage of `src/command/info/info.c` L976:
```
if (value != NULL)
    strCatFmt(annotationStr, "                %s: %s\n", strZ(key), strZ(value));
```

src/info/manifest.c
```diff
@@ -1542,6 +1543,25 @@ manifestBuildComplete(
...
+            const VariantList *const annotationKeyList = kvKeyList(annotation);
+
+            for (unsigned int keyIdx = 0; keyIdx < varLstSize(annotationKeyList); keyIdx++)
+            {
+                const Variant *const key = varLstGet(annotationKeyList, keyIdx);
+                const Variant *const value = kvGet(annotation, key);
+
+                // Skip empty values
+                if (!strEmpty(varStr(value)))
+                    kvPut(manifestAnnotationKv, key, value);
```
dwsteele, line 1561: Not entirely satisfied with this. If the user specifies a lot of empty values then we just store an empty KV? Shouldn't we error if the initial values are empty?

test/src/module/command/infoTest.c
dwsteele: If there are NULLs in the annotation data then we have a bug somewhere, right? Seems like this should be an error?

pgstef: If NULL always mean empty in our case, indeed that would mean we missed something somewhere. Would it worth changing `if` to `ASSERT` and fail the entire command rather than simply ignoring it?

pgstef: I played again with the current implementation and couldn't generate null "value". But just in case I switched the if to assert as discussed here in 0964d74ac1d885e663a9e85189c08bea42543d2e.

src/info/manifest.c
pgstef: I added some clean-up steps in 4db7d9a. 
With that clean-up, I don't think it is necessary to add the warning in the backup output (in case all provided annotations values are empty) given all empty values are skipped and the manifest/backup.info field isn't created anyway.

pgstef:
Hi,

I finally found the time to come back to this feature.
With the 2 last unsolved discussions, here's the current status :

--------------------------------------------------------------------------------
1) Add an `annotation` option to the backup command
The initial annotation is stored in backup.manifest and duplicated as "living" value into backup.info.
If the user only provide empty values, the annotation filed isn't created.

example:
```bash
$ pgbackrest backup --stanza=my_stanza --type=full --annotation=key1= --annotation=key2=
...
$ cat /var/lib/pgbackrest/backup/my_stanza/latest/backup.manifest |grep metadata -A 1
$ cat /var/lib/pgbackrest/backup/my_stanza/backup.info |grep annotation

$ pgbackrest backup --stanza=my_stanza --type=full --annotation=key1=value1 --annotation=key2=value2
...
$ cat /var/lib/pgbackrest/backup/my_stanza/latest/backup.manifest |grep metadata -A 1
[metadata]
annotation={"key1":"value1","key2":"value2"}
$ cat /var/lib/pgbackrest/backup/my_stanza/backup.info |grep annotation
..."backup-annotation":{"key1":"value1","key2":"value2"}...
```

2) Show the annotations in the info command text output for a specific backup set or globally with `--output=json`.

example:
```bash
$ pgbackrest info --stanza=my_stanza --set=20220801-164223F
...
full backup: 20220801-164223F
    timestamp start/stop: 2022-08-01 16:42:23 / 2022-08-01 16:42:26
    wal start/stop: 00000001000000000000001B / 00000001000000000000001B
    lsn start/stop: 0/1B000028 / 0/1B000138
    database size: 23.1MB, database backup size: 23.1MB
    repo1: backup set size: 2.8MB, backup size: 2.8MB
    database list: postgres (13319)
    annotation(s)
        key1: value1
        key2: value2

$ pgbackrest info --stanza=my_stanza --output=json
..."backup":[{"annotation":{"key1":"value1","key2":"value2"}...
```

3) Add an `annotate` command to add/modify/remove annotations.
Those living annotations will be updated/stored in backup.info and the initial ones created in backup.manifest won't change.
The annotation field is removed when all the annotations are removed too.

example:
```bash
$ pgbackrest annotate --stanza=my_stanza --set=20220801-164223F --annotation=key1= --annotation=key2="new value" --annotation=key3=value3
$ pgbackrest info --stanza=my_stanza --set=20220801-164223F
...
full backup: 20220801-164223F
    timestamp start/stop: 2022-08-01 16:42:23 / 2022-08-01 16:42:26
    wal start/stop: 00000001000000000000001B / 00000001000000000000001B
    lsn start/stop: 0/1B000028 / 0/1B000138
    database size: 23.1MB, database backup size: 23.1MB
    repo1: backup set size: 2.8MB, backup size: 2.8MB
    database list: postgres (13319)
    annotation(s)
        key2: new value
        key3: value3

$ cat /var/lib/pgbackrest/backup/my_stanza/latest/backup.manifest |grep metadata -A 1
[metadata]
annotation={"key1":"value1","key2":"value2"}
$ cat /var/lib/pgbackrest/backup/my_stanza/backup.info |grep annotation
..."backup-annotation":{"key2":"new value","key3":"value3"}...

$ pgbackrest annotate --stanza=my_stanza --set=20220801-164223F --annotation=key2= --annotation=key3=
$ cat /var/lib/pgbackrest/backup/my_stanza/backup.info |grep annotation
```
--------------------------------------------------------------------------------

With that in mind, I'm wondering if I couldn't add a `--set=latest` alias in the annotate command.
Of course we could speak about multiple repositories. But given that command is operating on one repo only (the one specified with --repo or the first one), that alias would refer to the latest backup set inside the specified (or default) repo. That would add an extra layer in the automation direction.

What's your thoughts about this?

Many thanks in advance,
Kind Regards

dwsteele:
OK, these changes look pretty good to me, but I have not reviewed the code yet.

> With that in mind, I'm wondering if I couldn't add a --set=latest alias in the annotate command.

Let's leave that for a future commit. I'm worried that it might lead to unpredictable results and I don't want to rush it in.

Is this ready for final review now?

pgstef:
> Let's leave that for a future commit. I'm worried that it might lead to unpredictable results and I don't want to rush it in.

Ok, sounds good 👍

> Is this ready for final review now?

I think so indeed 😊

dwsteele (changes requested):
OK, the code looks good to me now. I played around with the feature manually and now all the behavior is exactly as I expect WRT adding and removing keys, passing empty keys, etc.

Last but not least, I think we need something in the user guide. A small section under https://pgbackrest.org/user-guide.html#backup should do the trick. Not sure this is worth putting in Quick Start since it is more advanced usage.

pgstef:
Thanks for the review and testing!

> Last but not least, I think we need something in the user guide. A small section under https://pgbackrest.org/user-guide.html#backup should do the trick. Not sure this is worth putting in Quick Start since it is more advanced usage.

That's a good idea indeed. Added an example in 4af0842 (and fixed the command length in b6f5eb2)

dwsteele (approved):
Made some edits in 4d206fa70 (and reverted some of them in e3104f395cf).

The important change was "output by the <cmd>info</cmd> command text output when a backup" => "output by the <cmd>info</cmd> command when a backup". Since we now support annotation output in JSON, the second version seems more correct.

This now looks ready to commit unless you have objections.

dwsteele:
Oops, I see what you mean. Not ideal, but we can just repeat the text as in e3520035b6d.

pgstef:
e352003 LGTM :+1:

PR #1730, opened 2022-05-02
Error on all lock failures except another process holding the lock.

dwsteele:
The archive-get/archive-push commands would not error for, .e.g permissions errors, when attempting to get a lock before launching the async process. Since the async process was not launched there would be no error status file and the user would get a generic failure message. Also, there would be no async log.

Refactor lockAcquireFile() to throw an error when failOnNoLock = false unless the file is locked by another process. This seems to be the original intent of this parameter and there may have been a mistake when porting from Perl. In any case it looks wrong enough to be considered a bug.

Also see 03c71aa6 and 0055fa40 for other improvements made in this area.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -116,6 +116,28 @@
 
                         <p>Improve error message for invalid <br-option>repo-azure-key</br-option>.</p>
                     </release-item>
+
+                    <release-item>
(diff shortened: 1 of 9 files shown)


jreidthompson (approved):
This looks good to me

PR #1733, opened 2022-05-03
Document required base branch for pull requests.

pgguru:
Be explicit when submitting a PR about which branch to use as the base.

Side note: If this is always the branch to be targeted, is it worth considering changing the default branch in the repository?


dwsteele:
> Side note: If this is always the branch to be targeted, is it worth considering changing the default branch in the repository?

I wish, but that would be the branch the user sees, works from, etc. We frequently force push integration so that won't work. What we need is to set a default branch for PRs, but that is not supported.

So, basically, we *work* from main, but we *merge* to integration.

doc/xml/contributing.xml
```diff
@@ -787,6 +787,7 @@ pgbackrest/doc/doc.pl --out=html --include=user-guide --require=/quickstart --va
         <list>
             <list-item>Provide a short submission title.</list-item>
             <list-item>Write a detailed comment to describe the purpose of your submission and any issue(s), if any, it is resolving; a link to the GitHub issue is also helpful.</list-item>
+            <list-item>Ensure your PR is targeted against the <code>integration</code> branch.</list-item>
```
jreidthompson, line 790: Looking at the github instructions for creating a PR, the merge into/left side is called the base,  so should we try to match the terminology? Something along the lines of
Select the `integration` branch as the base in your PR, do not select `main` nor any other branch.

pgguru: WFM.

dwsteele:
This looks ready to commit. I updated the wording slightly.

PR #1735, opened 2022-05-03
Remove dependency on pg_database.datlastsysoid.

dwsteele:
This column has been removed in PostgreSQL 15. Rather than add a lot of special handling, it seems better just to update all versions to not depend on this column.

Add centralized functions to identify the type of database (i.e. system or user) by name and use FirstNormalObjectId when a name is not available.

The new query in the db module will still return the prior result for PostgreSQL <= 15, which will be stored in the manifest. This is important to preserve behavior when downgrading pgBackRest. There are no concerns here for PostgreSQL 15 since older versions of pgBackRest won't be able to restore backups for PostgreSQL 15 anyway.


pgstef (approved):
The code change looks good. Especially the new functions which makes it easier to read imo.
So +1 for me :-)

PR #1736, opened 2022-05-04
PostgreSQL 15 support.

dwsteele:
PostgreSQL 15 drops support for exclusive backup and renames the start/stop backup commands.

This is based on the pgdg-testing repo since beta1 has not been released yet, but it seems unlikely that breaking changes will be made at this point. beta1 should be tagged just before our next release so we'll retest before the release just to be sure.


pgstef (approved):
Code change (mostly in src/db/db.c) and the added test look good to me.
My usual tests on Rocky8 with PG 14 and PG 15 went fine.

So this PR looks good to me :-)

PR #1738, opened 2022-05-05
Strip extensions from history manifest before showing in error message.

dwsteele:
In cases where clock skew or timezone issues are preventing backup label generation the user could see an error like this:

new backup label '20220504-152308F' is not later than latest backup label '20220504-222042F_20220504-222141I.manifest.gz'

This will happen if the most recent label is drawn from the history. It is cleaner (and probably less confusing) to strip off the extensions so the user sees:

new backup label '20220504-152308F' is not later than latest backup label '20220504-222042F_20220504-222141I'

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -185,6 +185,17 @@
                     </release-item>
 
                     <release-item>
+                        <github-pull-request id="1738"/>
+
+                        <release-item-contributor-list>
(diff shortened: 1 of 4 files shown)


pgstef (approved):
Looks good to me

dwsteele:
Excellent, thanks!

PR #1739, opened 2022-05-05
Add ClockError for unexpected clock skew and timezone changes.

dwsteele:
Having a distinct result code should help with debugging clock skew and timezone issues.


pgstef (approved):
Looks good to me

turnstep (approved):

PR #1742, opened 2022-05-06
Simplify messaging around supported versions in the documentation.

dwsteele:
The version ranges given in the user guides caused confusion. For example, because the user guide for RHEL specified PostgreSQL 9.6-11, users questioned whether pgBackRest worked for PostgreSQL 12 on RHEL.

Remove these ranges and add more explanatory text to the introduction to try and make it clearer how the user guides work and which versions are covered (basically all of them).


pgstef (approved):
Makes sense and looks good to me.

jreidthompson (approved):
looks good to me

PR #1745, opened 2022-05-09
Remove integration expect log testing.

dwsteele:
Integration expect log testing was originally used as a rough-and-ready way to make sure that certain code paths were being executed before the unit tests existed. Now that we have 100% unit test coverage (with expect log testing) the value of the integration expect tests seems minimal at best.

But they do cause numerous issues:

- Maintenance of the expect code and replacements that are required to keep logs reproducible.
- Even a trivial change can cause massive churn in the expect logs, e.g. d9088b2e. These changes should be minutely audited but since the expect logs have little value now it is seldom worth the effort.
- The OS version used to do expect testing (RHEL7) can only be used to test one version of PostgreSQL. This makes it hard to balance the PostgreSQL version testing between OS versions.
- When a commit affects expect logs it is not clear (especially for new developers) how to regenerate them and our contributing guide is silent on the issue.

The goal is to migrate the integration tests to C and expect testing is not part of that plan. It seems best to get rid of them now.


test/lib/pgBackRestTest/Common/RunTest.pm
```diff
@@ -300,7 +247,6 @@ sub testResult
             {name => 'strExpected', required => false, trace => true},
             {name => 'strDescription', trace => true},
             {name => 'iWaitSeconds', optional => true, default => 0, trace => true},
-            {name => 'strLogExpect', optional => true, trace => true},
             {name => 'strLogLevel', optional => true, default => WARN, trace => true},
```
jreidthompson, line 250: mismatch strLogLevel exists on rhs of assignment but not left???

dwsteele: Fixed in 3e8942c47013d9. Interesting that Perl was cool with this.

dwsteele: Ah, I see. Perl is fine with the lvalue array being smaller than the rvalue array. Might be fine the other way, too.

test/lib/pgBackRestTest/Common/LogTest.pm
```diff
@@ -1,492 +0,0 @@
-####################################################################################################################################
```
jreidthompson, line 1: logTest.pm has been deleted but there is a reference at line 1862 for test/lib/pgBackRestTest/Common/LogTest.pm in code-count/file-type.yaml

dwsteele: This file is auto-generated with every release, but no harm in removing that file now. Done in 299158379.

test/lib/pgBackRestTest/Module/Real/RealAllTest.pm
```diff
@@ -350,8 +348,8 @@ sub run
         #---------------------------------------------------------------------------------------------------------------------------
         # Restart the cluster to check for any errors before continuing since the stop tests will definitely create errors and the
         # logs will to be deleted to avoid causing issues further down the line. This test is not version specific so is run on only
-        # the expect version.
-        if ($bExpectVersion)
+        # one version.
```
jreidthompson, line 351: happened to notice that either the word 'need' is missing or the word 'to' s/b deleted from line 350?

test/lib/pgBackRestTest/Common/RunTest.pm
jreidthompson: Interesting. I wasn't aware of that. Previous jobs had some perl portions, but not a lot of it overlapped with what I mainly worked on.

test/lib/pgBackRestTest/Module/Real/RealAllTest.pm
dwsteele: Fix in main at 6b98b3534e15a8d277547eb2932c085bc7515eec.

jreidthompson (approved):
looks good to me

PR #1751, opened 2022-05-13
Fix issues in improved path validation for repo-* commands.

dwsteele:
If the user requested the exact repo path then strSub() would be passed an invalid start value leading to an assertion:

$ pgbackrest --stanza=test repo-ls /var/lib/pgbackrest
ASSERT: [025]: start <= this->pub.size (on dev builds)
ASSERT: [025]: string size must be <= 1073741824 bytes (on prod builds)

Fix this by checking if the requested path exactly equals the repo path and returning an empty relative path in this case.

Another issue was that invalid subpaths were not detected if they started with the repo path. For example, /var/lib/pgbackrestsub would not generate an error if the repo path was /var/lib/pgbackrest. Fix this by explictly checking for a / between the repo path and the subpath. This also requires special handling when the repo path is /.

This is not classified as a bug since the issues were found in an unreleased feature introduced in 5ae84d5e.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -188,7 +188,12 @@
                     </release-item>
 
                     <release-item>
-                        <github-pull-request id="1697"/>
+                        <commit subject="Improve path validation for repo-* commands.">
+                            <github-pull-request id="1697"/>
+                        </commit>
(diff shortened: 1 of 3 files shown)


jreidthompson (approved):
Looks good.  Apologies for missing that.

PR #1759, opened 2022-05-25
Init pctCplt to zero at the start of backup

jreidthompson:
David noted that waiting to write percent complete status until the first file completes backup likely would result in a period of time where the backup is running without a status available to the user. Remedy this by initing the percent complete to 0 at the start of the backup.


dwsteele (approved):
Seems better to do this when copying starts or we could hang at zero for a very long time while a manifest is built. If we can include that in the percentage at some point it would be a good improvement.

jreidthompson:
Thanks-  looks good to me

PR #1760, opened 2022-05-25
Disable incremental manifest save when resume=n.

jreidthompson:
The manifest is saved on a regular basis during a backup allowing for a failed backup to be resumed.  For backups that the user has configured/invoked as not resumable, skip the incremental save of the manifest.


dwsteele:
The patch missed two places where the manifest was being save incrementally. I centralized the logic to make it a bit easier to test and maintain.

I'm not too satisfied with the weak test here, but not sure if is worth shimming `backupManifestSaveCopy()` to make it better.

dwsteele (approved):
For now I think we are OK without a better test. This should reduce I/O and for now that's all that is important.

When we start looking at WORM repos we'll shim the storage interface to make sure *nothing* is ever overwritten.

PR #1765, opened 2022-05-31
Add experimental Meson build.

dwsteele:
Meson is a new build system that offers simpler syntax and superior performance to autoconf/make. In addition, Windows is supported natively.

The meson build appears complete, but currently is used only for auto-generation of code and the host build of pgbackrest. Some container upgrades will be required before meson can be used for container builds.

Patch the Debian package to force autoconf/make rather than meson.


meson.build
```diff
@@ -0,0 +1,174 @@
...
+lib_openssl = dependency('openssl')
+
+# Find required pq library
+lib_pq = dependency('libpq')
+
+# Find required xml library
+lib_xml = dependency('libxml-2.0')
+
+# Find required yaml library (only used for build)
+lib_yaml = meson.get_compiler('c').find_library('yaml')
```
eli-schwartz, line 146: There is canonically a yaml-0.1.pc, so you should use `dependency('yaml-0.1')` here.

You can also use `meson wrap install libyaml` to install the Meson wrap for this as a fallback.

meson.build
```diff
@@ -0,0 +1,174 @@
...
+)
+
+####################################################################################################################################
+# OS-specific settings
+####################################################################################################################################
+if build_machine.system() == 'linux'
+    add_global_arguments('-D_POSIX_C_SOURCE=200809L', language : 'c')
+elif build_machine.system() == 'darwin'
+    add_global_arguments('-D_DARWIN_C_SOURCE', language : 'c')
+endif
```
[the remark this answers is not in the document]
dwsteele: Good point! Fixed in dcdbc96ecac.

meson.build
dwsteele: Thanks, fixed in 25f9a03e0c90. I've had surprisingly little luck finding examples for specific libs.

meson.build
```diff
@@ -0,0 +1,174 @@
...
+
+foreach warning : warning_enable
+    if meson.get_compiler('c').has_argument('-W' + warning)
+        add_project_arguments('-W' + warning, language: 'c')
+    endif
+endforeach
+
+####################################################################################################################################
+# Disable various unhelpful warnings by adding -Wno-warning switches. GCC won't complain about unrecognized -Wno-warning switches,
+# so we have to test for the positive form, and if that works, add the negative form.
```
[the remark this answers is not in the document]
dwsteele: At one point I was trying to target 0.45 and this was introduced in 0.51. For now I've added a note.

I'm waiting back from the Debian packagers to find out if it would be OK to specify a version of meson for the build. Their initial answer was no, but not sure they are aware it is a Python module.

If we must use the meson version that is available in the distro then we either need to figure out how to work with 0.45 or wait a few years to remove autoconf/make.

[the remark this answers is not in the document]
dwsteele: Ubuntu 18.04, https://packages.ubuntu.com/bionic/meson.

We've used backports to get a working version of the YAML Perl module before, so that might be an option.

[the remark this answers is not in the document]
dwsteele: Well, regardless of how I feel about Ubuntu, the apt packagers build for it so it needs to work.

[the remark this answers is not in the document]
dwsteele: So, yeah, I mean Debian-based. :)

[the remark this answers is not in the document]
dwsteele: I decided to leave the code as-is, but added a comment explaining why we are not using the new feature in 49ec515ed8f3b. Look OK?

meson.build
```diff
@@ -0,0 +1,176 @@
...
+if host_machine.system() == 'linux'
+    add_global_arguments('-D_POSIX_C_SOURCE=200809L', language : 'c')
+elif host_machine.system() == 'darwin'
+    add_global_arguments('-D_DARWIN_C_SOURCE', language : 'c')
+endif
+
+####################################################################################################################################
+# Enable as many additional warnings as possible to catch potential errors
+####################################################################################################################################
+warning_enable = [
```
shkshk90, line 35: Hello, 
I think that this array, (or maybe just the `68: foreach warning : warning_enable` below, needs to be checked against the OS, as this opens way to add similar warnings to other compilers, mainly msvc. 

Aside from that, I think it is great.

shkshk90 (approved):

meson.build
dwsteele: I'd prefer to hold off on that for now. We'll add that in your branch once we know what the Windows builds are going to look like. If we add it now it won't have a purpose and will probably end up looking different.

dwsteele: For example, I don't think we want to do an OS check here, but a compiler check. But we'll figure that out as we work on the Windows port.

eli-schwartz: You already use compiler.has_argument() here, which uses a configure probe to check whether the compiler supports it. So you can add both GCC and MSVC style arguments here and it will filter out whichever ones don't work.

It's possible to check the OS / compiler manually via host_machine.system() or compiler.get_id(), but all that will do is make the configure stage a bit faster as it skips compiling a test file to see if the command line argument triggered a fatal "invalid argument" error, on compilers that you statically code as "doesn't support this particular argument".

dwsteele: Excellent suggestion -- and exactly why I want to see how the Windows build works out before make changes to support it.

meson.build
```diff
@@ -0,0 +1,176 @@
...
+    'vla',
+
+    # Give string constants the type const char[length] so that copying the address of one into a non-const char * pointer produces
+    # a warning
+    'write-strings',
+]
+
+foreach warning : warning_enable
+    if meson.get_compiler('c').has_argument('-W' + warning)
+        add_project_arguments('-W' + warning, language: 'c')
```
eli-schwartz, line 70: Meson also has a `get_supported_arguments(list_of_warning_flags)` that allows you to do this:

```
add_project_arguments(cc.get_supported_arguments(warning_enable), language: 'c')
```
However, using it assumes that warning_enable is a list of flags which already have their leading `-W`. It's safe to do that for the enable flags, since you don't need to modify it to the `-Wno-*` form.

It's slightly inconsistent with the warning_disable handling below, but that may be okay since ultimately you want to use Meson 0.51.0 support for handling that automatically too :) and many people excuse inconsistent style if it's only a temporary backwards-compat measure.

This would also mean that you can later on extend warning_enable with MSVC style flags and it doesn't get the GCC-style leading `-W` treatment, so it simplifies adding multi-platform support.

meson.build
```diff
@@ -0,0 +1,176 @@
...
+        # Unroll loops whose number of iterations can be determined at compile time or upon entry to the loop
+        'unroll-loops',
+
+        # Perform loop vectorization on trees
+        'tree-vectorize',
+    ]
+
+    foreach optimization : optimization_enable
+        if meson.get_compiler('c').has_argument('-f' + optimization)
+            add_project_arguments('-f' + optimization, language: 'c')
```
eli-schwartz, line 111: Similarly here, you can list it with -f in the array and simplify the add_project_arguments call.

Actually you can also do something like:

```meson
add_project_arguments(cc.get_supported_arguments(warning_enable, optimization_enable), language: 'c')
```

eli-schwartz (approved):
One last optional style suggestion, but otherwise the Meson usage here LGTM.

meson.build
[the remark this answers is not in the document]
eli-schwartz: :+1:

Sure. The most important thing is to use whatever you need in order to handle the supported userbase. I generally recommend people either update to newer Meson usage, or add a TODO about the new feature, **whichever the project feels is right**. And what is right for you is going the TODO route, which you have done.

meson.build
eli-schwartz: What some people do in this situation is check `if get_option('foo') in ['2', '3']` then e.g. `all_checked_options += [...]` and then test+add them all at the end.

dwsteele: Hmmm, I can see that, but still prefer breaking things up a bit, as in ca61ddc93.

dwsteele: Fair enough. I think keeping the lists separate has some documentary value. Once these build files are written they won't get touched much so clearer is better. At least clearer to me. :)

PR #1767, opened 2022-06-01
Support OpenSSL 3.

dwsteele:
There are two changes:

1) Suppress deprecation warnings so we can build with -Werror and -Wfatal-errors. At some point we'll need to migrate to the new APIs but there does not seem to be a good reason to support two sets of code right now.

2) Update the handling for unexpected EOF to handle EOF or error. The error code for EOF has changed and become harder to identify, but we probably don't care whether it is an error or EOF.

NOTE: Fedora 36 is running many times slower than Fedora 33. Perhaps Fedora 36 is too bleeding edge for us and we should use u22 for those unit tests.


sfrost (approved):
approved with suggested changes

dwsteele:
I cut this PR down to just the OpenSSL 3 changes. The new VMs will be in the next commit.

PR #1773, opened 2022-06-08
Add FN_NO_RETURN macro.

shkshk90:
Change all instances of `__attribute__((__noreturn__))` to a macro in [`meson.build`](meson.build) / [`build.auto.h.in`](src/build.auto.h.in).

As compiler attributes written in the form of `__attribute__` is not supported by _msvc_, this is one of several pull requests to make the code-base more robust, and allows using _msvc_-style attributes later.


meson.build
```diff
@@ -155,6 +155,10 @@ endif
 # Set configuration path
 configuration.set_quoted('CFGOPTDEF_CONFIG_PATH', get_option('configdir'), description: 'Configuration path')
 
+# Set FN_NO_RETURN macro
+configuration.set('FN_NO_RETURN', '__attribute__((__noreturn__))', description: 'No return function attribute')
+
+
```
dwsteele, line 161: Remove the extra linefeed here. We never have more than one blank line for spacing.

src/common/error.h
```diff
@@ -314,28 +314,28 @@ jmp_buf *errorInternalJump(void);
...
     const ErrorType *errorType, const char *fileName, const char *functionName, int fileLine, const char *format, ...)
-    __attribute__((format(printf, 5, 6))) __attribute__((__noreturn__));
+    __attribute__((format(printf, 5, 6)));
 
 // Throw a system error
-void errorInternalThrowSys(
+FN_NO_RETURN void errorInternalThrowSys(
     int errNo, const ErrorType *errorType, const char *fileName, const char *functionName, int fileLine, const char *message)
-    __attribute__((__noreturn__));
+   ;
```
dwsteele, line 333: This semicolon should be at the end of the line above.

dwsteele (changes requested):
Looks pretty good overall but see comments.

dwsteele (approved):
Looks good, thanks!

In the future keep in mind that it is up to the reviewer to resolve comments, i.e. when the reviewer feels that the issue has been addressed. We just add a comment with the commit where the issue was fixed, then the reviewer checks and resolves.

meson.build
```diff
@@ -155,6 +155,9 @@ endif
 # Set configuration path
 configuration.set_quoted('CFGOPTDEF_CONFIG_PATH', get_option('configdir'), description: 'Configuration path')
 
+# Set FN_NO_RETURN macro
+configuration.set('FN_NO_RETURN', '__attribute__((__noreturn__))', description: 'Indicate that a function does not return')
```
eli-schwartz, line 159: Note for the future, you can do configure time checks to see if this is supported via:

https://mesonbuild.com/Reference-manual_returned_compiler.html#compilerhas_function_attribute

shkshk90: Thanks for this tip.

I went a bit through meson's reference, but didn't catch it.
I will definitely make use of that, when I start implementing the Windows port.

Thank you :).

PR #1781, opened 2022-06-14
Fix hard-coded WAL size assertion in walSegmentNext().

InnerLife0:
We use PG_WAL_SEGMENT_SIZE_DEFAULT to compare and check WAL size on pre-11
installations. However, there is an assertion with hard-coded WAL size value,
which doesn't respect PG_WAL_SEGMENT_SIZE_DEFAULT. This patch fixes this
hard-coded value.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -99,6 +99,17 @@
 
                         <p>Add <code>FN_NO_RETURN</code> macro.</p>
                     </release-item>
+
+                    <release-item>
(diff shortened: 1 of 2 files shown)


dwsteele (approved):
Looks good. Can you pull in the release note from https://github.com/dwsteele/pgbackrest/tree/fix_wal_size_assert_hardcode (since I don't have permission to modify your branch)?

InnerLife0:
Done. Thanks.

PR #1783, opened 2022-06-14
Use S3 IsTruncated flag to determine list continuation.

dwsteele:
Previously we were just checking for the existence of NextContinuationToken, which the S3 documentation indicates will not be present when the list is not truncated. However, recent versions of Scality send a blank NextContinuationToken when IsTruncated is false. Sending the blank continuation token back causes Scality to send another blank continuation token and an infinite loop occurs.

Instead use IsTruncated (which is required to be present) to determine whether NextContinuationToken should be present. Error if NextContinuationToken is then missing or empty, since an empty token caused an infinite loop with the Scality server (which arguably should have errored when passed an empty token).


Soulou:
Our provider using Scality has just announced they'll upgrade their ring the 5th of July. We have thousands of PostgreSQL deployments using pgbackrest (We're selling Database as a Service). Sooo.. stake is high for us.

@dwsteele, from your understanding of the issue, what's the change this patch WILL be the good one?

@MonChr Would you be able to test this patch about your already upgraded Scality cluster?

dwsteele:
> from your understanding of the issue, what's the change this patch WILL be the good one?

@Soulou This patch has been tested against Minio and AWS S3, so looks good from our perspective. We don't test against Scality, but of course you or @MonChr are welcome to do so.

Soulou:
@dwsteele Thanks a lot for your answer, hoping @MonChr will be available to do so

Soulou:
@dwsteele We'll be able to test the PR on an upgrade Scality Ring on 30/06 or 01/07 at worse.

MonChr:
Hello,
No problem to perform tests with Scality and pgBackRest, we are also waiting for this evolution (we have reverted some of our PostgreSQL nodes under barman to survive in the meantime).
See you soon.

Soulou:
I confirm, we've performed tests on our hand. Infinite loop before patch, working as expected with it.

So LGTM

Soulou:
@dwsteele Do you think a release can be perform quickly once this is merged?

dwsteele:
We are not planning to do an out-of-cycle release for this change. Of course, you are welcome to patch 2.39 with the change.

dwsteele:
@Soulou How would you like your name to appear in the release notes?

dwsteele:
@MonChr How would you like your name to appear in the release notes?

MonChr:
You can name me in the release notes.
For the tests, I will be absent from July 1st to July 4th, sorry. But if you have a beta version, I can test it today.

Soulou:
@dwsteele, you can us @Soulou  Thanks a lot!

precision-software:
This looks good to me.  You are covering the two important things.  1) use the truncate field as specified in the S3 documentation, and 2) there is an assertion check that the continuation token is present when truncate is set.

precision-software (approved):

MonChr:
> > You can name me in the release notes.
> 
> As `MonChr`?
> 
You can use Christian Montagne, thanks a lot.

dwsteele:
@MonChr Done in 326d152a14121cb298089847a3694aa9f50bcbd2.

MonChr:
Hello,
I have installed and tested version 2.40. It is absolutely functional with Scality S3 v8.5.3.1
Thank you for your reactivity and all your actions.

PR #1785, opened 2022-06-15
WIP: win32 port

shkshk90:
# Goal 

PR to track progress of porting pgbackrest to windows.

## Targets to port:

- [x] error
- [x] stack-trace
- [x] type-convert
- [x] assert-off
- [x] debug-off
- [x] assert-on
- [x] mem-context
- [x] time
- [x] encode
- [x] type-object
- [x] type-string
- [x] type-list
- [x] type-buffer
- [x] type-variant
- [x] reg-exp
- [x] log
- [x] debug-on
- [ ] fork
- [x] wait
- [x] type-json
- [x] type-key-value
- [x] type-xml
- [x] stat
- [x] user
- [x] io
- [x] type-pack
- [ ] io-tls
- [ ] io-http
- [ ] exec
- [x] ini
- [ ] compress
- [ ] exit
- [ ] lock
- [ ] protocol
- [ ] config
- [ ] storage
  * Depends on regex 
- [ ] postgres
- [ ] build
- [ ] info
- [ ] db
- [ ] command


meson.build
```diff
@@ -156,7 +161,19 @@ endif
 configuration.set_quoted('CFGOPTDEF_CONFIG_PATH', get_option('configdir'), description: 'Configuration path')
 
 # Set FN_NO_RETURN macro
-configuration.set('FN_NO_RETURN', '__attribute__((__noreturn__))', description: 'Indicate that a function does not return')
+if cc.has_function_attribute('__noreturn__')
```
shkshk90, line 164: @dwsteele: 
I wrapped ALL my changes inside if-s that check for msvc, except for this one.

However, it seems that my changes broke ALL platforms.
Any ideas?

dwsteele: This should work bd61e9d5ef. See the reference linked in that commit for valid values.

I also merged from main to cleanup cruft in the branch.

shkshk90:
@dwsteele: 

After the last commit, I have 2 warnings triggered by msvc:
* common\time.c(40,43): warning C4244: '=': conversion from 'time_t' to 'long', possible loss of data
  * This is due to `time_t` defined to long long, while `tv_sec` is defined as `long` on Windows.
  * I can workaround this with a `\#define`
* src\common\memContext.c(596,91): warning C4293: '<<': shift count negative or too big, undefined behavior
  * Not sure yet what is this about. It might be due to MSVC not accepting designated initializer syntax, need to check.

shkshk90:
> @dwsteele:
> 
> After the last commit, I have 2 warnings triggered by msvc:
> 
> * common\time.c(40,43): warning C4244: '=': conversion from 'time_t' to 'long', possible loss of data
>   
>   * This is due to `time_t` defined to long long, while `tv_sec` is defined as `long` on Windows.
>   * I can workaround this with a `\#define`
> * src\common\memContext.c(596,91): warning C4293: '<<': shift count negative or too big, undefined behavior
>   
>   * Not sure yet what is this about. It might be due to MSVC not accepting designated initializer syntax, need to check.

I fixed the first issue, 
Now the `log` and the `io` targets also compile.

I get couple of extra warnings, but they seem mostly fine with me. 
```console
pgbackrest\src\common\log.c(319,36): warning C4267: 'function': conversion from 'size_t' to 'unsigned int', possible loss of data
pgbackrest\src\common\log.c(434,54): warning C4018: '<': signed/unsigned mismatch
```

dwsteele:
The goal is to have no warnings so lets clean them up as we go.

> log.c(319,36): warning C4267: 'function': conversion from 'size_t' to 'unsigned int', possible loss of data

For now let's take care of this in the #define:
```
#define write(fd, buf, count) _write(fd, buf, (unsigned int)count)
```
Might be better to inline this, but we can make that decision later.

> log.c(434,54): warning C4018: '<': signed/unsigned mismatch

My guess is MSVC coerces to signed whenever there is subtraction, but we can fix this with a cast:
```
        for (unsigned int paddingIdx = 0; paddingIdx < (unsigned int)((logLevel - logLevelDebug + 1) * 4); paddingIdx++)
```

shkshk90:
Hello @dwsteele, 
In yaml.h, we have the following enum:

```c
typedef enum
{
    yamlEventTypeNone = STRID5("none", 0x2b9ee0),
    yamlEventTypeStreamBegin = STRID5("stream-begin", 0x724e516da12ca930),
    yamlEventTypeStreamEnd = STRID5("stream-end", 0x8e2eda12ca930),
    yamlEventTypeDocBegin = STRID5("doc-begin", 0xe49ca2d8de40),
    yamlEventTypeDocEnd = STRID5("doc-end", 0x11c5d8de40),
    yamlEventTypeAlias = STRID5("alias", 0x130a5810),
    yamlEventTypeScalar = STRID5("scalar", 0x241604730),
    yamlEventTypeSeqBegin = STRID5("seq-begin", 0xe49ca2dc4b30),
    yamlEventTypeSeqEnd = STRID5("seq-end", 0x11c5dc4b30),
    yamlEventTypeMapBegin = STRID5("map-begin", 0xe49ca2dc02d0),
    yamlEventTypeMapEnd = STRID5("map-end", 0x11c5dc02d0),
} YamlEventType;
```

These cause the following warning:
`warning C4309: 'initializing': truncation of constant value`. 

The enum seems to be `int`, and with large numbers, it has negative values.
Are the constant values needed? 
I can re-define them as macros on Windows.

What do you think?

dwsteele:
Check out this article: https://iangilham.com/2012/09/22/VS2003-compiler-bug-64bit-enum-truncation.html.

Looks like you might need to specify the type for the enum, like:
```
typedef enum : Int64
{
    ...
```

precision-software:
First, a question. Which version of C supports the typed enum?  Is it part of C99, and if not, is that a problem?

Another thought, perhaps for the future.

It is possible to define a C macro which computes a hash function at compile time.  We could simply specify STRID("teststring") and let the STRID macro compute an integer expression. It seems less error prone than calculating the hash offline.

Here is an example implementation which could probably be simplified using the Boost macros.
   https://stackoverflow.com/questions/28654707/how-to-calculate-the-hash-of-a-string-literal-using-only-the-c-preprocessor

dwsteele:
> First, a question. Which version of C supports the typed enum? Is it part of C99, and if not, is that a problem?

Enums are typed in C99 -- not sure what you mean?

> It is possible to define a C macro which computes a hash function at compile time.

You can, but the macros can expand to thousands of lines. It's painful -- and they can't be used in switch statements. It could certainly be done in a preprocessor but we don't generate that many StringIds so doesn't seem worth it.

Also, these are not hashes, they are encoded strings that can be decoded for logging, debugging, etc. The advantage is they can be used in switch statements and don't need allocations, but they can also be rendered as strings when needed. See src/common/type/stringId.h for more info.

precision-software:
In C99, all enums have "int" type.  The extension to general integer types comes from C++.  If you compile with gcc -pedantic, this code will fail.

I wonder, would this problem go away if we used the "/permissive" option of msvc?

precision-software:
As an experiment, I tried adding /permissive to the msvc compile options.  There was no change - msvc still truncates enum values.

I hesitate to suggest it, but we could use mingw as a workaround until we figure things out.

shkshk90:
I confirm that I tried specifying the type on my side, and it didn't work.

I knew the syntax from C++, but I didn't know that it might work for C.
It seems a non-standard extension from gcc, and msvc doesn't support it.

I did look into the compiler options also, but I don't see anything that can enable such feature.
I can define them using a constant _Int64 literals, what do you think?

shkshk90:
Quick look at the [standard](https://en.cppreference.com/w/c/language/enum), I don't see that typed enums are supported in any version of C.

It seems also, that this issue was faced [here](https://github.com/capstone-engine/capstone/issues/323), and they converted the values to `#define`s.

dwsteele:
>In C99, all enums have "int" type. The extension to general integer types comes from C++. If you compile with gcc -pedantic, this code will fail.

What happens if we remove `'c_std=c99'` from `meson.build`? We could keep C99 for unit-testing, but there are other reasons why we might not want to enforce it for building the binary, see https://www.phoronix.com/scan.php?page=news_item&px=Arch-Linux-Bizarre-Zstd.

Further, if this is something that is supported by C11 we might want to consider migrating to that standard, or even compiling under C++ for MSVC.

precision-software:
So far, we have 3 possible workaround/solutions.   In order of their appearance here,

1. Rewrite the enumerator constants as #defines.
2. Switch to the mingw compiler.
3. Define a macro "Int64_literals" as ": int64_t" on Windows and place it in select enum definitions.

Brainstorming, there is another option we haven't discussed.
4. Create a parallel table of strings. This table allows us to fetch the string without encoding the string in the enum constant.

We may need a quick+dirty solution to keep the port moving forward while we figure out a longer term solution.

Thoughts?

dwsteele:
> Rewrite the enumerator constants as #defines.

This breaks the compiler checking switch cases, which is a feature of this design. Also, these enums are everywhere so it's a pretty big change.

> Switch to the mingw compiler.

Might work. I assume you mean: https://www.mingw-w64.org/?

> Define a macro "Int64_literals" as ": int64_t" on Windows and place it in select enum definitions.

Samuel said he tried this and it did not work, but seems like the simplest solution if removing the C99 requirement allows it to work.

> Create a parallel table of strings. This table allows us to fetch the string without encoding the string in the enum constant.

This would break many things -- these are expected to be encoded strings. In fact, this was the system we moved away from because it was so cumbersome.

precision-software:
>> Define a macro "Int64_literals" as ": int64_t" on Windows and place it in select enum definitions.

>> Samuel said he tried this and it did not work, but seems like the simplest solution if removing the C99
>> requirement allows it to work.

Unstated, the macro would be empty on other platforms.

shkshk90:
Also @dwsteele, Would you kindly check why the current build is not working? 
All my changes should be behind a windows flag, so nothing should have been affected?

shkshk90:
I did check, and saw that, but there shouldn't be an issue here.

Anyway, I will re-check, thanks for the hint.

precision-software:
Switching to C++ is likely to be a morass.  C++ does a whole bunch of name mangling which will be incompatible with C code unless you declare everything  ‘extern “C” …’

Installing mingw is straightforward and I can give you a hand. Reciting from memory, 

*	Download msys2 and install it.
*	Run the msys2 terminal startup program.
*	You will now have a session similar to Linux.
*	Using the pacman package manager (similar to apt), install mingw.
It is a subdirectory under the main msys2 directory.
*	Restart your terminal using the terminal startup from msys2/mingw.
*	Your path should now include the mingw compilers. 
(I don’t know why the compilers are rooted under msys2/mingw rather than msys2.)
*	Continue installing packages as needed. Pacman is just like apt.

 

Postgresql built the first time I tried it under this environment.

src/common/memContext.c
```diff
@@ -40,13 +40,21 @@ typedef struct MemContextAlloc
 Contains information about the memory context
 ***********************************************************************************************************************************/
 // Quantity of child contexts, allocations, or callbacks
+#ifndef _WIN32
```
precision-software, line 43: Is it possible to test for MSVC rather than win32?  It would keep the options open for using the gnu compiler on windows.

precision-software (commented):
I think these ifdefs are great as a short-term hack to get things running, but they are not so good for production code.  Maybe we should create a new file system driver for native windows rather than force-fitting a non-Posix filesystem into the Posix code.

Another approach might be to use or create a small library translating some of the Posix calls into native windows. (Cygwin is too heavy handed.)

My concern comes from past experience with "ifdef'itis", the creeping flow of ifdefs into code which should be machine independent. 

Your thoughts?

dwsteele:
> I think these ifdefs are great as a short-term hack to get things running, but they are not so good for production code. 

We discussed this before you came onboard. This is only temporary and we definitely plan to have a Windows storage driver. For now it is easier to hack up the Posix driver to give us an idea of what the Windows driver will look like.

PR #1786, opened 2022-06-15
Add retry to storagePosixPathRemove().

dwsteele:
Add a retries to handle the case where readdir() skips over existing files when files are simultaneously being deleted.

This is a bit of a band aid. There are lots of other places where this readdir() issue could bite us.

NOTE: this could use better testing. There is 100% coverage because of the way to code is written, but a regression of the retry would likely not be detected.

### DIFF
--- a/src/storage/posix/storage.c
+++ b/src/storage/posix/storage.c
@@ -433,6 +433,17 @@ storagePosixPathRemoveCallback(void *const callbackData, const StorageInfo *cons
     FUNCTION_TEST_RETURN_VOID();
 }
 
+// Helper to determine if an error code indicates that a path is not empty. This can vary by implementation.
+static bool
+storagePosixErrorIsEmpty(const int error)
+{
+    FUNCTION_TEST_BEGIN();
+        FUNCTION_TEST_PARAM(INT, error);
+    FUNCTION_TEST_END();
+
+    FUNCTION_TEST_RETURN(BOOL, error == ENOTEMPTY || error == EEXIST);
+}
+
 static bool
 storagePosixPathRemove(THIS_VOID, const String *path, bool recurse, StorageInterfacePathRemoveParam param)
 {
@@ -466,7 +477,7 @@ storagePosixPathRemove(THIS_VOID, const String *path, bool recurse, StorageInter
                     removed = true;
                 }
                 // Else if not empty but recursion requested then remove all sub paths/files
-                else if (errno == ENOTEMPTY && recurse)
+                else if (storagePosixErrorIsEmpty(errno) && recurse)                                                // {vm_covered}
                 {
                     StoragePosixPathRemoveData data =
                     {
@@ -477,7 +488,7 @@ storagePosixPathRemove(THIS_VOID, const String *path, bool recurse, StorageInter
                     storageInterfaceInfoListP(this, path, storageInfoLevelExists, storagePosixPathRemoveCallback, &data);
                 }
                 // Else error
-                else
+                else                                                                                                // {vm_covered}
                     THROW_SYS_ERROR_FMT(PathRemoveError, STORAGE_ERROR_PATH_REMOVE, strZ(path));                    // {vm_covered}
             }
             // Else path was removed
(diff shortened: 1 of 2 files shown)


src/storage/posix/storage.c
```diff
@@ -452,28 +463,39 @@ storagePosixPathRemove(THIS_VOID, const String *path, bool recurse, StorageInter
...
-        if (rmdir(strZ(path)) == -1)
+        do
         {
-            if (errno != ENOENT)                                                                                    // {vm_covered}
-                THROW_SYS_ERROR_FMT(PathRemoveError, STORAGE_ERROR_PATH_REMOVE, strZ(path));                        // {vm_covered}
+            // If path could not be removed
+            if (rmdir(strZ(path)) == -1)
+            {
+                // Path does not exist
+                if (errno == ENOENT)
```
jreidthompson, line 474: In the original implementation, ENOENT would result in returning true.  New implementation is returning false for ENOENT?

jreidthompson: OK.  Ignore the above.

jreidthompson (approved):
This looks good to me. Sorry for the noise/initial comment. I definitely need more caffeine this morning, I completely misread the left hand side of the diffs in git

src/storage/posix/storage.c
```diff
@@ -452,28 +463,39 @@ storagePosixPathRemove(THIS_VOID, const String *path, bool recurse, StorageInter
...
+                {
+                    result = false;
+                    removed = true;
+                }
+                // Else if not empty but recursion requested then remove all sub paths/files
```
precision-software, line 479: I'm a bit confused by the comment and the name of the function. Should the function be named "storagePosixErrorIsNotEmpty"?

src/storage/posix/storage.c
```diff
@@ -452,28 +463,39 @@ storagePosixPathRemove(THIS_VOID, const String *path, bool recurse, StorageInter
...
-            };
+        bool removed = false;
 
-            // Remove all sub paths/files
-            storageInterfaceInfoListP(this, path, storageInfoLevelExists, storagePosixPathRemoveCallback, &data);
-        }
-
-        // Delete the path
-        if (rmdir(strZ(path)) == -1)
+        do
```
precision-software, line 468: This would be a great place to leave a comment 1) keep trying until the directory is fully removed, and 2) why it might not work the first time.

Can we guarantee it can't go into an infinite loop?  Perhaps a limit on how many times we retry?

precision-software:
Adding and deleting files while reading a directory with readdir() can be a bit unpredictable.  scandir() requires more memory, but it might be an alternative way to solve this problem. The retry approach used here is robust and should work across local and remote file systems, so I'd say go with it.

src/storage/posix/storage.c
dwsteele: It certainly should. Fixed in 5c5f59990.

src/storage/posix/storage.c
dwsteele: Comment updated in 8fcb15cf4f8 and b36d5afce.

> Can we guarantee it can't go into an infinite loop? Perhaps a limit on how many times we retry?

It seems unlikely, but not impossible. A retry limit would be a good here -- that's pretty standard practice elsewhere. That requires shimming something, but not that big a deal.

dwsteele:
I'm still not entirely happy with this approach. I think we'd have more control if we moved readdir() within this function, or better yet created an abstraction for readdir() that can be used at will in the module. I'm doing other work in this area so I'll think about that.

precision-software:
I agree a readdir abstraction a good idea.  One of the properties we might want from the abstraction is to take a consistent "snapshot" of the directory contents.  I don't think it matters for local Linux file systems, but it may matter for remote ones. I'm only hypothesizing, but the directory locking mechanisms used internally on local file systems don't exist for remote file systems such as NFS. Perhaps scanning should be part of the storage interface so we can tailor it as needed.

dwsteele:
We decided on a path snapshot approach as implemented by #1805. By not removing files during a scan we should be able to achieve the same functionality without the possible side effects of a retry approach, e.g succeeding even when another process is actively writing into the path we are trying to remove.

PR #1805, opened 2022-07-05
Create snapshot when listing contents of a path.

dwsteele:
Previously a callback was used to list path contents and if no sort was specified then a snapshot was not required. When deleting files from the path some filesystems could omit files that still existed, which meant the path could not be removed.

Filter . out of lists in the Posix driver since this special entry was only used by test code (and filtered everywhere in the core code).

Also remove callbacks from the storage interface and replace with an iterator that should be easier to use and guarantees efficient use of the snapshots.


src/command/backup/backup.c
```diff
@@ -489,147 +489,151 @@ backupBuildIncr(const InfoBackup *infoBackup, Manifest *manifest, Manifest *mani
...
+                // -----------------------------------------------------------------------------------------------------------------
+                case storageTypePath:
+                {
+                    // If the path was not found in the new manifest then remove it
+                    if (manifestPathFindDefault(manifest, manifestName, NULL) == NULL)
+                    {
+                        LOG_DETAIL_FMT("remove path '%s' from resumed backup", strZ(storagePathP(storageRepo(), backupPath)));
+                        storagePathRemoveP(storageRepoWrite(), backupPath, .recurse = true);
+                    }
+                    // Else recurse into the path
```
precision-software, line 544: The "else" is missing from the code.

dwsteele: Yep, that's in main as well. Not an issue since there will not be a path to recurse into, but not as efficient as it should be.

src/common/type/blob.h
```diff
@@ -0,0 +1,58 @@
+/***********************************************************************************************************************************
+Blob Handler
```
precision-software, line 2: Thanks for the excellent description!

precision-software:
Continuing my review and should finish up by lunch tomorrow (PST). I'm not sure if it better to have a series of smaller PRs or a single big one like this one. Either way it amounts to the same work for reviewing, but I find myself longing for bite-sized pieces. Will continue in the AM.

src/common/type/blob.h
```diff
@@ -0,0 +1,58 @@
...
+
+/***********************************************************************************************************************************
+Constructor
+***********************************************************************************************************************************/
+Blob *blbNew(void);
+
+/***********************************************************************************************************************************
+Functions
+***********************************************************************************************************************************/
+// Add data to the blob
```
precision-software, line 38: Since this is used mostly for efficient storage of strings, maybe we should have a "blbAddStr(*blob, *char) convenience function.

dwsteele: I thought about that, but would involve an extra call to strlen(). In this case the caller already knows the string size. Better might be blbAddStr(Blob *, String *) but that would be confusing because we would not being storing the entire string object, just the buffer holding the zero-terminated string.

precision-software (approved):
I think we can further refine the way iterators are used, but this code is pretty good. My thinking is the storage driver API should return a single level iterator (which would internally include the storage list in some cases). We would build the universal recursive scan on top of the driver's single level scan.  This is something to consider for the future to make the various drivers a bit more similar to each other.

Looks good.  I'm in favor of fixing that missing "else", not because it is doing any harm, but because it looks so much like a bug I'm uncomfortable leaving it in place.

src/command/backup/backup.c
dwsteele: Fixed in b54d25cf7a.

dwsteele:
> My thinking is the storage driver API should return a single level iterator (which would internally include the storage list in some cases).

Agreed. This would work for the object stores where we currently still have callbacks as well.

> Looks good. I'm in favor of fixing that missing "else", not because it is doing any harm, but because it looks so much like a bug I'm uncomfortable leaving it in place.

Done in b54d25cf7a. I was just holding it until you finished your review.

src/common/type/blob.h
precision-software: Someday, somebody (maybe me), is going to forget to add 1 to the strlen.  But you are right about efficiency and confusion.

dwsteele: I actually did forget to add 1 on my first pass and it blew up in a pretty spectacular fashion. Not only did the tests fail completely, valgrind complained that I had read into unitialized memory (since it never stopped reading from the first string). So, it's a concern but I think we have enough tools in place to catch that type of error.

sfrost (approved):

PR #1809, opened 2022-07-07
Add link to PostgreSQL configuration in repository host section.

dwsteele:
This should make the documentation clearer when starting from this section.


pgstef (approved):
Agreed. Adding a link seems reasonable indeed :+1:

PR #1814, opened 2022-07-13
Force target-timeline=current when restore type=immediate.

dwsteele:
If target-timeline is not current then the recovery may fail if the most recent timeline cannot be recovered to from the backup (current) timeline. This is arguably a bug in PostgreSQL since replay should stop before a timeline switch would need to occur.

In PostgreSQL < 12 the timeline defaults to current so nothing needs to be explicitly set.

### DIFF
--- a/src/command/restore/restore.c
+++ b/src/command/restore/restore.c
@@ -1663,11 +1663,15 @@ restoreRecoveryOption(unsigned int pgVersion)
         {
             kvPut(result, VARSTRZ(RECOVERY_TARGET_TIMELINE), VARSTR(cfgOptionStr(cfgOptTargetTimeline)));
         }
-        // Else set to current when type=immediate and PostgreSQL >= 12. The reason for this is that PostgreSQL will fail if the
-        // latest timeline is not recoverable from the backup timeline, even though replay should never go that far. This is
-        // arguably a bug in PostgreSQL but we'll handle it here until it can be fixed in PostgreSQL.
+        // Else explicitly set target timeline to "current" when type=immediate and PostgreSQL >= 12. We do this because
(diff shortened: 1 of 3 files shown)


src/command/restore/restore.c
```diff
@@ -1656,9 +1658,18 @@ restoreRecoveryOption(unsigned int pgVersion)
...
-        // Write recovery_target_timeline
+        // Write recovery_target_timeline if set
         if (cfgOptionTest(cfgOptTargetTimeline))
+        {
             kvPut(result, VARSTRZ(RECOVERY_TARGET_TIMELINE), VARSTR(cfgOptionStr(cfgOptTargetTimeline)));
+        }
+        // Else set to current when type=immediate and PostgreSQL >= 12. The reason for this is that PostgreSQL will fail if the
+        // latest timeline is not recoverable from the backup timeline, even though replay should never go that far. This is
+        // arguably a bug in PostgreSQL but we'll handle it here until it can be fixed in PostgreSQL.
+        //
```
sfrost, line 1669: A bit of wordsmithing/copy-editing on this:

Else explicitly set target timeline to "current" when type=immediate and PostgreSQL >= 12.  We do this because type=immediate means there won't be any actual attempt to change timelines, but if we leave the target timeline as the default of "latest" then PG might fail to restore because it can't reach the "latest" timeline in the repository from this backup.  This is really a PG bug and will hopefully be addressed there, but we'll handle it here for older versions, at least until they aren't really seen in the wild any longer.

sfrost (approved):

src/command/restore/restore.c
dwsteele: Looks better to me. See f8f8f4535e for update.

PR #1838, opened 2022-08-15
Do not allow CATCH() to catch a fatal error.

dwsteele:
Catching individual fatal errors was only used in testing so the tests have been updated to use other errors instead. CATCH_FATAL() is now the only way to catch fatal errors.

This simplifies the logic a bit for upcoming changes to error handling and cleanup.

Also fix an issue where passing errorMessage() directly to THROW*() would attempt to copy the message buffer instead of preserving it, which is undefined behavior. Since there were no instances of this behavior before this commit, this was not a live bug.

### DIFF
--- a/src/common/error.c
+++ b/src/common/error.c
@@ -302,6 +302,8 @@ errorInternalJump(void)
 bool
 errorInternalCatch(const ErrorType *const errorTypeCatch, const bool fatalCatch)
 {
+    assert(fatalCatch || !errorTypeFatal(errorTypeCatch));
+
     // If just entering error state clean up the stack
     if (errorInternalState() == errorStateTry)
     {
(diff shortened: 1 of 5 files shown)


precision-software (commented):
.

src/common/error.c
```diff
@@ -371,9 +373,12 @@ errorInternalThrow(
     errorContext.error.functionName = functionName;
     errorContext.error.fileLine = fileLine;
 
-    // Assign message to the error
-    strncpy(messageBuffer, message, sizeof(messageBuffer));
-    messageBuffer[sizeof(messageBuffer) - 1] = 0;
+    // Assign message to the error. If errorMessage() is passed as the message there is no need to make a copy.
```
precision-software, line 376: This isn't just an optimization; it fixes a potential (but likely harmless) bug. With strncpy, the strings are not supposed to overlap.

I've never liked the way strncpy handles EOS.  We could use the Berkeley strlcpy which makes that extra assignment unnecessary

precision-software (approved):
Looks good.

src/common/error.c
dwsteele: True enough, but we've never used it in this way before, so it was never a problem. Normally when you want the same message you also want the same error type so we just RETHROW().

Still, a bug of course, but not a live one as far as I can tell.

dwsteele: I amended the commit message to make it clear that a bug was being fixed to prevent confusion in the future.

PR #1842, opened 2022-08-17
Skip mem context cleanup in CATCH_FATAL() block.

dwsteele:
An error that gets raised all the way to the top TRY block might need to free a lot of resources and any of these callbacks could throw an error and mask the original error. In fact this is pretty likely since we are already in an error state. For example, the Db object will try to close the remote db connection, but if the protocol is in a bad state it will not be able to do so.

Solve this, for now, by not freeing memory or calling callbacks in the CATCH_FATAL() block. This gives us a better chance if being able to report the error without encountering another error first.

For the most part, we don't need to worry about freeing resources (file handles, TLS contexts, etc.) if the program is going to exit immediately. However, it is important to attempt to terminate all active protocol connections, which is done by protocolFree() in main() since the protocol objects live in the top context.

Another way to handle this would be to implement an error stack and that is probably something we will do in the future. But, in the case of a segfault the original error would still be lost. Yet another option would be to still do cleanup but defer it until after the CATCH_FATAL() block.


precision-software (approved):
Looks good to go. I'd summarize the changes as 1) add a "fatal" parameter to the error handlers, and 2) in the memcontext error hander, don't free memory if fatal.  (and update the unit tests).

Finally, an extra thought.  What if ... we appended to the error message buffer instead of overwriting it?  Could that be a quick way of getting some error stack info before we're actually ready to implement an error stack?

dwsteele:
> What if ... we appended to the error message buffer instead of overwriting it?

I've been thinking about something along these lines, but it would likely make the stack trace confusing since it is stored separately (since the stack trace is only dumped sometimes). I basically came to the conclusion that it would be 75% as much work and just implementing a real error stack, so seems better to just go for it.

PR #1847, opened 2022-08-22
Allow upload chunk size to be configured for object stores.

dwsteele:
There are performance benefits to increasing the upload chunk size as long as the tradeoff with memory additional memory consumption is acceptable.

Make the chunk size configurable for S3, GCS, and Azure, but don't attempt to do any validation of the chunk size beyond some sane limits. The defaults remain as is for each storage type to avoid any unintentional regressions.


pgstef (approved):
The code change and help text look good to me :+1:

PR #1854, opened 2022-08-30
Fix incorrect time expiration being used for non-default repositories.

dwsteele:
If a repo is not specified for the expire command then the lowest repo becomes the default. In one place we were querying the default rather than a specific repo which led to an incorrect expiration begin applied. Fix it and add a test.

It would be better if the default repo could not be queried in this case but it is not clear how to do that since the repo option is valid for expire (unlike, e.g., archive-push).

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -16,6 +16,21 @@
     <release-list>
         <release date="XXXX-XX-XX" version="2.41dev" title="UNDER DEVELOPMENT">
             <release-core-list>
+                <release-bug-list>
+                    <release-item>
+                        <github-issue id="1852"/>
+                        <github-pull-request id="1854"/>
+
+                        <release-item-contributor-list>
+                            <release-item-ideator id="adam.brusselback"/>
+                            <release-item-contributor id="david.steele"/>
+                            <release-item-reviewer id="stefan.fercot"/>
+                        </release-item-contributor-list>
+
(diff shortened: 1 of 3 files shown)


pgstef (approved):
The fix looks indeed pretty accurate.

I tried it:
```bash
# repo1: 20220831-073055F, 20220831-073254F
# repo2: 20220831-070232F, 20220831-071616F
$ pgbackrest expire --stanza=c7pg --dry-run 
2022-09-15 12:00:00.000 P00   INFO: [DRY-RUN] expire command begin 2.41dev: 
--repo1-retention-full=1 --repo2-retention-full=30 
--repo1-retention-full-type=time --repo2-retention-full-type=time 
2022-09-15 12:00:00.000 P00   INFO: [DRY-RUN] repo1: expire time-based backup 20220831-073055F

$ pgbackrest expire --stanza=c7pg --dry-run 
2022-10-15 12:00:00.000 P00   INFO: [DRY-RUN] expire command begin 2.41dev: 
--repo1-retention-full=1 --repo2-retention-full=30 
--repo1-retention-full-type=time --repo2-retention-full-type=time 
2022-10-15 12:00:00.000 P00   INFO: [DRY-RUN] repo1: expire time-based backup 20220831-073055F
2022-10-15 12:00:00.000 P00   INFO: [DRY-RUN] repo2: expire time-based backup 20220831-070232F
```

PR #1855, opened 2022-08-31
Update archive.info timestamps after a successful backup.

dwsteele:
Lifecycle policies can cause the archive.info file and its copy to be removed since they are only updated on a stanza-upgrade. Update the timestamps after a successful backup to prevent this.

This does not mean that lifecycle policies should be used as a replacement for expiration. However, in some cases there may be policies in place that are out of admin control. If the lifecycle expiration is less than pgbackrest expiration then corruption of the earliest backup will occur at the very least and there might be other corruption which would make the repo unrecoverable.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -61,6 +61,19 @@
                     </release-item>
 
                     <release-item>
+                        <github-issue id="1853"/>
+                        <github-pull-request id="1855"/>
+
+                        <release-item-contributor-list>
+                            <release-item-ideator id="alex.richman"/>
+                            <release-item-contributor id="david.steele"/>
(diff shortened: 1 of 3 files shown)


pgstef (approved):
The code change and test look good to me.
If that small trick can make some users happy I agree it worth it.
Ran my usual tests and didn't found any obvious issue.

To me, this PR is ready to go :+1:

dwsteele:
> If that small trick can make some users happy I agree it worth it.

At this point I'm only intending this to help users who have lifecycle rules beyond the horizon of their pgbackrest expiration that have been enforced beyond their control. There are lots of other gotchas with lifecycle expiration that are not addressed here.

PR #1859, opened 2022-09-06
List iterator

precision-software:
Create an interface for iterating through abstract containers.  Currently implemented for Lists, the next target is to iterate through different types of storage directories.

The concepts are similar to Rust containers and iterators, with some macro "syntactic sugar".

To cast a List into an abstract collection:
       ` Collection collection= NEWCOLLECTION(List, list)`

The macros can iterate on the List directly, or through the abstract Collection.

```
       FOREACH(ItemType, item, Collection, collection)                 FOREACH(ItemType, item, List, list)
           doSomething(*item);                                            doSomething(*item);
       ENDFOREACH;                                                     ENDFOREACH;
```
With only Lists implementing the interface, collections and iterators are not particularly interesting. The power comes when we implement other types of collections. The next goal is to create ReadDir iterators to scan directories in the various storage drivers. Each driver can choose the most effective implementation and package it as an absract Collection for others to use.


src/common/type/collection.h
```diff
@@ -0,0 +1,163 @@
...
+    ENDFOREACH;
+
+***********************************************************************************************************************************/
+#ifndef PGBACKREST_COLLECTION_H
+#define PGBACKREST_COLLECTION_H
+
+#include <stdbool.h>
+#include "common/assert.h"
+
+// TODO: Temporary. There must be a reason this hasn't been done. Perhaps test mocks need the "attribute()" syntax visible.
```
dwsteele, line 30: If "never got around to it" is a good reason, then yes. Here's a PR that creates a macro for this commonly used pattern: https://github.com/pgbackrest/pgbackrest/pull/1861.

PR #1861, opened 2022-09-06
Add FN_INLINE_ALWAYS macro.

dwsteele:
Eliminate a lot of useless repetition for a commonly used pattern.


sfrost (approved):

precision-software (commented):
Looks good - I was wondering why it hadn't been done earlier.  I'm not completely sure about the  name - FN_INLINE_ALWAYS.  I'd favor dropping the ALWAYS part. of the name. (I don't think there will ever be an FN_INLINE_SOMETIMES)

dwsteele:
I'd prefer to differentiate this from the `inline` keyword which equates to inline sometimes.

PR #1878, opened 2022-09-19
Fix memory leak in file bundle backup/restore.

dwsteele:
When converting `restoreFile()` to support file bundling there were some IO objects that were only freed at the end of the function that should have been freed at the end of each loop. Wrap the loops in temp mem contexts to fix this.

Do the same to `backupFile()` since it would have a similar leak when resuming a backup. Since file bundles cannot be resumed the leak would not be as severe, but still seems worth doing to protect against future leaks.

Note that this commit does a lot of indentation so it is best viewed without whitespace changes: https://github.com/pgbackrest/pgbackrest/pull/1878/files?w=1


precision-software (approved):
The memory management changes look good to me. 

As an aside, I like the way IoFilters are mixed+matched. There is a lot of flexibility while still providing a uniform file access API.

dwsteele:
> As an aside, I like the way IoFilters are mixed+matched. There is a lot of flexibility while still providing a uniform file access API.

Thanks! That's the idea. The goal is also to have the filters be relatively easy to write. Most of the truly hairy logic is contained in https://github.com/pgbackrest/pgbackrest/blob/main/src/common/io/filter/group.c. Luckily we don't need to change that module very often.

PR #1879, opened 2022-09-20
Move link creation to storage interface.

jreidthompson:
Migrate hard/soft link creation to driver and storage interface. Utilize an enum to signify whether link should be a hard or soft link. The posix driver was the only current driver utilizing links and has been updated. This will prepare the way for adding sftp softlink creation in the sftp driver.


precision-software:
Looks good. Some thoughts:

- For the other storage managers, I assume the jump table will be filled in with a dummy which simply throws an exception.
- Would it be useful to query a storage manager to find out if it supports links, hard or soft?

jreidthompson:
If my reading is correct...
In backup.c, and common.c invocations are wrapped by a storageFeature check.
In restore.c, if a symlink existed when the backup was taken, then an attempt will be made to create it on restore. If unable to create it, then an error will be thrown.                                          

Since the function is optional, we could utilize a
   if (this->pub.interface.linkCreate != NULL)
wrapper in storageLinkCreate() ala storagePathSync().  At one point I had this, but without it no tests failed.

Or, we could add to storageLinkCreate()
ASSERT(this->pub.interface.linkCreate != NULL && storageFeature(this, storageFeatureSymLink));
ASSERT(this->pub.interface.linkCreate != NULL && storageFeature(this, storageFeatureHardLink));
I think that either of these would prevent us from attempting to reference the function if it isn't implemented.                    

Thoughts?

storageFeature(this, storageFeatureSymLink) and storageFeature(this, storageFeatureHardLink) are used to query the storage manager if it supports links.

dwsteele:
I think we need two things here. First, there should be a check in the storageNew() to make sure the interface function has been implemented when links are supported, something like:

precision-software:
- On one hand, If we use a "throw exception" default value, then we need an additional interface to ask if the driver supports links.
 - On the other hand, we can treat it as an Option<pointer> where null is a valid entry and we always test it before use. 

But ... the possibility of jumping to zero can be a big problem. My recollection is, some machines can't do a valid traceback when that happens, so you are stuck looking at the return address in a register. 

What about stores which support one type of link but not the other?

I don't like the idea of adding another API just to query link capabilities, and i don't like having to wait until a link is requested before throwing an error, and I don't like risking a jump to address 0. In short, I don't have a good answer, just a perplexing question.

Could we slip a query into the interface?  Creating a hard/soft link from null to null would return an error if it didn't support that type of link. Maybe something like that.

src/storage/storage.h
```diff
@@ -241,6 +241,12 @@ typedef struct StorageRemoveParam
 
 void storageRemove(const Storage *this, const String *fileExp, StorageRemoveParam param);
 
+// Create a hard or soft link
```
dwsteele, line 244: I'm not sure that linkType deserves a required parameter since it we use symlink in all but one place. We'd need a StorageNewWriteParam struct to contain the parameter. We could either keep the enum with SymLink as the default or just use a boolean `hard`. I'm fine with either.

If we keep the enum we'll want to use symlink instead of softlink. Or change the current instances of "sym" to "soft" but not sure that's worth it. Either way let's keep the naming consistent.

dwsteele:
Ideally we would also implement this method in the remote storage driver. At the moment we don't support creating links remotely, but seems better to write the entire feature rather than leaving this as a landmine for later.

src/storage/posix/storage.c
```diff
@@ -528,6 +528,46 @@ storagePosixRemove(THIS_VOID, const String *file, StorageInterfaceRemoveParam pa
...
+            strZ(target));
+    }
+    else if (linkType == storageHardLink)
+    {
+        // Create the hard link
+        THROW_ON_SYS_ERROR_FMT(
+            link(strZ(target), strZ(linkPath)) == -1, FileOpenError, "unable to create hardlink '%s' to '%s'", strZ(linkPath),
+            strZ(target));
+    }
+    else
```
dwsteele, line 565: This is such an unlikely error (because of checks done before this) that I'd prefer to make it an assert at line 560 and remove the if else (just need else). i.e. `ASSERT(linkType == storageHardLink)`.

dwsteele (changes requested):
See comments.

precision-software:
I agree. Having a storageFeature query looks like a good way to go.

dwsteele:
> Having a storageFeature query looks like a better way to go.

`storageFeature()` already exists and it is the driver that determines what features are available. Unless I'm misunderstanding?

precision-software:
Thanks for clarifying. Early on, I didn't realize we already had the query. I agree with your suggestion.
 
As a corner case for the Posix driver, do we need to worry about remote SMB mounts? Those mounts may not support symlinks even though they otherwise act like Posix.

dwsteele:
> As a corner case for the Posix driver, do we need to worry about remote SMB mounts?

We have the `cifs` driver for this case, which is basically the Posix driver with links and directory fsync disabled.

jreidthompson:
> I think we need two things here. First, there should be a check in the storageNew() to make sure the interface function has been implemented when links are supported, something like:
> 
> ```
>     CHECK(
>         AssertError, !storageFeature(this, storageFeatureSymLink) &&  !storageFeature(this, storageFeatureHardLink) || interface.linkCreate != null,
>         "linkCreate required");
> ```
> 
> Second, some assertions in the function would be good. Similar to what you have above but checking to make sure that the requested link type is supported by the driver. Something like:
> 
> ```
> ASSERT(this->pub.interface.linkCreate != NULL);
> ASSERT(linkType == storageSoftLink && storageFeature(this, storageFeatureSymLink) || linkType == storageHardLink && storageFeature(this, storageFeatureHardLink) 
> ```

Added these in. The CHECK forced the implementation of the function for the remote driver, otherwise it wouldn't build.

src/storage/storage.h
jreidthompson: Followed the model of the other functions and created StorageInterfaceLinkCreateParam to contain the parameter. Kept enums. Changed soft to sym to stay with original naming and keep things consistent.

src/storage/posix/storage.c
jreidthompson: Implemented these changes

precision-software:
I noticed the remote driver doesn't implement .followLinks. (Or did I miss something?) It would be nice if we could assert "remote(localhost) is equivalent to the local file system", but I would favor making that a separate PR.

Looks good.

test/src/module/storage/remoteTest.c
```diff
@@ -489,17 +489,24 @@ testRun(void)
...
+        // Write the link to the repo via the remote
+        TEST_RESULT_VOID(storageLinkCreateP(storageRepoWrite, file, linkPath), "create link");
+
         // Check the repo via the local test storage to ensure the remote wrote it, then remove via the remote and confirm removed
         TEST_RESULT_BOOL(storageExistsP(storageTest, strNewFmt("repo128/%s", strZ(file))), true, "file exists");
+        TEST_RESULT_BOOL(storageExistsP(storageTest, linkPath), true, "link exists");
         TEST_RESULT_VOID(storageRemoveP(storageRepoWrite, file), "remote remove file");
+        TEST_RESULT_VOID(storageRemoveP(storageRepoWrite, linkPath), "remote remove link");
         TEST_RESULT_BOOL(storageExistsP(storageTest, strNewFmt("repo128/%s", strZ(file))), false, "file removed");
+        TEST_RESULT_BOOL(storageExistsP(storageTest, strNewFmt("repo128/%s", strZ(linkPath))), false, "link removed");
```
dwsteele, line 509: It not clear to me why the link tests were mixed up into the remove tests. Seems like they should get a separate test section like the Posix tests. I would also like to see info tests here to show that the correct type of links are being created, i.e. the parameters are not getting mixed up in the protocol.

dwsteele (changes requested):
I did some rearranging and renamed the enum. The code looks ready to me now.

Please have a look at the comment regarding tests.

dwsteele:
> I noticed the remote driver doesn't implement .followLinks

Not sure what you mean. `followLinks` is in `storageRemoteInfo()` and it does appear to be implemented.

> It would be nice if we could assert "remote(localhost) is equivalent to the local file system"

This might be true for some filesystems, but not for example S3. What would be the advantage?

precision-software:
>> I noticed the remote driver doesn't implement .followLinks. (Or did I miss something?)

I was looking at the storageRemoteList code, but looking closer I see .followLinks doesn't apply to  storage*List on any of the  storage drivers. Nothing is missing.

test/src/module/storage/remoteTest.c
jreidthompson: Per comment, moved the link tests to their own section.  Added info and additional tests to validate the links.

dwsteele (approved):

PR #1887, opened 2022-10-04
Do not store references for zero-length files when bundling.

dwsteele:
This appears to have been an oversight in 34d64957. Storing the reference is not really correct since the file is not stored in a prior backup. It also uses more space.

There is no real harm in storing the reference, since it is always ignored on restore, but the code is simpler if the zero-length files can be dealt with during the manifest and don't need additional handling later on. This is also an important part of some upcoming optimizations.


pgstef (approved):
This change looks good to me :+1:

dwsteele:
Thanks, @pgstef!

PR #1888, opened 2022-10-04
Improve manifest file updates.

dwsteele:
The prior manifestFileUpdate() function was pretty difficult to use since all the parameters had to specified. Instead, pass a ManifestFile struct that has all members set as needed.

When new struct members are added the manifestFileUpdate() call sites will still need to be reviewed, but this should make the process of adding members a bit simpler.


pgstef (approved):
This code change looks good to me too :+1:

dwsteele:
Thanks again, @pgstef!

PR #1889, opened 2022-10-04
Add manifest flags for file processing during backup.

dwsteele:
The prior method was to check a combination of fields to determine if a file needed to be copied, delta'd, or resumed. This was complicated and ultimately imposed a limitation on the number of operations that could be performed.

Introduce copy, delta, and resume flags in the manifest to make it clearer which operations need to be performed and to reduce complex and duplicated logic.

This also allows zero-length bundled files to be completed during manifest build rather than later on during backup processing.

This commit is best viewed with whitespace changes hidden: https://github.com/pgbackrest/pgbackrest/pull/1889/files?w=1


pgstef (approved):
The code change looks good to me and I feel like the added tests should cover most of the situations :+1:

src/info/manifest.c
```diff
@@ -125,6 +128,15 @@ manifestFilePack(const Manifest *const manifest, const ManifestFile *const file)
     // Flags
     uint64_t flag = 0;
```
precision-software, line 129: At first pass, I wasn't enthused about packing bits using explicit shifts - usually I prefer using bitfields in the first place. But this case it makes sense - bit fields are not portable, and the shifting only takes place in the pack and unpack routines. So ... looks good.

This  function also explains an earlier comment about using low bits for the most likely cases. It isn't because big shifts are slower (which isn't true on most architectures) , but because the varint format is more efficient when higher bits are zero. So ... the earlier comment makes sense now. 

I approve ... just being verbose in my thinking.

precision-software (approved):
Again, looks good.

sfrost:
Looking at https://github.com/pgbackrest/pgbackrest/pull/1889/files#diff-e6fb8861af6cea164972d9684914ca9152a9b0fda5a9bb6c70f9032ca2f2b4f6 specifically at src/info/manifest.h, it seemed a bit odd to only have the ':1' explicit bit count for the 'checksumPage' and 'checksumPageError' bool fields of ManifestFile but not have those for these newly introduced bool fields.

Seems like we should be at least consistent there?

dwsteele:
> Seems like we should be at least consistent there?

Hmm, yeah. We used to store a copy of this struct for each file so it was very important to be as compact as possible. Now we unpack into the struct when requested so compactness is not as important.

I'm happy to make the new ones bitfields, though.

dwsteele:
Done here 352cbf1d5738a72029.

PR #1890, opened 2022-10-04
Add persistent reference list to manifest.

dwsteele:
The reference list was previously built at load time from whichever references existed in the file list. This was sufficient since the list was for informational purposes only.

The block incremental feature will require a reference list that contains all prior backups, even those that are not explicitly referenced from the manifest. Therefore it makes sense to build and persist a manifest list rather than building it at load time.

This list can still be used for informational purposes, though it needs to be sorted since the list it sill built for older manifest versions and may not be in sorted order.

Add strLstFindIdx() to find references in the list.


src/common/type/stringList.h
```diff
@@ -95,6 +95,18 @@ strLstExists(const StringList *const this, const String *const string)
...
 }
 
+// Find string index in the list
+typedef struct StrLstFindIdxParam
+{
+    VAR_PARAM_HEADER;
+    bool required;
+} StrLstFindIdxParam;
+
+#define strLstFindIdxP(this, string, ...)                                                                                            \
```
jreidthompson, line 105: Define continuation '\' is at column 134 vs 132

jreidthompson:
Looks good to me. Other than the format note, I haven't noted any issues with this PR.

src/common/type/stringList.h
dwsteele: Good catch, fixed.

PR #1891, opened 2022-10-04
Add varint-128 encode/decode to IoRead/IoWrite.

dwsteele:
This makes it more efficient to read/write (especially read) varint-128 to/from IO.

Update the Pack type to take advantage of the more efficient read and remove some duplicate code.


precision-software (approved):
This looks good .

precision-software (commented):
I noticed there is a conversion function "cvtUInt64ToVarInt128" for converting output, but there is no equivalent routine used for input.  Symmetry is nice, but on a more practical level, some other project is probably going to optimize the heck out of both of them, and it would be nice to plug it in when that happens.

Looking around the web , I found a SIMD version which converts a batch of values in parallel - a possible solution if we need faster streaming in the future.

I was expecting to see a  "get byte" or fread style interface to hide the buffering. Was that a performance decision?

These are more questions than anything. The code looks good.

dwsteele:
> I noticed there is a conversion function "cvtUInt64ToVarInt128" for converting output, but there is no equivalent routine used for input.

There is, `cvtUInt64FromVarInt128()`, but it is not efficient in this case because it wants the entire buffer up front but we don't know how large the buffer is going to be.

> Looking around the web , I found a SIMD version which converts a batch of values in parallel - a possible solution if we need faster streaming in the future.

I'm not sure we need that much optimization. In this case the IO portion is the bottleneck, not the decoding. But who knows in the future...

> I was expecting to see a "get byte" or fread style interface to hide the buffering. Was that a performance decision?

We used something like that in the version in the Pack object that this code replaced, `ioReadSmall()`. This is faster though, since we have direct access to the IO object's buffer.

PR #1900, opened 2022-10-14
Fix protocol error on short read of remote file.

dwsteele:
If a remote file read was stopped before the read was complete or if an error occurred in the middle of the read then the protocol would end up in a bad state and produce this error:

ProtocolError: client state is 'data-get' but expected 'idle'

Prevent this by reading the rest of the file on close() or free() to leave the protocol in an idle state for the next command.

This was a possible issue for bundling because the amount to read is known in advance and therefore eof may not be reached. However, I was only able to reproduce this issue with unreleased code.

On error this issue would cause the original error to be lost. The process may still fail with this fix (if the error comes from another source) but hopefully we'll get better information about the original error.

### DIFF
--- a/src/storage/remote/read.c
+++ b/src/storage/remote/read.c
@@ -72,8 +72,7 @@ storageReadRemoteFreeResource(THIS_VOID)
                 // If binary then discard
                 if (pckReadType(read) == pckTypeBin)
                 {
-                    // Free the buffer immediately on read in case there are many buffers to read
-                    bufFree(pckReadBinP(read));
+                    pckReadBinP(read);
                 }
                 // Else read is complete so discard the filter list
                 else
--- a/src/storage/remote/read.c
+++ b/src/storage/remote/read.c
@@ -8,6 +8,7 @@ Remote Storage Read
 
 #include "common/compress/helper.h"
 #include "common/debug.h"
+#include "common/io/io.h"
 #include "common/io/read.h"
 #include "common/log.h"
 #include "common/type/convert.h"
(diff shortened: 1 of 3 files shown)


src/storage/remote/read.c
```diff
@@ -43,6 +44,54 @@ Macros for function logging
...
+            MEM_CONTEXT_TEMP_BEGIN()
+            {
+                PackRead *const read = protocolClientDataGet(this->client);
+                pckReadNext(read);
+
+                // If binary then discard
+                if (pckReadType(read) == pckTypeBin)
+                {
+                    // Free the buffer immediately on read in case there are many buffers to read
+                    bufFree(pckReadBinP(read));
```
sfrost, line 76: The very next thing after this is literally MEM_CONTEXT_TEMP_END when going through this code path, isn't it?  Feel like either we don't need to be creating/destroying a temp mem context on every iteration of the loop as we seem to be doing here and instead move the temp mem context outside the loop and keep the bufFree here, or we don't really need the bufFree here because the temp context destroy will handle it.

sfrost (commented):
Generally looks alright modulo the one item about the bufFree and the temp memory context.  Code matches up to what's in storageReadRemote also.

src/storage/remote/read.c
dwsteele: Oops. I copy-pasted this from `storageReadRemote()` and then did not think hard enough about the contexts.

I think it makes sense to keep the temp mem context since PackRead is also holding a copy of the buffer. Performance is not the main goal here but minimizing memory usage seems like a good idea. So, done that way at 7a233a8d0d2.

sfrost (approved):

PR #1901, opened 2022-10-14
Use more generic descriptions for pg_start_backup()/pg_stop_backup().

dwsteele:
The names were changed in PostgreSQL 15, so update the code and docs to make the naming more generic where needed to avoid using a version-specific name in the logs and documentation.


src/command/backup/backup.c
```diff
@@ -989,7 +989,7 @@ backupFilePut(BackupData *backupData, Manifest *manifest, const String *name, ti
 
             manifestFileAdd(manifest, &file);
 
-            LOG_DETAIL_FMT("wrote '%s' file returned from pg_stop_backup()", strZ(name));
+            LOG_DETAIL_FMT("wrote '%s' file returned from backup", strZ(name));
```
turnstep, line 992: As this is no longer clear this is a function by context, maybe say `returned from backup function`

Same for other lines below e.g. 2960

turnstep (changes requested):
Looks fine, one minor quibble.

src/build/help/help.xml
```diff
@@ -1190,7 +1190,7 @@
                         <summary>Force a checkpoint to start backup quickly.</summary>
 
                         <text>
-                            <p>Forces a checkpoint (by passing <id>y</id> to the <id>fast</id> parameter of <code>pg_start_backup()</code>) so the backup begins immediately. Otherwise the backup will start after the next regular checkpoint.</p>
+                            <p>Forces a checkpoint (by passing <id>y</id> to the <id>fast</id> parameter of backup start function so the backup begins immediately. Otherwise the backup will start after the next regular checkpoint.</p>
```
pgguru, line 1193:
```suggestion
                            <p>Forces a checkpoint (by passing <id>y</id> to the <id>fast</id> parameter of the backup start function so the backup begins immediately. Otherwise the backup will start after the next regular checkpoint.</p>
```

pgguru (changes requested):

src/build/help/help.xml
dwsteele: Fixed in a46d71797a.

src/command/backup/backup.c
dwsteele: Good point, fixed in a9e546c3dc90.

pgguru (approved):
Changes LGTM, thanks!

turnstep (approved):

PR #1916, opened 2022-11-03
Block-level incremental backup.

dwsteele:
The primary goal of the block incremental backup is to save space in the repository by only storing changed parts of a file rather than the entire file. This implementation is focused on restore performance more than saving space in the repository, though there may be substantial savings depending on the workload.

The repo-block option enables the feature (when repo-bundle is already enabled). The block size is determined based on the file size and age. Very old or very small files will not use block incremental.

In a test database created with pg_bench -s 2000 and 1000 additional small tables added (about 15GB) the full backup with block incremental enabled was within about .5% of a non block-incremental backup using gz. Backups with zst were actually a bit smaller with the block incremental, which is an interesting result. But, this is not the most realistic data set so it may just be an artifact of really repetitious data.

NOTE: This commit only deals with omitting headers from encrypted blocks. The same should be done for compression. Also, headers can be excluded for bundles when block incr is specified.


src/info/manifest.c
```diff
@@ -695,6 +718,83 @@ manifestLinkCheck(const Manifest *this)
     FUNCTION_LOG_RETURN_VOID();
 }
 
+/***********************************************************************************************************************************
```
precision-software, line 721: This is a very good explanation. Thanks.

src/common/io/filter/chunk.c
```diff
@@ -0,0 +1,172 @@
+/***********************************************************************************************************************************
```
precision-software, line 1: I look forward to seeing the overall description of chunking and its purpose.  Question: can it tie into the compression/encryption code as well?  With chunking, it becomes possible to append to a file which contains encrypted/compressed/signed data. It seems like a generally useful idea.

precision-software: (FYI - I'm assuming each chunk is individually encrypted/compressed or signed)

precision-software: (I'll update my comments as I read through ... still digesting it.)

dwsteele: > I look forward to seeing the overall description of chunking and its purpose.

There is a brief description in the header: `Split data up into chunks so it can be written without knowing the eventual length of the data.` Basically the same idea as HTTP chunked encoding. We don't know how much data we are going to write up front, but this allows us to write as we go and the reader of the data can reassemble the chunks when reading sequentially.

> Question: can it tie into the compression/encryption code as well?

You can add this filter after cmp/enc and the output will be chunked.

> With chunking, it becomes possible to append to a file which contains encrypted/compressed/signed data. It seems like a generally useful idea.

Possibly, but appending to files in object stores is generally not allowed so not something we could take advantage of.

src/command/backup/blockMap.h
```diff
@@ -0,0 +1,92 @@
+/***********************************************************************************************************************************
```
precision-software, line 1: This is a very compact representation of the block map when on disk. The limitation is the 20 byte checksum which can't be compressed. In memory, it takes around  48 bytes per entry, and I assume 25-30bytes when serialized to disk. Question: how many entries are we expecting the block maps to contain?  This is more doublechecking. From my first impressions, the map is reasonably sized with even a million entries.

Usually I would suggest ordering the fields by their alignment needs. It doesn't make any difference in this case, since SHA1 size is 20 bytes and all fields align without creating holes.

precision-software: Unless I've understated the likely size, this looks quite reasonable,

precision-software (approved):
This looks good to me.  It makes effective use of the "IoFilter" architecture so it can be packaged as "just another filter".  It also introduces solutions which may be applicable to other situations. 
For example, 

- chunking with a map provides the ability to seek into specific places in a compressed/encrypted stream.
- hash codes and resource fields implement a sort of chunk based "copy on write". 

I wonder if it is worth trying to separate those two mechanisms. - a chunk map indexing the chunks in that file, and a COW map pointing to chunks across the manifest. I'll leave that as a vague sort of question, not as anything needing resolution.

src/command/backup/blockIncr.h
```diff
@@ -0,0 +1,46 @@
...
+and a prior map (each subsequent time). The block list is stored first and the block map second. The filter returns the size of the
+block map so it can be extracted separately.
+
+The map is duplicated in each backup where the file has changed so the block map only needs to be retrieved from the most recent
+backup. However, the block map may reference block lists (or parts thereof) in any prior (or the current) backup.
+
+The block list consists of a series of blocks stored as chunks (see IoChunk filter) to make the format more flexible. Each block
+consists of the following, compressed and encrypted as required:
+
+- A varint-128 encoded block number stored as a delta of the previous block. So, if the first block is 138 it would be stored as 138
```
precision-software, line 14: Clarifying how sequential reads work, specifically how storing the block number at the beginning of the block provides enough information to scan a compressed file.

It looks like:
- the original blocks were fixed size cleartext, so we just keep reading cleartext up to that size,
- the IoFilters stop cleanly after providing the fixed size data.

These both seem like reasonable assumptions. Is that what you have in mind?

dwsteele:
> I wonder if it is worth trying to separate those two mechanisms. - a chunk map indexing the chunks in that file, and a COW map pointing to chunks across the manifest.

I think it is better to maintain one map, both for size efficiency and to simplify retrieval of the map. But at the same time the chunked format allows us to write incrementally and scan the blocks when the map is not available (as in verify, which has not been implemented yet).

> These both seem like reasonable assumptions. Is that what you have in mind?

Yes, but whether or not the data is plaintext (or encrypted/compressed) will depend on flags in the manifest. But in general the chunked data can be read without knowing that. Technically we are leaking info about the size of the changes here, but Stephen and I decided this was acceptable.

src/command/backup/blockIncr.h
dwsteele: The chunk headers provide information about the block size as stored, so we just read to the end of chunk marker. The block no is used to decide where the block goes in the destination file but is not used for decoding the block.

src/build/help/help.xml
```diff
@@ -506,6 +506,21 @@
...
+                        <summary>Enable block incremental backup.</summary>
+
+                        <text>
+                            <p>Block incremental allows for more granular backups by splitting files into blocks that can be backed up independently. This saves space in the repository and can improve delta restore performance because individual blocks can be fetched without reading the entire file from the repository.</p>
+
+                            <admonition type="note">The <br-option>repo-bundle</br-option> option must be enabled before <br-option>repo-block</br-option> can be enabled.</admonition>
+
+                            <p>The block size for a file is determined based on the file size and age. Generally, older/larger files will get larger block sizes. If a file is old enough, it will not be backed up using block incremental.</p>
+
+                            <p>Block incremental is most efficient when enabled for all backup types, including full. This makes the full a bit larger but subsequent differential and incremental backups can make use of the block maps generated by the full backup to save space.</p>
```
pgstef, line 519: Wouldn't it be worth it to force to have repo-block enabled for all backup types? Just as repo-bundle must also be enabled...

src/command/restore/restore.c
```diff
@@ -2327,15 +2329,31 @@ static ProtocolParallelJob *restoreJobCallback(void *data, unsigned int clientId
                 pckWriteStrP(param, restoreManifestOwnerReplace(file.user, jobData->rootReplaceUser));
                 pckWriteStrP(param, restoreManifestOwnerReplace(file.group, jobData->rootReplaceGroup));
 
-                if (file.bundleId != 0)
+                // If block incremental then modified offset and size to where the map is stored, since we need to read that first.
```
pgstef, line 2332: Comment typo? modified -> modify?

src/command/info/info.c
```diff
@@ -868,10 +876,19 @@ formatTextBackup(const DbGroup *dbGroup, String *resultStr)
...
-            strZ(strSizeFormat(varUInt64Force(kvGet(repoInfo, KEY_DELTA_VAR)))));
+            strZ(strSizeFormat(varUInt64Force(kvGet(repoInfo, KEY_SIZE_VAR)))));
+
+        if (kvGet(repoInfo, KEY_SIZE_MAP_VAR) != NULL)
+            strCatFmt(resultStr, " (%s map)", strZ(strSizeFormat(varUInt64Force(kvGet(repoInfo, KEY_SIZE_MAP_VAR)))));
+
+        strCatFmt(resultStr, ", backup size: %s", strZ(strSizeFormat(varUInt64Force(kvGet(repoInfo, KEY_DELTA_VAR)))));
+
+        if (kvGet(repoInfo, KEY_DELTA_MAP_VAR) != NULL)
+            strCatFmt(resultStr, " (%s map)", strZ(strSizeFormat(varUInt64Force(kvGet(repoInfo, KEY_DELTA_MAP_VAR)))));
```
pgstef, line 889: If I understood correctly, key_size_map = size of the map for the entire backup set but the map could reference a file from a previous backup so key_delta_map = only the size of the map for the files in this specific backup set.

Since that info will be displayed to users, wouldn't it be worth it to document it somewhere (in the user-guide perhaps)?

pgstef:
So, it took me some time to finally write down the result of the various [tests](https://github.com/pgstef/pgstef.github.io/blob/master/pages/pgbackrest-block-level-incremental-backup-preview.md) I've made. I decided to give it another round today after all the improvements made lately.

> Review here should focus on the proposed block incremental format and usage rather than of the current implementation.

The proposed implementation looks very good to me.

Even though the focus is on the restore performance first (and even if we're limited by the checksums calculation), the disk-space saving will worth it anyway imo, and obviously having to store less data should speed-up the backup time.

Overall :+1: then :-)

src/build/help/help.xml
dwsteele: We could do that, but I would like the user to be able to enable this without needing to wait until the next full backup. Even when enabled not all files get a block incremental (depending on size/age) so all block incremental backups are hybrids and it does not matter if it is turned on an off.

Also, we can't just force it as the default because some users depend on being able to snapshot and use the backups as-is.

src/command/info/info.c
dwsteele: Yes, we'll need to do that, but I can't add the docs yet since this will be release initially as an internal feature.

src/command/restore/restore.c
dwsteele: Fixed in 4842cfb2027.

src/command/info/info.c
dwsteele: Or maybe this is too much even for the detailed output? I added it mostly so we could analyze block size strategies and perhaps the JSON output is good enough for that.

src/build/help/help.xml
pgstef: OK, seems reasonable then :+1:

src/command/restore/restore.c
pgstef: :+1:

src/command/info/info.c
pgstef: Even when only displayed in the backup set text details, I believe we should document everything users can "see" or we'll probably have questions about it.

Probably step 1 would be to add it into JSON output and if we later see it can be used more often, then adding it to the backup set text with the documentation could be step 2?

dwsteele: Agreed, reverted this in eed6c4886.

pgstef: Looks good to me :-)

sfrost (approved):
Looked through the changes to the existing code and didn't see much to be concerned about.  Raised a couple of comments about things I did see though.  Also played around with the feature some and backups, full, incremental, incremental on older fulls, restores, et al, seemed to work for me.

src/info/manifest.c
```diff
@@ -726,6 +749,80 @@ manifestLinkCheck(const Manifest *this)
...
+};
+
+// File age to block multiplier map
+static struct ManifestBuildBlockIncrTimeMap
+{
+    uint32_t fileAge;
+    uint32_t blockMultiplier;
+} manifestBuildBlockIncrTimeMap[] =
+{
+    {.fileAge = 4 * 7 * 86400, .blockMultiplier = 0},
```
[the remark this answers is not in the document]
dwsteele: A block size of 0 means that block-incr is disabled. We do this for files that are old enough that they are not likely to change.

src/info/manifest.c
```diff
@@ -726,6 +749,80 @@ manifestLinkCheck(const Manifest *this)
...
+    // If block size > 0 then search age map for a multiplier
+    if (result != 0)
+    {
+        const time_t fileAge = timeStart - file->timestamp;
+
+        for (unsigned int timeIdx = 0; timeIdx < LENGTH_OF(manifestBuildBlockIncrTimeMap); timeIdx++)
+        {
+            if (fileAge >= (time_t)manifestBuildBlockIncrTimeMap[timeIdx].fileAge)
+            {
+                result *= manifestBuildBlockIncrTimeMap[timeIdx].blockMultiplier;
```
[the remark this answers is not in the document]
dwsteele: See above.

dwsteele:
@sfrost Wondering if we should rename the option from `repo-block` to `repo-block-incr`?

sfrost:
We've been calling it block-level-incremental ... Not sure what else might be "repo-block", but I'm good either way really and haven't a very strong opinion.

I guess 'repo-block' could possibly be confused as "repo block size" or something along those lines, so maybe "repo-block-incr" is better.  Don't feel strongly about it though.

dwsteele:
I left the option name as-is for now. Since this is being committed as an internal feature we can always change our minds before release.

PR #1929, opened 2022-11-17
Add hint when an option is specified without an index.

dwsteele:
Hopefully this will make it a little clearer to the user what is wrong when a user specifies an indexed option without an index.

Also fix an ambiguous use of cfgParseOptionP(). The prior code worked in that it set prefixMatch = true but it was not very readable.


pgstef:
Hm, even though the HINT looks good, I'm not sure what I'm doing wrong yet but I've got segmentation faults when using `pg-path` with this branch :thinking:

dwsteele:
> I've got segmentation faults when using pg-path with this branch

Oops! cfgOptionIdxName() can't be used before the config is initialized, so a bit of a chicken/egg issue here. The unit tests miss this because the config system has already been initialized by other tests. I'm thinking about how that can be made better.

Anyway, should be fixed now.

pgstef:
> Anyway, should be fixed now.

Indeed, working now :+1: 

LGTM then ;-)

pgstef (approved):

PR #1935, opened 2022-11-25
Fix missing reference in diff/incr backup.

dwsteele:
When loading prior manifests without the new reference list, the code failed to add the current backup to the reference list. Since the current backup can never be referenced, building references from the file list was not sufficient to generate a complete list.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -16,6 +16,23 @@
     <release-list>
         <release date="XXXX-XX-XX" version="2.43dev" title="UNDER DEVELOPMENT">
             <release-core-list>
+                <release-bug-list>
+                    <release-item>
+                        <github-issue id="1933"/>
+                        <github-pull-request id="1935"/>
+
+                        <release-item-contributor-list>
+                            <release-item-ideator id="marcel.borger"/>
+                            <release-item-ideator id="ulfedf"/>
+                            <release-item-ideator id="jaymefSO"/>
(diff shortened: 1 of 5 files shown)


pgstef (approved):
The PR seems to fix the mentioned issue:

```
$ pgbackrest version
pgBackRest 2.42

$ pgbackrest backup --type=incr --stanza=ro8pg --log-level-console=info
...
P00   INFO: last backup label = 20221125-123310F_20221125-123345I, version = 2.41
...
ASSERT: [025]: unable to find '20221125-123310F_20221125-123345I' in string list

$ pgbackrest version
pgBackRest 2.43dev

$ pgbackrest backup --type=incr --stanza=ro8pg --log-level-console=info
...
P00   INFO: last backup label = 20221125-123310F_20221125-123345I, version = 2.41
...
P00   INFO: new backup label = 20221125-123310F_20221125-124858I
...
P00   INFO: backup command end: completed successfully (4864ms)
```

dwsteele:
I made some modifications after thinking about this some more.

The main problem here was a bad test which I fixed in 28f660482aa5abb97850a9d7c89c2e5e9f23a4c9. This masked the issue and prevented it from being found. I fixed that separately and now it is clear in the test that the current label is missing.

I moved the code to add the current label so it will be a noop if there is nothing to do, i.e. the reference list was read from the manifest.

pgstef:
@dwsteele, looks good to me :+1:

PR #1941, opened 2022-12-01
Remove support for PostgreSQL 9.0/9.1/9.2.

dwsteele:
Our new policy is to support ten versions of PostgreSQL, the five supported releases and the last five EOL releases. As of PostgreSQL 15, that means 9.0/9.1/9.2 are no longer supported by pgBackRest.

Remove all logic associated with 9.0/9.1/9.2 and update the tests.

Document the new support policy.

Update InfoPg to read/write control versions for the history in backup.info, since we can no longer rely on the mappings being available. In theory this could have been an issue after removing 8.3/8.4 if anybody was using a version that old. Another option here would be to filter and remove old versions on load.

NOTE: VMs will need to be rebuilt to remove the old versions before this can be merged.


pgstef (approved):
> Update InfoPg to read/write control versions for the history in backup.info, since we can no longer rely on the mappings being available. In theory this could have been an issue after removing 8.3/8.4 if anybody was using a version that old. Another option here would be to filter and remove old versions on load.

I've tested the following situation:
- Install Postgres 9.2 and take some backups with pgBackRrest 2.43
- Install Postgres 15, install pgBackRest 2.44dev, perform a stanza-upgrade
- Restore a 9.2 backup doesn't work (as expected): `ASSERT: [025]: invalid PostgreSQL version 90200`
- Expire and expire --set seems to be working though

```bash
$ pgbackrest_dev --stanza=main info --set=20221201-132519F
...
db (prior)
    wal archive min/max (9.2): 00000001000000000000000F/00000001000000000000001B

    full backup: 20221201-132519F
        timestamp start/stop: 2022-12-01 13:25:19 / 2022-12-01 13:25:27
        wal start/stop: 000000010000000000000014 / 000000010000000000000015
        lsn start/stop: 0/14000020 / 0/15000048
        database size: 18.4MB, database backup size: 18.4MB
        repo1: backup set size: 2.1MB, backup size: 2.1MB
        database list: postgres (12059)

$ pgbackrest_dev --stanza=main restore --set=20221201-132519F --pg1-path=/tmp/test
ASSERT: [025]: invalid PostgreSQL version 90200
        --------------------------------------------------------------------
        If SUBMITTING AN ISSUE please provide the following information:
        
        version: 2.44dev
        command: restore
        ...
        --------------------------------------------------------------------

$ pgbackrest_dev --stanza=main expire --set=20221201-132519F
$ pgbackrest_dev --stanza=main info --set=20221201-132456F
status: error (requested backup not found)
```

If users can't restore a 9.2 backup (because not supported anymore), it might be confusing to see it in the info command; but in the other hand, they would need a way to know that the backups are there and expire them. I think the suggested change makes sense, filtering out the old versions on load could be even more confusing for some users imho.

dwsteele:
> I think the suggested change makes sense, filtering out the old versions on load could be even more confusing for some users imho.

The change to allow control versions for unsupported changes is already in this commit, see `info/infoPg.c`. And agreed, I don't like the filtering idea very much. It would probably make things more confusing.

If they have backups older than 9.3 then they are going to have problems. The logic I added was just to make sure that old versions in the history would not cause errors and that expiration would still work.

pgstef:
Yep, that's what I meant by "the suggested change" --> the current PR :-)

> The logic I added was just to make sure that old versions in the history would not cause errors and that expiration would still work.

This looks good to me :+1:

PR #1953, opened 2022-12-16
Store manifest checksums in memory more efficiently.

dwsteele:
Manifest checksums were stored as hex-encoded strings due to legacy compatibility with Perl. Storing the checksums as binary in memory uses half the space and avoids many conversions.

There is no change to the on-disk manifest format which stores the checksum as a hex-encoded string.

Also see preparatory commits: 9a9ee8e640, 010efffb, 17ce738ff.


pgstef (approved):
I did play a bit with this change (and ran my usual test suite) but didn't notice anything obvious.
Had a look at the code too and overall the PR looks good to me :+1:

PR #1956, opened 2022-12-21
Add repository checksum to make verify and resume more efficient.

dwsteele:
Calculate a checksum of the data stored in the repository when a file is transformed (e.g. compressed). This allows resume and verify to operate without needing to decompress/decrypt the data.

This can also be used to verify more complex formats such as block incremental and allow backups from the repository without needing to decompress the data to verify the checksum.

Add some basic encrypted tests to maintain coverage. These will be expanded in a future commit.


pgstef (approved):
I did play a bit with this change (and ran my usual test suite) and I agree this can indeed be a very interesting information to have.

Had a look at the code too and, except the comment which I might have misunderstood, overall the PR looks good to me +1

PR #1957, opened 2022-12-21
Create harnessIo.* shim files

jreidthompson:
Create harnessIo shim files with shimmed  ioSessionFd() that returns an arbitrarily defined file descriptor value. This will be required by the SFTP driver test script,


test/src/common/harnessIo.c
```diff
@@ -0,0 +1,75 @@
+/***********************************************************************************************************************************
+Harness for Testing
```
sfrost, line 2: Would be good to have this as a better comment ...

test/src/common/harnessIo.h
```diff
@@ -0,0 +1,16 @@
+/***********************************************************************************************************************************
+Harness for Testing
```
sfrost, line 2: Another comment that should be more descriptive.

test/src/module/common/ioTlsTest.c
```diff
@@ -949,5 +950,52 @@ testRun(void)
...
+
+        struct addrinfo hints = (struct addrinfo)
+        {
+            .ai_family = AF_UNSPEC,
+            .ai_socktype = SOCK_STREAM,
+            .ai_protocol = IPPROTO_TCP,
+        };
+
+        const char *port = "7777";
+        const char *hostBad = "172.31.255.255";
```
sfrost, line 967: We seem to be defining these over and over.. just wonder if maybe we should put them in one place?  Probably a question for @dwsteele really.

test/src/common/harnessIo.c
```diff
@@ -0,0 +1,75 @@
...
+#include "common/fork.h"
+
+#include "common/harnessConfig.h"
+#include "common/harnessDebug.h"
+#include "common/harnessIo.h"
+#include "common/harnessLog.h"
+#include "common/harnessProtocol.h"
+#include "common/wait.h"
+#include "storage/storage.h"
+
```
sfrost, line 20: Sure seems like we have a lot of things include'd above- do we really need all of these?

test/src/module/common/ioTlsTest.c
```diff
@@ -949,5 +950,52 @@ testRun(void)
         TEST_RESULT_PTR_NE(statToJson(), NULL, "check");
     }
 
+    // *****************************************************************************************************************************
+    if (testBegin("harnessIo"))
```
sfrost, line 954: We don't have any other tests names 'harness ...' instead we talk about testing the function and we may use a harness/shim for that, right?  Maybe this should be named differently?

sfrost (changes requested):

test/src/common/harnessIo.c
jreidthompson: looks like I can reduce this to
  #include "common/harnessConfig.h"
  #include "common/harnessDebug.h"
  #include "common/harnessIo.h"
will run through tests to verify. I must have copied another file for template and failed to comment the headers out and add back in as needed.

test/src/module/common/ioTlsTest.c
jreidthompson: will update to (testBegin("verify shimmed ioSessionFd()"))

test/src/common/harnessIo.h
jreidthompson: Updating to Harness for Io Testing would be in line with the others.

test/src/common/harnessIo.c
jreidthompson: Updating to Harness for Io Testing would be in line with the others.

test/src/module/common/ioTlsTest.c
dwsteele: Perhaps, but this test in general seems to be overkill. Seems like it would be better to just pass a bogus fd (e.g. HRNIO_FILE_DESCRIPTOR - 1) to `sckSessionNew()` and test to be sure the shim works that way.

Honestly not sure we need this test at all, since it should be pretty obvious in the test module if it is not working.

dwsteele:
I'm not sure it is a good idea to commit this separately (yet). It has no current uses and seems to me there is a good chance it may be changed during the review of the main patch. I'd rather see it in the context of the main patch and commit it separately once the main patch is looking pretty solid.

dwsteele:
Closing since it makes more sense to commit this with #1963.

PR #1960, opened 2022-12-23
Clarify target-action behavior on various PostgreSQL versions.

### LINKED ISSUE #987: Clarity around target-action in the documentation
The [`restore --target-action` documentation](https://pgbackrest.org/command.html#command-restore/category-command/option-target-action) states:

> This option is effective when `hot_standby=on` is configured in `postgresql.conf`, otherwise the cluster will be promoted when the target is reached or there is no more WAL in the archive.

However, I don't believe this is accurate. In recent versions of PostgreSQL, [1][2] `recovery_target_action` is considered anytime one of `recovery_target`, `recovery_target_name`, `recovery_target_time`, `recovery_target_xid`, or `recovery_target_lsn` is set. When `hot_standby` is enabled and the target is reached,
- `promote` ends recovery, picks a new timeline, and accepts connections.
- `pause` leaves the cluster in recovery and accepts connections.
- `shutdown` stops the cluster.

When `hot_standby` is disabled, the only difference is that `pause` acts like `shutdown`.

[1]\: https://www.postgresql.org/docs/12/runtime-config-wal.html
[2]\: https://www.postgresql.org/docs/11/recovery-target-settings.html

On PostgreSQL 12, I have observed that when the recovery target is reached:
- `hot_standby=on recovery_target_time={timestamp}` accepts connections and `pg_is_in_recovery() is true`. The default is to pause, as documented.

- `hot_standby=off recovery_target_time={timestamp}` stops the cluster. The cluster is not promoted, seemingly contrary to this documentation.

- `hot_standby=off recovery_target_time={timestamp} recovery_target_action=promote` ends recovery and accepts connections. The option is effective, seemingly contrary to this documentation.

cbandy:
The behavior of `pause` depends on the `hot_standby` parameter and the PostgreSQL version so mention both.

I verified the [behavior](https://github.com/CrunchyData/postgres-operator/blob/master/internal/pgbackrest/restore.md?plain=1) of PostgreSQL 9.6–15. PostgreSQL 12 is an inflection point because the behavior of an unset `recovery_target_action` with `hot_standby=off` changed in [this commit](https://git.postgresql.org/gitweb/?p=postgresql.git;h=2dedf4d9a899b36d1a8ed29be5efbd1b31a8fe85).

This fixes #987, in my opinion.


dwsteele (approved):
This looks good to me. We'll give it another few days to see if anyone else has a comment.

pgstef:
> This looks good to me. We'll give it another few days to see if anyone else has a comment.

Looks good to me too :+1:

PR #1962, opened 2022-12-26
Add compress-level range checking for each compress-type.

dwsteele:
The prior range checking was done based on the valid values for gz. While this worked it was a subset of what is available for lz4 and zst.

Allow the range to be specified for each compress-type. Adding this functionality to the parse module would be a better solution but that is a bigger project than this fix deserves, at least for now.


src/config/load.c
```diff
@@ -357,12 +357,26 @@ cfgLoadUpdateOption(void)
...
+            cfgOptionSet(cfgOptCompressLevel, cfgSourceDefault, VARINT64(compressLevelDefault(compressType)));
+        }
+        else if (compressType != compressTypeNone)
+        {
+            if (cfgOptionInt(cfgOptCompressLevel) < compressLevelMin(compressType) ||
+                cfgOptionInt(cfgOptCompressLevel) > compressLevelMax(compressType))
+            {
+                THROW_FMT(
+                    OptionInvalidValueError,
+                    "'%d' is out of range for '" CFGOPT_COMPRESS_LEVEL "' option when '" CFGOPT_COMPRESS_TYPE " option ' = '%s'",
```
pgstef, line 376: This gives `ERROR: [032]: '21' is out of range for 'compress-level' option when 'compress-type option ' = 'gz'`

Shouldn't it be `when '" CFGOPT_COMPRESS_TYPE "' option = '%s'`?

pgstef (changes requested):
Except for the error message, this PR looks good to me :+1:

src/config/load.c
pgstef: Hmm, even after a77a016, I still think there's an `'` too much:

Shouldn't `when '" CFGOPT_COMPRESS_TYPE "' option ' = '%s'` become `when '" CFGOPT_COMPRESS_TYPE "' option = '%s'` ?

I can try to change it myself if you wish ;-)

dwsteele: Wow, tough day! Fixed (I think) in aada0a951e367dbf.

pgstef: Indeed, looks good to me now :-) :+1:

pgstef (approved):

PR #1963, opened 2022-12-26
Add replacement for linefeeds in monitoring example.

dwsteele:
The copy command was converting \n to a linefeed, which the json conversion did not like. In a healthy repo there won't be any linefeeds but certain errors can contain them.

Fix by loading into a text field and then replacing the linefeed when converting to jsonb.


pgstef (approved):
Tried to reproduce the issue reported and this fix seems to solve it.
Looks good to me :+1:

PR #1966, opened 2022-12-28
Updates and clarifications to index page.

dwsteele:
In particular the section about other backup solutions not supporting parallel processing was no longer accurate, so reword it.

Also update some other sections that used older nomenclature, had awkward wording, or needed clarification.


pgstef (approved):
Looks good to me

Issue #1942, opened 2022-12-01
How to fail restoring a backup if there is only base backups from previous major versions?


Soulou:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version:

```
$ pgbackrest version
pgBackRest 2.39
```

2. PostgreSQL version:

13.9

4. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each:

Official Docker Image (Debian based)

5. Did you install pgBackRest from source or from a package?

From source

7. Please attach the following as applicable:

n/a

9. Describe the issue:

We're deploying PostgreSQL clusters and automating their major upgrades, including major upgrades. Our clusters are managed with Patroni.

When we trigger a major upgrade on a 2-nodes cluster, it works the following way: (example with v12 to v13)

* Cluster is running with v12 for a while, and at least 1 base backup exists
* Stop both members
* Run pg_upgrade on previous leader
* Restart previous leader with v13
* Restart previous follower with v13 (which has an empty datadir at that point)

As mentionned we're using Patroni which does handle the restart of the follower and has 2 methods to create a new replica out of it
```
      create_replica_methods:
        - pgbackrest
        - live_basebackup
```

→ It means that if pgbackrest fails to restore, then we want to stream a base backup from existing primary
    
State of base backups (You can see there is one v12 full base backup):
    
```
$ pgbackrest --stanza $STANZA info
stanza: ced4_5102
    status: ok
    cipher: aes-256-cbc

    db (prior)
        wal archive min/max (12): 000000010000000000000004/000000010000000000000005

        full backup: 20221201-150422F
            timestamp start/stop: 2022-12-01 15:04:22 / 2022-12-01 15:05:34
            wal start/stop: 000000010000000000000004 / 000000010000000000000004
            database size: 31MB, database backup size: 31MB
            repo1: backup set size: 3.8MB, backup size: 3.8MB
```

* pgbackrest is executed and succeeds, while we would expect it to fail since we're running in v13 and only `prior` backups are present for v12

```
2022-12-01 16:06:20.686664676 +0100 CET [postgresql-2] ----&gt; PostgreSQL version: postgres (PostgreSQL) 13.9 (Debian 13.9-1.pgdg110+1)
2022-12-01 16:06:20.687295845 +0100 CET [postgresql-2] Start Patroni
[...]

PR #1967, opened 2022-12-28
Restore errors when no backup matches the current version of PostgreSQL.

pgstef:
As discussed in issue #1942, it might be dangerous to restore the latest backup set when it's not part of the current db history.
Even if that could be a breaking change, there's a simple workaround by providing the backup set to restore (or provide a target time/lsn).

Not sure about the error message/hint though.


pgstef:
[023e856](https://github.com/pgbackrest/pgbackrest/pull/1967/commits/023e8561c009303bc821d21acd908c60405fc7ef) looks good to me. Thanks @dwsteele!

dwsteele (approved):

dwsteele:
I updated this again to remove the word "history" which is an internal term for us an unlikely to have meaning to a user.

pgstef:
> I updated this again to remove the word "history" which is an internal term for us an unlikely to have meaning to a user.

Looks great, thanks !

PR #1968, opened 2022-12-29
Refactor common/ini module to remove callbacks and duplicated code.

dwsteele:
The callbacks in iniLoad made the downstream code more complicated than it needed to be so use an iterator model instead.

Combine the two functions that were used to load the ini data to remove code duplication. In theory it would be nice to use iniValueNext() in the config/parse module rather than loading a KeyValue store but this would mean a big change to the parser, which does not seem worthwhile.


pgstef (approved):
Hi,
I've built the sources and tried some basic actions, read the code a few times and didn't spot anything obvious. So, looks good to me.

PR #1970, opened 2023-01-01
Improve libbacktrace implementation.

dwsteele:
The libbacktrace feature has not been working since the move to meson because libbacktrace detection was not added to the meson build. Add libbacktrace to meson and improve the feature so that it can be compiled into release builds. 

The prior implementation fetched line numbers with each stack trace push. Not only was this slow but it missed any functions that were not being tracked on our stack.

Instead just examine the backtrace when an error happens and merge it with the info we have on our stack. If the backtrace is not available then the output remains as before.

Also remove --backtrace from test.pl since the library is now auto-detected.


dwsteele:
Most of this was committed in c8264291238e. Use in the release build still needs to be evaluated.

sfrost (approved):
LGTM

PR #1980, opened 2023-01-20
Add warning about enabling "hierarchical namespace" on Azure storage.

dwsteele:
If this feature is enabled expire will fail since directories need to be deleted separately.

Ideally we would add support for this feature but for now we'll just document the issue.


pgstef (approved):
The warning looks good to me :+1:

Issue #1774, opened 2022-06-08
Documentation for upgrading pgbackrest binaries 


Krysztophe:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

7. Describe the issue:

In https://pgbackrest.org/user-guide-rhel.html#upgrading,
nothing is said about an upgrade from recent versions, like 2.35 to 2.39.
v1 to v2 is already so old.

Obviously, it is just a matter of upgrading the binaries, but that would be nice to write it.
For example, I don't know if which order binaries on primary and secondary servers must be upgraded: the one which writes into the repo must be the last one, I assume? Or maybe is it not important. I assume that the repos are backward-compatible.

PR #1993, opened 2023-02-07
Add explicit instructions for upgrading between 2.x versions.

Krysztophe:
Following https://github.com/pgbackrest/pgbackrest/issues/1774, simply add the statement that there is nothing special to do when upgrading between two 2.x versions. That's better when it is explicit.

I've left in the previous paragraph the bit about the default location that changed between 2.00 and 2.02, as it is more a matter of transitioning from 1.x to 2.x.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -36,6 +36,17 @@
             <release-doc-list>
                 <release-improvement-list>
                     <release-item>
+                        <github-pull-request id="1993"/>
+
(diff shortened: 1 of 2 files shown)


dwsteele (approved):
Made some edits so if they look OK to you we are ready to commit.

dwsteele:
@Krysztophe Do these edits look OK?

Krysztophe:
Sure!

PR #1996, opened 2023-02-13
Block-level incremental backup super blocks.

dwsteele:
Small blocks sizes can lead to reduced compression efficiency, so allow multiple blocks to be compressed together in a super block. The disadvantage is that the super block must be read sequentially to retrieve blocks. However, different super block sizes can be used for different backup types, so the full backup super block sizes are large for compression efficiency and diff/incr are smaller for retrieval efficiency.

For review, it is probably easier to look at blockMap.c from scratch rather than trying to look at the diff. Same for blockIncrProcess(). For the rest diffs should be OK.


src/build/help/help.xml
```diff
@@ -542,6 +542,26 @@
                         <example>16KiB=8KiB</example>
                     </config-key>
 
+                    <config-key id="repo-block-size-super" name="Block Incremental Super Block Size">
+                        <summary>Block incremental super block size.</summary>
+
```
precision-software, line 547: Another possible approach is to explicitly manage lz4 dictionaries.  In that case, a "superblock" would be defined as a dictionary plus a list of blocks, and each block could be decompressed individually rather than decompressing the entire superblock.  

Did you consider this approach as well? I can see some advantages and some disadvantages, but smaller blocks are more efficient for capturing small changes to a table and explicit dictionaries mean the compression ratios can be the same as for large blocks.

dwsteele: I did consider this approach (zstd also supports this), but I was concerned about the complexity of a per compression-method block format. bzip/gzip don't support this feature so we need a generic super block format anyway, which seemed like a good place to start.

I found that 1mb super blocks pretty much negate compression efficiency loss and since they are usually just a few hundred kb compressed fetching them is not too expensive. For object stores (S3, etc) the per request cost is high enough that fetching smaller blocks is not always the advantage that it might seem.

Dictionaries could certainly be implemented in a future release and enabled by something like `repo-block-strategy=dictionary` for the compression methods that support it. lz4 and zstd are not yet available for all pgbackrest builds but that should improved in the future.

dwsteele: The header/frame overhead for lz4/zstd is large enough that I think we'd still still want superblocks for very small block sizes (e.g. 8k) even if we used dictionaries. That's just a feeling, though, I have not tested it.

sfrost: Reviewing lz4 at least... looks like the overhead is only along the lines of tens-of-bytes and the same seems to be true for zstd.  Given an 8k block, that seems unlikely to be that much of an issue.  Being able to pull down a much smaller amount of a superblock than the whole thing strikes me as potentially helpful for keeping restores as performant as possible.

On the other hand, I agree that we could implement this later using a configuration option, so at this point, I'd rather have block level incremental in this release than wait for a new approach to be implemented, especially one that can be implemented in the future.

dwsteele: And we would likely want to train the dictionary on the super block and then use that dictionary just for the blocks in the super block. I'm worried that training the dictionary on the entire file could lead to poor results. It would certainly require two passes over each file, which I would prefer to avoid.

precision-software: I agree that we would have a separate dictionary per super block.  Note you don't have to train on the entire superblock - it might be reasonable to train on a reasonably sized sample of the superblock.

precision-software:
As an overall view, I see this as well coded and well thought out.  It took me a bit longer than I like to work my way through the data structures.  I think this would be a great place to show a picture or diagram of the data structures and how the pieces of the incremental backups relate to each other. I wouldn't hold up the review for that reason, and in fact, I'd be willing to translate a hand drawn picture into something like a mermaid diagram (as supported by GitHub) as a near-term enhancement for the documentation.

I didn't understand the purpose of the "continued" flag. Is it a method of allowing a superblock to contain additional blocks? I would also like to confirm the block offset in the data structures refers the offset of the compressed block. I'd be happy with a bit of clarification, but it isn't a showstopper. In the meantime, I am quite willing to give approval to avoid delays.

dwsteele:
> I'd be willing to translate a hand drawn picture into something like a mermaid diagram (as supported by GitHub) as a near-term enhancement for the documentation.

This might be a good thing to include in the block post that we will be writing for this feature.

> I didn't understand the purpose of the "continued" flag. Is it a method of allowing a superblock to contain additional blocks?

Pretty much. Say you have a super block in the first backup:
`ref 0 super block offset 999, block 0-3`

In the next backup blocks 1-2 are changed:
`ref 0 super block offset 999, block 0`
`ref 1 super block offset 10345, block 1-2`
`ref 0 super block continue, block 3`

The advantage here is we don't need to store any info about the continued super block in the map (e.g. offset, which can be quite large) because we already have the info about the current super block for each reference in memory.

> I would also like to confirm the block offset in the data structures refers the offset of the compressed block.

Yes, this is the offset to the beginning of the compressed block that contains the block. If ref/offset are equal for two blocks then they are in the same superblock.

PR #2011, opened 2023-02-26
Keep only one all-default group index.

dwsteele:
It is possible that a group index was created for an option that was later found to not meet dependencies. In this case all values will be default leading to a phantom group, which can be quite confusing. Remove all group indexes that are all default (except the final one) and make sure the key for the final all default group index is 1.

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -33,6 +33,19 @@
                         <p>Block-level incremental backup.</p>
                     </release-item>
                 </release-feature-list>
+
+                <release-improvement-list>
+                    <release-item>
+                        <github-pull-request id="2011"/>
+
+                        <release-item-contributor-list>
+                            <release-item-contributor id="david.steele"/>
+                            <release-item-reviewer id="stefan.fercot"/>
+                        </release-item-contributor-list>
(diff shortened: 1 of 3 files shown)


pgstef (approved):
Hi,

I ran my usual test suite on this branch and didn't saw anything obvious.

I also tried to play with it a little bit and the index seems to default correctly to repo1 now:

```bash
$ cat /etc/pgbackrest.conf
...
repo1-s3-key=test
repo3-type=s3
repo3-path=/repo3
repo3-s3-endpoint=xxx
repo3-s3-region=eu-west-2
repo3-s3-bucket=bucket
repo3-s3-key=xxx
repo3-s3-key-secret=xxx
...

$ pgbackrest info
...
full backup: 20230301-081500F
    ...
    repo1: backup set size: 12.9MB, backup size: 12.9MB
```

Compared to 2.44:

```bash
status: mixed
    repo1: error (missing stanza path)
    repo3: ok

full backup: 20230301-090631F
    ...
    repo3: backup set size: 12.9MB, backup size: 12.9MB
```


So overall, this PR looks good to me :+1:

PR #2012, opened 2023-02-27
Add pg-version-force option for fork integration.

pgstef:
- Stanza-create and stanza-upgrade
    Get the version and system information ultimately from `pgControlFromFile()`.
    However, `server_version_num` must match the `pgInterface[].version` selected with the `pg-version` option. So, changing `checkDbConfig()` to relax this condition.

- Archive-push and archive-get
    Relies on `pgControlFromFile()` or directly on the info written in the repo.
    Adapt `pgWalFromFile()` to avoid ArchiveMismatchError in `archivePushFile()`.

- Restore
    Only adding `pg-version` validity to let the restore_command use that option if specified in the command line.

I'm not sure it was necessary to allow using `pg-version` not matching `server_version_num` (which then need to change `pgWalFromFile`). It can probably work with only `pgControlFromFile` modified if we only want to limit this option to force pg_control and pg_catalog version.

If we want to go further, I'd have to adapt the functions in db/db.c so dbBackupStart,... will also use pg-version rather than the version queried. What would be better?

Thanks in advance for the thoughts and review,
Kind Regards


dwsteele:
I played around with this a bit and it seems to be heading in the right direction. I got some of the tests working just to make sure and updated the harness to support the tests (we'll probably want to commit that separately).

I wonder if we should consider setting the version in cfgLoadUpdateOption() rather than passing it every time (e.g. pgVersionSet() in interface.c). This would also be helpful if we start forcing the contents of pg_control as it seems the Greenplum folks need.

How about writing some tests for stanza-create/update to see how that works?

pgstef:
> I played around with this a bit and it seems to be heading in the right direction. I got some of the tests working just to make sure and updated the harness to support the tests (we'll probably want to commit that separately).

Thanks a lot for this

> How about writing some tests for stanza-create/update to see how that works?

I added some basic tests for it. I hope that's enough? 
Since we only pick the interface from pg-version and update the control-version (but everything else is still loaded from the pg_control file), I also updated the help message to reflect that info.

> I wonder if we should consider setting the version in cfgLoadUpdateOption() rather than passing it every time (e.g. pgVersionSet() in interface.c). This would also be helpful if we start forcing the contents of pg_control as it seems the Greenplum folks need.

The easiest way the better imho. Since the user asks to force a PG version, I'm in favor to force that value everywhere and not rely on server_version_num at all if we can :thinking:

pgstef:
> How about writing some tests for stanza-create/update to see how that works?

Reworked some tests in 28d7f4d, fed7b0c and ce000d2.
There's probably something simpler with HrnPgControl in the backup command test but at least it works.

Not sure why the code-format fail though :thinking: 
```
FAIL: pgbackrest/src/command/verify/verify.c (File size changed from 84346 to 84314)
FAIL: pgbackrest/src/postgres/interface/version.intern.h (File size changed from 11057 to 11049)
FAIL: pgbackrest/test/src/module/command/backupTest.c (File size changed from 221064 to 221218)
```

dwsteele:
> There's probably something simpler with HrnPgControl in the backup command test but at least it works.

I simplified the harness to make this less awkward.

> Not sure why the code-format fail though

This is new -- use `test.pl --code-format` to format your code before pushing to CI.

test/src/module/command/stanzaTest.c
```diff
@@ -512,6 +512,39 @@ testRun(void)
...
             "P00   INFO: stanza-create for stanza 'db' on repo1");
+
+        // Clean-up file in archive directory
+        HRN_STORAGE_REMOVE(storageRepoIdxWrite(0), STORAGE_REPO_ARCHIVE "/somefile", .comment = "remove old test file");
+
+        // -------------------------------------------------------------------------------------------------------------------------
+        TEST_TITLE("stanza-create forcing another PG version");
+
+        argList = strLstDup(argListBase);
+        hrnCfgArgRawZ(argList, cfgOptPgVersion, "10");
```
dwsteele, line 523: I think these tests would do better to write a v10 pg_control with a different control version. This tests reads a 9.6 pg_control file as 10, which seems to work, but is not what we are trying accomplish here.

pgstef: I initially thought it would be a good idea to check across PG version to see how it goes with extra fields in pg_control.
But you're right, it's best to keep it as clean as possible. Changed in ba1ba84 :-)

dwsteele:
>  I initially thought it would be a good idea to check across PG version to see how it goes with extra fields in pg_control.

Yeah, this looks better now. So, I think now we just need a test in backup, archive-push, archive-get, and verify.

pgstef:
> So, I think now we just need a test in backup, archive-push, archive-get, and verify.

Added for backup, archive-push and archive-get. I hope those ones are enough.
Still trying to figure out the best place for verify :thinking:

src/command/check/common.c
```diff
@@ -59,7 +59,9 @@ checkDbConfig(const unsigned int pgVersion, const unsigned int pgIdx, const Db *
         const String *dbPath = dbPgDataPath(dbObject);
 
         // Error if the version from the control file and the configured pg-path do not match the values obtained from the database
-        if (pgVersion != dbVersion || strCmp(cfgOptionIdxStr(cfgOptPgPath, pgIdx), dbPath) != 0)
+        // Skip the db version check if the pg-option is provided since we want to force that value
+        if ((pgVersion != dbVersion && !cfgOptionTest(cfgOptPgVersion)) ||
+            strCmp(cfgOptionIdxStr(cfgOptPgPath, pgIdx), dbPath) != 0)
```
dwsteele, line 64: It is not clear to me why we need this check. If the version was already forced then it should match, right?

dwsteele:
This is starting to look pretty good, but have a look at the check command comment.

I'm very tempted to rename this option to `pg-version-force` to make it a little clearer what is happening. If the wrong version is forced the results could be very unexpected. Pretty much the same as if pg_control or a WAL header was corrupted.

src/command/check/common.c
pgstef: Well, the `pgVersion != dbVersion` check was there before the pg-version option and checks if `server_version_num` matches the `pgInterface[].version`.

What we are doing with pg-version is forcing the `pgInterface[].version`. But the queried `server_version_num` might still not match (I don't know how other forks work with that number but it could be changed to match the fork version and not the PG version itself). We are not forcing 

I believe that check is important when we don't use the pg-version option. The skip part is there because we are not forcing `dbVersion`. If we did, then pgVersion and dbVersion would match indeed.

dwsteele: Hmmm, I'm not convinced. As I wrote in the docs, right now we are expecting the server to report the same version even if it has been forced. If we decide to allow the server version to be forced, we should put that in the db module rather than trying to figure out where it would need to be ignored.

pgstef: Oh, I see what you meant now.

> If we decide to allow the server version to be forced, we should put that in the db module rather than trying to figure out where it would need to be ignored.

Agreed. I'll remove the `&& !cfgOptionTest(cfgOptPgVersion)) ` part.

> As I wrote in the docs, right now we are expecting the server to report the same version even if it has been forced.

Hm, I might have missed it but that's not clear to me from the help content :thinking: 
`The specified <postgres/> version will be used instead of the version automatically detected` couldn't mean from a "beginner" point of view that we're already forcing the db part too?

pgstef:
Option renamed to pg-version-force and extra check removed ;-)

I however still think we might want to make it a bit more clearly in the docs that right now we are expecting the server to report the same version even if it has been forced. Just to avoid any confusion. Or would we want to keep that option internal so only fork maintainers who actually tested it would tell about it ? :thinking:

dwsteele:
> I however still think we might want to make it a bit more clearly in the docs that right now we are expecting the server to report the same version even if it has been forced.

Updated docs to reflect this.

> Or would we want to keep that option internal so only fork maintainers who actually tested it would tell about it ?

That seems like a good way to start, so I have marked the option internal.

PR #2019, opened 2023-03-04
Skip writing recovery.signal by default for restores of offline backups.

dwsteele:
When restoring an offline backup on PostgreSQL >= 12, skip writing `recovery.signal` by default since this will error if the backup was made with wal_level=minimal. If the user explicitly sets the type option to something other than none, then write `recovery.signal` as usual since it is possible to do PITR from an offline backup as long as wal_level was not minimal.


pgstef (approved):
Hi,

I've just tested to restore an offline backup with type=none (using this branch) and PG could indeed start without any other manual change. I then agree about the default restore type for offline backup.

I'm just a bit concerned about combining --no-online backup with --exclude=pg_wal. That could then easily lead to `PANIC:  could not locate a valid checkpoint record`. But I guess that's up to the user...

So overall this change looks good to me :+1: 

Shouldn't we document it somewhere too? Perhaps in the `--online` backup option description?

m-borger:
Hi,
I think the concern of @pgstef about the combining `--no-online`  and `--exclude=pg_wal` is quite valid, however not particularly related to this change. This combination can lead to this disaster regardless of the recovery.signal writing behavior every time the server hasn't shut down cleanly (I think). Maybe off-topic for this fix, but you might want to consider protecting users against this bad decision - perhaps by not allowing pg_wal to be excluded in case of `--no-online`? On the other hand, by doing this you are really explicitly asking for trouble...

m-borger:
Ah, I hadn't thought of that use case, I totally agree.

dwsteele:
> Shouldn't we document it somewhere too?

Docs added. I put them in the `type` option.

pgstef:
> Docs added. I put them in the `type` option.

Looks good to me :-)

PR #2044, opened 2023-04-13
fix: allow pd_upper to be zero in a checksummed page

pgguru:
When using pgBackRest with pages that utilize an encrypted page format, pgBackRest would on occasionally hit the case where `pd_upper` would be 0x0000 in the encrypted data.  The current code would skip checking the actual checksum to see if it matched and instead mark the block as invalid after the normal storage retry.

This commit changes the checksum calculations to compare against the computed checksum even when pd_upper is zero.

In passing, remove an unneeded copying of the page data, as the new and old routines both restored the original pd_checksum in the page header when they were called, so the comment was inaccurate and this copy did not need to be made.


pgguru:
Closing mine in lieu of the option-based solution.

PR #2047, opened 2023-04-18
Support IP-based SANs for TLS certificate validation.

pgguru:
The existing SAN code only recognizes DNS-based SANs, which means that it won't properly validate if using a SAN with an IP-based one. This PR adds support for IPv4, and if the OS supports it, IPv6. The current support is exact matching only.

This support would have simplified testing in a few occasions where many cert generation tools have trouble generating a DNS:1.2.3.4-style address, preferring to include the SAN as IP:1.2.3.4.

Needs tests, but this is POC of the feature.


pgguru:
And doing a little more research on this, looks like `inet_pton` can also map IPv4 addresses; we followed what Postgres does here as far as the IP SAN resolution, and they'd determined they needed additional flexibility in representation for IPv4 that `inet_aton()` supports.  I do wonder if that is actually needed though and if we could get away with `inet_pton(AF_INET)` for IPv4 as well; would simplify this feature.

dwsteele:
> And doing a little more research on this, looks like inet_pton can also map IPv4 addresses; we followed what Postgres does here as far as the IP SAN resolution, and they'd determined they needed additional flexibility in representation for IPv4 that inet_aton() supports.

It might be worth checking with the author of that patch so see what their motivation was. I don't have a strong opinion either way. If we use inet_aton we'll need to grab the configure/meson code that makes inet_aton/inet_pton work. Defining _GNU_SOURCE here is not going to work (yes, I get that this is temporary).

pgguru:
> > And doing a little more research on this, looks like inet_pton can also map IPv4 addresses; we followed what Postgres does here as far as the IP SAN resolution, and they'd determined they needed additional flexibility in representation for IPv4 that inet_aton() supports.
> 
> It might be worth checking with the author of that patch so see what their motivation was. I don't have a strong opinion either way. If we use inet_aton we'll need to grab the configure/meson code that makes inet_aton/inet_pton work. Defining _GNU_SOURCE here is not going to work (yes, I get that this is temporary).

Yeah, looks like it was Peter E; will shoot him an email.

dwsteele:
I think it looks reasonable to just use `inet_pton()`. It is in the Posix 2001 spec so should be supported on all the systems we test on. Give it a try and see how it goes, i.e. remove `HAVE_INET_PTON` and see if anything breaks without requiring a configure check.

pgguru:
Okay, tweaked that and fixed a few other things; looks like it's at least compiling without complaints, which is a start... 😁.  Will review after existing checks run.

dwsteele:
Yeah, looks like coverage is next. The main test cert will need to be rebuilt, see `test/certificate/README.md`. 

Then we'll need some new tests right around line 610 in `ioTlsTest.c`.

dwsteele (approved):
I added tests and updated the code to project style. Might be good to test this once more in a real env, if possible.

dwsteele:
I realized that treating the address as a strings in tlsClientHostVerifyIPAddr() was a pretty bad idea (certainly for logging) so I fixed that in 80abf6bf2c42. I also merged in the change to require OpenSSL >= 1.1.1.

pgguru:
I setup/verified handling w/IPv4 setup and was able to validate that we were not allowed to connect if the server's IP was not included in the cert it presented.

Full disclosure: I did get tired of fighting self-sign cert validations, so just commented out the check for validating the cert chain here: https://github.com/pgbackrest/pgbackrest/blob/f8b40f8003b4d32cd2d7eb13da8eb25247d9bd56/src/common/io/tls/client.c#L263

I also just validated that we passed the `tlsClientAuth()` portion, didn't get a full system setup due to more self-sign issues like:

```
2024-07-10 16:53:37.724 P00   WARN: unable to check pg1: [ServiceError] TLS error [1:167773208] tlsv1 alert unknown ca
```

That said, my testing validated the changes in this PR, so considering this sufficient for now.

** Methodology

I generated two `pgbackrest` builds, one for my patched version and one for the 2.52 with symlinked `/usr/local/bin/pgbackrest` on each of two VMs so I could compare the patched vs unpatched behaviors here.

I also generated 2 sets of certificates: one which included IP SAN entries for both nodes and one which included just the pgbackrest server node and not the Postgres node.  Validated the expected IP SANs by revie4wing the output of `openssl x509 -in cert.pem -noout -text` and looking for the appropriate SAN output, verifying it was IP and not DNS.

Set up a basic pgbackrest server and postgres server using largely the recipe from Stefan's blog article about the setup.  Tested using the unpatched pgbackrest and validated that we just were unable to connect when verification was required.

Switched to the patched pgbackrest; verified with the cert that did not include the postgres server that we got the expected `WARN: unable to check pg1: [CryptoError] unable to find hostname '192.168.122.16' in certificate common name or subject alternative names`.

I then swapped the cert on the postgres server to the ones that included, restarted pgbackrest server, and again ran the attempted `stanza-create` from the pgbackrest node, which progressed past the server host check and got a `WARN: unable to check pg1: [ServiceError] TLS error [1:167773208] tlsv1 alert unknown ca`, related to the self-signed cert and CA I was using for testing.

QED

pgguru:
In any case, I approve; since this is my PR and I'm not an authorized merger I guess I'll have to leave it to @dwsteele to do the merge itself.

dwsteele:
@pgguru Nice testing! This has now been committed.

PR #2051, opened 2023-04-25
PostgreSQL 16 Support.

dwsteele:
Add catalog version and WAL magic for PostgreSQL 16.

The GUC to force parallel mode has be renamed so update that in the tests.


pgstef (approved):
Looks good to me.
Ran some tests against the `postgresql16-server-16-alpha_20230425_PGDG.rhel8.1.x86_64` package and didn't spot anything obvious.
:+1:

Issue #2028, opened 2023-03-16
Feature request: expose S3_STS_HOST_STR as a configurable environment variable


oadekoya:
While trying to configure pgbackrest to use AWS-integrated identity pprovider, we found out that the STS endpoint is hard-coded as **sts.amazonaws.com** [here](https://github.com/pgbackrest/pgbackrest/blob/7e5adc03594a3b7a975001e80902a51df3ddd577/src/storage/s3/storage.c#L388) and that prevents us from using a regional endpoint that resolves to private IP (in an airgapped cluster environment). We even tried to do a manual DNS re-write but the certificate validation failed since pgbackrest expects the CN name to match exactly with the sts endpoint hard-coded. 

So, is it possible to expose the S3 STS credential host such that its configurable for different use cases?

dwsteele:
It is not clear to me from the documentation (https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRoleWithWebIdentity.html) that another host can be substituted here.

Have you tried changing the constant and rebuilding to see if it works?

oadekoya:
STS can be isolated to a specific region and that's our use case. We could modify the code and make the constant configurable but we didn't want to be managing the overhead of keeping up with the upstream considering the crunchydata operator we are using also requires building the containers.

dwsteele:
I was only suggesting that you try it once to be sure it works before we look at making this configurable.

PR #2058, opened 2023-05-02
Use STS regional endpoints when configured as web-id

brenix:
This is an effort to solve https://github.com/pgbackrest/pgbackrest/issues/2028.

When using `web-id` repo type, this will check the presence of `AWS_STS_REGIONAL_ENDPOINTS`. If this is set, then it will further check that the `AWS_REGION` is set as well, then set the endpoint to the regional hostname. If the variable is not set, then we fallback to the original endpoint of `sts.amazonaws.com`.

Unfortunately, we are unable to test this easily as it seems that all of the crunchydata images have pgbackrest under various locations. It has been extremely difficult to build the pgbackrest binary statically in order to replace it in the various crunchydata images. Therefore we had to rebuild one from scratch, but then realized that pgbackrest is placed in the various locations.

I am not well versed in C, nor the test framework. Therefore I welcome any changes to the code to make this work/pass tests. Any support is appreciated.

### DIFF
--- a/src/storage/s3/storage.c
+++ b/src/storage/s3/storage.c
@@ -4,6 +4,7 @@ S3 Storage
 #include "build.auto.h"
 
 #include <string.h>
+#include <stdlib.h>
 
 #include "common/crypto/hash.h"
 #include "common/debug.h"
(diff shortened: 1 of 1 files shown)


dwsteele:
Hmmm, nevermind -- looks like these endpoints need to be explicitly enabled by account. So it can't just be enabled by default.

brenix:
> Hmmm, nevermind -- looks like these endpoints need to be explicitly enabled by account. So it can't just be enabled by default.

Ah, good catch - This would mean the recent updates I made today would not work well as a default.

We are going to drop this effort for now as we have other plans and priorities at the moment. Therefore I am going to close this PR for now, but would consider revisiting it in the future if the need resurfaces.

I appreciate the quick feedback and assistance on this!

gricey432:
AWS have changed the routing behaviour for the default domain to route to your source region, which might reduce the need for this. It also handles the 'not enabled' problem for us.

https://aws.amazon.com/about-aws/whats-new/2025/04/aws-sts-global-endpoint-requests-locally-regions-default/

I know this PR is old but it comes up in search + this change would be good to know if anyone thought about revisiting this in the future.

PR #2069, opened 2023-05-16
Add warning about using recovery type=none.

dwsteele:


dwsteele:
> This will bring confusion and can i.e. easily lead to...?

I was thinking that "disk fills up and PG panics" was sufficient to discourage people. What did you have in mind here?

> This could also prevent pg_rewind to work too so there's some serious drawbacks doing this 

Updated to add pg_rewind.

pgstef:
> What did you have in mind here?

Blocked archiving is pretty easy to solve on test systems (push to another repo, turn off archive_mode, archive-push-queue-max...).
But having the possibility to write again a complete history on the same timeline (between the end of the backup and now for example) without having any possibility to distinguished the data/wals generated on each cluster feels way scarier to me.

The updated comment looks good enough to me though.
So :+1: for this PR ;-)

pgstef (approved):

PR #2070, opened 2023-05-16
Skip recovery.signal for PostgreSQL >= 12 when recovery type=none.

dwsteele:
Bring PostgreSQL >= 12 behavior in line with other versions when recovery type=none.

We are fairly sure this did not work correctly when PostgreSQL 12 was released, but apparently the issue has been fixed since then. Either way, after testing we have determined that the behavior is now as expected.


pgstef (approved):
The suggested code change looks good to me based on our investigations in the gh issue.

But I'm wondering, couldn't we add an extra "real" restore test to detect any potential behavior change in PG? :thinking:

PR #2076, opened 2023-05-25
Error when pull request target branch is not integration.

cbandy:
[Submitting a Pull Request](https://github.com/pgbackrest/pgbackrest/blob/main/CONTRIBUTING.md#submitting-a-pull-request) says to open PRs against the `integration` branch. This PR check fails any PR opened against the wrong branch. You can see how it looked when it [ran in my fork](https://github.com/cbandy/pgbackrest/actions/runs/5076020716?pr=1).

### DIFF
--- a/.github/workflows/target-branch.yml
+++ b/.github/workflows/target-branch.yml
@@ -1,0 +1,31 @@
+name: target-branch
+
+# Run this workflow *in the context of the target branch* when a pull request
+# is opened, reopened, or the target branch is changed to anything other than
+# the integration branch.
+#
+# NOTE: Pull requests that change this file do not run those changes during the
+# "pull_request_target" event.
+on:
+  pull_request_target:
+    types: [synchronize, opened, reopened, edited]
+    branches-ignore:
+      - integration
+
+jobs:
+  check:
+    # Only run this job on the main pgBackRest repository. Forks can do what
+    # they want with branches.
+    if: github.repository == 'pgbackrest/pgbackrest'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+
+      # Show pull request instructions on the workflow run page, print an error
+      # to the job log, and exit non-zero to fail the pull request check.
+      - name: Incorrect Base Branch
+        run: |-
+          >> "${GITHUB_STEP_SUMMARY}" cat .github/pull_request_template.md
+          echo '::error ::Please change the base branch of this pull request'
+          exit 1
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -73,6 +73,19 @@
                     </release-item>
                 </release-improvement-list>
             </release-doc-list>
+
+            <release-test-list>
+                <release-improvement-list>
+                    <release-item>
+                        <release-item-contributor-list>
+                            <release-item-contributor id="chris.bandy"/>
+                            <release-item-reviewer id="david.steele"/>
+                        </release-item-contributor-list>
+
+                        <p>Error when pull request target branch is not integration.</p>
+                    </release-item>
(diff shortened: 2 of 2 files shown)


.github/workflows/target-branch.yml
```diff
@@ -0,0 +1,29 @@
...
+
+# Run this workflow *in the context of the main branch* when a pull request is
+# opened or reopened targeting the main branch.
+#
+# NOTE: Pull requests that change this file do not run those changes during the
+# "pull_request_target" event.
+on:
+  pull_request_target:
+    branches:
+      - main
```
cbandy, line 11: This differs slightly from the instructions, which say, "do not select `main` nor any other branch." I believe we can match that here [using `branches-ignore`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpull_requestpull_request_targetbranchesbranches-ignore):

```yaml
    branches-ignore:
      - integration
```

.github/workflows/target-branch.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  check:
+    # Only run this job on the main pgBackRest repository. Forks can do what
+    # they want with branches.
+    if: github.repository == 'pgbackrest/pgbackrest'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
```
cbandy, line 21: This is only necessary to `cat .github/pull_request_template.md` in the following step.

.github/workflows/target-branch.yml
dwsteele: Yeah, seems like this would be the way to go.

However, I have not found any reasonable way to test this yet. I tried a fork and it didn't work out well -- maybe I need to create an independent repo to trick it -- basically the problem is I can never get this workflow to run.

cbandy: The [`pull_request_target` event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) is slightly different from the typical `pull_request` event. In general, it does things according to the _base_ or _target_ branch of the pull request, rather than the contents of the pull request itself.

The benefit is that it runs according to what is already committed, and there's no way for someone to submit a pull request that bypasses it. The drawback is that there's no way to trigger it until _after_ it is committed.

I'll make the change and figure out the behavior on my fork. I suspect that the workflow definition needs to be committed to the `main` branch before PRs to the `main` branch will be rejected.

dwsteele: > I suspect that the workflow definition needs to be committed to the main branch before PRs to the main branch will be rejected.

That's pretty much what I found -- I think... It can always be patched if needed. It wouldn't be the first bad thing we committed to main.

But if you can get it as right looking as possible that would be helpful. Thanks!

cbandy: I tried some combinations in https://github.com/cbandy/pgbackrest/pull/2. The target branch (`main`) needs the workflow committed.

cbandy:
Rebased. I found that the default triggers do not include the `edited` event, which [occurs when someone re-targets](https://github.com/actions/runner/issues/980#issuecomment-777057271) a PR. The `types` field catches that now, so this should be good to go.

dwsteele (approved):

dwsteele:
Unfortunately this did not work well so I backed it out. The problem was that if the PR got into an error state because the target branch was wrong it would stay that way even after the branch was changed. This means we'd have to merge PRs that did not have all passing tests.

I think I'd rather just leave this alone. Github has beefed up the branch protections a bit, so PRs with a target branch of main cannot be merged since not all the tests will run and the branch can't be merged without the tests. It used to be that admins bypassed this check but it seems that Github has fixed that.

Thanks for the effort, Chris!

PR #2077, opened 2023-05-25
Update NEWS.md for new version and features.

dwsteele:


doc/BLOG.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+This time there are far fewer files. The small files have been bundled together and zero-length files are stored only in the manifest.
+
+The `repo-bundle-size` option can be used to control the maximum size of bundles before compression and other operations are applied.
+The `repo-bundle-limit` option limits the files that will be added to bundles. It is not a good idea to set these options too large because any failure in the bundle on backup or restore will require the entire bundle to be retried. The goal of file bundling is to combine small files -- there is very seldom any benefit in combining larger files.
+
+## Block Incremental Backup
+
+Block incremental backup saves space in the repository by storing only the parts of the file that have changed since the last backup. The block size depends the file size and when the file was last modified, i.e. larger, older files will get larger block sizes. Blocks are compressed and encrypted into super blocks that can be retrieved independently to make restore more efficient.
```
pgstef, line 69: `The block size depends ON the file size...`

doc/NEWS.md
```diff
@@ -1,15 +1,25 @@
-**February 10, 2022**: [Crunchy Data](https://www.crunchydata.com) is pleased to announce the release of [pgBackRest](https://pgbackrest.org/) 2.37, the latest version of the reliable, easy-to-use backup and restore solution that can seamlessly scale up to the largest databases and workloads.
+**May !!!, 2022**: [Crunchy Data](https://www.crunchydata.com) is pleased to announce the release of [pgBackRest](https://pgbackrest.org/) 2.46, the latest version of the reliable, easy-to-use backup and restore solution that can seamlessly scale up to the largest databases and workloads.
```
pgstef, line 1: May 2023? :-)

doc/BLOG.md
dwsteele: Fixed in b7ec9f102fb.

doc/NEWS.md
dwsteele: This is not ready for review yet, but I did update the year so I don't forget to. :)

doc/BLOG.md
```diff
@@ -0,0 +1,140 @@
...
+    database size: 995.7MB, database backup size: 995.7MB
+    repo2: backup size: 56MB
+
+incr backup: 20230520-082438F_20230520-083027I
+    database size: 995.7MB, database backup size: 972.8MB
+    repo2: backup size: 943.3KB
+```
+Block incremental also improves the efficiency of the restore command. For example, on repo1 restoring the full backup would require reading the entire 30.4MB (compressed) `base/13427/16501` file, but with block incremental only 3.MB of blocks are required to restore the file from the full backup.
+
+It is best to avoid long chains of incremental backups since they can have a negative impact on restore performance. In this case pgBackRest may be forced to pull from many backups to restore a file.
```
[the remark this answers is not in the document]
dwsteele: Fixed in 240b34e6.

doc/BLOG.md
```diff
@@ -0,0 +1,140 @@
...
+## Block Incremental Backup
+
+Block incremental backup saves space in the repository by storing only the parts of the file that have changed since the last backup. The block size depends on the file size and when the file was last modified, i.e. larger, older files will get larger block sizes. Blocks are compressed and encrypted into super blocks that can be retrieved independently to make restore more efficient.
+
+To demonstrate block incremental we need to make some changes to the database. With `pgbench` we can update 100 random rows in the main table, which is about 1GB in size.
+```
+$ /usr/lib/postgresql/12/bin/pgbench -n -b simple-update -t 100
+```
+
+On repo1 the time to make a incremental backup is very similar to making a full backup. That's because most of the data is contained in a single table, so when the table is updated it must be copied in full.
```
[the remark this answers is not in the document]
dwsteele: I have tried to make this much more explicit.

doc/BLOG.md
```diff
@@ -0,0 +1,140 @@
...
+
+full backup: 20230520-082438F
+    database size: 995.7MB, database backup size: 995.7MB
+    repo2: backup size: 56MB
+
+incr backup: 20230520-082438F_20230520-083027I
+    database size: 995.7MB, database backup size: 972.8MB
+    repo2: backup size: 943.3KB
+```
+Block incremental also improves the efficiency of the restore command. For example, on repo1 restoring the full backup would require reading the entire 30.4MB (compressed) `base/13427/16501` file, but with block incremental only 3.MB of blocks are required to restore the file from the full backup.
```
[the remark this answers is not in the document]
dwsteele: > we don’t have to spend any time reading the existing file if this is a delta restore (most likely)

Even normal delta restore requires a full scan of the file to generate a SHA1 checksum (unless the file is smaller than the repo size). Delta restore for block incremental adds the block checksums but they are far faster. So even in my env where IO latency is very low block incremental saves some time on restore.

>  maybe also include actual restore runs and timing results, etc, to demonstrate?

Good idea, done.

doc/BLOG.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+And then perform a delta restore back to the full backup in repo 2:
+```
+pgbackrest --stanza=demo --delta --repo=2 --set=20230526-053406F restore
+
+<...>
+INFO: restore command end: completed successfully (1536ms)
+```
+This is noticeably faster even on our fairly small demo database. When storage latency is high (e.g. S3) the performance improvement will be more pronounced. With block incremental enabled delta restore only had to copy 3.5MB of the main table file from repo 2, as compared to 30.4MB from repo 1.
```
cbandy, line 142:
```suggestion
This is noticeably faster even on our fairly small demo database. When storage latency is high (e.g. S3) the performance improvement will be more pronounced. With block incremental enabled, delta restore only had to copy 3.5MB of the main table file from repo 2, as compared to 30.4MB from repo 1.
```

dwsteele: I was on the fence about the comma, so this works for me.

doc/BLOG.md
```diff
@@ -0,0 +1,166 @@
...
+```
+$ pg_ctlcluster 12 demo stop
+$ pgbackrest --stanza=demo --delta --repo=1 --set=20230526-053458F restore
+
+<...>
+INFO: restore command end: completed successfully (3697ms)
+```
+As we saw above the main table is contained in a single file, so the restore must copy and decompress the entire file from repo 1 (compressed size 30.4MB) because it was changed since the full backup.
+
+To test the full backup restore in repo 2 we need to first restore the cluster to the most recent backup in repo 2:
```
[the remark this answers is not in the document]
dwsteele: Well, I mean to test a delta restore of the the full backup, but it is obviously not clear. How about d4e66e3?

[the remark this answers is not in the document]
sfrost: Yup, that's what I was trying to say.  Thanks.

sfrost (approved):

sfrost (approved):

PR #2089, opened 2023-06-09
Improve performance of SFTP storage driver.

sfrost:
The initial implementation used simple waitNew's when having to loop due to getting a LIBSSH2_ERROR_EAGAIN, but we don't want to just wait some amount of time, we want to wait until we're able to read or write on the fd that we would have blocked on.

This change removes all of the wait code from the sftp driver and changes the loops to call the newly introduced storageSftpWaitFd() which in turn checks with libssh2 to determine the appropriate direction to wait on (read, write, or both) and then calls fdReady() to perform the wait, using the provided timeout (which, if hit, will throw an exception which may be caught above and retried).

This also removes the need to pass ioSession or timeout down into the Sftp read and write code, so remove those.


sfrost (commented):
Looks good to me, thanks.

dwsteele (approved):
I did a bit of cleanup. If that all looks good this is ready to merge.

dwsteele:
Got confirmation from @sfrost that the final changes look OK. Thank you @sfrost and @jreidthompson!

PR #2096, opened 2023-06-19
Fix GCC strict aliasing issues

red1452:
There is error on Ubuntu 18.04 with GCC 7.4 and ppc64le at building:
```
gcc -std=gnu11   -std=c99 -Wall -Wextra -Wno-missing-field-initializers -Wno-implicit-fallthrough -Wno-clobbered -Wwrite-strings -Wconversion -Wformat=2 -Wformat-nonliteral -Wstrict-prototypes -Wpointer-arith -Wvla -Werror -Wfatal-errors -g -D_POSIX_C_SOURCE=200809L -I. -I/usr/lib/gpdb/include -I/usr/include/libxml2   -I. -I. -c -o .build/common/crypto/xxhash.o common/crypto/xxhash.c -MMD -MP -MF .build/common/crypto/xxhash.dep
In file included from common/crypto/xxhash.c:17:0:
./common/crypto/xxhash.vendor.c.inc: In function 'XXH3_accumulate_512_vsx':
./common/crypto/xxhash.vendor.c.inc:4148:9: error: invalid parameter combination for AltiVec intrinsic
         xxh_u64x2 acc_vec        = vec_xl(0, xacc + 2 * i);
         ^
compilation terminated due to -Wfatal-errors.
make: *** [.build/common/crypto/xxhash.o] Error 1
```
This commit is adapted and applyed from
https://github.com/Cyan4973/xxHash/commit/ad63946825f89c69ffac3861d8156fb2f10b5e17

This commit fixes the same problem with build on ppc64le:

This fixes GCC 12 on macOS ARM64, as well as the existing s390x zVector bug that was worked around with a `__sync_synchronize()` by using `__attribute__((may_alias))` which excludes a type from strict aliasing optimizations.

`vldq`/`vst1q` are still avoided due to older versions having issues with load-store optimizations.

This is not needed on x86 because the SSE types are impossible to not alias.

### DIFF
--- a/src/common/crypto/xxhash.vendor.c.inc
+++ b/src/common/crypto/xxhash.vendor.c.inc
@@ -2908,6 +2908,12 @@ enum XXH_VECTOR_TYPE /* fake enum */ {
 #  define XXH_SEC_ALIGN 8
 #endif
 
+#if defined(__GNUC__) || defined(__clang__)
+#  define XXH_ALIASING __attribute__((may_alias))
+#else
+#  define XXH_ALIASING /* nothing */
+#endif
+
 /*
  * UGLY HACK:
  * GCC usually generates the best code with -O3 for xxHash.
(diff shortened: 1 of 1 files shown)


dwsteele:
I tested this by merging this change into the xxHash v0.8.1 release and running the tests: https://github.com/dwsteele/xxHash/actions/runs/5343723524. Ubuntu 18.04 tests don't run anymore since the OS is deprecated, but not too worried out that (anyway they were all Intel tests which were not changed by this patch).

I then introduced an error to make sure tests would fail (at least for architectures modified by this patch): https://github.com/dwsteele/xxHash/actions/runs/5343824809

I'm not super-excited about this change since I would prefer to be running an official released version of xxHash, but https://github.com/Cyan4973/xxHash/issues/843 indicates that a release may be some time away.

Thoughts on this @sfrost?

dwsteele:
It seems inaction sometimes pays off! There has been a new release of xxHash (which should fix these issues) and I have merged it into main (6c75c99d0).

Can you test from main and see if your issues are resolved?

dwsteele:
Closing this since it should be resolved. If you still have an issue we can reopen it.

red1452:
I checked current version of pgbackrest (commit af4621894a20ca0333726d459e2e6409bd4ee1e0): There is not any problem at build on ubuntu 18.04 and ppc64le. Thank you!

Issue #1873, opened 2022-09-16
incr backup references latest backup label of an expired backup in a check and fails


ikzelf:
1. pgBackRest version: 2.40
2. PostgreSQL version: 12.1
3. Operating system/version: rhel-8.6
4. Did you install pgBackRest from source or from a package? yum package
5. conf + log extract in [pgbackrest-2.40-expired.log](https://github.com/pgbackrest/pgbackrest/files/9583177/pgbackrest-2.40-expired.log)

7. Describe the issue:

We did a 
full       20220915-130608F
full       20220915-130650F
incr      20220915-130650F_20220915-130846I
expire  20220915-130650F 
incr      fail

2022-09-16 08:13:07.943 P00  ERROR: [106]: new backup label '20220915-130608F_20220916-081305I' is not later than latest backup label '20220915-130650F_20220915-130846I'

it looks like pgbackrest-2.40 has some memory about the -manually- expired incremental backup that was part of 20220915-130650F 
2022-09-15 13:25:50.529 P00   INFO: repo1: remove expired backup 20220915-130650F_20220915-130846I
2022-09-15 13:25:50.642 P00   INFO: repo1: remove expired backup 20220915-130650F
The new incr backup is correctly based on the full of 0220915-130608F but a consistency check goes wrong.

I think this is a bug. In case this is not a bug, please educate me.

dwsteele:
Looks like a bug to me as well, but not sure how to go about fixing it yet. It could be that we would ignore the history in this case but that may have its own pitfalls.

For now you can make a new full backup to get past this issue.

ikzelf:
Thanks for the confirmation. I expected that the info returned by the info command would be leading here .... That info seems to be correct.

dwsteele:
Info should be correct. This has to do with the manifests stored in `backup.history`. We expect backup labels to be ascending (even vs history), but the introduction of adhoc expire has introduced a bit of an issue with that.

Really not sure how to go about fixing this, but we'll definitely need to review past changes in this area before proceeding.

ikzelf:
In the backup.history I see that the manifest files of the expired backups are available. All of them (since start usage of pgbackrest?)
-rw-r-----. 1 pgbackrest pgbackrest 65431 Sep 12 08:13 20220912-081305F.manifest.gz
-rw-r-----. 1 pgbackrest pgbackrest 65223 Sep 13 08:13 20220912-081305F_20220913-081304I.manifest.gz
[...]

PR #2104, opened 2023-06-28
Fix unique label generation for diff/incr backup.

andr-sokolov:
If there were at least 2 full backups and then the last one was expired, then it was impossible to create either a differential or an incremental backup. The backupLabelCreate function identified this situation as clock skew, because the new backup label was compared with label of the expired full backup. If the new backup is differential or incremental, then its label should be compared with labels of differential or incremental backups related to the same full backup.
Replace numbers with macro.
https://github.com/pgbackrest/pgbackrest/issues/1873


dwsteele (approved):
Looks good. Thank you for the fix!

PR #2109, opened 2023-07-05
Fix time-based archive expiration when no backups are expired.

dwsteele:
In the case that no backups were expired but time-based retention was met no archive expiration would occur and the following would be logged:

INFO: time-based archive retention not met - archive logs will not be expired

In most cases this was harmless, but when retention was first met or if retention was increased, it would require one additional backup to expire earlier WAL. After that expiration worked as normal.

Even once expiration was working normally the message would continue to be output, which was pretty misleading since retention had been met, even though there was nothing to do.

Bring this code in line with count-based retention, i.e. always log what should be expired at detail level (even if nothing will be expired) and then log info about what was expired (even if nothing is expired). For example:

DETAIL: repo1: 11-1 archive retention on backup 20181119-152138F, start = 000000010000000000000002
    INFO: repo1: 11-1 no archive to remove

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -41,6 +41,18 @@
 
                         <p>Fix unique label generation for diff/incr backup.</p>
                     </release-item>
+
+                    <release-item>
+                        <github-issue id="2103"/>
+                        <github-pull-request id="2109"/>
+
+                        <release-item-contributor-list>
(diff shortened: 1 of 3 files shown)


pgstef (approved):
Looks good to me :+1:

dwsteele:
> Looks good to me

Thanks, @pgstef!

PR #2120, opened 2023-07-12
Fix block incremental file names in verify command.

dwsteele:
The verify command was not appending the .pgbi extension instead of the compression extension when verifying block incremental files stored outside a bundle.

Originally the idea was that verify would not need any changes (since it just examines repo-size and checksum) but at some point the new extension was added and broke that assumption.

Use backupFileRepoPathP() to generate the correct filename (Just like backup, restore, etc).

### DIFF
--- a/doc/xml/release.xml
+++ b/doc/xml/release.xml
@@ -18,6 +18,19 @@
             <release-core-list>
                 <release-bug-list>
                     <release-item>
+                        <github-issue id="2112"/>
+                        <github-pull-request id="2120"/>
+
+                        <release-item-contributor-list>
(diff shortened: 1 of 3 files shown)


jreidthompson (commented):
OK I've done a first run through this and it looks good to me.                                                                      
Setup and ran local debug tests with original code, noted invalid filename/failure of verify.                                        
Setup and ran local debug tests with PR code, noted valid filename/success of verify.

jreidthompson (approved):
All looks good to me.

dwsteele:
Thank you, @jreidthompson!

PR #2125, opened 2023-07-20
Add known hosts checking for SFTP storage driver.

jreidthompson:
Add an option to require a known_hosts match as part of the SFTP module's libssh2 authentication process. The match is expected to be found in one of the default know hosts files  $HOME/.ssh/known_hosts, $HOME/.ssh/known_hosts2, /etc/ssh/ssh_known_hosts, /etc/ssh/ssh_known_hosts2 or in a list of known hosts files provided by the user. Match handling should mimic ssh_config StrictHostKeyChecking.


src/common/user.c
```diff
@@ -24,6 +24,7 @@ static struct
 
     gid_t groupId;                                                  // Real group id of the calling process from getgid()
     const String *groupName;                                        // Group name if it exists
+    const String *userHome;                                         // User HOME directory
```
sfrost, line 27: Should this also be under a #ifdef HAVE_LIBSSH2 similar to the other code..?  Otherwise some part of the code might try to use it without being compiled with HAVE_LIBSSH2 and it'd be uninitialized...

src/common/user.c
```diff
@@ -114,6 +118,34 @@ groupNameFromId(gid_t groupId)
     FUNCTION_TEST_RETURN(STRING, NULL);
 }
 
+#ifdef HAVE_LIBSSH2
```
sfrost, line 121: Would be good to comment that we only use this when building with libssh2 currently, at least imv, so people aren't wondering why this seemingly independent thing is under this #ifdef.

src/storage/sftp/storage.c
```diff
@@ -877,6 +878,102 @@ storageSftpNew(
             }
         }
 
+        // Check for a matching entry in the .ssh/known_hosts file
+        if (param.requireKnownHostsMatch)
+        {
+            LIBSSH2_KNOWNHOSTS *nh;
```
sfrost, line 884: "nh" seems like a pretty terrible variable name for this...

src/build/help/help.xml
```diff
@@ -1085,6 +1085,16 @@
                         <example>~/.ssh/id_ed25519.pub</example>
                     </config-key>
 
+                    <config-key id="repo-sftp-req-known-hosts-match" name="Require a known_hosts match for the SFTP host">
```
sfrost, line 1088: Instead of a yes/no flag and a hard-coded .ssh/known_hosts file, we should be following closer to what OpenSSH does and instead have an option where the user can tell us of the known_hosts files to use, and then also have a default which will automatically be used if nothing has been specified.

eg: repo-sftp-known-hosts-file="/etc/ssh/ssh_known_hosts /etc/ssh/ssh_known_hosts2 ~/.ssh/known_hosts ~/.ssh/known_hosts2"

would be the default, while explicitly specifying the file is supported in the same way and specifying it but leaving it blank would disable the feature.

Also- we don't need both this and the explicit host fingerprint.  The explicit fingerprint should be used if it's set and perhaps we should throw a warning if a known hosts file and an explicit fingerprint are configured because it could lead to confusing results.  In other words- if *both* are explicitly set, then throw a warning or error saying that it doesn't make sense to have them both set.  If *only* the fingerprint is set, then use only that and treat the known hosts file list as empty.  If *only* the known host file list is set, then use that and we shouldn't require the host fingerprint to be set (I don't think we do, but just trying to be clear).

sfrost: If *neither* are set, then we should use the default known_hosts files that I included in the above comment, matching how OpenSSH works today.

sfrost (changes requested):

src/common/user.c
jreidthompson: It should. I'll make it so.

src/common/user.c
jreidthompson: Will add a comment to note this.

src/storage/sftp/storage.c
jreidthompson: :slightly_frowning_face: you are correct, that was a straight carryover from the example code w/o thinking on my part. Will replace with a descriptive name.

src/build/help/help.xml
jreidthompson: Implemented in commit ccacee0b3640f9fa173b70190540927695b911bd. I believe that the only discrepancy is that repo-sftp-known-hosts-files can't be configured as empty to disable, so implemented as an explicit disable repo-sftp-known-hosts-files=disable

src/storage/sftp/storage.c
```diff
@@ -850,16 +905,28 @@ storageSftpNew(
...
                 break;
         }
 
-        const char *binaryFingerprint = libssh2_hostkey_hash(this->session, hashType);
+        const char *const binaryFingerprint = libssh2_hostkey_hash(this->session, hashType);
 
         if (binaryFingerprint == NULL)
             THROW_FMT(ServiceError, "libssh2 hostkey hash failed: libssh2 errno [%d]", libssh2_session_last_errno(this->session));
 
+        // Is known_hosts search enabled
```
jreidthompson, line 917: Will update this comment to
// Enable known_hosts search unless param.knownHostsFiles (--repo-sftp-known-hosts-files) is explicitly set to disable it. NULL results in searching a default known_hosts file list.

jreidthompson: Updated in aab080ae3528d998b46ab7b6905d1fb9e8083519

src/storage/sftp/storage.c
```diff
@@ -37,6 +38,199 @@ struct StorageSftp
...
+            // successfully loaded hosts or a negative value on error, an empty known hosts file will return 0.
+            if ((rc = libssh2_knownhost_readfile(userKnownHostsList, userKnownHostsFile, LIBSSH2_KNOWNHOST_FILE_OPENSSH)) < 0)
+            {
+                // Missing known_hosts file will return LIBSSH2_ERROR_FILE. Possibly issues other than missing may return this.
+                if (rc == LIBSSH2_ERROR_FILE)
+                {
+                    // If user's known_hosts file is non-existant, create an empty one for libssh2 to operate on.
+                    Storage *sshStorage =
+                        storagePosixNewP(
+                            strNewFmt("%s%s", strZ(userHome()), "/.ssh"), .modeFile = 0644, .modePath = 0700, .write = true);
```
sfrost, line 177: We should probably be doing modeFile = 0600, right?  That seems to be what ssh does when it creates a new .ssh and .ssh/known_hosts file.

src/storage/sftp/storage.c
```diff
@@ -37,6 +38,199 @@ struct StorageSftp
...
+    MEM_CONTEXT_TEMP_BEGIN()
+    {
+        LIBSSH2_KNOWNHOSTS *userKnownHostsList;
+        char *libSsh2ErrMsg;
+        int libSsh2ErrMsgLen;
+        int rc;
+
+        // Init a known host collection for the user's known_hosts file
+        const char *const userKnownHostsFile = strZ(strNewFmt("%s%s", strZ(userHome()), "/.ssh/known_hosts"));
+        userKnownHostsList = libssh2_knownhost_init(this->session);
```
sfrost, line 153: Callers to this function have already done an init, why aren't we passing down that into this function and using it..?

src/storage/sftp/storage.c
```diff
@@ -37,6 +38,199 @@ struct StorageSftp
...
+
+                LOG_WARN_FMT(
+                    "libssh2 unable to read '%s' for update: libssh2 errno [%d] %s\n"
+                    "HINT: does '%s' exist with proper permissions?", userKnownHostsFile, rc, libSsh2ErrMsg, userKnownHostsFile);
+            }
+        }
+
+        // Free the user's known hosts list
+        if (userKnownHostsList)
+            libssh2_knownhost_free(userKnownHostsList);
```
sfrost, line 227: Seems like all callers init their own and will handle free'ing it too, so maybe we should just use that one instead?  See earlier comment.

sfrost (commented):
Couple of minor comments but generally looking pretty good.

src/build/help/help.xml
```diff
@@ -1057,6 +1057,18 @@
                         <example>pg-backup</example>
                     </config-key>
 
+                    <config-key id="repo-sftp-known-hosts" name="SFTP known hosts file">
+                        <summary>a known hosts file to search.</summary>
+
```
dwsteele, line 1062: "Known hosts file to search." is fine here.

src/storage/sftp/storage.c
```diff
@@ -37,6 +38,199 @@ struct StorageSftp
...
+        case LIBSSH2_KNOWNHOST_CHECK_NOTFOUND:
+            matchFailMsg = "not found in known hosts files: LIBSSH2_KNOWNHOST_CHECK_NOTFOUND";
+            break;
+
+        case LIBSSH2_KNOWNHOST_CHECK_MISMATCH:
+            matchFailMsg = "mismatch in known hosts files: LIBSSH2_KNOWNHOST_CHECK_MISMATCH";
+            break;
+
+        default:
+            matchFailMsg = "experienced an unknown failure:";
```
dwsteele, line 122: How about "unknown failure" here? Not sure we need the : at the end (it doesn't really fit with the eventual formatting).

src/storage/sftp/storage.c
```diff
@@ -67,7 +261,7 @@ storageSftpLibSsh2SessionFreeResource(THIS_VOID)
                 THROW_FMT(
                     ServiceError, "failed to close sftpHandle: libssh2 errno [%d]%s", rc,
                     rc == LIBSSH2_ERROR_SFTP_PROTOCOL ?
-                        strZ(strNewFmt(": sftp errno [%lu]", libssh2_sftp_last_error(this->sftpSession))) : "");
+                    strZ(strNewFmt(": sftp errno [%lu]", libssh2_sftp_last_error(this->sftpSession))) : "");
```
dwsteele, line 264: Bring back indent.

src/storage/sftp/storage.c
```diff
@@ -88,7 +282,7 @@ storageSftpLibSsh2SessionFreeResource(THIS_VOID)
                 THROW_FMT(
                     ServiceError, "failed to shutdown sftpSession: libssh2 errno [%d]%s", rc,
                     rc == LIBSSH2_ERROR_SFTP_PROTOCOL ?
-                        strZ(strNewFmt(": sftp errno [%lu]", libssh2_sftp_last_error(this->sftpSession))) : "");
+                    strZ(strNewFmt(": sftp errno [%lu]", libssh2_sftp_last_error(this->sftpSession))) : "");
```
dwsteele, line 285: Bring back indent.

src/storage/sftp/storage.c
```diff
@@ -99,7 +293,7 @@ storageSftpLibSsh2SessionFreeResource(THIS_VOID)
     {
         do
         {
-            rc = libssh2_session_disconnect_ex(this->session, SSH_DISCONNECT_BY_APPLICATION, "pgbackrest instance shutdown", "");
+            rc = libssh2_session_disconnect_ex(this->session, SSH_DISCONNECT_BY_APPLICATION, "pgBackRest instance shutdown", "");
```
dwsteele, line 296: Use PROJECT_NAME define here.

src/storage/sftp/storage.c
```diff
@@ -897,6 +1146,140 @@ storageSftpNew(
                     fingerprint, strZ(param.hostFingerprint));
             }
         }
+        else
+        {
+            LIBSSH2_KNOWNHOSTS *knownHostsList;
+            const char *hostKey;
+            size_t hostKeyLen;
+            int hostKeyType;
```
dwsteele, line 1154: Generally we declare variables as we assign them, rather than all at the top.

src/storage/sftp/storage.c
```diff
@@ -897,6 +1146,140 @@ storageSftpNew(
...
+                    // If this flag is set to yes, never automatically add host keys to the ~/.ssh/known_hosts file, and refuse to
+                    // connect to hosts whose host key has changed. This option forces the user to manually add all new hosts. If
+                    // this flag is set to “accept-new” then automatically add new host keys to the user known hosts files, but do
+                    // not permit connections to hosts with changed host keys. If this flag is set to “no” or “off”, automatically
+                    // add new host keys to the user known hosts files and allow connections to hosts with changed hostkeys to
+                    // proceed.
+                    switch (param.sftpStrictHostKeyChecking)
+                    {
+                        case SFTP_STRICT_HOSTKEY_CHECKING_YES:
+
```
dwsteele, line 1221: All of these cases should get braces. See https://github.com/pgbackrest/pgbackrest/blob/main/CODING.md#braces

src/storage/sftp/helper.c
```diff
@@ -32,7 +32,9 @@ storageSftpHelper(const unsigned int repoIdx, const bool write, StoragePathExpre
             .pathExpressionFunction = pathExpressionCallback, .modeFile = STORAGE_MODE_FILE_DEFAULT,
             .modePath = STORAGE_MODE_PATH_DEFAULT, .keyPub = cfgOptionIdxStrNull(cfgOptRepoSftpPublicKeyFile, repoIdx),
             .keyPassphrase = cfgOptionIdxStrNull(cfgOptRepoSftpPrivateKeyPassphrase, repoIdx),
-            .hostFingerprint = cfgOptionIdxStrNull(cfgOptRepoSftpHostFingerprint, repoIdx)));
+            .hostFingerprint = cfgOptionIdxStrNull(cfgOptRepoSftpHostFingerprint, repoIdx),
+            .sftpStrictHostKeyChecking = cfgOptionIdxStrId(cfgOptRepoSftpStrictHostKeyChecking, repoIdx),
+            .sftpKnownHosts = cfgOptionIdxLst(cfgOptRepoSftpKnownHosts, repoIdx)));
 }
```
dwsteele, line 38: Passing known hosts around as a variant list is a bit awkward -- perhaps use strLstNewVarLst() to transform it into a string list?

src/storage/sftp/storage.c
```diff
@@ -37,6 +38,199 @@ struct StorageSftp
...
+        FUNCTION_LOG_PARAM(STRINGZ, hostKey);
+        FUNCTION_LOG_PARAM(SIZE, hostKeyLen);
+    FUNCTION_LOG_END();
+
+    MEM_CONTEXT_TEMP_BEGIN()
+    {
+        LIBSSH2_KNOWNHOSTS *userKnownHostsList;
+        char *libSsh2ErrMsg;
+        int libSsh2ErrMsgLen;
+        int rc;
```
dwsteele, line 149: Again, declare on assignment with possible.

src/storage/sftp/storage.c
dwsteele: I just realized code formatting is failing -- better to format the code before we start reviewing. I thought these were cases that the formatter had somehow missed.

src/storage/sftp/helper.c
jreidthompson: Switched to passing a StringList in 
 https://github.com/pgbackrest/pgbackrest/pull/2125/commits/0716d6fa556caac34864f9a370575ae643c53f01

src/storage/sftp/storage.c
jreidthompson: Updates in
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/4fd017c80716d71a7ffff1f25379b9da196f7001
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/6811405f150c8f1095c3da31d952f0ddebc58be1
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/a3a82046feab0a915e0c116f867818b86ae42621

src/build/config/config.yaml
```diff
@@ -2399,6 +2399,14 @@ option:
     command: repo-type
     depend: repo-sftp-host
 
+  repo-sftp-known-hosts:
```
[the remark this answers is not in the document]
jreidthompson: Updates in
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/4fd017c80716d71a7ffff1f25379b9da196f7001
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/6811405f150c8f1095c3da31d952f0ddebc58be1
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/a3a82046feab0a915e0c116f867818b86ae42621

src/build/help/help.xml
jreidthompson: Updates in
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/4fd017c80716d71a7ffff1f25379b9da196f7001
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/6811405f150c8f1095c3da31d952f0ddebc58be1
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/a3a82046feab0a915e0c116f867818b86ae42621

src/storage/sftp/read.c
```diff
@@ -199,7 +199,7 @@ storageReadSftpClose(THIS_VOID)
                     FileCloseError,
                     STORAGE_ERROR_READ_CLOSE ": libssh2 errno [%d]%s", strZ(this->interface.name), rc,
                     rc == LIBSSH2_ERROR_SFTP_PROTOCOL ?
-                        strZ(strNewFmt(": sftp errno [%lu]", libssh2_sftp_last_error(this->sftpSession))) : "");
+                    strZ(strNewFmt(": sftp errno [%lu]", libssh2_sftp_last_error(this->sftpSession))) : "");
```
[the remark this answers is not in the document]
jreidthompson: Updates in
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/4fd017c80716d71a7ffff1f25379b9da196f7001
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/6811405f150c8f1095c3da31d952f0ddebc58be1
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/a3a82046feab0a915e0c116f867818b86ae42621

src/storage/sftp/storage.c
jreidthompson: Updates in
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/4fd017c80716d71a7ffff1f25379b9da196f7001
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/6811405f150c8f1095c3da31d952f0ddebc58be1
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/a3a82046feab0a915e0c116f867818b86ae42621

src/storage/sftp/storage.c
jreidthompson: Updates in
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/4fd017c80716d71a7ffff1f25379b9da196f7001
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/6811405f150c8f1095c3da31d952f0ddebc58be1
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/a3a82046feab0a915e0c116f867818b86ae42621

src/storage/sftp/storage.c
jreidthompson: Updates in
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/4fd017c80716d71a7ffff1f25379b9da196f7001
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/6811405f150c8f1095c3da31d952f0ddebc58be1
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/a3a82046feab0a915e0c116f867818b86ae42621

src/storage/sftp/storage.c
jreidthompson: Updates in
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/4fd017c80716d71a7ffff1f25379b9da196f7001
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/6811405f150c8f1095c3da31d952f0ddebc58be1
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/a3a82046feab0a915e0c116f867818b86ae42621

src/storage/sftp/storage.c
jreidthompson: Updates in
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/4fd017c80716d71a7ffff1f25379b9da196f7001
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/6811405f150c8f1095c3da31d952f0ddebc58be1
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/a3a82046feab0a915e0c116f867818b86ae42621

src/storage/sftp/storage.c
jreidthompson: Updates in
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/4fd017c80716d71a7ffff1f25379b9da196f7001
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/6811405f150c8f1095c3da31d952f0ddebc58be1
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/a3a82046feab0a915e0c116f867818b86ae42621

src/storage/sftp/storage.c
jreidthompson: Updates in
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/4fd017c80716d71a7ffff1f25379b9da196f7001
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/6811405f150c8f1095c3da31d952f0ddebc58be1
https://github.com/pgbackrest/pgbackrest/pull/2125/commits/a3a82046feab0a915e0c116f867818b86ae42621

src/storage/sftp/storage.c
jreidthompson: The initial collection will load up the entries in either the default hosts file list or the file(s) provided by the repo-sftp-known-host values. The storageSftpUpdateKnownHostsFile() function needs to operate on only the user's known_hosts file. If we use the initial collection then we'd write out all entries from all files to the user's known_hosts when we add a new host entry.

src/storage/sftp/storage.c
jreidthompson: The initial init will load up the entries in either the default hosts file list or the file(s) provided by the repo-sftp-known-host values. The storageSftpUpdateKnownHostsFile() function needs to operate on only the user's known_hosts file. If we use the initial collection then we'd write out all entries from all files to the user's known_hosts when we add a new host entry.

src/storage/sftp/storage.c
sfrost: Ah, alright, well, 0600 is probably fine if we're going to set a mode.  I'd be alright with letting umask apply too if that's easy to do.

dwsteele: By default we set umask to 0: https://github.com/pgbackrest/pgbackrest/blob/release/2.47/src/config/load.c#L459

sfrost: Ok, so then setting mode to 0600 probably is best plan.  Resolving.

src/build/config/config.yaml
```diff
@@ -2423,6 +2431,19 @@ option:
     command: repo-type
     depend: repo-sftp-host
 
+  repo-sftp-strict-host-key-checking:
+    section: global
```
dwsteele, line 2435: @sfrost I would prefer this to be `repo-sftp-strict-host-key-check` to be more consistent with our option naming, but I realize this name was selected to match up with the OpenSSH setting. What do you think?

sfrost: I'm fine with either naming.  More concerned about people being able to understand quickly how it works than about the exact name, which is never going to match what the OpenSSH config option is named anyway.

jreidthompson: I have no objection to changing it.

dwsteele: OK, renamed, but still not too satisfied. The option is one we'd usually end with "-type", which I can live with, but it also appears that repo-sftp-strict-host-key-check depends on whether repo-sftp-host-fingerprint is specified, which is exactly backward from the way we usually do it.

I'm guessing this was done to (try to) avoid breakage, but it's not ideal. Usually the way we would handle this is to have a type (which is default) that enables repo-sftp-host-fingerprint and other types that enable repo-sftp-known-host. That preserves the current behavior, but my guess is you want to auto change the current behavior, which may cause breakage anyway if they do not have repo-sftp-host-fingerprint set.

Thoughts? Am I missing something?

src/build/help/help.xml
```diff
@@ -1087,6 +1097,16 @@
                         <example>~/.ssh/id_ed25519.pub</example>
                     </config-key>
 
+                    <config-key id="repo-sftp-strict-host-key-check" name="SFTP host key check type">
+                        <summary>SFTP host key check type.</summary>
+
+                        <text>
+                            <p>The default setting is <quote>yes</quote>. If this flag is set to <quote>yes</quote>, <backrest/> will never automatically add host keys to the ~/.ssh/known_hosts file, and refuses to connect to hosts whose host key has changed or is not found in the known hosts files. This option forces the user to manually add all new hosts. If this flag is set to <quote>accept-new</quote> then <backrest/> will automatically add new host keys to the user's known hosts file, but will not permit connections to hosts with changed host keys. If this flag is set to <quote>no</quote> or <quote>off</quote>, <backrest/> will automatically add new host keys to the user's known hosts file and allow connections to hosts with changed host keys to proceed.</p>
+                        </text>
```
dwsteele, line 1105: Why do no and off do the same thing? Is this to emulate something in OpenSSH?

src/build/config/config.yaml
sfrost: I'd much prefer to have the known-host checking stuff get automatically picked up, if no fingerprint has been set (if a fingerprint has been set ... then we haven't really got much business groveling around through known hosts files, imv).

By default, ssh has 'ask' for StrictHostKeyChecking which will ask about not-yet-known hosts, but for hosts who have a known fingerprint will refuse to connect if the fingerprint from the server doesn't match what's in the known hosts.  This is basically what we want to get to and yes, that does mean we might break things for people who didn't specify a fingerprint and who don't have the host known yet in known-hosts.

Having an option for repo-sftp-strict-host-key-checking which is "use the provided fingerprint" could be alright but seems a bit silly- if it's provided, we should use it, not ignore it, regardless of what this other option is set to.  Still, I won't push back too hard if that's what you want to do.  What I would want, however, is a way to set the *default* for repo-sftp-strict-host-key-checking to be "use the provided fingerprint, but if there isn't one, then use the known hosts file".  This all seems like it's getting pretty far from how that option works in ssh_config though and I'm not super thrilled with that.

Hopefully this at least helps clear up some of the intention here.  I'm open to different ways of getting there.

src/build/help/help.xml
sfrost: That is how the StrictHostKeyChecking option in OpenSSH works, yes.

dwsteele: Not sure if I'm excited about having legacy cruft from OpenSSH transferred to pgBackRest.

Also @jreidthompson, usually we document these options as lists (see repo-type). Any reason not to do that here?

sfrost: Didn't really see it as legacy, just how openssh options generally work.  I'm not completely set on it being this way, but it seemed to be reasonable and to match openssh.  If you really don't like that then I suppose we can change it to whatever you feel would be more pgbackrest-y.

dwsteele: I'd prefer just `no`. That's still a valid OpenSSH setting. If we accept `off` shouldn't we accept `on`? :)

sfrost: If you'd like to just accept 'no' then that's fine with me.

src/build/config/config.yaml
dwsteele: OK, let me think on it. I'm not very content with the dependency checking being backward -- and with no dependencies in the config. We don't have any way for a type option to depend on a string because we don't do it that way in general...

src/build/help/help.xml
dwsteele: OK -- @jreidthompson can you remove off as an option? I'm done in this branch for the day.

jreidthompson: Yes. I will remove it. Will also look at the options noted above as a list.

jreidthompson: Removed off as an option. Updated the options to be presented as a list. Updated comments, tests, help entries.
919fac824d11bf47c4f3704a2fad72787565bea4

src/build/help/help.xml
```diff
@@ -1087,6 +1097,22 @@
...
+                    <config-key id="repo-sftp-strict-host-key-check" name="SFTP host key check type">
+                        <summary>SFTP host key check type.</summary>
+
+                        <text>
+                            <p>The following SFTP host key check types are supported:</p>
+                            <list>
+                                <list-item><id>yes</id> - <backrest/> will never automatically add host keys to the ~/.ssh/known_hosts file, and refuses to connect to hosts whose host key has changed or is not found in the known hosts files. This option forces the user to manually add all new hosts</list-item>
+                                <list-item><id>accept-new</id> - <backrest/> will automatically add new host keys to the user's known hosts file, but will not permit connections to hosts with changed host keys</list-item>
+                                <list-item><id>no</id> - <backrest/> will automatically add new host keys to the user's known hosts file and allow connections to hosts with changed host keys to proceed</list-item>
+                            </list>
```
dwsteele, line 1109: Wow, this seems a bit crazy. Why would you add the host keys at all if you are just ignoring them? Is this how OpenSSH operates?

jreidthompson: I think this defines the restrictions https://github.com/openssh/openssh-portable/blob/b6b49130a0089b297245ee39e769231d7c763014/sshconnect.c#L1298

dwsteele: Well, this kind of makes it look like we should not be updating host keys in this case: https://github.com/openssh/openssh-portable/blob/b6b49130a0089b297245ee39e769231d7c763014/sshconnect.c#L1342?

dwsteele: But this does `If this flag is set to “accept-new” then ssh will automatically add new host keys to the user known hosts files, but will not permit connections to hosts with changed host keys.  If this flag is set to “no” or “off”, ssh will automatically add new host keys to the user known hosts files and allow connections to hosts with changed hostkeys to proceed, subject to some restrictions.` Seems dumb as bricks to me, and I wonder what the `restrictions` are.

src/build/config/config.yaml
dwsteele: I believe I have resolved this in 672a3130c5. Basically, update the dependencies to be the (mostly) correct but add some rules in `cfgLoadUpdateOption()` to error on new cases where `repo-sftp-strict-host-key-check` is explicitly set incorrectly, but allow the current config of fingerprint to pass with a warning to update the config.

dwsteele: @sfrost I believe we have an unresolved issue from our discussion about whether to rename `repo-sftp-strict-host-key-check` to something like `repo-sftp-host-key-check-type`? I guess the values would be `strict`, `fingerprint`, accept-new` and `none`?

src/build/help/help.xml
dwsteele: I think I don't care what OpenSSH does here -- no means no. Perhaps it's OK for OpenSSH to muck about with these files but I don't think we should.

src/build/config/config.yaml
sfrost: I'm fine with the rename.

src/build/help/help.xml
sfrost: I'm fine with that.  I agree it seems pretty crazy and I'm not sure why OpenSSH does it that way.  I would prefer to know why they do it that way, but we don't and so I'm thinking that we go ahead and do what we think makes the most sense here.

jreidthompson: > Well, this kind of makes it look like we should not be updating host keys in this case: https://github.com/openssh/openssh-portable/blob/b6b49130a0089b297245ee39e769231d7c763014/sshconnect.c#L1342?

Just to note.
UpdateHostKeys is different from add a new hostkey. 
 UpdateHostKeys                                                                                                                     
 Specifies whether ssh(1) should accept notifications of additional hostkeys from the server sent after authentication has completed and add them to UserKnownHostsFile.  The argument must be yes, no or ask.  This option allows learning alternate hostkeys for a server and supports graceful key rotation by allowing a server to send replacement public keys before old ones are removed.  Additional hostkeys are only accepted if the key used to authenticate the host was already trusted or explicitly accepted by the user.                                        
                                                                                                                                    
 UpdateHostKeys is enabled by default if the user has not overridden the default UserKnownHostsFile setting, otherwise UpdateHostKeys will be set to ask.
                                                                                                                                    
 If UpdateHostKeys is set to ask, then the user is asked to confirm the modifications to the known_hosts file.  Confirmation is currently incompatible with ControlPersist, and will be disabled if it is enabled.
                                                                                                                                    
 Presently, only sshd(8) from OpenSSH 6.8 and greater support the "hostkeys@openssh.com" protocol extension used to inform the client of all the server's hostkeys.

Re adding new host keys.
For  case HOST_NEW:

https://github.com/openssh/openssh-portable/blob/b6b49130a0089b297245ee39e769231d7c763014/sshconnect.c#L1191

src/storage/sftp/storage.c
```diff
@@ -897,6 +1142,120 @@ storageSftpNew(
                     fingerprint, strZ(param.hostFingerprint));
             }
         }
+        else if (param.sftpStrictHostKeyChecking == SFTP_STRICT_HOSTKEY_CHECKING_NONE)
+        {
+            LOG_WARN_FMT("host key checking disabled (repo-sftp-host-key-check-type=none), connections are not secure!");
```
dwsteele, line 1147: Why did we add this warning? We don't want to emit warnings that the user has no ability to stop. This is not the default setting so they made a conscious choice -- not our problem.

dwsteele: I removed this in fd992dd9fbb9e.

dwsteele (approved):

dwsteele:
I'm done with my changes and review @jreidthompson, @sfrost. Let me know if everything looks OK.

jreidthompson:
> I'm done with my changes and review @jreidthompson, @sfrost. Let me know if everything looks OK.

Looks good to me

sfrost (approved):

PR #2131, opened 2023-07-31
Retry reads of pg_control until checksum is valid.

dwsteele:
On certain file systems (e.g. ext4) pg_control may appear torn if there is a concurrent write while reading the file. To prevent an invalid read, retry until the checksum matches the control data.

Special handling is required for the `pg-version-force` feature since the offset of the checksum is not known. In this case, scan from the default position to the end of the data looking for a checksum match. This is a bit imprecise, but better than nothing, and the chance of a random collision in the control data seems very remote.


pgstef:
> This causes a problem for the `pg-version-force` feature, though, since we don't know exactly where the checksum is located. A possible solution would be to scan forward from the default crc location to handle forks that add additional fields to pg_control.
> 
> Any thoughts on this, @sfrost, @pgstef?

Hi,
As we've discussed last week, the crc is usually the last field of the ControlFileData struct. So scanning to the end to look for the crc would indeed make it work imho :+1:

sfrost:
You mean just move forward one byte at a time until a checksum works or ... we get to the point where everything is just zeros?  idk that it'd make sense to just keep going all the way through the remaining 8k that is a pg_control file size.  Also seems a bit.. sketchy to just scan through and hope that when a valid match is found that it's actually the correct answer.  Given this is an internal feature that's more intended for development, we could ask users what the offset for the crc is, perhaps, or have an option to disable the crc check..

dwsteele:
> You mean just move forward one byte at a time until a checksum works or ... we get to the point where everything is just zeros?

Well, four bytes at a time since it is `uint32`.

>  idk that it'd make sense to just keep going all the way through the remaining 8k that is a pg_control file size.

Well, we'd do 512 bytes since that is the max size the control data can be. Also we'd only enable this when `pg-version-force=y`.

> Given this is an internal feature that's more intended for development, we could ask users what the offset for the crc is, perhaps, or have an option to disable the crc check..

Well, its a bit more than a developer option since it is being used in the field. I was thinking of adding a new flag to mark it as "maintainer" to make make sure users know it is special.

I though about the offset -- but in that end that's not much easier and a whole lot harder for the user. I'd rather not disable the CRC check in this case. Either we support this or we don't...

sfrost:
Alright, with those constraints maybe it won't be too bad.  Is there a good way for us to test that this is sensible and doesn't end up giving us the wrong answer..?  How do we know?

dwsteele:
@sfrost @pgstef This is ready for review. I removed the exclusion for `pg_control.tmp` since we don't know if this will be the final solution PG uses. The rest can be committed now regardless of how PG addresses the issue.

pgstef (approved):
Hi,

First of all, thanks for keeping the pgVersionForce compatibility.

I don't have lots of comments on the code itself, it looks good to me. The added tests seems enough.

I ran my usual test suite on this branch (including EPAS 15 tests) and didn't notice any trouble :+1:

sfrost (approved):

PR #2136, opened 2023-08-03
Accept leading tilde in paths for SFTP public/private keys.

jreidthompson:
Bug fix. Accept leading tilde file paths for public/private keys. Documentation indicates that leading tilde file paths for public/private keys is valid. It was an oversight on my part to not implement this originally.


dwsteele (approved):
Made a few minor adjustments, but overall this looks good to me. Unless you have objections @sfrost I am going to commit this.

sfrost:
No objections here, sounds good to me.

jreidthompson:
All looks good to me also

PR #2140, opened 2023-08-08
Report backup size and copy progress in info json

pgstef:
Report total size of the backup and copy progress (in bytes) to the lock file.
Then, add that to the info command json output.


dwsteele:
Duplicate of #2141.

PR #2141, opened 2023-08-08
Add backup size completed/total to info command JSON output.

pgstef:
Report total size of the backup and copy progress (in bytes) to the lock file.
Then, add that to the info command json output.


dwsteele (approved):
Updated the names of the fields. In `info` we already use size to mean size-total so I don't think we need size-total (also it is awkward to abbreviate). Also change size-prg to size-cplt to match pct-cplt, since they are talking about the same thing.

The backup module is still using progress and total, so we might consider making that match, but I think that should be a separate commit.

dwsteele:
> I've just STATUS_KEY_LOCK_BACKUP_SIZE_PROGRESS_VAR to reflect the new field names too.

Good catch, thanks!

PR #2144, opened 2023-08-11
pg_control version check FAIL in function pgControlFromBuffer() when PG version is 14

Monkey-LaoHan:
make  function `pgControlFromBuffer()'s` pg_control version check to support PG14 & PG15


dwsteele:
For the record, the control version for 13, 14, 15, and 16 is `1300`. Perhaps you are using a fork?

Monkey-LaoHan:
> For the record, the control version for 13, 14, 15, and 16 is `1300`. Perhaps you are using a fork?

Yes, I am using a fork.
Thanks for your reply

dwsteele:
You can try the `pg-version-force` option. It is internal so you'll need to get help from the command line (or peek in the doc source).

PR #2159, opened 2023-08-29
Add --repo-storage-tag option to create object tags.

dwsteele:
This new option allows tags to be added to objects in S3, GCS, and Azure repositories.

This was fairly straightforward for S3 and Azure, but GCS does not allow tags for a simple upload using the JSON interface. If tags are required then the resumable interface must be used even if the file falls below the limit that usually triggers a resumable upload (i.e. size < repo-storage-upload-chunk-size).

This option is structured so that tags must be specified per-repo rather than globally for all repos. This seems logical since the tag keys and values may vary by service. However, we can decide to configure tags differently if that makes more sense.

I have decided not to tie the storage tags the backup annotations since they seem to serve different purposes, but this decision is up for debate.


sfrost (commented):
Generally looks good to me, though want to make sure it works in all the cases we need it to.

dwsteele:
> be sure to set them correctly before running <cmd>stanza-create</cmd> to ensure uniform tags across the entire repository

This means if you want *tags* in the repo to be uniform you need to set them before stanza-create, .e.g. if you want all objects to be tagged with the same `CustomerId`.

You are correct that you can tag parts of the repo differently if you want, but seems unlikely to be very useful in this case. You never know, though.

> But the other files (backup files, archives,..) will eventually be expired and get new tags

Sure, eventually...

pgstef:
> This means if you want _tags_ in the repo to be uniform you need to set them before stanza-create, .e.g. if you want all objects to be tagged with the same `CustomerId`.

Makes sense in that context then :+1: 
Thanks for the explanation :-)

pgstef:
> You are correct that you can tag parts of the repo differently if you want, but seems unlikely to be very useful in this case. You never know, though.

I've been able to test it over an S3 bucket and it seems to add tags correctly.

However, probably not tight to this specific feature but repeating multiple times the same config key gets me confused when you define it in different places.

```ini
[global]
repo1-storage-tag=project=xXx

[global:archive-push]
repo1-storage-tag=classification=wal-archives
```
and `pgbackrest --stanza=X backup --repo1-storage-tag=classification=backup`

Result: files created by stanza-create get the `project=xXx`, archives only get `classification=wal-archives` and backup files only get `classification=backup`. I actually expected that the `project=xXx` tag would also be added to every files.

To achieve this, I had to:
```ini
[global]
repo1-storage-tag=project=xXx

[global:archive-push]
repo1-storage-tag=project=xXx
repo1-storage-tag=classification=wal-archives
```
and  `pgbackrest --stanza=X backup --repo1-storage-tag=project=xXx --repo1-storage-tag=classification=backup`

When you think about it, it sounds logical to overwrite the initial setting even though it doesn't feel very natural to have to duplicate it. When the same setting is possible multiple times, I would have felt more natural to have to provide an empty value to remove the key (eg. `repo1-storage-tag=project=`). It's working like that for annotations iirc.

Doesn't need to change anything, just saying because I thought it's good to know / expect that behavior.

pgstef (approved):

PR #2162, opened 2023-08-29
Improve GCS multi-part performance.

dwsteele:
The prior code avoided uploading a chunk if it was not clear whether the write was complete or not. This was primarily due to the GCS documentation being very vague on what to do in the case of a zero-size chunk.

Now chunks are uploaded as they are available. This should improve performance and also reduces the diff against a future commit that absolutely requires zero-size chunks.


jreidthompson (approved):
This looks good to me.

PR #2165, opened 2023-09-01
Fix configuration reference example for the tls-server-address option.

htgoebel:
The old example said
```
example: tls-server-address=::*
```
anyhow when using this, it results in an error:
```
ERROR: [049]: unable to get address for '::*': [-2] Name or service not known
```
Correct values are either `*` for IPv4 or `::` for IPv6.


htgoebel:
This PR is basically fixing a typo. I don't understand why some arm64 task fails.

dwsteele:
Not sure why the arm64 tests are failing, but this patch is not the cause. They suddenly started failing on Friday so I suspect a base image change is to blame. I find these issues often resolve themselves in a few days, but I'll probably try a new image to see what happens.

dwsteele (approved):
The arm64 image issue has resolved itself, ready to commit. Thanks!

PR #2169, opened 2023-09-04
Optimize WAL segment check after successful backup.

dwsteele:
The prior code did one list command against the storage for each WAL segment. This led to a lot of lists and was especially inefficient when the WAL (or the majority of it) was already present.

Optimize to keep the the contents of a WAL directory and use them on a subsequent search. Leave the optimizations for a single WAL segment since other places still use that mode.


src/command/archive/find.c
```diff
@@ -0,0 +1,230 @@
...
+        const String *const path = strNewFmt(STORAGE_REPO_ARCHIVE "/%s/%s", strZ(this->archiveId), strZ(prefix));
+        const String *const expression = strNewFmt(
+            "^%s%s-[0-f]{40}" COMPRESS_TYPE_REGEXP "{0,1}$", strZ(strSubN(walSegment, 0, 24)),
+            walIsPartial(walSegment) ? WAL_SEGMENT_PARTIAL_EXT : "");
+        RegExp *regExp = NULL;
+
+        do
+        {
+            // Get a list of all WAL segments that match the directory (and prefix when finding a single WAL)
+            if (this->list == NULL || !strEq(prefix, this->prefix))
```
sfrost, line 92: Given that this is testing for this->list being NULL and not merely empty, isn't it the case that eventually the list will be empty due to the removals and then a later search for a segment (for whatever reason, if it happens with this WalSegmentFind...) is attempted, it won't be found even if it actually existed?

More broadly, I'd say this function should probably have a more specific name if it's going to be "grab the list of segments with this prefix and then go through them but only allow any to be checked once" or if it's intended as "general walSegmentFind but with an optimization to grab a whole list and then go through it in order, but will still work properly for any requested segment".

Also just generally, not sure if the delete is really necessary anyway?  One could imagine that you just remember the last position and step forward from there when searching, which would allow you to wrap around if an earlier segment was asked for again for some reason, and also might be cheaper than the routine delete's and associated memmove's.

src/command/archive/find.h
```diff
@@ -0,0 +1,36 @@
...
+***********************************************************************************************************************************/
+FN_EXTERN WalSegmentFind *walSegmentFindNew(const Storage *storage, const String *archiveId, bool single, TimeMSec timeout);
+
+/***********************************************************************************************************************************
+Functions
+***********************************************************************************************************************************/
+// Find a WAL segment in the repository. The file name can have several things appended such as a hash, compression extension, and
+// partial extension so it is possible to have multiple files that match the segment, though more than one match is not a good
+// thing.
+FN_EXTERN String *walSegmentFind(WalSegmentFind *this, const String *walSegment);
```
sfrost, line 28: Not sure about the naming here, see other comments, might be better to use a new name for this and keep the existing for the 'fine one segment' case.  Depends on where things are taken with the other comment though.

sfrost (commented):
See in-line comments.  Think that the overall design might need a bit of work or at least some consideration for what happens in some edge cases.  Definitely like the overall idea though.

src/command/archive/find.c
dwsteele: > Given that this is testing for this->list being NULL and not merely empty, isn't it the case that eventually the list will be empty due to the removals and then a later search for a segment (for whatever reason, if it happens with this WalSegmentFind...) is attempted, it won't be found even if it actually existed?

The list is cleared and set to NULL at line 178 when there is nothing left in it, but there was no explicit test for this so added at 5cc0030a50.

> More broadly, I'd say this function should probably have a more specific name

As far as I can see, the functionality has not changed, except that it is possible to optimize it for a single lookup or multiple lookups. Single lookups still work in the general case, they just have to check more results. Conversely, multiple lookups still work if single lookup is specified, but they will be no faster than the prior code.

Also, when doing multiple lookups the optimization only works well if segments are requested in order, but in any case the results should be correct.

> Also just generally, not sure if the delete is really necessary anyway?

In the List object, deleting the first item just moves the pointer (https://github.com/pgbackrest/pgbackrest/blob/release/2.47/src/common/type/list.c#L339) so no reason for us to reinvent that in this function.

sfrost (approved):
Alright, looks good, thanks for adding the test.

PR #2174, opened 2023-09-07
Refactor address list lookup to include all returned addresses.

dwsteele:
sckHostLookup() only returned the first address record returned from getaddrinfo(). The new AddressInfo object provides a full list of values returned from getaddrinfo(). Freeing the list is also handled by the object so there is no longer a need for FINALLY blocks to ensure the list is freed.

Add the selected address to the client/server names for debugging purposes.

This code does not attempt to connect to multiple addresses. It just lays the groundwork for a future commit to do so.


src/common/io/socket/address.c
```diff
@@ -0,0 +1,187 @@
...
+            {
+                .ai_family = AF_UNSPEC,
+                .ai_flags = AI_PASSIVE,
+                .ai_socktype = SOCK_STREAM,
+                .ai_protocol = IPPROTO_TCP,
+            };
+
+            // Convert the port to a zero-terminated string for use with getaddrinfo()
+            char portZ[CVT_BASE10_BUFFER_SIZE];
+            cvtUIntToZ(port, portZ, sizeof(portZ));
```
sfrost, line 79: Given we're only supporting numeric ports, why not also set .ai_flags = AI_NUMERICSERV ; ?

src/common/io/socket/address.c
```diff
@@ -0,0 +1,187 @@
...
+    FUNCTION_LOG_END();
+
+    ASSERT(this != NULL);
+
+    freeaddrinfo(*(struct addrinfo **)lstGet(this->pub.list, 0));
+
+    FUNCTION_LOG_RETURN_VOID();
+}
+
+/**********************************************************************************************************************************/
```
sfrost, line 42: Presumably could use a description ... along with a few others in here.

sfrost (approved):
Couple of minor comments and maybe something to consider changing, but overall looks good to me.

src/common/io/socket/address.c
dwsteele: Make sense, done in a76d4ffbf138f09.

src/common/io/socket/address.c
dwsteele: Added more description of what is being freed here in 8a8e58c4b6a. extern'd functions have their comment in the header.

PR #2175, opened 2023-09-10
Support for dual stack connections.

dwsteele:
The prior code would only connect to the first address provided by getaddrinfo().

Instead try each address in the list. If all connections fail then wait and try them all again until timeout.

Currently a round robin approach is used where each connection attempt must fail before the next connection is attempted. This works fine, for example, when an ipv6 address has no route to the host, but will work less well when a host answers but doesn't respond in a timely fashion.

We may consider a Happy Eyeballs approach in the future, but since pgBackRest is primarily a background process, it is not clear that slightly improved response time (in the case of failed connections) is worth the extra complexity.


sfrost (approved):
LGTM

PR #2176, opened 2023-09-13
Add report option to check command.

dwsteele:
This option is intended to eventually create a comprehensive report about the state of the pgBackRest configuration based on the results of the check command.

Implement a detailed report of the configuration options in the environment and configuration files. This should be useful information when debugging configuration errors, since invalid options and configurations are automatically noted. While custom config locations will not be found automatically, it will at least be clear that the config is not in a standard location.

For now keep this option internal since there is a lot of work to be done, but commit it so that it can be used when needed and tested in various environments.

Note that for now when --report is specified, the check command is not being run at all. Only the config report is generated. This behavior will be improved in the future.


src/command/check/check.c
```diff
@@ -178,53 +182,58 @@ cmdCheck(void)
...
-        // Build stanza list based on whether a stanza was specified or not
-        StringList *stanzaList;
-        bool stanzaSpecified = cfgOptionTest(cfgOptStanza);
-
-        if (stanzaSpecified)
-        {
-            stanzaList = strLstNew();
-            strLstAdd(stanzaList, cfgOptionStr(cfgOptStanza));
-        }
+        if (cfgOptionBool(cfgOptReport))
```
sfrost, line 185: If the idea is for the report to be based on the results of the check command... then it seems kind of odd to entirely skip the actual check command run when this option is enabled?  Or at least, that's what it looks like is happening here.

Other than that ... the actual report generation and such seems pretty reasonable.  I do wonder if the rendering of all the options and such might be something we want to support calling from places other than check, but I suppose we can move the code around in the future if that ends up being the case.

dwsteele: Operative phrase here is `intended to eventually create a comprehensive report`. This started as a separate `support` command that I decided to merge into check instead. So no, it doesn't really do what is advertised, which is why it will remain internal for now. But since it touches stuff in the config module (a pretty fluid area) seems good to go ahead and get it in. Also, this way we can play with it and iterate.

sfrost: I'd at least make a note in the commit message and/or possibly in the comments around this that we currently aren't actually running 'check' as part of this, but that we will be in the future.

dwsteele: OK, update commit message above. Also keep in mind this option is internal and not user-facing.

sfrost (approved):

Issue #2163, opened 2023-08-30
How to disable SFTP feature ad tests if don't install libssh2


50wu:
Hi Team,

Since the `libssh2` package relies on the EPEL repo, we cannot include the EPEL repo to the customer. Is there any way to disable the SFTP feature and related tests if we don't want to install libssh2?

Right now build can succeed if don't install libssh2, but the test failed when running the following way:

```
pushd source/pgbackrest-release-2.47
      git init
popd
export PKG_CONFIG_PATH=/opt/vmware/postgres/15/lib64/pkgconfig
export LD_LIBRARY_PATH=/opt/vmware/postgres/15/lib64
./source/pgbackrest-release-2.47/test/test.pl --vm=none --min-gen --no-valgrind --no-coverage
```
And got the following error

```
FAILED: test-unit.p/test.c.o 
                                        cc -Itest-unit.p -I. -I.. -I../../../repo/src -I../../../repo/test/src -I/opt/vmware/postgres/15/include -I/usr/include/libxml2 -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -Werror -std=c99 -g -Wconversion -Wduplicated-cond -Wduplicated-branches -Wformat-nonliteral -Wformat=2 -Wformat-signedness -Wpointer-arith -Wstrict-prototypes -Wvla -Wwrite-strings -Wno-clobbered -Wno-missing-field-initializers -Wno-implicit-fallthrough -Wfatal-errors -fmacro-prefix-map=../src/= -fmacro-prefix-map=../test/src/=test/ -D_POSIX_C_SOURCE=200809L -DFN_EXTERN=extern -DVR_EXTERN_DECLARE=extern -DVR_EXTERN_DEFINE= -DERROR_MESSAGE_BUFFER_SIZE=131072 -DHRN_FEATURE_ERROR -DHRN_FEATURE_STACKTRACE -DHRN_FEATURE_ASSERT -DHRN_FEATURE_MEMCONTEXT -DHRN_FEATURE_STRING -DHRN_FEATURE_LOG -DHRN_FEATURE_DEBUG -DHRN_FEATURE_STAT -DHRN_FEATURE_IO -DHRN_FEATURE_STORAGE -DHRN_FEATURE_SOCKET -MD -MQ test-unit.p/test.c.o -MF test-unit.p/test.c.o.d -o test-unit.p/test.c.o -c ../test.c
[...]

PR #2177, opened 2023-09-14
Allow storage/sftp unit test to run without libssh2 installed.

jreidthompson:
Add missing #ifdefs and update tests to prevent test compilation failure and test run issues when libssh2 is not present as noted in issue https://github.com/pgbackrest/pgbackrest/issues/2163


dwsteele (approved):
Removed OptionalStorageRepoType. Since it was based on HAVE_LIBSSH the tests for it could never fail, so they were just dead code.

Fixed some errors that occurred when running the test module without libssh2 installed.

dwsteele:
> Fixed some errors that occurred when running the test module without libssh2 installed.

Looks like this was caused by merging the new SFTP code from integration. Even so -- fixed.

jreidthompson:
Looks good to me - Thanks

PR #2197, opened 2023-10-03
Add support for 1, 2, 4, 16 and 32 KB pages

KnightMurloc:
PostgreSql supports page sizes other than 8 KB. This patch adds support for other page sizes.

this patch is based on this PR https://github.com/pgbackrest/pgbackrest/pull/1833


src/command/backup/backup.c
```diff
@@ -2101,6 +2105,7 @@ backupProcess(
             .compressLevel = cfgOptionInt(cfgOptCompressLevel),
             .cipherType = cfgOptionStrId(cfgOptRepoCipherType),
             .cipherSubPass = manifestCipherSubPass(manifest),
+            .pageSize = backupData->dbPrimary ? dbPgControl(backupData->dbPrimary).pageSize : PG_PAGE_SIZE_8,
```
dwsteele, line 2108: Why not always just use `backupData->pageSize` here?

dwsteele:
I've been through this a reworked it: https://github.com/dwsteele/pgbackrest/tree/ADBDEV-3895-2-cig

1) Make the enum fit our standard and use it consistently throughout.
2) There's no point in having `pageChecksum.vendor.c.inc` if it is not a direct copy of the code from Postgres. So I consolidated the logic into `page.c`, which results in less duplication of logic in the binary.
3) Minor other corrections.

It will still need to be tested to ensure that there is no regression in performance, but seems better for you to have a look first and pull into this branch if it looks OK with you.

src/command/backup/backup.c
dwsteele: I see you changed this -- looks good to me.

dwsteele:
Have you been able to test this for performance regression?

I'm not quite ready to commit yet -- I'll look at doing that mid-December.

KnightMurloc:
Where do you suppose there could be a performance degradation? In the calculation of checksums or somewhere else?

KnightMurloc:
I did a little test and did not see a difference in the old and new version of the function for 100,000,000 pages. https://github.com/KnightMurloc/checksum-benchmark

dwsteele:
Don't you want to add `-funroll-loops -ftree-vectorize`?

dwsteele:
> still the same

OK, good. That's what I expected, but always good to be sure.

dwsteele:
I've made a few final changes in https://github.com/dwsteele/pgbackrest/tree/ADBDEV-3895-2-ci and this is now ready to commit. Please pull the changes into the PR.

I would appreciate it in the future if you would use a personal fork to submit PRs rather than the arenadata fork. It just wastes time if I can't update the PR myself.

PR #2202, opened 2023-10-10
Document maintainer options.

dwsteele:
Document maintainer options in a separate section with appropriate explanation and caveats.

Also make the pg-version-force option user visible now that maintainer caveats have been documented.


pgstef (approved):
Looks good to me. Having a maintainer dedicated section makes sense :+1:

PR #2224, opened 2023-11-14
The storageReadRemote method should return actual read bytes

RekGRpth:
The storageReadRemote method should return actual read bytes

All interface read methods should return actual read bytes.
This patch refactors storageReadRemote method to simplify it,
to eliminate duplicate code and to return actual read bytes.
The return value is calculated as the number of bytes written
to the passed buffer.

### DIFF
--- a/src/storage/remote/read.c
+++ b/src/storage/remote/read.c
@@ -154,21 +154,16 @@ storageReadRemote(THIS_VOID, Buffer *buffer, bool block)
             // Read if not eof
             if (!this->eof)
             {
-                // If the buffer can contain all remaining bytes
-                if (bufRemains(buffer) >= this->remaining)
+                size_t remains = bufRemains(buffer);
+                if (this->remaining < remains)
+                    remains = this->remaining;
+                bufCatSub(buffer, this->block, bufUsed(this->block) - this->remaining, remains);
+                result += remains;
+                this->remaining -= remains;
+                if (this->remaining == 0)
                 {
-                    bufCatSub(buffer, this->block, bufUsed(this->block) - this->remaining, this->remaining);
-
-                    this->remaining = 0;
                     bufFree(this->block);
                     this->block = NULL;
-                }
-                // Else read what we can
-                else
-                {
-                    size_t remains = bufRemains(buffer);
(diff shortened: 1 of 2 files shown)


dwsteele:
Thanks for catching this. Anything in particular you were trying to implement that you had to fix this? These read return values are not really used so I had been considering removing them...

I've made some minor edits here https://github.com/dwsteele/pgbackrest/tree/ADBDEV-4632-cig. Please pull them into this PR.

I must admit it is a minor irritation that I can't push directly to any PR branches from arenadata. Would be great if you could allow pushes from PRs.

RekGRpth:
> Anything in particular you were trying to implement that you had to fix this? These read return values are not really used so I had been considering removing them...

I plan to use this in another patch.

RekGRpth:
> I've made some minor edits here https://github.com/dwsteele/pgbackrest/tree/ADBDEV-4632-cig. Please pull them into this PR.

Applied

dwsteele (approved):

RekGRpth:
> I must admit it is a minor irritation that I can't push directly to any PR branches from arenadata. Would be great if you could allow pushes from PRs.

What kind of access do you need and how to set it up?

dwsteele:
> What kind of access do you need and how to set it up?

I wish I knew. Normally this just works, so I feel it is something that arenadata has modified in the repository or org permissions.

FWIW this is the error I get:
```
Pushing to https://github.com/arenadata/pgbackrest
remote: Permission to arenadata/pgbackrest.git denied to dwsteele.
fatal: unable to access 'https://github.com/arenadata/pgbackrest/': The requested URL returned error: 403
```
There is a check box on the PR that says something like "allow maintainers to push" Is that selected on your PRs?

RekGRpth:
> There is a check box on the PR that says something like "allow maintainers to push" Is that selected on your PRs?

Unfortunately, I don't have such a flag in the repository arenadata/pgbackrest.

dwsteele:
Ultimately I think the problem here is that you are submitting PRs from your org. I think you should be submitting PRs from a personal fork, e.g. `https://github.com/RekGRpth/pgbackrest`.

PR #2227, opened 2023-11-17
Skip files truncated during backup when bundling.

RekGRpth:
In bundle mode pgbackrest skips files of zero size, that is, it does not queue them for copying.

After splitting the files into bundles, pgbackrest launches one or more processes that directly perform the backup, namely, read the files and, if necessary, write them to the bundles.

If during the time between the distribution of all files among bundles and the direct copying of a file to a bundle, this file of non-zero size was truncated to zero size (for example, when the table was truncated), then pgbackrest still unconditionally places such a zero-size file in the bundle, taking up space in it equal to the size of the headings, and additionally writes the original file size to the manifest.

In debug build an assertion was added, that does not allow zero-size files to be written to bundles, which leads to an error.

To solve the problem, this patch, when reading the next file, separately calculates its size without taking into account filters. It creates a bundle and copies a file into it only if the file size is greater than zero. Otherwise, it just writes information about such a file in the manifest, indicating the expected size.

The advantages of the solution are that, firstly, the assert will not fire on debug builds, and secondly, we will not place zero-size files in bundles, which exactly corresponds to the specification.

The patch adds the backupCopyResultTruncate value to the BackupCopyResult enumeration to use it to indicate the result when a non-zero size file is truncated to zero size during the backup process.


src/command/backup/file.c
```diff
@@ -294,61 +294,84 @@ backupFile(
...
-                        // context because it will live longer than a single loop when more than one file is being written.
-                        if (write == NULL)
+                        Buffer *const buffer = bufNew(ioBufferSize());
+
+                        do
                         {
-                            MEM_CONTEXT_PRIOR_BEGIN()
+                            ioRead(readIo, buffer);
+                            // Copy the file in non-bundling mode or if the file size is not zero
+                            if (bundleId == 0 || ioReadInterface(readIo)->totalBytesRead != 0)
```
dwsteele, line 306: Why not just use `!bufEmpty(buffer)` here? I'm not sure that `totalBytesRead` is needed.

src/command/backup/file.c
```diff
@@ -294,61 +294,84 @@ backupFile(
...
+                                    {
+                                        write = storageNewWriteP(
+                                            storageRepoWrite(), repoFile, .compressible = compressible, .noAtomic = true,
+                                            .noSyncPath = true);
+                                        ioWriteOpen(storageWriteIo(write));
+                                    }
+                                    MEM_CONTEXT_PRIOR_END();
+                                }
+
+                                ioWrite(storageWriteIo(write), buffer);
```
dwsteele, line 325: Seems like we could just use ioCopy() after the first buffer is written and get rid of the loop. This would make the code a bit more compact.

test/src/module/command/backupTest.c
```diff
@@ -2270,6 +2299,60 @@ testRun(void)
...
+            "log truncate result");
+        TEST_RESULT_VOID(lockRelease(true), "release backup lock");
+
+        ManifestFile file = manifestFileFind(manifest, STRDEF("pg_data/test1"));
+        TEST_RESULT_UINT(file.bundleId, 0, "check bundle");
+        TEST_RESULT_UINT(file.size, 0, "check size");
+        TEST_RESULT_BOOL(bufEq(HASH_TYPE_SHA1_ZERO_BUF, BUF(file.checksumSha1, HASH_TYPE_SHA1_SIZE)), true, "checksum eq");
+
+        TEST_RESULT_LOG("P00 DETAIL: backup file host:" TEST_PATH "/test1 (10B->0B, 100.00%)");
     }
```
dwsteele, line 2356: These tests are more indirect than i would like, and we are trying to get rid of the tests directly against backupFile().

I'm working on a test harness to inject changes during the backup, but for now the way we do this is to put these types of tests at 3082 since a different file on the standby will appear to have changed after the backup started. See examples starting at 3121.

dwsteele (changes requested):
Looks reasonable, see review comments.

src/command/backup/file.c
dwsteele: Hmm, this may not work because of where the eof test is in ioCopy(). But that test could be moved to before the loop. Maybe getting too complicated.

src/command/backup/file.c
RekGRpth: because buffer contains filters data, not only file data

dwsteele: Ah, I see. Good point. Moving the filters to the write side does not seem like a good option here.

src/common/io/read.intern.h
```diff
@@ -15,6 +15,7 @@ Constructors
 typedef struct IoReadInterface
 {
     bool block;                                               // Do reads block when buffer is larger than available bytes?
+    size_t totalBytesRead;                                    // Total bytes read
```
dwsteele, line 18: This should be in the interface object (https://github.com/pgbackrest/pgbackrest/blob/release/2.48/src/common/io/read.c#L17) and exposed via a getter function, e.g. `ioReadTotal()`.

RekGRpth: I did exactly this in one of the previous versions of the patch.

RekGRpth: > This should be in the interface object

in IoReadInterface or in IoRead?

dwsteele: In IoRead, here: https://github.com/pgbackrest/pgbackrest/blob/release/2.48/src/common/io/read.c#L17

RekGRpth: done

test/src/module/command/backupTest.c
dwsteele: But that's not going to work because this test does not do bundling. I'll need to think about it.

dwsteele: I have committed the new backup test harness, which allows us to modify files during the backup. I've added a test to show the current behavior (337da35a), you can update the test as needed for your fix. Don't forget to put back the ASSERT() that I removed.

Be sure to merge all the commits from main as I have migrated all the backupFile() tests to use the new harness.

RekGRpth: > Be sure to merge all the commits from main as I have migrated all the backupFile() tests to use the new harness.

Do I need to make a pull request to the main branch or pgbackrest:integration?

dwsteele: You don't need to create a new pull request -- just merge the commits from pgbackrest:main into this branch so you get the new test harness.

RekGRpth: done

dwsteele: Yes, this is much better.

src/command/backup/file.c
dwsteele: I may adjust this before commit, but it seems OK for now.

src/command/backup/backup.c
```diff
@@ -1573,7 +1573,8 @@ backupJobResult(
                     file.checksumPageError = checksumPageError;
                     file.checksumPageErrorList =
                         checksumPageErrorList != NULL ? jsonFromVar(varNewVarLst(checksumPageErrorList)) : NULL;
-                    file.bundleId = bundleId;
+                    // Truncated file is not put in bundle
+                    file.bundleId = copyResult != backupCopyResultTruncate ? bundleId : 0;
```
dwsteele, line 1577: I think we should also change the log message when the file is truncate. Something like:
```
store truncated file pg1/grow-to-block-incr (4B->0B, [PCT])
```
No need for the checksum or bundle info in this case, but knowing the original size before it was truncated might be useful for debugging.

RekGRpth: done

dwsteele (changes requested):
I was doing my final final review of this patch when I realized it would not work for all cases. When there is a remote ioReadTotalBytes() will end up returning the number of bytes read after all filters are applied because the filters are pushed to the remote and *this* IoRead object is doing reads of that filtered data.

I think historically this is why we have resisted having a function like  ioReadTotalBytes(), because it will return different results depending on which host it is called from. However, a size filter will return the same result in all configurations.

Here's an implementation that uses the size filter to determine if a file has been truncated: https://github.com/dwsteele/pgbackrest/tree/ADBDEV-4647-cig

RekGRpth:
I updated PR, thanx

dwsteele:
> I updated PR, thanx

OK -- I'm going to let this sit for a while since I don't want to commit right now and I'll be mostly out until Dec 13.

src/command/backup/file.c
```diff
@@ -294,61 +294,99 @@ backupFile(
...
-                        // Setup the repo file for write. There is no need to write the file atomically (e.g. via a temp file on
-                        // Posix) because checksums are tested on resume after a failed backup. The path does not need to be synced
-                        // for each file because all paths are synced at the end of the backup. It needs to be created in the prior
-                        // context because it will live longer than a single loop when more than one file is being written.
-                        if (write == NULL)
+                        Buffer *const buffer = bufNew(ioBufferSize());
+                        bool readEof = false;
+
+                        // Read the first buffer to determine if the file was truncated. Detecting truncation matters only when
+                        // bundling is enabled as otherwise the file will be stored anyway.
```
RekGRpth, line 304: Why do we still need to save a zero size file when not bundling?
Can't we then restore such a file using the information from the manifest?

dwsteele: In some situations, users may do manual restores of their backups, usually because they copied a full backup out of the repo for long term storage. Or, an uncompressed backup can be used directly on ZFS with snapshots. In any case, it would break backward compitility so not something we are going to change. Going forward, most people should use bundling. I wish we could make it the default, but again, backward compatibility.

RekGRpth:
> Please pull in the release note from https://github.com/dwsteele/pgbackrest/tree/ADBDEV-4647-ci.

done

PR #2232, opened 2023-11-23
Prevent invalid recovery when backup_label removed.

dwsteele:
If backup_label is removed from a restored backup then PostgreSQL will instead use checkpoint information from pg_control to attempt (what is thinks is) crash recovery. This will nearly always result in a corrupt cluster because the checkpoint will not be from the beginning of the backup, and even if it is, the end point will not be specified, which could lead to recovery stopping too early.

To prevent this, invalidate the checkpoint LSN in pg_control on restore. If backup_label is removed then recovery will still fail because PostgreSQL will not be able to find the invalid checkpoint. The LSN of the checkpoint is not logged but it will be visible in pg_controldata output as 0/DEAD. This value is invalid because PostgreSQL always skips the first WAL segment when initializing a cluster.


sfrost (approved):
I'm generally on-board with these changes ... my only hesitation is if we need to worry about the PG version force stuff and non-regular-PG pg_control files ...

dwsteele:
> my only hesitation is if we need to worry about the PG version force stuff and non-regular-PG pg_control files

We do handle pg-version-force with this patch (the version I previewed internally did not). There's still the problem of fields being jumbled in pg_control without us knowing it, though. This would cause pgBackRest all kinds of problems, anyway, so it would be best if we detected it.

We do have some sanity checks like making sure page size and segment size are set appropriately. We could also add a check for FLOATFORMAT_VALUE and/or validate that checkpoint >= redo. Any other thoughts?

dwsteele:
Although `floatFormat` is right next to `blcksz` so not sure how helpful it is as a canary since we already validate `blcksz`. I guess it couldn't hurt, though.

sfrost:
Haven't a lot of other thoughts.  Wish we didn't have to deal with folks whacking around pg_control, but seems that ship has sailed.

dwsteele:
@sfrost Here's an improvement (7225b98b1) that throws an error on archive-get when the invalid checkpoint is detected. This only works when `recovery_target_timeline != current` but `latest` is the default after PG12.

The reason this works only in that case is that WAL is not fetched from the archive when backup_label is not present, even though `restore_command` is set and `recovery.signal` is present. However, history files *are* fetched -- which just overall seems broken to me.

dwsteele:
Added new validations in 63541b22 and 960b4358 to help ensure the format of pg_control is as expected.

PR #2235, opened 2023-11-28
Improved support for dual stack connections.

dwsteele:
Connections are established using the "happy eyeballs" approach from RFC 8305, i.e. new addresses (if available) are tried if the prior address has already had a reasonable time to connect. This prevents waiting too long on a failed connection but does not try all the addresses at once. Prior connections that are still waiting are rechecked periodically if no subsequent connection is successful.

This improves substantially on 39bb8a0d, which failed to take into account connection attempts that do not fail (but never connect) and use up all the available time.

[Some more real-world testing would be good to do here, but I'd like to get some review first to avoid testing multiple times].


sfrost (approved):
While there seems to always be some tension between gai.conf and 'happy eyeballs', based on my review, I can't find anything wrong with the approach we're taking here and it should give us a better result overall.  We also aren't in exactly the same boat as web browsers as a short delay with starting up a pgbackrest session is unlikely to cause problems given how long our sessions last.

PR #2246, opened 2023-12-16
Do not store block map when timestamp is updated but file is unchanged.

dwsteele:
If backup delta is not enabled, then the timestamp is used to determine if a file has changed. If the timestamp changes but the file is the same then the prior map would be stored unchanged in the new backup. This is not quite as bad as storing the entire file but obviously not ideal.

Instead detect if there are no new/changed blocks and simply reference the prior backup in this case.

In passing add a missing const keyword.


dwsteele:
Closing for now since I see a way to combine this with checking for equality in small files.

PR #2249, opened 2023-12-18
Improve SFTP storage error messages.

jreidthompson:
Utilize storageSftpEvalLibSsh2Error in more locations to provide better error information for debugging libssh2 issues.


src/storage/sftp/storage.c
```diff
@@ -1098,7 +1127,7 @@ storageSftpNew(
             THROW_FMT(ServiceError, "timeout during libssh2 handshake [%d]", rc);
 
         if (rc != 0)
-            THROW_FMT(ServiceError, "libssh2 handshake failed [%d]", rc);
+            THROW_FMT(ServiceError, "libssh2 handshake failed [%d]; man libssh2_session_handshake to map error", rc);
```
dwsteele, line 1130: Shouldn't the ; be a :? Also, not sure what `man libssh2_session_handshake to map error` means.

dwsteele (changes requested):

src/storage/sftp/storage.c
jreidthompson: I updated the handshake error handling to get  the error message from libssh2, replacing the poor attempt to refer to the man page to map the error code

src/storage/sftp/storage.c
```diff
@@ -1098,7 +1127,15 @@ storageSftpNew(
             THROW_FMT(ServiceError, "timeout during libssh2 handshake [%d]", rc);
 
         if (rc != 0)
-            THROW_FMT(ServiceError, "libssh2 handshake failed [%d]", rc);
+        {
+            char *libSsh2ErrMsg;
+            int libSsh2ErrMsgLen;
+
+            // Get the libssh2 error message
+            rc = libssh2_session_last_error(this->session, &libSsh2ErrMsg, &libSsh2ErrMsgLen, 0);
```
dwsteele, line 1135: libssh2_session_last_error() is already called three times and now we are adding another call. Since it is awkward to call I think it deserves a wrapper function.

src/storage/sftp/storage.c
```diff
@@ -1158,7 +1195,11 @@ storageSftpNew(
             LIBSSH2_KNOWNHOSTS *const knownHostsList = libssh2_knownhost_init(this->session);
 
             if (knownHostsList == NULL)
-                THROW_FMT(ServiceError, "failure during libssh2_knownhost_init");
+            {
+                THROW_FMT(
+                    ServiceError,
+                    "failure during libssh2_knownhost_init: libssh2 errno [%d]", libssh2_session_last_errno(this->session));
```
dwsteele, line 1201: Don't we want the error text here as in other places?

dwsteele (changes requested):

src/storage/sftp/storage.c
jreidthompson: Created storageSftpLibSsh2SessionLastError() wrapper to get libssh2 generated error message

src/storage/sftp/storage.c
jreidthompson: Updated to utilize storageSftpLibSsh2SessionLastError() to add libssh2 generated error message to the service error

Issue #2251, opened 2023-12-20
Bug Report: Backup progress not reporting via info command in version 2.48


rob-donovan:
pgBackRest version: 2.48

PostgreSQL version: 14

Operating system/version - Alpine Linux 3.19.0

Did you install pgBackRest from source or from a package? Source

Hello, I believe there is an issue in reporting the progress of backups via the info command in version 2.48, introduced by this change:
https://github.com/pgbackrest/pgbackrest/commit/e3fcd8cd9e0be42a32fc4cbd2ae3db2c072eabcd

The symptom we have seen is that we no longer receive reports of in-progress backups using the info command in version 2.48.

The writing of the new size parameters to the lock file in this change uses jsonWriteUInt64 whereas the reads use jsonReadUInt. For backups larger than UINT_MAX, this means that lock file writes are successful but reads fail silently, as any exceptions are swallowed in lockReadFile in lock.c. I verified this by manually editing the contents of the lock file and reducing the size parameter to below UINT_MAX, which resulted in backup progress being reported again. Happy to raise a pull request if this can be verified as an issue. Many thanks.

dwsteele:
Yeah, I have verified that this is a bug. Making this change in commonLockTest.c:
```
-                lockTypeBackup, .percentComplete = VARUINT(8888), .sizeComplete = VARUINT64(2807719), .size = VARUINT64(3159000)),
+                lockTypeBackup, .percentComplete = VARUINT(8888), .sizeComplete = VARUINT64(2807719000000), .size = VARUINT64(3159000000000)),
```
Leads to the following error:
```
unable to read lock file '/home/vagrant/test/test-0/test-backup.lock': 3159000000000 is out of range for uint
```
I don't think we want to do anything about this error getting swallowed -- we don't want to fail info because something is wrong with the lock file. A bit unfortunate that it gets hidden like this, though.

Go ahead and submit a PR, if you don't mind.

PR #2252, opened 2023-12-21
Read correct type for sizes in lock file

rob-donovan:
As per issue #2251 

The writing of the sz and szCplt parameters in the lock file use jsonWriteUInt64 but the reading of these parameters use jsonReadUInt. This causes a silent exception for any backups larger than MAX_UINT and prevents the info command from reporting progress. Correcting this so the reads are symmetric and verified before/after with the test change.


joewildish (approved):
+1 LGTM. We are also affected by this bug so appreciate the fix.

dwsteele:
Also looks good to me so committed. Thanks!

PR #2253, opened 2023-12-21
Refactor backup incremental manifest generation.

dwsteele:
This refactor should provide more clarity on what factors affect an incremental, rather that just having one big expression do it all. Overall this may be slightly more efficient since some values are reused that before were recalculated.

No behavioral changes are introduced.


src/info/manifest.c
```diff
@@ -1663,42 +1663,81 @@ manifestBuildIncr(Manifest *this, const Manifest *manifestPrior, BackupType type
...
-                    file.blockIncrMapSize = filePrior.blockIncrMapSize;
-
-                    // Perform delta if the file size is not zero
-                    file.delta = delta && file.size != 0;
-
-                    // Copy if the file has changed or delta
-                    file.copy = (file.size != 0 && file.timestamp != filePrior.timestamp) || file.delta;
+                    // If file size is equal to prior size then the file can be referenced instead of copied if it has not changed
+                    // (this must be determined during the backup).
+                    const bool fileEqual = file.size == filePrior.size;
```
sfrost, line 1684: This would really be better named 'fileSizeEqual'...

sfrost (approved):

src/info/manifest.c
dwsteele: Agreed. Done in d40d79b4316.

PR #2257, opened 2023-12-27
Detect files that have not changed during non-delta incremental backup.

dwsteele:
02eea555 added code to load a buffer of data from a file being backup up to detect files that have been truncated to zero after manifest generation. This mechanism can also be used to detect files that have not changed since the prior backup.

If the result of the file copy fits into a single buffer, then the size and checksum can be compared to the prior file before anything gets stored. If the file matches then it is referenced to the file in to prior backup.

The size that can be compared for normal copies is limited by the buffer size but for block incremental it works with any size file since there is no output from block incremental when the file is identical.


src/info/manifest.c
```diff
@@ -1712,8 +1728,9 @@ manifestBuildIncr(Manifest *this, const Manifest *manifestPrior, BackupType type
 
                     // Preserve values if the file will not be copied, is possibly equal to the prior file, or will be stored with
```
sfrost, line 1729: This comment looks like it needs updating?

sfrost (approved):
Just a minor comment update but the rest of it looks solid, so approving.

src/info/manifest.c
dwsteele: The comment is still true, but I added a comment to clarify why in 02f5415.

PR #2263, opened 2024-01-18
Allow --version and --help for version and help.

dwsteele:
It is a bit confusing that --help and --version do not work like most command-line programs. For example, git allows either --help or help.

Make these work by making them shortcuts (not actual options) to the applicable commands.

The user will still need to use help (not --help) to get help on specific commands/options, but at least they can get to the main help (which will tell them this) via --help.


turnstep (approved):
Looks good, thanks for this!

dwsteele:
At long last, right? :)

But I was in this code recently and I realized we could cheat a bit -- more or less what you were trying to do, but (hopefully) without side effects. We are in feature freeze right now, but this will go into the next release.

PR #2264, opened 2024-01-22
Make meson the primary build system.

dwsteele:
Meson has a lot of advantages over autoconf/make, primarily in ease-of-use and performance. Make meson the only build system used for testing and building the Debian documentation, but leave the RHEL documentation using autoconf/make for now so it gets some testing.


sfrost (approved):
Looks like this needs a rebase but otherwise it seems fine to me- I tested building with meson locally and with configure/make after these changes and all seemed to be fine.

PR #2268, opened 2024-01-28
Limit resume functionality to full backups.

dwsteele:
Resume does not work correctly with delta diff/incr backups because the presence of a reference causes it to remove the file with the idea that it can just be referenced again. This is true for timestamp-based backups but for deltas all existing files need to be rechecked (which requires a reference).

This is fixable but not without significant effort and new tests and it calls into question the usefulness of non-full resumes. For diff/incr, if the file was changed since the prior backup there is a good chance it will be modified again before the resume occurs.

In order to keep this feature as useful as possible for the most valuable case, limit resumes to full backups.

The patch is best viewed without white space changes: https://github.com/pgbackrest/pgbackrest/pull/2268/files?diff=unified&w=1


pgstef:
From a user point-of-view, the resume feature is very handy for big db and full backups. diff/incr are usually shorter anyway, so this makes sense :+1:

dwsteele:
> Maybe we could adjust the expire command message to tell the users that this was an un-resumable backup to remove?

This would be more complicated that it is worth, I think. We could adjust the resume message instead, but not sure what wording would work there.

This wording and behavior have been around for a pretty long time. If we do decide to adjust it, I think it should be in a separate commit.

pgstef:
> This would be more complicated that it is worth, I think. We could adjust the resume message instead, but not sure what wording would work there.

I guess it at least would worth to let users know that this particular non-resumable backup will/should be removed by the expire command. Maybe even detect if expire-auto is disabled (but I agree it would probably not worth that amount of effort).

> This wording and behavior have been around for a pretty long time. If we do decide to adjust it, I think it should be in a separate commit.

The current state of the PR is totally fine by me. Adjusting the messages could be done afterwards if we get more people confused/asking about it ;-)

sfrost (approved):

dwsteele:
OK, I've added a card for the wording change: https://github.com/pgbackrest/pgbackrest/projects/2#card-92022329. Still not very happy with anything I've though of.

PR #2281, opened 2024-02-11
Update references to recovery.conf.

dwsteele:
Update references to recovery.conf to include postgresql.auto.conf used in newer versions.

Also update a broken recovery URL and point it to the current version (with a hint to select the proper version of PostgreSQL).


pgstef (approved):
Looks good to me :+1:

Issue #2279, opened 2024-02-08
Enhancement request:  Add support for S3-CMK (SSE-C)


udf2457:
My understanding is that a present pgbackrest only supports SSE-S3 and SSE-KMS modes of encryption.

Please add support for SSE-C mode whereby the key can be provided client-side, i.e. by pgbackrest (https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html).

This is important for two reasons:

 1. sometimes regulatory or contractual reasons means that it is not possible to trust the cloud provider key management (or there are requirements for segregation of duties whereby the cloud provider cannot be both the data store and the key store)
 2. there are implementations of S3 out there that only support SSE-C mode, for example Exosale Simple Object Storage: https://community.exoscale.com/documentation/storage/encryption/

dwsteele:
It's hard to see how this is better than just using pgBackRest encryption: https://pgbackrest.org/user-guide.html#quickstart/configure-encryption.

It is probably not something that we'll work on (at least for now) but we would very likely accept a patch.

udf2457:
Fair enough.

Unfortunatley I don't speak C, otherwise I would. 😉

dwsteele:
> Unfortunatley I don't speak C, otherwise I would. 

Well, even if you just hacked it up that would be fine. The biggest barrier for us is setting up a test env that correctly reproduces what you want to do.

If you can hack up storage/s3/storage.c to do what you need, we can make it pretty and write unit tests, then have you retest the final result.

udf2457:
I know what you mean.  

In Go it's easy because you can just embed `minio` to mock up any S3 stuff.

The other thing that came to mind was `testcontainers` but looking at their website they don't have `c` integration.

I'll have a look at `storage.c` and see if it looks feasible.

dwsteele:
Basically, you need to put your parameters here: https://github.com/pgbackrest/pgbackrest/blob/main/src/storage/s3/storage.c#L491. Doesn't matter if you hard-code them -- we'll add the proper options later.

If you can get that to work, submit a PR with what you have (remember to redact the hard-coded key) and we'll have a look.

dwsteele:
> In Go it's easy because you can just embed minio to mock up any S3 stuff.

We have a way to mock up anything we want -- including features not supported by Minio. But first we need a working implementation.

dwsteele:
Closing this since there is PR #2282.

[...]

PR #2282, opened 2024-02-11
S3 SSE-C encryption support.

udf2457:
As per #2279, suggestion for SSE-C support.

Note on implementation detail:

I left it as an excercise for the user to provide the base64-encoded 128-bit MD5 digest of the encryption key. Because otherwise it would mean doing `md5(base64decode(userProvidedKey))`,  which would all get very messy, especially as you don't have existing base64 decoding in place.

Also from an integrity checking perspective, enforcing it on the user to provide both parameters means we are providing for end-to-end integrity (i.e. the user is the source of truth of the integrity check, not a `pgbackrest` function).


dwsteele:
> Also from an integrity checking perspective, enforcing it on the user to provide both parameters means we are providing for end-to-end integrity

AWS' boto does not require this and nor should we. The purpose of the field according the the docs is to ensure transmission integrity not to make the user's life hard. All the necessary function are present in pgBackRest so generating the field automatically is no problem.

> suggestion for SSE-C support.

The goal here was to get a mockup that had actually been compiled and tested. However, I have provided an implementation based on the documentation with zero real-world testing. Please compile and run through at least all the steps in the quickstart using the new feature (https://pgbackrest.org/user-guide.html#quickstart) at a minimum.

udf2457:
Apologies for the delay on this @dwsteele, have been busy at $work, but am hoping to test it over the next couple of days. 

Test and compile off main branch, right ?

dwsteele:
> Test and compile off main branch, right ?

Yes, main branch in the https://github.com/udf2457/pgbackrest repo.

udf2457:
Apologies, I don't know my way around `meson`.

I cloned my repo but `meson` is refusing to have to do anything with it:

`build/pgbackrest/doc/src/meson.build:5:4: ERROR: Problem encountered: meson is currently not supported for release builds
`

dwsteele:
This branch has not been updated in a while so it predates us enabling meson. I've updated the branch and it should work with meson now.

Alternately you could use the autoconf/make build, which is deprecated but still works: https://pgbackrest.org/user-guide-rhel.html#build

udf2457:
@dwsteele 

Tested using quickstart with the following config:

```
[demo]
pg1-path=/var/lib/postgresql/16/main

[global]
repo1-retention-full=2
repo1-type=s3
repo1-s3-endpoint=<REMOVED FOR SECURITY>
repo1-s3-bucket=<REMOVED FOR SECURITY>
repo1-s3-key=<REMOVED FOR SECURITY>
repo1-s3-key-secret=<REMOVED FOR SECURITY>-ixYnisz4e4Vy57sVZP4
repo1-s3-region=ch-gva-2
repo1-s3-sse-customer-key=<REMOVED FOR SECURITY>
```

It was tested against [Exoscale SOS](https://community.exoscale.com/documentation/storage/encryption/)

udf2457:
As an aside @dwsteele , you need to do whatever you need to do to ensure gets replaced with `<redacted>` for `repo1-s3-sse-customer-key`, like you already do for `s3-key-secret` etc.

dwsteele:
> you need to do whatever you need to do to ensure the SSE key gets replaced with <redacted>

Ah yes, good point. Done in 74562d5a.

> but your docs don't mention the requirement for head-*

These are not required on standard S3. My guess is HEAD is implied by GET, which makes a lot of sense since HEAD is just GET without the content.

dwsteele (approved):
OK, this looks good to me now, and works fine on regular S3.

dwsteele:
@udf2457 How would you like your name to appear in the release notes?

udf2457:
@dwsteele Tim Jones

Thanks again for your help making this happen !

dwsteele:
> Thanks again for your help making this happen !

You are welcome!

PR #2303, opened 2024-03-15
Allow alternative WAL segment sizes for PostgreSQL <= 10.

KnightMurloc:
For PostgreSQL before 11, alternate WAL segment sizes can be selected only at compile-time. While compile-time settings are generally not well tested by core, some established forks such as Greenplum use them.
This patch removes the limitation of the default WAL segment size for PostgreSQL before 11.


KnightMurloc:
@dwsteele, you asked to make PRs from personal forks, not from the company's one. My team lead says that we cannot use personal forks, because there is a possibility of losing access to the PR in case the author of the PR is unavailable (has resigned, gone on vacation, on sick leave, etc.). Therefore, the company policy requires making PRs from the corporate fork. As a solution, if you want to make cosmetic changes, you can form final commit with your changes in the integration branch, and then close the PR.

dwsteele:
I'm not too excited about adding an option for this. If we are going to let people bypass this then I think it makes more sense to simply remove the constraint at https://github.com/pgbackrest/pgbackrest/blob/release/2.50/src/postgres/interface.c#L176.

> you asked to make PRs from personal forks, not from the company's one

You and your colleagues are making valuable contributions so I'm not going to make a big fuss. I still prefer to merge the commit rather than closing it so we'll just continue what we've been doing.

KnightMurloc:
That is, would you prefer the option without checking for Postgres < 11 that the size of the WAL segment is the default, leaving only the check that the value is valid?

dwsteele:
> That is, would you prefer the option without checking for Postgres < 11 that the size of the WAL segment is the default, leaving only the check that the value is valid?

I can't quite parse these sentence, but I think so? All versions <= 11 are EOL so I'm OK with just removing the constraint on the value being default, which should just mean removing that check and adjusting the test.

KnightMurloc:
@dwsteele Hello, I would like to know the status of this PR. Do I need to make any other edits?

src/command/archive/common.c
```diff
@@ -419,7 +419,7 @@ walSegmentNext(const String *walSegment, size_t walSegmentSize, unsigned int pgV
     ASSERT(walSegment != NULL);
     ASSERT(strSize(walSegment) == 24);
     ASSERT(UINT32_MAX % walSegmentSize == walSegmentSize - 1);
-    ASSERT(pgVersion >= PG_VERSION_11 || walSegmentSize == PG_WAL_SEGMENT_SIZE_DEFAULT);
+    ASSERT(pgWalSegmentSizeValid(pgVersion, walSegmentSize));
```
dwsteele, line 422: I'd prefer to just remove this assert().

src/postgres/interface.c
```diff
@@ -165,6 +165,23 @@ pgDbIsSystemId(const unsigned int id)
...
+    FUNCTION_TEST_END();
+
+    if (pgVersion < PG_VERSION_10 && walSegmentSize > PG_9_MAX_WAL_SEGMENT_SIZE)
+    {
+        FUNCTION_TEST_RETURN(BOOL, false);
+    }
+
+    FUNCTION_TEST_RETURN(BOOL, IsValidWalSegSize(walSegmentSize));
+}
+
```
dwsteele, line 184: I don't think it is worth adding this exception for EOL versions of PostgreSQL. If somebody somehow convinces older versions of PostgreSQL to use large WAL segments sizes, then that is on them.

src/postgres/interface.c
```diff
@@ -176,19 +193,12 @@ pgWalSegmentSizeCheck(unsigned int pgVersion, unsigned int walSegmentSize)
         FUNCTION_TEST_PARAM(UINT, walSegmentSize);
     FUNCTION_TEST_END();
 
-    if (pgVersion < PG_VERSION_11 && walSegmentSize != PG_WAL_SEGMENT_SIZE_DEFAULT)
-    {
-        THROW_FMT(
-            FormatError, "wal segment size is %u but must be %u for " PG_NAME " <= " PG_VERSION_10_Z, walSegmentSize,
-            PG_WAL_SEGMENT_SIZE_DEFAULT);
-    }
-
```
dwsteele, line 185: I think we should just be removing this and then updating tests.

dwsteele (changes requested):
Sorry, I didn't see that this had been updated. I'd like to keep the implementation simpler since we are dealing with EOL versions of Postgres. See comments.

dwsteele (approved):
I renamed HRN_PG_WAL_SEGMENT_SIZE_DEFAULT in c57496a483b but otherwise looks good to me. Please pull https://github.com/dwsteele/pgbackrest/tree/ADBDEV-4788-ci with these changes and the release note.

dwsteele:
Updated to 2cf5a46663d to fix commit message.

PR #2321, opened 2024-04-09
Allow explicit disabling of optional dependencies in meson builds.

mschout:
I'm the maintainer of the FreeBSD port of pgbackrest.

One of the challenges of porting the FreeBSD port over to meson builds is that there needs to be a way to ensure that the build is not accidentally linked against one of the "optional" dependencies, even if it happens to be present on the system.

In the FreeBSD port, user can configure their package builds on a per-port basis.  For the work for the 2.51 release, this configuration looks like this:

image

LZ4 is enabled by default since the port has always depended on that.  Then later support for ZSTD was added.  Now, we have libssh2 support as well.      So if the port is configured with just the `LZ4` option enabled, the build should *NOT* link against `libssh2` nor should it link `libzstd`, even if it happens to already be installed on the system.  Conversely, `meson setup` should *fail* if the configured features are not found.

Some searching, my understanding is that the current best way to accomplish this is by using meson "features".  The default behaviour is "auto" which does the same thing that `required: false` does.  But it can be overridden with `-Dliblz4=disabled` which means it is explicitly turned off.  Also you could say `-Dliblz4=enabled` which means the depenency is required.

I am incorporating these patches into the FreeBSD port for now, but I think it would be useful to have in the upstream build files.

See [meson PR where this was added](https://github.com/mesonbuild/meson/pull/3376)

See: [FreeBSD ticket to update to 2.51](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=276559)


dwsteele:
I agree that this looks like the best way to handle this in meson and we are happy to accept it into upstream.

I did some wordsmithing on the option comments, but left the commit otherwise unchanged so there should be no functional difference. Please note that libssh2 is required for SFTP support but *not* SSH support. We still support SSH by exec'ing `ssh` as we always have.

Once the meson dust settles we are planning to make lz4 a build requirement as there are several places that we would like to use it internally (where we now use gzip).

Finally, we appreciate you jumping right on the meson build. We are eager to retire autoconf/make but want to be sure everyone is safely on meson first.

dwsteele (approved):

PR #2326, opened 2024-04-16
Add GCS batch delete support.

dwsteele:
The GCS driver sent a single file delete request for each file while deleting a path. Depending on latency this could lead to rather long delete times, especially noticeable during expiration.

Improve GCS delete to use batches, which require multipart HTTP, so also add multipart HTTP infrastructure.


src/common/io/http/request.c
```diff
@@ -302,6 +344,153 @@ httpRequestError(const HttpRequest *const this, HttpResponse *const response)
...
+
+        // Add first boundary
+        bufCat(result, boundary);
+
+        // Add content and boundaries
+        for (unsigned int contentIdx = 0; contentIdx < lstSize(this->contentList); contentIdx++)
+        {
+            bufCat(result, *(Buffer **)lstGet(this->contentList, contentIdx));
+            bufCat(result, boundary);
+        }
```
jreidthompson, line 464: Hey - I may be looking at the wrong reference(s), but examples that I've seen seem to have a two dash trailer on the last boundary entry in multicontent batch requests/responses?

Found this https://www.ietf.org/rfc/rfc2046.txt,
Page 19 paragraph 4
   The boundary delimiter line following the last body part is a
   distinguished delimiter that indicates that no further body parts
   will follow.  Such a delimiter line is identical to the previous
   delimiter lines, with the addition of two more hyphens after the
   boundary parameter value.
     --gc0pJq0M:08jU534c0p--

dwsteele: No, I think you have it right and not sure how I missed it. Add in 3b2d778849.

I've decided to leave this check out of the parsing side, though, since GCS clearly does. There are other ways for us to know if the response is complete, i.e. content-length or content-encoding. My guess is this is a holdover from the HTTP 1.0 days when EOF was frequently used to denote end of content.

dwsteele:
Committed, thanks for the review!

PR #2338, opened 2024-04-23
Dynamically find python in meson build.

mschout:
Another small MR to make meson builds work without any fuss on FreeBSD.  My understanding is that this is the more reliable thing to do everywhere.

See: [Upstream FreeBSD Port Review](https://reviews.freebsd.org/D44791#1023359)


mschout:
bit more explanation: on FreeBSD, there probably wont be a `python3` binary.  Rather its probably going to be called `python3.9`.  The patch here works in both scenarios.

dwsteele (approved):
This turned out to be a pain because some meson versions were reporting that they needed python3-setup tools installed, but this turned out to be a bug in those versions. They actually need python3-distutils installed.

mschout:
Thanks for all the work getting this merged @dwsteele !

PR #2352, opened 2024-05-19
Update start/stop documentation to reflect actual functionality.

dwsteele:
The exact functionality of start/stop have evolved over time and have become a bit confusing. It may be appropriate to make the behavior more consistent, but for now at least document the behavior correctly. The documentation for start/stop was fairly inaccurate.

### DIFF
--- a/doc/xml/user-guide.xml
+++ b/doc/xml/user-guide.xml
@@ -2948,7 +2948,7 @@
     <section id="start-stop" depend="/repo-host/config">
         <title>Starting and Stopping</title>
 
-        <p>Sometimes it is useful to prevent <backrest/> from writing to repositories. For example, when failing over from primary to standby it is best to prevent <backrest/> from writing on the old primary in case <postgres/> gets restarted or cannot be completely killed. If a standby is setup for testing then it is also a good idea to prevent it from writing to repositories.</p>
+        <p>If a standby is promoted for testing, or a test cluster is restored from a production backup, then it is a good idea to prevent those clusters from writing to <backrest/> repositories. This can be accomplished with the <cmd>stop</cmd> command.</p>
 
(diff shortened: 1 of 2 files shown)


doc/xml/user-guide.xml
```diff
@@ -2948,17 +2948,19 @@
     <section id="start-stop" depend="/repo-host/config">
         <title>Starting and Stopping</title>
 
-        <p>Sometimes it is useful to prevent <backrest/> from running on a system. For example, when failing over from a primary to a standby it's best to prevent <backrest/> from running on the old primary in case <postgres/> gets restarted or can't be completely killed. This will also prevent <backrest/> from running on <id>cron</id>.</p>
+        <p>Sometimes it is useful to prevent <backrest/> from writing to repositories. For example, when failing over from primary to standby it is best to prevent <backrest/> from writing on the old primary in case <postgres/> gets restarted or cannot be completely killed. If a standby is setup for testing then it is also a good idea to prevent it from writing to repositories.</p>
```
pgstef, line 2951: Hm, the `standby` term used here might be confusing for the users. Most often, they want to restore a production backup onto a test server. And a standby wouldn't push things unless promoted.

Maybe to be even more complete, something like this:

> If a standby is promoted for testing, or a test cluster is restored from the production backup, then it is also a good idea to prevent those clusters from writing to repositories.

:thinking:

dwsteele: Good idea. How about 73485485402f3e60?

pgstef: Looks great!

pgstef (approved):

doc/xml/user-guide.xml
dwsteele: Excellent, thank you!

PR #2367, opened 2024-05-31
Fix issue with files larger on the replica than on the primary.

dwsteele:
If a file on the primary was larger than on the replica then the next diff/incr backup would store the primary size instead of the replica size when block incremental was enabled. On the next diff/incr backup this would lead to a `repo size must be > 0 for file` error when validating the manifest.

Fix this by limiting copy based on sizeOriginal rather than size so size can be set to the value expected to be stored in the manifest. As a bonus sizePrior is no longer needed since size can be used to the same purpose.


src/command/backup/backup.c
```diff
@@ -1474,6 +1474,18 @@ backupJobResult(
                 {
                     LOG_DETAIL_PID_FMT(
                         processId, "match file from prior backup %s (%s)%s", strZ(fileLog), strZ(logProgress), strZ(logChecksum));
+
+                    // File size may need to be updated because the file original file size may have been larger than what was found
```
pgstef, line 1478: `file original file size` sounds weird to me :thinking: `because the original file size` should be enough?

pgstef (approved):
> If a file on the primary was larger than on the primary then the next diff/incr backup would store the primary size instead of the standby size when block incremental was enabled.

You meant if a file on the primary was larger than on the standby?

The fix looks simple enough and the added tests are quite good cases to cover :+1: 
So this PR looks good to me :-)

src/command/backup/backup.c
dwsteele: Definitely a typo. Fixed in 69780b86063ff3.

dwsteele:
> You meant if a file on the primary was larger than on the standby?

Haha, yes. Fixed in the PR description so it will be right in the commit message.

src/command/backup/backup.c
pgstef: cool, Thanks. LGTM now :+1:

dwsteele:
@pgstef I've updated this with what I think is a better implementation (see updated description) and also less prone to regression.

pgstef:
> @pgstef I've updated this with what I think is a better implementation (see updated description) and also less prone to regression.

I've tried to play a bit with this but couldn't really easily generate a replication use-case with fsm larger on the primary.

Anyway, using sizeOriginal sounds pretty clever, adding a nice bonus :+1: 
So this PR still looks good to me :-)

dwsteele:
> I've tried to play a bit with this but couldn't really easily generate a replication use-case with fsm larger on the primary.

Yeah, I have not been able to reproduce it either, but it is happening. My guess is it is rare or this bug would be more prevalent.

PR #2371, opened 2024-06-07
Refactor lock module.

dwsteele:
Refactor the lock module to split command-specific logic from the basic file locking functionality. Command specific logic is now in command/lock.c. This will make it easier to implement new features such as repository locking and updating lock file contents on remotes.

This implementation is essentially a drop-in replacement but there are a few differences. First, the lock names no longer require a path (the path is added in the lock module). Second, the timeout functionality has been removed since it was not being used.


test/src/module/command/infoTest.c
```diff
@@ -236,16 +236,17 @@ testRun(void)
         {
             HRN_FORK_CHILD_BEGIN()
             {
-                lockInit(cfgOptionStr(cfgOptLockPath), STRDEF("999-ffffffff"), STRDEF("stanza1"), lockTypeBackup);
-                TEST_RESULT_INT_NE(lockAcquireP(), -1, "create backup/expire lock");
+                lockInit(cfgOptionStr(cfgOptLockPath), STRDEF("999-ffffffff"));
+                TEST_RESULT_INT(
+                    lockAcquireP(cmdLockFileName(STRDEF("stanza1"), lockTypeBackup)), true, "create backup/expire lock");
```
jreidthompson, line 241: Is there a reason we compare lockAcquireP's result to INT in infoTest.c vs BOOL in the other test files?

test/src/module/command/lockTest.c
```diff
@@ -0,0 +1,83 @@
...
+/***********************************************************************************************************************************
+Test Run
+***********************************************************************************************************************************/
+static void
+testRun(void)
+{
+    FUNCTION_HARNESS_VOID();
+
+    // Create default storage object for testing
+    Storage *storageTest = storagePosixNewP(TEST_PATH_STR, .write = true);
```
jreidthompson, line 18: Should this be const Storage *const storageTest?

CONTRIBUTING.md
```diff
@@ -470,15 +470,15 @@ HRN_FORK_BEGIN()
 {
     HRN_FORK_CHILD_BEGIN()
     {
-        TEST_RESULT_INT_NE(lockAcquireP(), -1, "create backup/expire lock");
+        TEST_RESULT_INT_NE(cmdLockAcquireP(), -1, "create backup/expire lock");
```
jreidthompson, line 473: Should this be TEST_RESULT_BOOL, true?

doc/xml/contributing.xml
```diff
@@ -526,15 +526,15 @@ HRN_FORK_BEGIN()
 {
     HRN_FORK_CHILD_BEGIN()
     {
-        TEST_RESULT_INT_NE(lockAcquireP(), -1, "create backup/expire lock");
+        TEST_RESULT_INT_NE(cmdLockAcquireP(), -1, "create backup/expire lock");
```
jreidthompson, line 529: Should this be TEST_RESULT_BOOL, true?

src/command/lock.c
```diff
@@ -0,0 +1,149 @@
...
+    ASSERT(lockType == lockTypeBackup);
+    ASSERT(stanza != NULL);
+
+    FUNCTION_AUDIT_STRUCT();
+
+    LockReadResult result = {0};
+
+    MEM_CONTEXT_TEMP_BEGIN()
+    {
+        String *const lockFileName = cmdLockFileName(stanza, lockType);
```
jreidthompson, line 115: String should be const?

test/src/module/command/lockTest.c
dwsteele: I'd rather leave this for now. In general the way we use const for Storage objects is not consistent with how were use const for other objects. This is because Storage was one of the first objects created and never got a refactor (but I have added it to my list).

test/src/module/command/infoTest.c
dwsteele: No reason -- just incautious search and replace. Fixed in 95f3baad267.

CONTRIBUTING.md
dwsteele: Fixed in 95f3baad267.

doc/xml/contributing.xml
dwsteele: Fixed in 95f3baad267.

src/command/lock.c
dwsteele: Fixed in 4f9bc4a8d27f.

jreidthompson:
This looks good to me.

jreidthompson (approved):
This looks good to me

Issue #2376, opened 2024-06-14
Backup to Sftp not working


ahmed112212:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version:
2.52

2. PostgreSQL version:
16
3. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each:
ubuntu 24.04
4. Did you install pgBackRest from source or from a package?
package
5. Please attach the following as applicable:
    - `pgbackrest.conf` file(s)
[pgbackrest.conf.txt](https://github.com/user-attachments/files/15837663/pgbackrest.conf.txt)

    - `postgresql.conf` settings applicable to pgBackRest (`archive_command`, `archive_mode`, `listen_addresses`, `max_wal_senders`, `wal_level`, `port`)
    - errors in the postgresql log file before or during the time you experienced the issue
    - log file in `/var/log/pgbackrest` for the commands run (e.g. `/var/log/pgbackrest/mystanza_backup.log`)

7. Describe the issue:

pgbackrest --stanza=logs --type=full backup --compress-type=none --process-max=4
2024-06-13 12:10:31.741 P00   INFO: backup command begin 2.52: --compress-type=none --db-timeout=3600 --delta --exec-id=3631932-6aa3143e --log-level-console=info --log-level-file=debug --pg1-path=/etc/postgresql/16/main --process-max=4 --repo1-bundle --repo1-path=/backup/postgresql-logs-main --repo1-retention-diff=7 --repo1-retention-full=2 --repo1-sftp-host=USERNAME.your-storagebox.de --repo1-sftp-host-key-hash-type=sha1 --repo1-sftp-host-port=22 --repo1-sftp-host-user=USERNAME --repo1-sftp-private-key-file=/var/lib/pgbackrest/USERNAME_rsa --repo1-type=sftp --stanza=logs --type=full
2024-06-13 12:10:32.994 P00   INFO: execute non-exclusive backup start: backup begins after the next regular checkpoint completes
2024-06-13 12:10:33.495 P00   INFO: backup start archive = 0000000100002396000000A5, lsn = 2396/A5000028
2024-06-13 12:10:33.495 P00   INFO: check archive for prior segment 0000000100002396000000A4
[...]

PR #2377, opened 2024-06-17
Fix SFTP renaming failure when file already exists.

jreidthompson:
Update error handling for libssh2_sftp_rename_ex in storageWriteSftpClose for file already exists. During development and testing the sftp servers never returned file already exists (#define LIBSSH2_FX_FILE_ALREADY_EXISTS 11UL), rather they returned LIBSSH2_FX_FAILURE 4UL when a file already existed.  Issue https://github.com/pgbackrest/pgbackrest/issues/2376 reflects returning  file already exists. Also made an update where storageSftpPathCreate may similarly return file already exists.


dwsteele (approved):
This looks OK to me. Seems like you want to try the same server/version the user is using before committing this?

jreidthompson:
> This looks OK to me. Seems like you want to try the same server/version the user is using before committing this?

If possible yes, to see if I can flush out any other cases where it differs from the original dev/test environments.

jreidthompson:
Per comment in the issue, the sftp storage is https://www.hetzner.com storage-box.  I created an account and verified that prior to code changes the issue error occurred and that after code changes the following commands complete successfully. 

stanza-create
check
info
backup full
backup incr
backup diff
info
repo-ls
repo-ls recurse
verify
restore

dwsteele:
So this is ready for commit?

jreidthompson:
> So this is ready for commit?

Yes, it's ready.

Issue #2054, opened 2023-04-28
How to take pgbackrest backup without failure when 1 node in the patroni 2 node cluster fails.


ashwinsirigi:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version: 2.37

2. PostgreSQL version: 13.9

3. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each: CentOS Stream release 8 

4. Did you install pgBackRest from source or from a package? Source

5. Please attach the following as applicable:
    - `pgbackrest.conf` file(s)
[pgbackrest@Pr-IoTDB4 ~]$ cat /etc/pgbackrest/pgbackrest.conf 
[cbdemo]
pg1-host=192.168.232.1
pg1-host-port=22
pg1-port=25432
pg1-path=/disk01/postgresql
pg2-host=192.168.232.2
pg2-host-port=22
pg2-port=25432
pg2-path=/disk01/postgresql

[global]
buffer-size=4MiB
compress-type=gz
backup-standby=y
process-max=4
repo1-path=/nfs/pgsql
repo1-retention-full-type=time
repo1-retention-full=365
repo1-retention-diff=7
repo1-retention-archive=14
repo1-retention-history=0
start-fast=y
log-level-console=info
log-level-file=info

[global:archive-push]
compress-level=6

    - `patroni.conf` settings applicable to pgBackRest (`archive_command`, `archive_mode`, `listen_addresses`, `max_wal_senders`, `wal_level`, `port`)
Below config is from one of the node   
 archive_mode: 'on'
        archive_command: '/usr/bin/pgbackrest --config=/etc/pgbackrest/pgbackrest.conf --stanza=cbdemo archive-push %p'
listen: 192.168.232.1:25432
max_wal_senders: '10'
wal_level: 'replica'
port: 25432

    - errors in the postgresql log file before or during the time you experienced the issue
    - log file in `/var/log/pgbackrest` for the commands run (e.g. `/var/log/pgbackrest/mystanza_backup.log`)

7. Describe the issue:
When 1 one of node in 2 node patroni cluster is getting failed. Our pgbackrest backups are getting failed with below errors. I am aware  that pgbackrest checks for standby node everytime before performing backup. I have limitation that i cannot have extra node to make 3 node cluster I have to handle this case with 2 nodes. I have tested with --no-backup-standby option which is working. But I am looking for an better alternative solution avoid backup failure?

[...]

Issue #1550, opened 2021-10-27
Have pgbackrest exit silently when run against a standby node


df7cb:
We have a two-node Patroni cluster with the backup repository on S3. To keep the setup symmetrical I would like to be able to run the same cronjob/systemd timer on both nodes, but have the backup only suceed on the primary node. Since we don't need it (and generally don't want for security), there is no password-less ssh connection between the two nodes, so each of the pgbackrest.conf files has only `pg1` configured.

At the moment, this setup fails on the standby side:
```
$ pgbackrest --stanza=prod backup
ERROR: [056]: unable to find primary cluster - cannot proceed
```

My expectation would be that there would be a way to tell pgbackrest not to do anything when invoked on a non-primary node. At the moment I'm hacking around that by asking the Patroni API if the local node is a primary:
```
ExecStart=/bin/sh -c 'if curl -o /dev/null -sf localhost:8008; then pgbackrest --stanza=prod --type=full --log-level-console=info backup; else echo "Not running on primary node, skipping pgbackrest"; fi'
```
I believe the functionality should be built into pgbackrest so this command line shouldn't have to be that ugly.
We aren't asking for pgbackrest to talk to the Patroni API - just connecting to the local PG instance and exiting if it's not a primary would be perfect.

A side issue here is that I can't be explicit about not wanting to backup from the primary:
```
$ pgbackrest --stanza=valantic_prod --backup-standby=n backup
ERROR: [031]: option 'backup-standby' does not allow an argument
```

Thanks for considering.

dwsteele:
Everything starts as a question and we update when we decide for sure that it will be an enhancement. Just never got to this one.

mbanck:
Is there any update on it? Having a symmetric setup for a replication pair and only running backups on the primary strikes me as a valid use-case, and it seems to be really complicated with pgbackrest

dwsteele:
No update. The feature looks viable but we have not had time to schedule it yet.

You are welcome to put together a patch yourself.

dwsteele:
@df7cb @mbanck I know this issue is old as the hills but Stefan has implemented in it #2860. Please have a look if you can.

PR #2378, opened 2024-06-18
Allow requested standby backup to proceed with no standby.

dwsteele:
Add a "prefer" value to the backup-standby option to allow the backup to proceed when no standby is found. Note that this will not help if the standby is responding but fails to sync with the primary after the backup has started.

Introduce a new option modifier, bool-like, that allows a boolean option to be converted to a string or string-id option.


src/build/config/config.yaml
```diff
@@ -215,6 +215,10 @@ optionGroup:
 #      defined for the option command override, then they override command roles defined for the option (rule 2) above)
 #      and all roles defined for the command (rule 1) above).
 #
+# bool-like:
+#   Used to convert a boolean to a string or string-id type. It can been used as a boolean (e.g. --option or option=y) and the
```
pgstef, line 219: "it can be"?

dwsteele: I think it is just as good as "it may be".

pgstef: I was just nitpicking about the "it can be**en**" ;-)

pgstef:
I've just tested this feature and using `--backup-standby=prefer` will indeed make a backup with standby if reachable, and with primary if standby can't be reached.

I get the error just like what is described in the test.
```
WARN: unable to check pg1: [DbConnectError] unable to connect to 'dbname='postgres' port=5432 user='postgres' host='/var/run/postgresql'': connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory
      	Is the server running locally and accepting connections on that socket?
```

As far as I can read from the code change, the all logic is in dbGet. But we don't really do anything with the "prefer" value. This is just a "not no, not yes". The word in itself is totally meaningful and it guess it totally would worth documenting it for the users too (unless you want to keep it "internal" for now).

From the users perspective btw, the WARN isn't very meaningful I think. Couldn't we do something when "prefer" is used and no standby is found, just catch the WARN and adjust the error message? Just to let the user know that no standby was found and since prefer was used, the backup will be performed against primary? Or do you think it doesn't worth being noted in the logs?

src/build/config/config.yaml
dwsteele: Oh, I see it *now*! Fixed in cff6695.

dwsteele:
> it guess it totally would worth documenting it for the users too 

Agreed.

> From the users perspective btw, the WARN isn't very meaningful I think.

I'd like to keep the warn, but yeah I also think it makes sense to emit another message when we fail back from the primary.

I marked this WIP for now. Honestly, we are a bit skeptical of whether this is a good feature to have at all. If the standby(s) are down then putting sudden load on the primary is pretty much always a bad idea. The thing to do would be to fix the standbys and then run a manual backup if it seems like one is needed. Or, if you really need a backup and it is not possible to get the standbys backup up you can run the backup with --no-backup-standby.

The question is -- is there any time this feature would be beneficial? Is it worth degrading performance on the primary just to get a backup as long as WAL archiving is still working? Of course, the backup will fail anyway if it isn't.

pgstef:
> I marked this WIP for now. Honestly, we are a bit skeptical of whether this is a good feature to have at all. If the standby(s) are down then putting sudden load on the primary is pretty much always a bad idea. The thing to do would be to fix the standbys and then run a manual backup if it seems like one is needed. Or, if you really need a backup and it is not possible to get the standbys backup up you can run the backup with --no-backup-standby.
> 
> The question is -- is there any time this feature would be beneficial? Is it worth degrading performance on the primary just to get a backup as long as WAL archiving is still working? Of course, the backup will fail anyway if it isn't.

I guess there are 2 parts in that question. The first one is the new bool-like type. It is pretty interesting feature in itself imho.

`backup-standby=prefer` is one example of that bool-like possibility. Does it sounds bad to suddenly put pressure on the primary? Sure. But that's up to the user to take that decision, and imo, that would be adding a good flexibility in some automated fail-over cases. I at least could find 1 [GH issue](https://github.com/pgbackrest/pgbackrest/issues/2054) requesting for such feature...

And while enhancing the backup-standby mode, we could even probably think about [this](https://github.com/pgbackrest/pgbackrest/issues/1550) use-case. 

Anyway, fwiw, I'd be in favor of this feature since you still can keep the old behavior with `--no-backup-standby` ;-)

dwsteele:
> it guess it totally would worth documenting it for the users too

Added the docs.

> I'd like to keep the warn, but yeah I also think it makes sense to emit another message when we fail back from the primary.

I left the current warning in place but added another warning to indicate that the backup is being run from the primary because a standby could not be found.

This is ready for final review now.

src/command/check/check.c
```diff
@@ -101,7 +101,7 @@ checkStandby(const DbGetResult dbGroup, const unsigned int pgPathDefinedTotal)
         dbFree(dbGroup.standby);
     }
     // If backup from standby is true then warn when a standby not found
-    else if (cfgOptionBool(cfgOptBackupStandby))
+    else if (cfgOptionStrId(cfgOptBackupStandby) != CFGOPTVAL_BACKUP_STANDBY_N)
     {
         LOG_WARN("option '" CFGOPT_BACKUP_STANDBY "' is enabled but standby is not properly configured");
```
pgstef, line 106: Should we emit this warning if backup-standby=prefer? :thinking: 
By setting "prefer", the user accepts that there's no standby reachable, but in the other ends it's good to know that there's no standby available. It's just a small question, I don't have a strong opinion on this :shrug:

pgstef (approved):
Tested again and `--backup-standby` seems to work as expected.

```bash
$ pgbackrest backup --backup-standby=y --reset-pg2-host --reset-pg3-host ...

WARN: unable to check pg2: [DbConnectError] unable to connect to 'dbname='postgres' port=5432 host='/var/run/postgresql'': connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory
	Is the server running locally and accepting connections on that socket?
WARN: unable to check pg3: [DbConnectError] unable to connect to 'dbname='postgres' port=5432 host='/var/run/postgresql'': connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory
	Is the server running locally and accepting connections on that socket?
ERROR: [056]: unable to find standby cluster - cannot proceed
```
```bash
$ pgbackrest backup --backup-standby=prefer --reset-pg2-host --reset-pg3-host ...

WARN: unable to check pg2: [DbConnectError] unable to connect to 'dbname='postgres' port=5432 host='/var/run/postgresql'': connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory
	Is the server running locally and accepting connections on that socket?
WARN: unable to check pg3: [DbConnectError] unable to connect to 'dbname='postgres' port=5432 host='/var/run/postgresql'': connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory
	Is the server running locally and accepting connections on that socket?
WARN: unable to find a standby to perform the backup, using primary instead
...
INFO: backup command end: completed successfully
```

> Added the docs.

That part looks really good to me :+1: 

Except for the small question about the warning message in the check command, I think this PR is very nice :muscle:

src/command/check/check.c
dwsteele: I think check should definitely be warning here. Even when using prefer backing up from the primary is not ideal so we want to let the user know it will happen.

pgstef: Okay, good for me. I just wanted to make sure we thought about it :-)

PR #2379, opened 2024-06-20
Default log-level-stderr option to off.

dwsteele:
Writing warnings and errors to stderr by default leads to error messages being lost when the user does not correctly redirect stderr while generating logs for analysis. This happens so often that it seems worth changing the default to increase the quality of the logs we receive.

### DIFF
--- a/test/src/module/config/parseTest.c
+++ b/test/src/module/config/parseTest.c
@@ -1517,7 +1517,7 @@ testRun(void)
         TEST_RESULT_BOOL(cfgCommandHelp(), true, "help is set");
         TEST_RESULT_INT(cfgCommand(), cfgCmdNone, "command is none");
         TEST_RESULT_INT(hrnLogLevelStdOut(), logLevelWarn, "console logging is warn");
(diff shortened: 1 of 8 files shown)


turnstep (approved):
Looks fine to me

test/src/module/config/parseTest.c
```diff
@@ -1517,7 +1517,7 @@ testRun(void)
         TEST_RESULT_BOOL(cfgCommandHelp(), true, "help is set");
         TEST_RESULT_INT(cfgCommand(), cfgCmdNone, "command is none");
         TEST_RESULT_INT(hrnLogLevelStdOut(), logLevelWarn, "console logging is warn");
-        TEST_RESULT_INT(hrnLogLevelStdErr(), logLevelWarn, "stderr logging is warn");
+        TEST_RESULT_INT(hrnLogLevelStdErr(), logLevelOff, "stderr logging is warn");
```
pgstef, line 1520: small nitpick, shouldn't the comment also reflect the level off?

dwsteele: It should -- fixed in 5d27cdaa6051.

Issue #2381, opened 2024-06-20
Unable to setup sftp due to public key authentication failed: libssh2 error [-18]


thecatontheflat:
Hey team,

I can't seem to be able to configure sftp setup. 

Here is my config:

```
[xxx]
pg1-path=/xxx/xxx
pg1-port=11338

[global]
process-max=48

repo1-block=y
repo1-bundle=y
repo1-retention-full=1
repo1-path=/xxx
repo1-sftp-host=XX.XX.XX.XX
repo1-sftp-host-key-hash-type=sha1
repo1-sftp-host-user=root
repo1-sftp-private-key-file=/var/lib/postgresql/.ssh/id_rsa_sftp
repo1-sftp-public-key-file=/var/lib/postgresql/.ssh/id_rsa_sftp.pub
repo1-sftp-private-key-passphrase=XXX
repo1-sftp-host-key-check-type=none
repo1-type=sftp

compress-level=3
log-level-console=debug
archive-async=y
```

Debug info below when executing the `stanza-create` command:

```
ERROR: [101]: public key authentication failed: libssh2 error [-18]
       HINT: libssh2 compiled against non-openssl libraries requires --repo-sftp-private-key-file and --repo-sftp-public-key-file to be provided
       HINT: libssh2 versions before 1.9.0 expect a PEM format keypair, try ssh-keygen -m PEM -t rsa -P "" to generate the keypair
       --------------------------------------------------------------------
       If SUBMITTING AN ISSUE please provide the following information:

       version: 2.51
       command: stanza-create
       options: --exec-id=3976607-ce7e6c40 --log-level-console=debug --pg1-path=/xxx/xxx --pg1-port=11338 --repo1-path=/xxx --repo1-sftp-host=XX.XX.XX.XX --repo1-sftp-host-key-check-type=none --repo1-sftp-host-key-hash-type=sha1 --repo1-sftp-host-user=root --repo1-sftp-private-key-file=/var/lib/postgresql/.ssh/id_rsa_sftp --repo1-sftp-private-key-passphrase=<redacted> --repo1-sftp-public-key-file=/var/lib/postgresql/.ssh/id_rsa_sftp.pub --repo1-type=sftp --stanza=xxxxxx

       stack trace:
       storage/sftp/storage.c:storageSftpEvalLibSsh2Error:363:(test build required for parameters)
           ... function(s) omitted ...
[...]

PR #2385, opened 2024-06-24
Add hint to check SFTP authorization log.

thecatontheflat:
This is a follow up to this issue https://github.com/pgbackrest/pgbackrest/issues/2381


src/storage/sftp/storage.c
```diff
@@ -1343,7 +1343,8 @@ storageSftpNew(
                     "HINT: libssh2 compiled against non-openssl libraries requires --repo-sftp-private-key-file and"
                     " --repo-sftp-public-key-file to be provided\n"
                     "HINT: libssh2 versions before 1.9.0 expect a PEM format keypair, try ssh-keygen -m PEM -t rsa -P \"\" to"
-                    " generate the keypair"));
+                    " generate the keypair\n"
+                    "HINT: You may also want to check /var/log/auth.log of the SFTP server"));
```
dwsteele, line 1347: I'll defer to @jreidthompson on this, but I wonder if this message is a bit too specific RE the log file name. Maybe something like:
```
Check the authorization log on the SFTP server.
```
On the other hand, water it down this much I wonder if it is very useful.

jreidthompson: I agree that /var/log/auth.log is too specific, the log name and location can vary by OS, sftp server, or configuration, etc. I do think it would be useful. A quick check indicates that a google search for "$OS ssh authorization log file" or  "$DISTRO ssh authorization log" provides decent references for persons that may not be familiar with where logging is directed.

PR #2389, opened 2024-06-28
Allow backups to run concurrently on different repositories.

dwsteele:
The prior locking only allowed one backup per stanza, which was required by PostgreSQL <= 9.5 and didn't present a problem when only one stanza could be created.

Now that multiple stanzas are allowed relax this restriction so that backups can run concurrently for PostgreSQL > 9.5. To do this update the locking to be per stanza and repo rather than per stanza. Remotes are not aware of the repos that require locking so send an explicit list of files to be locked to the remote. Also remove the advisory lock for PostgreSQL > 9.5.

For info output the running backups are combined for progress output in order to avoid changing the JSON format. It probably makes sense to have per repo progress as well but I believe that would be better as a separate commit to avoid holding up this commit with bikeshedding.


src/command/info/info.c
```diff
@@ -1295,20 +1297,19 @@ infoUpdateStanza(
...
-                    if (stanzaRepo->backupLockHeld)
+                    if (lockResult.data.size != NULL)
                     {
-                        stanzaRepo->percentComplete = lockResult.data.percentComplete;
-                        stanzaRepo->sizeComplete = lockResult.data.sizeComplete;
-                        stanzaRepo->size = lockResult.data.size;
+                        ASSERT(lockResult.data.size != NULL);
+
+                        stanzaRepo->sizeComplete += varUInt64(lockResult.data.sizeComplete);
+                        stanzaRepo->size += varUInt64(lockResult.data.size);
```
jreidthompson, line 1312: These changed from '=' to '+=', do these run the risk of adding to random data in sizeComplete and size?

dwsteele: We'll hope there is no random data in the lock file. But the change to += was intentional since there can be multiple backups running we add together to the progress so we maintain the current format. We'll likely break these out in the future -- see PR comment.

jreidthompson (approved):
This looks good to me

pgstef:
> Now that multiple stanzas are allowed relax this restriction so that backups can run concurrently for PostgreSQL > 9.5. To do this update the locking to be per stanza and repo rather than per stanza. Remotes are not aware of the repos that require locking so send an explicit list of files to be locked to the remote. Also remove the advisory lock for PostgreSQL > 9.5.

```bash
# on PG host
$ cat /tmp/pgbackrest/ro9pg-backup-1.lock ; cat /tmp/pgbackrest/ro9pg-backup-2.lock 
{"execId":"54104-ae082574","pid":38138}
{"execId":"54110-8d0fb942","pid":38146}

# on repo host
$ cat /tmp/pgbackrest/ro9pg-backup-1.lock ; cat /tmp/pgbackrest/ro9pg-backup-2.lock 
{"execId":"54104-ae082574","pctCplt":1113,"pid":54104,"sz":9643689300,"szCplt":1073741824}
{"execId":"54110-8d0fb942","pctCplt":1113,"pid":54110,"sz":9643689300,"szCplt":1073741824}
```

Lock is successfully sent to the remote, and the % is only showed on the repo host, which is kinda expected. But maybe we could document that somehow? So people know they have to monitor that on the host where the backup is running?

> For info output the running backups are combined for progress output in order to avoid changing the JSON format. It probably makes sense to have per repo progress as well but I believe that would be better as a separate commit to avoid holding up this commit with bikeshedding.

% report combined for all running backups of the same stanza is fine as starting point imho

The PR looks good to me. It's a very nice improvement, Thanks!

pgstef (approved):

PR #2395, opened 2024-07-09
Simplify defaults and remove current in command-line help.

dwsteele:
The current values were misleading for any indexed option (e.g. pg-path) since it would only show the current value for the lowest index, e.g. pg1-path. This might be fixable but it would become quite verbose and probably was not that useful in the first place. The best way to get the values of all options used to run a command is to check the log.

The defaults were correct, but we are planning more advanced rules for defaults that would make them very difficult to display once defaults are not the same for all indexes in an option. More or less the same problem as above.

To solve these problems remove the current values everywhere and remove defaults in the option list. Defaults remain in the option description but they are stored in the help data rather than being fetched from the current configuration.

In addition to removing features that were of dubious value and allowing future improvements, this simplifies the config parser and everywhere the "none" command needed to be handled.

NOTE: There are a lot of indentation changes so best to review with whitespace changes turned off.


test/src/module/config/parseTest.c
```diff
@@ -87,7 +87,7 @@ testRun(void)
         // -------------------------------------------------------------------------------------------------------------------------
         TEST_TITLE("config command defaults to none before cfgInit()");
 
-        TEST_RESULT_UINT(cfgCommand(), cfgCmdNone, "command is none");
+        TEST_RESULT_UINT(cfgCommand(), cfgCmdHelp, "command is helps");
```
pgstef, line 90: typo? `helps` -> `help`

dwsteele: Whoops -- I fixed that and a few other comments but forgot to push it. Fixed in cd5eec846.

dwsteele:
> now the new way to get the default value ... would be to run help command option, right?

Yes. You could always get the default this way, but now it is the only way to get the default from the command-line.

> but also have something like 'pgbackrest help [options]' directly

I'm not too excited by this since a lot of stuff can vary by command, including defaults. I think making this context sensitive is better and something the web docs don't do as well. The options that vary by command are under "command" instead of "configuration" which I think is a bit confusing for users.

> would be happy to approve the PR if necessary

Approve away. I'm also hoping to get at least one more opinion to make sure I'm not upsetting people by removing this. I kind of doubt it, but you never know.

pgstef:
> I'm not too excited by this since a lot of stuff can vary by command, including defaults. I think making this context sensitive is better and something the web docs don't do as well. The options that vary by command are under "command" instead of "configuration" which I think is a bit confusing for users.

Right, that make sens to give the command context in the command line, and refer to the online doc for the global config.
Maybe it would be interesting to reduce a bit the commands online docs then? Like i.e. only keeping the command options and the options which would have a different default value for that specific command?

I'm fine with the current docs since it's quite complete and I'm used to it, but I can see why new users can be confused :shrug:

dwsteele:
> Maybe it would be interesting to reduce a bit the commands online docs then? Like i.e. only keeping the command options and the options which would have a different default value for that specific command?

This is on the list. As least to just link to the "configuration" page when it matches what would be on the "command" page.

src/config/config.c
```diff
@@ -113,7 +104,7 @@ cfgCommandName(void)
     FUNCTION_TEST_VOID();
 
     ASSERT(configLocal != NULL);
-    ASSERT(configLocal->command < cfgCmdNone);
+    ASSERT(configLocal->command < CFG_COMMAND_TOTAL);
```
pgguru, line 107: The previous `Assert()` on :65 had been:

```diff
-    ASSERT(configLocal->command <= cfgCmdNone);
+    ASSERT(configLocal->command < CFG_COMMAND_TOTAL);
```
Here, we're changing `< cfgCmdNone` to the same value.  I would assume that the difference here is that the argument here could never have been `cfgCmdNone`, so practically-speaking these are equivalent, but just wanted to call this out in case this `<` vs `<=` is significant.

pgguru (commented):
Playing a bit, seems like I can get an Assert failure, though admittedly with garbage input:

```console
$ ./src/pgbackrest help info --repo-type
ERROR: [031]: option '--repo-type' requires an argument

Uncaught AssertError: assertion 'configLocal != NULL' failed
    thrown at config/config.c:215
```

cbandy:
> > The best way to get the values of all options used to run a command is to check the log.
> 
> I agree that getting the current values is hard and best way it to actually run the command (maybe if we could have a default "dry-run" or "init-only" sub-command, it could help us here?).

I have my configuration arranged into multiple files in `conf.d`, and I've used the `help backup` command to see that they combine how I expect. For example,

```console
$ cat > 'pgbackrest.conf' <<'EOF'
[global]
exclude = main
exclude = main
EOF

$ cat > 'conf.d/one.conf' <<'EOF'
[global]
exclude = 1
EOF

$ pgbackrest --config-path="$(pwd)" help backup
…
[current=main, main, 1]
```

🤔 I would appreciate some way to see the resulting configuration that does not "actually take a backup."

src/config/config.c
dwsteele: The < vs <= difference is down to using a total vs the max enum. We could keep the same logic with `<= cfgCmdVerify` but then we'd need to worry about keeping that updated since there is no longer a deterministic last command. So we used the total and < instead.

dwsteele:
> Playing a bit, seems like I can get an Assert failure, though admittedly with garbage input

It was possible to get the assert with any command and an invalid option, not just help. And in a non-assert build it would be a segfault, which is pretty bad. Even garbage inputs should never result in a segfault.

I have fixed this in 440ce05541288 by reviving the logic that only runs cmdEnd() if the config was initialized. Since we don't have cfgCmdNone anymore I added cfgInited() to do that.

dwsteele:
> I have my configuration arranged into multiple files in conf.d, and I've used the help backup command to see that they combine how I expect.

So somebody *was* using this! Even so, the output is incorrect/misleading for any indexed option (pg*, repo*).

Another thing you could try is to do:
```
test/build/none/src/pgbackrest check --report
```
In your case this returns (you can make it more readable with jq or your favorite JSON tool):
```
{"cfg":{"env":{},"file":{"global":{"exclude":{"val":["main","main","1"]}}}}}
```
This does not give you the config from the point of view of a command, but it does show you where everything is coming from, i.e. env vs file.

pgstef:
> This does not give you the config from the point of view of a command, but it does show you where everything is coming from, i.e. env vs file.

`check --report` is definitely interesting since check can run on pg and repo host. but based on that idea, wouldn't it be nice to have some kind of `--dump-config` option which would be valid for all commands and return the config list like we have in the 1st INFO output of the command? :thinking: that idea kinda sound appealing to me :shrug:

dwsteele:
> wouldn't it be nice to have some kind of --dump-config option

Pretty much exactly what I was thinking about today. It could run with all commands and just dump out the config and exit. How about `config-dry-run`?

So instead of `backup command begin: ...` You'd get `backup command config: ...` logged. Or something like that. Seems pretty simple if we can manage not to bikeshed it to death.

dwsteele:
Maybe `dry-run-config` is better.

dwsteele:
> wouldn't it be nice to have some kind of --dump-config option

#2397 is a possible solution to this. Not sure if I am satisfied with it, though.

dwsteele:
Closing this since it represents a loss of functionality.

PR #2397, opened 2024-07-11
WIP: Add config-dry-run option.

dwsteele:
This option allows debugging of the configuration for any command:
```
pgbackrest --config-dry-run backup --stanza=test --log-level-console=info
```
outputs:
```
2024-07-11 14:38:34.082 P00   INFO: [DRY-RUN] backup command begin 2.53dev: --config-dry-run --exec-id=23275-1b2d660f --log-level-console=info --log-level-file=off --pg1-path=/pg1 --stanza=test
2024-07-11 14:38:34.082 P00   INFO: [DRY-RUN] backup command end: completed successfully (5ms)
```
This would seem to provide the functionality that is missing after #2395.

Not sure if I am entirely satisfied with this, but I figured I would get it out there for discussion.

### DIFF
--- a/src/build/config/config.yaml
+++ b/src/build/config/config.yaml
@@ -281,6 +281,12 @@ option:
     default: CFGOPTDEF_CONFIG_PATH
     default-literal: true
 
+  config-dry-run:
+    type: boolean
+    default: false
+    command-role:
+      main: {}
+
   dry-run:
     type: boolean
     default: false
--- a/src/build/help/help.xml
+++ b/src/build/help/help.xml
@@ -1970,6 +1970,16 @@
                     <example>/conf/pgbackrest/pgbackrest.conf</example>
                 </option>
 
+                <option id="config-dry-run" name="Config Dry Run">
+                    <summary>Configuration dry run.</summary>
+
+                    <text>
+                        <p>Log command begin/end and exit. Configuration errors and warnings will also be logged.</p>
+                    </text>
+
+                    <example>y</example>
+                </option>
+
                 <option id="config-include-path" name="Config Include Path">
                     <summary>Path to additional <backrest/> configuration files.</summary>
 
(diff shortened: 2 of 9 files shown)


pgstef:
I'm kinda afraid that only logging wouldn't be very convenient to handle, specifically since `log-level-console` isn't `info` by default. I initially thought we could try to dump out the options in some json format just like what you did for the report. That would be easier to manipulate (for humans or scripts/monitoring).

Maybe we could add something in `cmdBegin', reusing some logic of `cmdOption` to build the list of options and report it as json output too? Or would that be too much overhead?

dwsteele:
> I initially thought we could try to dump out the options in some json format just like what you did for the report.

I think that should be a separate command, not just an option to a regular command. Logging options, etc, need to be different and it gets pretty complicated -- not to mention changing some of the output.

However, I'm not too happy with this feature as is, either.

pgstef:
> I think that should be a separate command, not just an option to a regular command. Logging options, etc, need to be different and it gets pretty complicated -- not to mention changing some of the output.

Was there a specific reason why you've done `check --report` rather than having the report command directly?

> However, I'm not too happy with this feature as is, either.

In itself I think the ` --config-dry-run` is interesting, because it adds a basic dry-run for some commands that didn't had it yet. Like for backup and restore, getting the exact command that would be run can be very handy imo. But we might need to filter out the commands that aren't really meaningful with this option - i.e. `info` wouldn't return anything.

```bash
$ pgbackrest info --stanza=my_stanza --config-dry-run --log-level-console=info
$ pgbackrest check --stanza=my_stanza --config-dry-run --log-level-console=info
P00   INFO: [DRY-RUN] check command begin 2.53dev: --config-dry-run --exec-id=35029-7d9594c4 --log-level-console=info --log-level-file=detail --pg1-path=/var/lib/pgsql/16/data --repo1-path=/var/lib/pgbackrest --stanza=my_stanza
P00   INFO: [DRY-RUN] check command end: completed successfully (2ms)
```

Probably backup, expire and restore commands would be enough? Not sure about stanza-create, stanza-delete, stanza-upgrade and verify :thinking:

dwsteele:
> Was there a specific reason why you've done check --report rather than having the report command directly?

The check results will be integrated into the report, eventually, so this will be something user can send to us in the future. So, there would be a lot of overlap of there was a dedicated command.

Overall, I'm not too happy with this so I'm going to close it. I have some ideas on how we can handle this in help so I'll open a new PR for that after feature freeze.

PR #2398, opened 2024-07-15
Clarify archive-push multi-repo behavior.

dwsteele:
Clarify that archive-push will not be able to push ahead in other repositories if one is failing unless archive-async in enabled.


doc/xml/user-guide.xml
```diff
@@ -2395,7 +2395,9 @@
...
 
-        <p>The <cmd>archive-push</cmd> command will always push WAL to the archive in all configured repositories but backups will need to be scheduled individually for each repository. In many cases this is desirable since backup types and retention will vary by repository. Likewise, restores must specify a repository. It is generally better to specify a repository for restores that has low latency/cost even if that means more recovery time. Only restore testing can determine which repository will be most efficient.</p>
+        <p>The <cmd>archive-push</cmd> command will always push WAL to the archive in all configured repositories. When a repository cannot be reached, WAL will still be pushed to other repositories. However, for this to be work effectively <br-setting>archive-aync=y</br-setting> must be enabled, otherwise the other repositories can only get one WAL segment ahead of the unreachable repository. Also note that if WAL cannot be pushed to any repository then <postgres/> will not remove it from the <path>pg_wal</path> directory, which can cause the volume to run out of space.</p>
```
pgstef, line 2398: Hi,

I would suggest these little readability changes:
> However, for this to work effectively, <br-setting>archive-aync=y</br-setting> must be enabled; otherwise, the other repositories can only get one WAL segment ahead of the unreachable repository. Also, note that if WAL cannot be pushed to any repository, then <postgres/> will not remove it from the <path>pg_wal</path> directory, which can cause the volume to run out of space.

Otherwise this looks good to me :+1:

pgstef (changes requested):

doc/xml/user-guide.xml
dwsteele: This looks good. Updated in f63337b7a with minor edits, including bringing back the word PostgreSQL.

pgstef: Perfect :ok_hand:

doc/xml/user-guide.xml
```diff
@@ -2395,7 +2395,9 @@
...
 
-        <p>The <cmd>archive-push</cmd> command will always push WAL to the archive in all configured repositories but backups will need to be scheduled individually for each repository. In many cases this is desirable since backup types and retention will vary by repository. Likewise, restores must specify a repository. It is generally better to specify a repository for restores that has low latency/cost even if that means more recovery time. Only restore testing can determine which repository will be most efficient.</p>
+        <p>The <cmd>archive-push</cmd> command will always push WAL to the archive in all configured repositories. When a repository cannot be reached, WAL will still be pushed to other repositories. However, for this to work effectively, <br-setting>archive-aync=y</br-setting> must be enabled; otherwise, the other repositories can only get one WAL segment ahead of the unreachable repository. Also, note that if WAL cannot be pushed to any repository, then <postgres/> will not remove it from the <path>pg_wal</path> directory, which may cause the volume to run out of space.</p>
```
roberto-mello, line 2398: I feel like this could be worded more clearly and emphatically. Here's an attempt:
```
        <p>The <cmd>archive-push</cmd> command will always push WAL to the archive in all configured repositories. When a repository cannot be reached, only one WAL segment ahead of the unreachable repository will be pushed to the other repositories. When WAL cannot be pushed to any of the configured repositories, <postgres/> will not remove it from the <path>pg_wal</path> directory, which may cause the volume to run out of space.</p>
        <p>For <cmd>archive-push</cmd> to always push WAL to repositories even if one repository cannot be reached - which is the most likely desired behavior - <br-setting>archive-aync=y</br-setting> must be enabled. This is not default behavior for historical reasons, but it is recommended.</p>
```

roberto-mello (commented):
Suggestions for readability.

doc/xml/user-guide.xml
dwsteele: I prefer Stefan's version. If this were existing text that was being misunderstood I'd be OK with being more emphatic, but I think this is good for new text.

roberto-mello: Ok. Thanks for the review.

pgstef: To me, the suggested new phrasing suggests that the problem with archiving only 1 WAL archive is a pgBackRest problem, while it is actually depending on Postgres. And we don't state that clearly either. 
That's mostly why I don't like this new suggestion either tbh :worried:

PR #2403, opened 2024-07-19
Save backup.info only when contents have changed.

dwsteele:
It is (slightly) more efficient to avoid this save when the file has not changed.

More importantly, as we look to support versioning on object stores, it will be much easier to determine a good point-in-time to use for restore if there are no extraneous saves of backup.info. The idea is if corruption occurs in the repo (either accidental or malicious) the user would have the ability to restore from the repo as it looked at a prior point-in-time. Determining the point-in-time to use will be critical so the cleaner we can make the backup.info version history the easier it will be for the user.

In the short term this will also make development of the versioning feature a bit easier.


pgstef (commented):
Hm, what if there's something to fix in the backup.info without having a backup to expire?

A simple example:

```
P00   WARN: backup '20240719-080316F' missing manifest removed from backup.info
P00   INFO: expire command end: completed successfully
```

The output from `infoBackupLoadFileReconstruct` wouldn't be saved. Should we add a test for that btw? I'm surprised it wasn't catch already.

I'm not totally against this change but I kinda feel it would have bigger impacts. I know some people used this trick to re-import a backup set into the repo/backup.info too for example.

dwsteele:
Yeah, you are right. After I wrote this I started to think it would be better to have the InfoBackup object track changes (both add and remove) and then optionally skip the save (as a parameter to infoBackupSaveFile()). I'll work up something along those lines next week.

dwsteele:
> Should we add a test for that btw? I'm surprised it wasn't catch already.

Hmm, yeah, this seems like a good idea.

> I know some people used this trick to re-import a backup set into the repo/backup.info too for example.

I'm not to excited by the implications of that (i.e. they are copying backups around), but if we make the above work then I guess that works, too.

dwsteele:
> After I wrote this I started to think it would be better to have the InfoBackup object track changes

I've updated the code to work this way. It is ready for another look.

pgstef (approved):
In the use-case I've mentioned above, we then get:

```
WARN: [DRY-RUN] backup '20240904-091806F' missing manifest removed from backup.info
INFO: [DRY-RUN] expire command end: completed successfully
```
where the backup.info isn't updated, which is good since we're in dry-run mode, and without dry-run it gets updated:

```
WARN: backup '20240904-091806F' missing manifest removed from backup.info
INFO: expire command end: completed successfully
```

So the change looks good to me :+1:

dwsteele:
I forgot the release note so recommitted as b454b9e4035.

PR #2404, opened 2024-07-22
Allow repositories on versioned storage to be read at a target time.

dwsteele:
This feature allows the archive-get, info, repo-get, repo-ls, restore, and verify commands to operate at a point-in-time on versioned buckets in Azure, GCS, and S3. This allows recovery even if a repository has been accidentally or maliciously deleted or corrupted.

Documentation for the feature is here (https://github.com/dwsteele/pgbackrest/blob/dev-version/src/build/help/help.xml#L315) and here (https://github.com/dwsteele/pgbackrest/blob/dev-version/doc/xml/user-guide.xml#L2619).

This feature can be tested by using the documentation. First build the docs far enough to test:
```
pgbackrest/doc/doc.pl --out=html --include=user-guide --require=/quickstart/check-configuration --pre --no-cache --var=host-mem-limit=n --var=s3-all=y --var=encrypt=n
```
The commands can be run against the container `doc-pg-primary` container to test, e.g.:
```
docker exec -it doc-pg-primary pgbackrest --stanza=demo --repo=1 --repo-target-time="2024-08-29 03:18:00+00" info
```


turnstep:
Looks pretty good. Maybe have `--backup-version` instead of `--vers`?

This part looks the most wonky:
```
pgbackrest --vers --filter=^backup.info$ repo-ls backup/demo
```
So yes, I think we need some other way to store version information. I like solution 2. As you say, this is probably a rare and manual event that somebody needs to know the versions, so no big deal if we have to store a lot of them. But I wouldn't want that regex above as part of the official solution. :)

dwsteele:
> Maybe have --backup-version instead of --vers?

Well, this is general purpose, not just for backups. You could use it to look at any part of the tree, e.g. archive. I'm thinking `--versions` is best but we can't use that right now because we cheated with the `--version` option. I'm working on fixing that, though.

> pgbackrest --vers --filter=^backup.info$ repo-ls backup/demo

It's a bit primitive to be sure. `--filter` is optional but it narrows down the list a bit. So this works just fine but returns more data to sift through:
```
$ pgbackrest --vers repo-ls backup/demo
                                    20240720-085949F
                                    20240720-090112F
2024-07-20 09:01:14+00       1168   backup.info
2024-07-20 08:59:54+00       1168   backup.info
2024-07-20 08:59:30+00        480   backup.info
2024-07-20 09:01:14+00       1168   backup.info.copy
2024-07-20 08:59:54+00       1168   backup.info.copy
2024-07-20 08:59:30+00        480   backup.info.copy
```
I think if we document that as a quick way to find the appropriate time then it's acceptable, at least to start.

> So yes, I think we need some other way to store version information.

This is a separate issue from the above. Above we are finding a time where the repository is as we want it. That allows us to load the manifest for the backup we need. That manifest will have references to a bunch of files and we need to know their versions to read them. The question is do we use the time to find the correct version or always store the version in the manifest just in case we need it. At this point I'm leaning towards the former, storing versions is going to require pretty major manifest changes to store block incremental versions and in reality there generally only be one version of a file in a backup path. Except in the case of resume, of course. The path may have later been polluted but those timestamps should be measurably in the future, i.e. not in the same second as the good version that we want.

We could add a `--find-this-backup-set` option but not sure that's how this will work in practice. My guess is either the time the corruption happened will be known and you can go directly to running `info` with that timestamp, or the time won't be know and poking around will be required, which means searching for a specific backup set will not be useful. This ends up looking a lot like finding the appropriate time target for PostgreSQL PITR.

Maybe the answer is to add a command (or option to info using your idea) that just lists out all backups and the earliest timestamps at which they are available:
```
pgbackrest --stanza-demo --backup-version info
2024-07-20 09:01:14+00 20240720-090112F
2024-07-20 08:59:54+00 20240720-085949F
```
Maybe that's what you were suggesting above? However, it might be best to provide the basic functionality then add more focused options/tools as we see how people use the feature.

Thoughts?

dwsteele:
The `--vers` option has been renamed to `--versions`.

turnstep:
`pgbackrest --stanza-demo --backup-version info`  <-- I like that one a lot.

turnstep:
Random thought: should we have some sort of check/warning about somebody using what is not the latest version available? That's a pretty unusual use case. but maybe burying this in the power user section is sufficient.

Maybe not quite as obnoxious as patroni's "Please type 'Yes I am aware'" but a quick confirmation.

dwsteele:
> pgbackrest --stanza-demo --backup-version info <-- I like that one a lot.

We'll keep that in mind, then.

> Random thought: should we have some sort of check/warning about somebody using what is not the latest version available?

You'll automatically be using the latest version if `--limit-time` is not set. So you mean a warning in the log when `--limit-time` is set? That seems reasonable. Something like:
```
WARN: command is being performed on repoX as it appeared at 'SOME DATE/TIME'
```

pgstef:
> pgbackrest --stanza-demo --backup-version info <-- I like that one a lot.

Since WAL archives may also have versions, wouldn't `--object-versions` be more accurate? Even though I personally like `--versions`, that would be more consistent with the `list-object-versions`  s3api :thinking: 

> So you mean a warning in the log when `--limit-time` is set? That seems reasonable. Something like:
> 
> ```
> WARN: command is being performed on repoX as it appeared at 'SOME DATE/TIME'
> ```

That's a very good idea imho.

I played a bit with `repo-ls --versions` and `info --limit-time`, and that seems very promising!

```bash
$ pgbackrest info
<snip>
wal archive min/max (16): 00000002000000000000003D/000000030000000000000041

full backup: 20240731-152007F
    timestamp start/stop: 2024-07-31 15:20:07+00 / 2024-07-31 15:20:09+00
    wal start/stop: 00000002000000000000003D / 00000002000000000000003D
    database size: 181.2MB, database backup size: 181.2MB
    repo1: backup size: 11.6MB

incr backup: 20240731-152007F_20240731-152038I
    timestamp start/stop: 2024-07-31 15:20:38+00 / 2024-07-31 15:20:40+00
    wal start/stop: 000000030000000000000040 / 000000030000000000000040
    database size: 181.5MB, database backup size: 152.8MB
    repo1: backup size: 7.3MB
    backup reference list: 20240731-152007F

$ pgbackrest --versions --filter=^backup.info$ repo-ls backup/ro9pg
2024-07-31 15:20:41+00       2080   backup.info
2024-07-31 15:20:10+00       1248   backup.info
2024-07-31 15:19:56+00       2080   backup.info
2024-07-31 15:19:43+00       1248   backup.info
2024-07-31 15:19:25+00       2944   backup.info
2024-07-31 15:19:22+00       2064   backup.info
2024-07-31 15:19:20+00       1248   backup.info
2024-07-31 13:52:55+00       2080   backup.info

$ pgbackrest --limit-time="2024-07-31 15:20:10+00" info
<snip>
wal archive min/max (16): 00000002000000000000003D/00000002000000000000003D

full backup: 20240731-152007F
    timestamp start/stop: 2024-07-31 15:20:07+00 / 2024-07-31 15:20:09+00
    wal start/stop: 00000002000000000000003D / 00000002000000000000003D
    database size: 181.2MB, database backup size: 181.2MB
    repo1: backup size: 11.6MB

$ pgbackrest --limit-time="2024-07-31 15:19:56+00" info
<snip>
wal archive min/max (16): 000000020000000000000037/00000002000000000000003A

full backup: 20240731-151941F
    timestamp start/stop: 2024-07-31 15:19:41+00 / 2024-07-31 15:19:42+00
    wal start/stop: 000000020000000000000037 / 000000020000000000000037
    database size: 179.2MB, database backup size: 179.2MB
    repo1: backup size: 10.2MB

incr backup: 20240731-151941F_20240731-151954I
    timestamp start/stop: 2024-07-31 15:19:54+00 / 2024-07-31 15:19:56+00
    wal start/stop: 00000002000000000000003A / 00000002000000000000003A
    database size: 180.6MB, database backup size: 151.5MB
    repo1: backup size: 7.3MB
    backup reference list: 20240731-151941F
```
Being able to `verify` the repository state at a specific time/version will be handy to track corruption and find out the best time to go back to.

I'm wondering how it would react to multi-repo though. If we're restoring an old backup from last week's version and `archive-get` replays all the WAL archives from that repo, would it be able to see the latest version of the second repo to go ahead in WAL replay? (Or maybe we'll have to `--target-action=shutdown` and change `archive-get` time limit to go further)

dwsteele:
> Since WAL archives may also have versions, wouldn't --object-versions be more accurate? Even though I personally like --versions, that would be more consistent with the list-object-versions s3api

WAL archives may have versions, but it's unlikely since each WAL segment is written once. Anyway the WAL to recover is driven pretty much entirely by the restored backup, so that would be the focus here. I'm OK with just `--versions` as the option, though.

> I'm wondering how it would react to multi-repo though.

I think we need to force the user to select a single repo when using --limit-time. Some repo storage won't support versions (so you'd get errors) and other repos might not have it enabled (so inconsistent results which we should make errors).

> Being able to verify the repository state at a specific time/version will be handy to track corruption and find out the best time to go back to.

We have to assume that verify will not be able to detect all kinds of corruption. An attacker could corrupt/modify files and update the manifest to match. If this is just being used to recover a backup that was accidentally expired then verify would be fine, but I don't see that as the main use case of this feature.

In fact, I'm not sure that any query tools we can provide will be useful in the case of malware. In that situation the security team would need to do forensics and figure out what time it would be safe to read the repo at, i.e. by determining when the first breach happened, etc.

dwsteele:
@turnstep @pgstef I think this is done now. The big change is that I removed the ability to list versions with the `repo-ls` command. For now at least I think it makes sense to do this type of investigation with the vendor tools (aws, az, mc, etc). I'd rather see how people are using the tool rather than guess and provide not-quite-useful features that we need to maintain forever.

See the updated PR description for documentation and testing examples.

test/src/module/command/restoreTest.c
```diff
@@ -3229,14 +3256,12 @@ testRun(void)
...
-        hrnCfgArgRawZ(argList, cfgOptStanza, "test1");
-        hrnCfgArgRaw(argList, cfgOptRepoPath, repoPath);
-        hrnCfgArgRaw(argList, cfgOptPgPath, pgPath);
-        hrnCfgArgRawZ(argList, cfgOptSpoolPath, TEST_PATH "/spool");
-        hrnCfgArgRawBool(argList, cfgOptDelta, true);
-        hrnCfgArgRawZ(argList, cfgOptRepoCipherType, "aes-256-cbc");
-        hrnCfgEnvRawZ(cfgOptRepoCipherPass, TEST_CIPHER_PASS);
+        // For this restore to succeed we need to set limit time using the timestamp on backup.info
+        String *limitTime = strNewTimeP(
+            "%Y-%m-%d %H:%M:%S+00", storageInfoP(storageRepo(), INFO_BACKUP_PATH_FILE_STR).timeModified, .utc = true);
```
pgguru, line 3261: Any concerns about sub-second timing here and missing any files changed in the floor() of the second? (i.e., do we need second + 1 or something else involving milli-/microseconds here?)

test/src/module/integration/allTest.c
```diff
@@ -381,6 +381,11 @@ testRun(void)
             TEST_HOST_SQL_ONE_STR_Z(pg1, "select message from status", TEST_STATUS_TIME);
         }
 
+        // Store the time where the entire timeline captured in xidTimeline above exists so it can be used in a later test. We need
+        // to capture this here (instead of above) to be sure that all the timeline WAL has been archived.
+        const char *const xidTime = strZ(pckReadStrP(hrnHostSqlValue(pg1, "select current_timestamp::text")));
```
pgguru, line 386: What guarantees the TLI completion, just that we're using the same pg connection so it's guaranteed to be serial?

test/src/module/storage/azureTest.c
```diff
@@ -556,7 +556,7 @@ testRun(void)
                     "content-length: 0\n"
                     "date: <redacted>\n"
                     "host: %s\n"
-                    "x-ms-version: 2019-12-12\n"
+                    "x-ms-version: 2021-06-08\n"
```
pgguru, line 559: I assume this is some API version change that enables the features we're using for their storage protocol?

test/src/module/storage/azureTest.c
```diff
@@ -1026,6 +1026,113 @@ testRun(void)
...
                 TEST_RESULT_VOID(storagePathRemoveP(storage, STRDEF("/path"), .recurse = true), "remove");
 
+                // -----------------------------------------------------------------------------------------------------------------
+                TEST_TITLE("switch to time limited");
+
+                hrnServerScriptClose(service);
+
+                hrnCfgArgRawZ(argList, cfgOptPgPath, "/pg1");
+                hrnCfgArgRawZ(argList, cfgOptRepo, "1");
+                hrnCfgArgRawZ(argList, cfgOptLimitTime, "2024-08-04 02:54:09+00");
```
pgguru, line 1036: This is static because we're mocking the output from storage server?

src/build/config/config.yaml
```diff
@@ -867,6 +867,23 @@ option:
       restore: {}
       verify: {}
 
+  limit-time:
```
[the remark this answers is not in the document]
pgguru: `--storage-at-time`?  `--storage-pit`?  `--storage-snapshot`? Not really sure what the term of art is here, but might be worth a few more naming suggestions.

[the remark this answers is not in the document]
dwsteele: I'm not that excited about limit-time but not sure I like any of these options better. Is `repo-limit-time` any better?

src/command/repo/ls.c
```diff
@@ -62,6 +62,9 @@ storageListRenderInfo(const StorageInfo *const info, IoWrite *const write, const
         {
             ioWriteStr(write, strNewFmt(",\"size\":%" PRIu64, info->size));
             ioWriteStr(write, strNewFmt(",\"time\":%" PRId64, (int64_t)info->timeModified));
+
+            if (info->versionId != NULL)
+                ioWriteStr(write, strNewFmt(",\"version\":%s", strZ(jsonFromVar(VARSTR(info->versionId)))));
```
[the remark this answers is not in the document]
dwsteele: > so maybe this is just a catch-all for this sort of thing.

src/storage/list.c
```diff
@@ -138,11 +141,16 @@ storageLstInsert(StorageList *const this, const unsigned int idx, const StorageI
             }
 
             case storageInfoLevelBasic:
+            {
                 listInfo.basic.size = info->size;
                 listInfo.basic.timeModified = info->timeModified;
 
+                if (info->versionId != NULL)
```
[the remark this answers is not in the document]
dwsteele: This is an allocator to make storage of many files names more efficient. See common/type/blob.h. Arguably this should be built into memContext.c but we have not gotten there yet.

src/storage/posix/storage.c
```diff
@@ -356,6 +357,8 @@ storagePosixNewRead(THIS_VOID, const String *const file, const bool ignoreMissin
 
     ASSERT(this != NULL);
     ASSERT(file != NULL);
+    ASSERT(!param.version);
+    ASSERT(param.versionId == NULL);
```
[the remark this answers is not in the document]
dwsteele: We support that now, provided the user mounts a snapshot and presents it as the repo. But no, this commit does not introduce any special support.

test/src/module/command/restoreTest.c
dwsteele: We don't support subsecond timing since not all storage supports it. Added a clarification in c1176f227a3d.

test/src/module/integration/allTest.c
dwsteele: Yes, since we are on a new timeline here the old one should be archived. If something goes wrong with that then the later test will fail when it tries to follow that timeline. Since this is a test I'm not too worried about the that happening. What we need there is a time where we are pretty sure the prior timeline has been archived in its entirety.

test/src/module/storage/azureTest.c
dwsteele: Yes, this the minimum version that supports version listing with a delimiter, i.e. not having to list out every file in every subpath to get versions (yes, that's the way they implemented it originally).

test/src/module/storage/azureTest.c
dwsteele: Exactly. Makes the tests far simpler.

src/command/repo/ls.c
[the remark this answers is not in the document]
dwsteele: > Is the version the same thing as "the time to which we are limited"?

src/build/config/config.yaml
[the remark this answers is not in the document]
dwsteele: `repo-snapshot-time`?

test/src/module/command/restoreTest.c
pgstef: Hm, I've got some troubles to understand the clarification:

> Note comparisons to the storage timestamp are &lt;= the timestamp provided and milliseconds are truncated from the timestamp when provided.

Maybe I'm struggling with the grammar or punctuation, i.e. `Note: comparisons...` or `Note that comparisons...` would feel more natural to me :thinking:

src/build/config/config.yaml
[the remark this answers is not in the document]
pgguru: `repo-snapshot-time` sounds the best to me so far.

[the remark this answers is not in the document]
pgstef: if possible, I'd prefer to avoid using `snapshot here. While I understand why it might seem appropriate (viewing the repository as a capture at a specific point in time), I believe it could confuse users, as it could easily suggest that we support filesystem snapshots, a term widely used in sysadmin tasks :worried:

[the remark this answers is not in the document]
dwsteele: I think I agree with Stefan about `snapshot`. How about `repo-target-time`?

test/src/module/command/restoreTest.c
pgguru: So the `backup.info` file is the last thing written, so worst case it wouldn't see this as a valid backup (so would skip restoring) if you selected the timestamp that you completed at with partial milliseconds truncated—i.e., nothing inconsistent. 

I do wonder if a little stronger wording in the docs might be warranted, since if you saw a backup with milliseconds attached to the timestamp you'd need to choose a time at least a second after this, but maybe it's a similar issue with other PITR-related things so not worth stressing about.

src/build/config/config.yaml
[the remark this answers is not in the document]
pgguru: Yeah, that works for me, and seems a little clearer.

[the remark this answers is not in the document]
pgstef: Would work for me too :+1:

test/src/module/command/restoreTest.c
dwsteele: > Maybe I'm struggling with the grammar or punctuation

Definitely a typo on my part. Fixed in c8d450bd109a.

dwsteele: > so worst case it wouldn't see this as a valid backup (so would skip restoring) if you selected the timestamp that you completed at with partial milliseconds truncated—i.e., nothing inconsistent.

You would see it since we in effect round up. But yeah, in general either the backup will all be there or not be in backup.info.

> since if you saw a backup with milliseconds attached to the timestamp

src/build/config/config.yaml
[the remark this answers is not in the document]
dwsteele: Name change done here: 548ebfba014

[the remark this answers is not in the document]
pgguru: Reviewed that commit in isolation and it looked good (not on a computer so didn’t check for anything missed). Thanks!

test/src/module/command/restoreTest.c
pgguru: Ok so if we in effect round up not down I think my concern here is moot, but thanks for considering it. :)

src/storage/storage.c
```diff
@@ -71,9 +82,20 @@ storageNew(
...
         };
 
+        // Create storage cache list
+        if (this->targetTime != 0)
+            this->cacheList = lstNewP(sizeof(StorageListCache), .comparator = lstComparatorStr);
+
+        // If time limit is requested then versioning must be supported
+        CHECK(AssertError, targetTime == 0 || storageFeature(this, storageFeatureVersioning), "versioning required for time limit");
+
+        // If time limit is requested then storage must be write-only
```
pgguru, line 96:
```suggestion
        // If time limit is requested then storage must be read-only
```

src/storage/storage.intern.h
```diff
@@ -118,6 +118,10 @@ typedef struct StorageInterfaceNewReadParam
 
     // Limit bytes read from the file. NULL for no limit.
     const Variant *limit;
+
+    // File version to read (NULL for current version)
```
pgguru, line 122: nit: maybe have this right before the `String` def?

pgguru (approved):
Approved pending minor comment nits.

src/storage/storage.intern.h
dwsteele: The comment dates from before the boolean was added, so redid the comments in 3f02a3fea9e.

pgguru (approved):

PR #2416, opened 2024-07-31
Fix permissions when restore run as root user.

dwsteele:
When restore was run as the root user the pg_control file would end up with root permissions. This bug was introduced in e634fd85. Fix this by directly overwriting the pg_control temp file rather than doing an atomic write that updates permissions. Also update other parameters to more closely match similar calls.

There was also an adjacent error where restore as the root user would fail if the base path did not exist. Fix this by ignoring the missing path since it will be created later and this logic is just trying to find an alternate user for permissions if the user in the manifest does not exist.


test/src/module/integration/allTest.c
```diff
@@ -117,8 +117,48 @@ testRun(void)
...
+            // If the repo is running on pg2 then switch the user to root to test restoring as root. Use this configuration so the
+            // repo is local and root does not need to be configured for TLS or SSH.
+            const char *user = NULL;
+
+            if (pg2 == repo)
+            {
+                user = "root";
+
+                // Create the pg/ts directory so it will not be created with root permissions
+                HRN_STORAGE_PATH_CREATE(hrnHostDataStorage(pg2), strZ(hrnHostPgTsPath(pg2)), .mode = 0700);
```
pgstef, line 129: Is that really needed? Wouldn't this fix also make sure to create the missing path with proper user/perms?

pgstef (approved):
Tested a root restore with missing pgdata and confirmed the PR fix the described issue:

```
# pgbackrest restore --stanza=ro9pg
P00   INFO: restore command begin 2.53:
...
P00  ERROR: [041]: unable to get info for missing path/file '/var/lib/pgsql/16/main/data'
P00   INFO: restore command end: aborted with exception [041]

# sudo -iu postgres mkdir /var/lib/pgsql/16/main/data
# pgbackrest restore --stanza=ro9pg --type=immediate
P00   INFO: restore command begin 2.53:
...
P00   INFO: restore command end: completed successfully (1679ms)
# ls -l /var/lib/pgsql/16/main/data/global/pg_control 
-rw------- 1 root root 8192 Jul 31 10:22 /var/lib/pgsql/16/main/data/global/pg_control
```

```
# rm -rf /var/lib/pgsql/16/main/data
# pgbackrest restore --stanza=ro9pg
P00   INFO: restore command begin 2.54dev:
...
P00   INFO: restore command end: completed successfully (1816ms)
# ls -l /var/lib/pgsql/16/main/data/global/pg_control 
-rw------- 1 postgres postgres 8192 Jul 31 10:22 /var/lib/pgsql/16/main/data/global/pg_control
```

Except from one small question in the tests, this PR looks good to me :+1:

test/src/module/integration/allTest.c
dwsteele: This is not the base directory that pgbackrest will create (so we are testing that). This is higher level structure that we don't want created as root, which is what will happen if we are running as root.

PR #2422, opened 2024-08-14
Fix segfault on delayed connection errors.

dwsteele:
Connection errors could cause a segfault if the error was delayed enough to pass the initial call to sckClientOpenWait() and the error was instead thrown by a subsequent call to sckClientOpenWait(), which was not correctly initializing a variable required for error handling.

While this can be produced fairly easily in a test environment, I was unable to craft a unit test to hit this exact condition, probably due to timing. The new code still has full coverage and I added several comments to help prevent regressions.

### DIFF
--- a/doc/xml/release/2024/2.54.xml
+++ b/doc/xml/release/2024/2.54.xml
@@ -15,6 +15,19 @@
             </release-item>
 
             <release-item>
+                <github-issue id="2420"/>
+                <github-pull-request id="2422"/>
+
+                <release-item-contributor-list>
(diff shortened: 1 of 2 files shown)


pgguru (approved):
Makes sense; the issue is `openData` is required, but the call on the outer condition could fail, in which case it wouldn't be set.  So you reorganized to set it before the call that could error but had to unset afterwards since it is also used for signaling connection availability.

dwsteele:
Exactly that. Thanks for having a look!

PR #2429, opened 2024-08-28
Refresh web-id token for each S3 authentication.

### LINKED ISSUE #2428: pgbackrest doesn't refresh S3 tokens during execution
When using an S3 repo with web-id authentication, using time-limited tokens, the token can expire in the middle of a backup or restore. It looks like the S3 configuration is populated on first-use and is never updated. When running pgbackrest in kubernetes for example, the `AWS_WEB_IDENTITY_TOKEN_FILE` contains the token information, and is automatically updated when the token is renewed.

If pgbackrest gets an authentication error, it would be good to re-read the file and retry at least once in order to handle rotated tokens. Other options could be to re-read the file periodically, or stat it when building S3 requests and re-reading it if it's changed since the last time it was read?

If this seems reasonable, we could probably come up with a PR. Possibly adding an optional StorageInterface function to re-read credentials?

graveland:
The token file pointed to by the `AWS_WEB_IDENTITY_TOKEN_FILE` env var was read once at startup, but for long operations the token might expire before we're complete. This switches to reading it on each S3 request so we immediately use the renewed token. This is especially likely when running in kubernetes using a projected service account token.

Fixes #2428

I was having tests in GH CI timing out- I may have more work to do with testing. I've also still got to build and deploy a version with this PR in our k8s and see how that goes too.

### DIFF
--- a/doc/xml/release/2024/2.54.xml
+++ b/doc/xml/release/2024/2.54.xml
@@ -11,6 +11,18 @@
 
                 <p>Summarize backup reference list for <cmd>info</cmd> command text output.</p>
             </release-item>
+
+            <release-item>
+                <github-issue id="2428"/>
+                <github-pull-request id="2429"/>
+
+                <release-item-contributor-list>
+                    <release-item-contributor id="brent.graveland"/>
+                    <release-item-reviewer id="david.steele"/>
+                </release-item-contributor-list>
+
+                <p>Refresh web-id token for each <proper>S3</proper> authentication.</p>
+            </release-item>
         </release-improvement-list>
 
         <release-development-list>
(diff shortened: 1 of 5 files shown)


dwsteele:
Nice! That was simpler than I thought it would be. I fixed the test -- it was just a missing / in the path. If a test hangs run it with `--vm-out` and you can usually see what is going wrong.

graveland:
> Nice! That was simpler than I thought it would be. I fixed the test -- it was just a missing / in the path. If a test hangs run it with `--vm-out` and you can usually see what is going wrong.

That's a good thing to know, --vm-out would've made it much simpler no doubt.

Thanks!

dwsteele:
I'm just waiting on the results of your testing before committing this.

graveland:
> I'm just waiting on the results of your testing before committing this.

I'll do my best to get results to you today.

graveland:
> > I'm just waiting on the results of your testing before committing this.
> 
> I'll do my best to get results to you today.

A few more hours of my backup loop with short token lifetimes and I'll be able to say it's good.

graveland:
> > > I'm just waiting on the results of your testing before committing this.
> > 
> > 
> > I'll do my best to get results to you today.
> 
> A few more hours of my backup loop with short token lifetimes and I'll be able to say it's good.

200G database, 1/2 a CPU to keep it slow, one backup process:

2.53.1: failed 10 backups in a row with short token expiration times, zero successes
2.53.1+PR: succeeded 5 full backups with the exact same setup/same db as the first test, the only thing changed is this PR.

I think it's good to go!

dwsteele (approved):
Looks good! And thank you for the testing.

PR #2435, opened 2024-09-01
Remove limitation on reading files in parallel in restore file.

dwsteele:
This restriction prevented multiple files being read from a remote simultaneously, which was not supported. Although the limitation only applied to remotes, it was applied in all cases for testing purposes and because we planned to fix it.

Protocol command multiplexing added in df8cbc91 allows files to be read simultaneously from a remote so this restriction is no longer required.

Note that there is a test for this condition since the prior code had coverage. It might be tricky to ensure that test doesn't go away, but in general we should have enough tests in place to ensure simultaneous reads function as expected.


dwsteele:
For some reason Github got this PR conflated with a dev branch in my repo so it wasn't possible to get tests to pass. Closed in favor of #2436.

PR #2436, opened 2024-09-02
Remove limitation on reading files in parallel during restore.

dwsteele:
This restriction prevented multiple files being read from a remote simultaneously, which was not supported. Although the limitation only applied to remotes, it was applied in all cases for testing purposes and because we planned to fix it.

Protocol command multiplexing added in df8cbc91 allows files to be read simultaneously from a remote so this restriction is no longer required.

Note that there is a test for this condition since the prior code had coverage. It might be tricky to ensure that test doesn't go away, but in general we should have enough tests in place to ensure simultaneous reads function as expected.


pgguru (approved):
So just sanity-checking my understanding: it looks like this is just eliminating the check for the `blockIncrMapSize` (which was previously the limiting condition for preventing the restore).  Also since we don't need to consider the block incremental maps for this purpose any more, we can eliminate the close/subsequent reread of the file to get updated incrementals list, since we don't care at this point.

(The indentation viewing here is a bit wonky, so let me know if I've missed more changes/subtlety here, but that's my read on it.)

Assuming that's all the case, I think this makes sense.  Are there other implications from not closing/refreshing the repo file?

dwsteele:
> we can eliminate the close/subsequent reread of the file to get updated incrementals list, since we don't care at this point.

Not quite. This change means that block incremental maps no longer need to stop the larger sequential read on the repo bundle file. In practice this might not be helpful since if the next file is also bi we'll need to skip the block list but it still makes the code simpler. We construct these sequential reads containing many files to restore when possible to reduce the number of GETs on the repo, which is especially efficient for small files on object stores.

> Are there other implications from not closing/refreshing the repo file?

I think the large file is the more likely it is to have errors a force a retry of the entire bundle. I working on read retry to help with this, but for now it will still work just maybe be a bit slower when there are transmission errors.

dwsteele:
@pgguru Thoughts on my response? Make sense?

pgguru:
> @pgguru Thoughts on my response? Make sense?

I think I understand the reason for the code simplification (which is really what the PR is about), still chewing a bit on understanding:

> In practice this might not be helpful since if the next file is also bi we'll need to skip the block list

So I should definitely read more about bundles/incrementals; they just haven’t been on my radar to this point as a reviewers. 

So I am assuming a bundle file is effectively like a git pack file, consolidating multiple small files into a larger one so there are less required storage api API calls (which is again where I assume the 0-length bundle file optimization happens).

And incrementals basically are block range replacements, so I don’t need to re-store a 1 GB segment file if only a handful of pages have changed, just store the parent file reference and the block list/new blocks and handle it on object restore. (ISTR that block size for incrementals was more like 1MB and not strictly PG page size, tradeoff between redundant data and size of bi map I assume.)

I don’t think my current lack of detailed understanding needs to hold this up, will review the code for a good overview.

dwsteele:
> So I am assuming a bundle file is effectively like a git pack file, consolidating multiple small files into a larger one so there are less required storage api API calls

Correct.

> And incrementals basically are block range replacements, so I don’t need to re-store a 1 GB segment file if only a handful of pages have changed, just store the parent file reference and the block list/new blocks and handle it on object restore. (ISTR that block size for incrementals was more like 1MB and not strictly PG page size, tradeoff between redundant data and size of bi map I assume.)

Yes, though the block size varies from 8KiB to 88KiB based on the size/age of the file. The 1MiB you are thinking of is the superblock size for full backups (256KiB is used for diff/incr backups). The superblock combines blocks to improve compression efficiency. For that reason data can only be retrieved at the superblock level, even if only one block is needed.

> I don’t think my current lack of detailed understanding needs to hold this up, will review the code for a good overview.

OK, I'll go ahead and commit this, then. Thank you for the review!

PR #2439, opened 2024-09-11
Improve SFTP error messages.

jreidthompson:
Add output of the error number definition to error output. The numbers by themselves weren't informative and required looking into the libssh2_sftp.h header file for the definition.


src/storage/sftp/common.c
```diff
@@ -0,0 +1,114 @@
...
+
+        default:
+            errorMsg = "unknown error";
+            break;
+    }
+
+    return errorMsg;
+}
+
+#endif // HAVE_LIBSSH2
```
dwsteele, line 114: I don't think this function deserves its own module since it is only called in storage.c.

src/storage/sftp/common.c
```diff
@@ -0,0 +1,114 @@
...
+        case LIBSSH2_FX_NOT_A_DIRECTORY:
+            errorMsg = "LIBSSH2_FX_NOT_A_DIRECTORY";
+            break;
+
+        case LIBSSH2_FX_INVALID_FILENAME:
+            errorMsg = "LIBSSH2_FX_INVALID_FILENAME";
+            break;
+
+        case LIBSSH2_FX_LINK_LOOP:
+            errorMsg = "LIBSSH2_FX_LINK_LOOP";
```
dwsteele, line 103: Might be nice to translate these error constants into something a little more human readable. We already have the error code if we want to look it up.

src/storage/sftp/common.c
jreidthompson: ok - i'll move it

src/storage/sftp/common.c
jreidthompson: i'll update these. It was already in the back of my mind. Thanks

src/storage/sftp/common.c
dwsteele: Moved.

src/storage/sftp/common.c
dwsteele: Updated.

PR #2446, opened 2024-09-17
Add performance tuning section to user guide.

dwsteele:
We frequently tell users to enable to these options but they are spread through the documentation and not at all obvious. Hopefully putting them in quickstart will make them more visible, and it also gives us an easy place to link.


pgstef:
That's a good idea :+1: 

Should we also mention:
* repo1-cipher-type and repo1-cipher-pass
* start-fast
* log-level-console=info and log-level-file=detail ?

Those are the options I regularly see used but I understand everyone has its own default preferences ;-)

dwsteele:
> repo1-cipher-type and repo1-cipher-pass

These already have their own section in the quick-start.

> start-fast

This is also covered in quick start (6.8). It also doesn't fit in with the two groups in the new section so we'd need a separate group for it. I'd rather just leave it where it is since it is not as important (and can be changed at any time).

> log-level-console=info and log-level-file=detail 

I think the defaults are fine as is. `log-level-console=info` can fill up the postgres log with archive push messages and `log-level-file=detail` is orders of magnitude larger than `log-level-file=info`. If detail level were enough for debugging I might feel differently, but in general it is not.

dwsteele:
But since these options are performance oriented in general I wonder if we should call the section "Performance Configuration" or "Configuring for Performance". Might help get people's attention.

pgstef:
> I think the defaults are fine as is. `log-level-console=info` can fill up the postgres log with archive push messages and `log-level-file=detail` is orders of magnitude larger than `log-level-file=info`. If detail level were enough for debugging I might feel differently, but in general it is not.

I agree console=info is chatty for archive-push but it is pretty interesting for everything else. I get a lot of questions about "the process isn't doing anything" since it is not displaying anything in the console output and newcomers don't necessarily think about log files. I kinda find debug very chatty, so file=detail helped me a few times to avoid having to re-run the command with different log-level. 

But we probably all have our own preferences and usual recommendations. So I'm fine with only the options you suggested ;-)

> But since these options are performance oriented in general I wonder if we should call the section "Performance Configuration" or "Configuring for Performance". Might help get people's attention.

Makes sense indeed. To keep a direct tone as it would be between "Check the Configuration" and "Perform a Backup" iiuc, maybe "Tune settings for better performance"? (with or without "better" btw)

pgstef (approved):
`Performance Tuning` sounds good to me, good catch ! :+1:

dwsteele:
> Makes sense indeed. To keep a direct tone as it would be between "Check the Configuration" and "Perform a Backup" iiuc, maybe "Tune settings for better performance"? (with or without "better" btw)

How about just "Performance Tuning"? Updated as such.

dwsteele:
@pgstef Thank you for the review! I also pushed this up to the website so we don't have to wait to start using it.

PR #2453, opened 2024-09-24
Correctly display current values for indexed options in help.

dwsteele:
The current value for an indexed option was always for the first index, e.g. pg1-path. This is likely legacy from before indexing was added (and faithfully copied over from Perl, apparently).

Fix this by enumerating the current values in the option help and displaying <multi> in the option list when more than one value exists.

### DIFF
--- a/src/command/help/help.c
+++ b/src/command/help/help.c
@@ -240,7 +240,6 @@ helpRenderValue(const ConfigOption optionId)
             for (unsigned int optionIdx = 1; optionIdx < cfgOptionGroupIdxTotal(cfgOptionGroupId(optionId)); optionIdx++)
             {
                 const String *const value = helpRenderValueIdx(optionId, optionIdx);
-                // fprintf(stdout, "!!!RESULT %s VALUE %s\n", strZNull(result), strZNull(value));fflush(stdout);
 
                 if (!strEq(result, value))
                 {
--- a/src/command/help/help.c
+++ b/src/command/help/help.c
@@ -157,10 +157,10 @@ helpRenderText(
 }
 
 /***********************************************************************************************************************************
-Helper function for helpRender() to output values as strings
+Helper functions for helpRender() to output values as strings
 ***********************************************************************************************************************************/
 static String *
-helpRenderValue(const ConfigOption optionId, const unsigned int optionIdx)
+helpRenderValueIdx(const ConfigOption optionId, const unsigned int optionIdx)
 {
     FUNCTION_LOG_BEGIN(logLevelTrace);
         FUNCTION_LOG_PARAM(ENUM, optionId);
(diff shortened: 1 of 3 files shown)


dwsteele:
@pgstef @pgguru @cbandy This addresses the issues with current values brought up in #2395. However, instead of just removing them, I fixed them.

src/command/help/help.c
```diff
@@ -224,6 +224,42 @@ helpRenderValue(const ConfigOption optionId, const unsigned int optionIdx)
...
+            for (unsigned int optionIdx = 1; optionIdx < cfgOptionGroupIdxTotal(cfgOptionGroupId(optionId)); optionIdx++)
+            {
+                const String *const value = helpRenderValueIdx(optionId, optionIdx);
+
+                if (!strEq(result, value))
+                {
+                    MEM_CONTEXT_PRIOR_BEGIN()
+                    {
+                        strFree(result);
+                        result = strNewZ("<multi>");
```
pgguru, line 249: So this is the ultimate output if the values are different?  I guess I'd assumed it'd show all of the values in this case, but if we're able to query individually it's probably fine.

pgguru: (I see later that this is a sentinel value, so disregard.)

test/src/module/command/helpTest.c
```diff
@@ -584,6 +585,61 @@ testRun(void)
...
+        hrnCfgArgKeyRawZ(argList, cfgOptRepoHost, 1, "backup.example.net");
+        hrnCfgArgKeyRawZ(argList, cfgOptRepoHost, 3, "dr.test.org");
+        strLstAddZ(argList, "help");
+        strLstAddZ(argList, "restore");
+        strLstAddZ(argList, "repo-host");
+        TEST_RESULT_VOID(testCfgLoad(argList), "help for restore command, repo-host option");
+        TEST_RESULT_STR_Z(helpRender(helpData), optionHelp, "check text");
+
+        // -------------------------------------------------------------------------------------------------------------------------
+        TEST_TITLE("multiple current values (with one unset)");
```
pgguru, line 618: Not sure I understand the "unset" option here—just showing repo3's option?  Is the default for other repos supposed to be shown as well?

dwsteele: Just to make sure gaps are rendered correctly, e.g. `repo1-path`, `repo3-path`.

dwsteele: In other words, repo2 is skipped.

pgguru: Is it possible to set a global default per setting and override it on a repo-by-repo basis?  Is there a test exercising that if so? (Didn't seem to be in the new ones, which I'd expect if we're outputting for all of the known repos.)

dwsteele: > s it possible to set a global default per setting and override it on a repo-by-repo basis?

Issue #2458, opened 2024-10-08
Backup timeout issue with lot of databases (~6000)


NaPs:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version: 2.50, also tried on 2.53.1

2. PostgreSQL version: 13.6.1

3. Operating system/version: Ubuntu 20.04

4. Did you install pgBackRest from source or from a package? From PGDG repository

5. Please attach the following as applicable: n/a

7. Describe the issue:

PgBackRest fails to backup a PostgreSQL cluster with approximately 6000 databases due to a timeout error.

Environment:

- xx-db-1, xx-db-2 (PostgreSQL instances managed by pg_autofailover)
- xx-db-monitor (pg_autofailover cluster monitor and PgBackRest backup runner)

Error:

```
2024-10-07 01:30:01.586 P00   INFO: backup command begin 2.50: --backup-standby --compress-type=zst --exec-id=1834694-c57010f3 --pg1-host=xx-db-1 --pg2-host=xx-db-2 --pg1-host-ca-file=/etc/pgbackrest/cert/CA-cert.pem --pg2-host-ca-file=/etc/pgbackrest/cert/CA-cert.pem --pg1-host-cert-file=/etc/pgbackrest/cert/xx-db-monitor-crt.pem --pg2-host-cert-file=/etc/pgbackrest/cert/xx-db-monitor-crt.pem --pg1-host-key-file=/etc/pgbackrest/cert/xx-db-monitor-key.pem --pg2-host-key-file=/etc/pgbackrest/cert/xx-db-monitor-key.pem --pg1-host-type=tls --pg2-host-type=tls --pg1-path=/var/lib/postgresql/13/main --pg2-path=/var/lib/postgresql/13/main --protocol-timeout=3600 --repo1-block --repo1-bundle --repo1-path=/ --repo1-retention-archive=7 --repo1-retention-archive-type=diff --repo1-retention-diff=7 --repo1-retention-full=30 --repo1-retention-full-type=time --repo1-s3-bucket=xx-pg-pgbackrest --repo1-s3-endpoint=xxx.xxx.xxx.xxx --repo1-s3-key=<redacted> --repo1-s3-key-secret=<redacted> --repo1-s3-region=minio --repo1-s3-uri-style=path --repo1-storage-port=9000 --no-repo1-storage-verify-tls --repo1-type=s3 --stanza=xx --start-fast --stop-auto --type=diff
[...]

PR #2459, opened 2024-10-08
Fix PostgreSQL query performance for large datasets.

npdgm:
The asynchronous logic used to implement the query timeout is misusing `PQisBusy()`, which causes the wait handler throttling the consumption of command results. It can introduce a huge delay on a backup job up to `db-timeout` because of the back-off sequence. `PQconsumeInput()` appears to be reading 16KB at a time.

Following the recommendation of libpq, the fix is to watch the client socket for data availability and keep consuming results and checking for command busyness.


npdgm:
This is a proposed and tested fix for #2458

dwsteele:
We have a function available to do socket polling so I used that instead. The trick now will be getting the tests working but that shouldn't take too long.

npdgm:
> We have a function available to do socket polling so I used that instead. The trick now will be getting the tests working but that shouldn't take too long.

Neat! Thanks a lot for swift response.

npdgm:
We deployed this PR and are running tests. My coworker will keep you updated.

pgguru (approved):
The basic approach seems sane to me; look forward to hearing back on the status from the field.

dwsteele:
> look forward to hearing back on the status from the field.

See the linked issue (#2458). For their use case the patch is over a million times faster. Really just a comment on how inefficient the current code is.

PR #2470, opened 2024-10-17
Remove autoconf/make build.

dwsteele:
2.54.0 is the last release to support the autoconf/make build. Remove the autoconf/make build to reduce maintenance going forward.


pgguru (approved):
Looks sane here (code review only). The original Debian vs red hat logic was just related to using one of each for coverage on build systems I assume?

dwsteele:
> The original Debian vs red hat logic was just related to using one of each for coverage on build systems I assume?

Correct -- we had already cut over to mostly using meson for testing but we had to test autoconf/make *somewhere*.

That's really the biggest issue with keeping it around. Maintenance is not all that hard but testing both either blows up our test cases or gives short shrift to testing on one or both. All the packages that I am aware of have already made the switch so hopefully this is non-issue except for those building from source, and they'll just have to adapt.

dwsteele:
> Looks sane here (code review only).

This review is primarily a formality but extra eyeballs are always a good thing.

PR #2472, opened 2024-10-18
Remove support for PostgreSQL 9.4.

dwsteele:
Per our policy to support five EOL versions of PostgreSQL, 9.4 is no longer supported by pgBackRest. Remove all logic associated with 9.4 and update the tests.

This includes a small fix in infoPg.c to allow backup.info files with old versions to be saved. This allows expire to function when old versions are present. Even though those older versions cannot be used, they can be expired.

Tests for 9.4 are left in the expire/info tests to demonstrate that these commands work with old versions present.


pgstef (approved):
> This includes a small fix in infoPg.c to allow backup.info files with old versions to be saved. This allows expire to function when old versions are present. Even though those older versions cannot be used, they can be expired.

Would that change already be helpful today to manage i.e. `9.3`?

> Tests for 9.4 are left in the expire/info tests to demonstrate that these commands work with old versions present.

Would it make sense to add a comment somewhere in those tests to remember to update those next year?

The PR looks good to me otherwise :+1:

dwsteele:
> Would that change already be helpful today to manage i.e. 9.3?

Should work with 9.3 as well, though I don't want to explicitly test all old versions.

> Would it make sense to add a comment somewhere in those tests to remember to update those next year?

Yeah, good point. I have done that in 9ead1f50ba1e.

pgstef:
> Should work with 9.3 as well, though I don't want to explicitly test all old versions.

Sure, it makes sense to test `latest_supported_version - 1` but what I meant is that currently as we don't support 9.3 already, users with that version would have issues to expire backups and the suggested fix in this PR would improve the situation, not only for 9.4 but for the previous ones too. So that's a very good thing imho.

> Yeah, good point. I have done that in [9ead1f5](https://github.com/pgbackrest/pgbackrest/commit/9ead1f50ba1e54876d7130227b84af9add503e75).

Nice :+1:
LGTM then :-)

PR #2477, opened 2024-10-30
Fix issue with version/help commands attempting to load pgbackrest.conf.

dwsteele:
8d6bceb5 refactored version/help to operate more like regular commands partly to simplify the implementation of --version and --help. Unfortunately this had the side effect of these commands also loading pgbackrest.conf which would lead to an error if the file could not be read or parsed.

Add a filter to prevent version or help from loading pgbackrest.conf. Also prevent reads from the env to stop any warnings or errors from that source.


pgstef:
To be honest, I don't know what to think about this issue/use-case. I still fail to understand why people would want to run pgbackrest commands without having access to the config file **if it exists**.

I've tried to play a bit with this to understand the consequences:

```bash
# with 2.54.0
$ ls -l /etc/pgbackrest*
-rw-r----- 1 postgres postgres 1147 Oct 31 08:38 /etc/pgbackrest-custom.conf
-rw-r----- 1 root     root     1147 Oct 31 10:48 /etc/pgbackrest.conf

$ pgbackrest
P00  ERROR: [041]: unable to open file '/etc/pgbackrest.conf' for read: [13] Permission denied

$ pgbackrest version
P00  ERROR: [041]: unable to open file '/etc/pgbackrest.conf' for read: [13] Permission denied

$ pgbackrest help annotate                              
P00  ERROR: [041]: unable to open file '/etc/pgbackrest.conf' for read: [13] Permission denied

$ pgbackrest help annotate --config=/etc/pgbackrest-custom.conf | grep repo-cipher-type -A 1
  --repo-cipher-type                  cipher used to encrypt the repository
                                      [current=aes-256-cbc, default=none]
$ pgbackrest info
P00  ERROR: [041]: unable to open file '/etc/pgbackrest.conf' for read: [13] Permission denied

$ pgbackrest info --config=/etc/pgbackrest-custom.conf 
stanza: d11pg
    status: ok
    cipher: aes-256-cbc
...

$ sudo rm /etc/pgbackrest.conf 
$ pgbackrest info
No stanzas exist in the repository.
```
```bash
# with 2.55dev
$ ls -l /etc/pgbackrest*
-rw-r----- 1 postgres postgres 1147 Oct 31 08:38 /etc/pgbackrest-custom.conf
-rw-r----- 1 root     root     1147 Oct 31 11:03 /etc/pgbackrest.conf

$ pgbackrest
pgBackRest 2.55dev - General help
...

$ pgbackrest version
pgBackRest 2.55dev

$ pgbackrest help annotate                              
P00  ERROR: [041]: unable to open file '/etc/pgbackrest.conf' for read: [13] Permission denied

$ pgbackrest info
P00  ERROR: [041]: unable to open file '/etc/pgbackrest.conf' for read: [13] Permission denied
```

The PR effectively fixes the `pgbackrest`, `pgbackrest version` and `pgbackrest help` commands when the file exists.
It makes sense to throw the error with other commands like `info` because we need to load the config.

Although I'm unsure about `pgbackrest help subcommand`. It becomes logical to say that we through an error because we want to display the current values. But what if we don't want the current values, and just have the help message?
Couldn't we in this case just ignore the error and not load/display current values, only the default ones?
It would be easier to explain to people complaining to not see the current values to check their config file permissions than to explain to people complaining about the perm error why it makes sense to have the error?

dwsteele:
> I still fail to understand why people would want to run pgbackrest commands without having access to the config file if it exists.

I can see a use for monitoring or other purposes to make sure that pgbackrest is installed and has the correct version. In any case it is a regression from current behavior that is breaking tests downstream so I think it makes sense to fix it.

> Although I'm unsure about pgbackrest help subcommand. It becomes logical to say that we through an error because we want to display the current values. But what if we don't want the current values, and just have the help message?

This is a pre-existing issue so I am less worried about it. I tried to fix this by removing current value but that was not a popular move. In general fixing it is hard without making parsing a lot more complicated.

> Couldn't we in this case just ignore the error and not load/display current values, only the default ones?

Unfortunately no, because help depends on the config system to let it know which options are valid for which commands. This could be added separately to the help data but again that's a large-ish change.

> It would be easier to explain to people complaining to not see the current values to check their config file permissions than to explain to people complaining about the perm error why it makes sense to have the error?

Maybe? In this case I think we'd still want to display an error but then also continue displaying the help. Either way it is confusing.

pgstef (approved):
> This is a pre-existing issue so I am less worried about it. I tried to fix this by removing current value but that was not a popular move. In general fixing it is hard without making parsing a lot more complicated.

Thanks for the explanation (and sorry for the late reply). Given your feedback, the suggested fix looks good to me then :+1:

Issue #2463, opened 2024-10-11
Instability of allTest.c


KnightMurloc:
I have investigated the frequent test failures in the allTest.c file and found the reason for their instability.
The following confluence of circumstances leads to a fail:
1. After restoring the Postgres instance, the recovery process starts. Which calls the asynchronous archive-get.
2. After archive-get checks the existence of the queue directory, but before writing the WAL file, there are restores in the next test is begun. Which leads to the deletion of the queue directory.
3. Since the directory no longer exists, writing the WAL file will fail, and archive-get will write the error file to the queue.
4. A new Postgres instance will start, and the recovery process will begin, which requests the WAL file.
5. The new archive-get looks into the queue directory, finds the error file, and throws out the error, after which the Postgres recovery fails. Because the previous archive-get background process has not finished yet.

I managed to reproduce this problem manually on the main branch. A recent patch (https://github.com/pgbackrest/pgbackrest/commit/1020bc117ab2e116abb41eebd40216726a3d2561) may have masked the problem.
It seems that the problem is not related to the Postgres version, so it may be worth making more general fix. Would you be interested in accepting such a patch?
Perhaps it is also worth prohibiting the simultaneous work of archive-get and restore on the same stanza?

dwsteele:
I've been trying for forever to reproduce this in a test env and finally just gave up and shuffled versions around so it would work -- guessing (wrongly it sounds) that new postgres on older Debian was causing some timing issues.

> It seems that the problem is not related to the Postgres version, so it may be worth making more general fix. Would you be interested in accepting such a patch?

I would definitely be interested in a patch.

> Perhaps it is also worth prohibiting the simultaneous work of archive-get and restore on the same stanza?

Maybe -- but we would need a lock and we don't currently lock on read-only ops. For this we'd need a different lock architecture I think, since it would be valid to do multiple restores from a single stanza as long as the target paths are different.

dwsteele:
Closing since there is a PR.

PR #2481, opened 2024-11-06
Stabilize async archiving in integration tests.

KnightMurloc:
In the next set of circumstances, the test could fail:

1. After restoring the Postgres instance, the recovery process starts. Which calls the asynchronous archive-get.
2. After archive-get checks the existence of the queue directory, but before writing the WAL file, there are restores in the next test is begun. Which leads to the deletion of the queue directory.
3. Since the directory no longer exists, writing the WAL file will fail, and archive-get will write the error file to the queue.
4. A new Postgres instance will start, and the recovery process will begin, which requests the WAL file.
5. The new archive-get looks into the queue directory, finds the error file, and throws out the error, after which the Postgres recovery fails. Because the previous archive-get background process has not finished yet.

This patch fixes the problem by using a separate spool directory for each test.

Related issue: https://github.com/pgbackrest/pgbackrest/issues/2463

### DIFF
--- a/test/src/module/integration/allTest.c
+++ b/test/src/module/integration/allTest.c
@@ -44,9 +44,6 @@ Test statuses
 
 #define TEST_RESTORE_POINT                                          "pgbackrest"
 
-#define UPDATE_SPOOL_PATH() \
-    hrnHostConfigUpdateP(.archiveAsync = BOOL_TRUE_VAR, .spoolPathPostfix = VARSTR(STRDEF(STRINGIFY(__COUNTER__))))
-
 /***********************************************************************************************************************************
 Test Run
 ***********************************************************************************************************************************/
@@ -291,9 +288,6 @@ testRun(void)
             // Stop the cluster
             HRN_HOST_PG_STOP(pg1);
 
-            // Set unique spool path
-            UPDATE_SPOOL_PATH();
-
             // Restore
             TEST_HOST_BR(pg1, CFGCMD_RESTORE, .option = zNewFmt("--force --repo=%u", hrnHostRepoTotal()));
             HRN_HOST_PG_START(pg1);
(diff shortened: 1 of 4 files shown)


test/src/module/integration/allTest.c
```diff
@@ -288,6 +291,9 @@ testRun(void)
             // Stop the cluster
             HRN_HOST_PG_STOP(pg1);
 
+            // Set unique spool path
+            UPDATE_SPOOL_PATH();
+
```
dwsteele, line 296: I think it would be simpler to build this logic into TEST_HOST_BR() when command is CFGCMD_RESTORE. That will make it less likely to be forgotten.

test/src/module/integration/allTest.c
```diff
@@ -44,6 +44,9 @@ Test statuses
 
 #define TEST_RESTORE_POINT                                          "pgbackrest"
 
+#define UPDATE_SPOOL_PATH() \
+    hrnHostConfigUpdateP(.archiveAsync = BOOL_TRUE_VAR, .spoolPathPostfix = VARSTR(STRDEF(STRINGIFY(__COUNTER__))))
+
```
dwsteele, line 49: I'm not sure how I feel about using a GCC extension here. In the core code this would definitely not be OK, but even in the tests I feel like we should avoid it. If we build the logic into TEST_HOST_BR() (see below) I think the macro is less compelling overall.

dwsteele (changes requested):
Thanks for doing this! I like the idea but see comments about implementation.

dwsteele (approved):
I made some changed in https://github.com/dwsteele/pgbackrest/tree/ADBDEV-6654-ci and added a release note. Go ahead and pull the changes in if they look good to you.

dwsteele:
I was about to commit this when I realized that the spool path was not being changed for archive-push since it was only being passed to restore (which passes it to archive-get). I think it is better to update the config as I have done in the new commit here: https://github.com/dwsteele/pgbackrest/tree/ADBDEV-6654-ci.

Please have a look and pull this in.

PR #2488, opened 2024-11-14
Use lz4 for protocol compression.

dwsteele:
Now that lz4 is required it makes sense to use it for protocol compression.


src/build/config/config.yaml
```diff
@@ -767,8 +767,8 @@ option:
   compress-level-network:
     section: global
     type: integer
-    default: 3
-    allow-range: [0, 9]
+    default: 1
+    allow-range: [-5, 12]
```
pgstef, line 771: `[-5, 12]` this meets the boundaries we have in `compress.h`. I've struggled to find this information on the lz4 repo but out of curiosity, why the `-5`?

pgstef:
> Now that lz4 is required it makes sense to use it for protocol compression.

Linked to the c678543 commit which removed conditional compilation for lz4 I assume. This change makes sense then :+1: 
The code change LGTM, I'll try to do some tests to see if I can easily spot a performance difference ;-)

src/build/config/config.yaml
dwsteele: I found these values by digging pretty deep into the lz4 code a few years ago (we are already using these values for compress-level). The max value was fairly easy to find, the min value much less so. Anyway, the negative values enable a special high-speed mode, which could be useful for protocol compression especially.

pgstef: Thanks for the explanation :+1:

pgstef (approved):
I've played with this a little bit but given my test environment overall stays local, network compression should be barely noticeable. I still found out an improvement (a few seconds on >10G dataset) so it's quite impressive IMHO. I haven't spotted any regression either on my usual test cases and as mentioned previously already, the code change LGTM.

So this PR is good to go for me :+1:

dwsteele:
In my tests using lz4 for protocol compression is twice as fast as gz for a 14.6GB database (21 seconds vs 42 seconds). Protocol compression is only enabled when `compress-type=none` since it would not make sense to compress already compressed data.

pgstef:
> In my tests using lz4 for protocol compression is twice as fast as gz for a 14.6GB database (21 seconds vs 42 seconds). Protocol compression is only enabled when `compress-type=none` since it would not make sense to compress already compressed data.

That's impressive indeed. I've tried with `compress-type=none` and the same `compress-level-network`. With a backup taken in less than 1 minute on fast storage and test vms on the same host, seeing a difference of several seconds was already quite impressive imho, but that will definitely be even better on real use-cases :+1:

dwsteele:
> the same compress-level-network

That could be some of the difference.  I used the default `compress-level-network` value in both cases. I think this is likely to be how it is set in the field.

PR #2517, opened 2025-01-07
Improve full/incr backup limit time for preliminary copy.

dwsteele:
Two enhancements to improve limit time:

1) If start-fast is enabled then checkpoint before the preliminary copy. This handles cases where checkpoint-timeout is very large and would force the preliminary copy to operate far in the past.

2) Calculate limit time from the current time and move it before the checkpoint only if it would fall after the checkpoint. This results in a more accurate limit time than before.


dwsteele:
Closing since #2588 reverts the feature this is based on.

Issue #2410, opened 2024-07-25
Request for consistent semantic version number and server_version_num equivalent


keithf4:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version: 2.52.1, 2.53

2. PostgreSQL version: N/A

3. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each: N/A

4. Did you install pgBackRest from source or from a package? Package

5. Please attach the following as applicable: N/A

7. Describe the issue:

In pgMonitor we have a metric for the pgBackRest version just in case that was something people wanted to monitor for (watching for one system to upgrade while others didn't and breaking things). With version 2.52.1 that broke because previously all version numbers had been ##.## and easily converted to a float. With an actual 3 digit semantic version, that doesn't work, but that can, and will, be fixed on pgMonitor's end.

However, I see that version 2.53 went back to a 2 digit version. If semantic versioning is going to be used, can it stay consistent with the 3 version number? Ex. 2.53.0. It appears version 2 kept with the two digit version at that time. Ex. 2.00

As an added bonus, maybe also add something like the equivalent of the `server_version_num` in core postgres that always gives back an integer value for the version. This is what we plan to do in pgMonitor to handle things going forward, but a value in the json output like this would be great.

```
postgres=# SELECT (split_part('2.52.1', '.', 1) || lpad(split_part('2.52.1', '.', 2), 2, '0') || lpad(coalesce(nullif(split_part('2.52.1', '.', 3), ''), '00'), 2, '0'))::int AS backrest_repo_version;
-[ RECORD 1 ]---------+------
backrest_repo_version | 25201

postgres=# SELECT (split_part('2.53', '.', 1) || lpad(split_part('2.53', '.', 2), 2, '0') || lpad(coalesce(nullif(split_part('2.53', '.', 3), ''), '00'), 2, '0'))::int AS backrest_repo_version;
-[ RECORD 1 ]---------+------
backrest_repo_version | 25300

postgres=# SELECT (split_part('3.0', '.', 1) || lpad(split_part('3.0', '.', 2), 2, '0') || lpad(coalesce(nullif(split_part('3.0', '.', 3), ''), '00'), 2, '0'))::int AS backrest_repo_version;
-[ RECORD 1 ]---------+------
backrest_repo_version | 30000

[...]

PR #2526, opened 2025-01-20
Add numeric output to version command.

pgstef:
Hi :wave: 

This is a WIP ;-)

- Added `info --output=json` fields `"backrest":{"version":"2.55dev","version-num":"025500"}`
- Not really knowing where it would be best to place it, the convert function is still is the info command
- Tried adding a `version --output=num` option to return the same and use the above function

At the moment, I can build the code manually but CI seems to fail on `error: ‘cfgOptOutput’ undeclared`. Not sure what I missed regarding the help/version command :-(


dwsteele:
>  I can build the code manually but CI seems to fail on error: ‘cfgOptOutput’ undeclared

This failure happened when building `test-pgbackrest`. The test and doc code also use the help/version code so they need the options added as well.

I also simplified the code by adding a constant for version num.

I'm not sure I'm entirely happy with the info output, but go ahead and fix the coverage for the help module while I think about it.

src/version.h
```diff
@@ -31,13 +31,14 @@ will be invalid unless migration functions are written.
...
-Project version components. PROJECT_VERSION is automatically generated from the component parts.
+Project version components. PROJECT_VERSION and PROJECT_VERSION_NUM are automatically generated from the component parts.
 ***********************************************************************************************************************************/
 #define PROJECT_VERSION_MAJOR                                       2
 #define PROJECT_VERSION_MINOR                                       55
 #define PROJECT_VERSION_PATCH                                       0
 #define PROJECT_VERSION_SUFFIX                                      "dev"
 
 #define PROJECT_VERSION                                             "2.55.0dev"
+#define PROJECT_VERSION_NUM                                         2055000
```
pgstef, line 42: you're counting 1 digit for major, and 3 digits for minor and patch then?

pgstef:
Thanks for having a look at this :pray: 

> This failure happened when building `test-pgbackrest`. The test and doc code also use the help/version code so they need the options added as well.

I remember I tried adding it to the doc config.yaml/help.xml but not to the help one :facepalm:

> I also simplified the code by adding a constant for version num.

That was my initial idea but I wasn't sure you'd want to keep that up to date so I went for the extract function. But obviously, that makes it much simpler with the constant.

> I'm not sure I'm entirely happy with the info output, but go ahead and fix the coverage for the help module while I think about it.

As the json output loops over each stanza, it felt weird to me to have the same output in all the stanzas. That's why I went for the version option initially. Monitoring the version installed on your hosts separately with the version command seemed more reliable to me. But as the initial request was to add to the json info, I guess having it all in one command might feel easier to handle.

Regarding the help test, I tried adding something simple ([31ac9f4](https://github.com/pgbackrest/pgbackrest/pull/2526/commits/31ac9f4493f996fe3ad349412589c6fd21e979fc)) but that broke more tests with `BUT GOT AssertError: expected return type 'String' already found but also found 'String'` so I reverted it and will need to have a closer look at it next week :sweat_smile:

src/version.h
dwsteele: Yes -- for future proofing. We are already at 2.55.

dwsteele: Also, there are many digits as we like to major version -- we just don't prepend 0s. You'll notice that I also changed version-num to a number in the JSON. This makes comparisons easier and is equivalent to what Postgres does.

pgstef: Cool. Thanks for the clarification :+1:

dwsteele:
> That was my initial idea but I wasn't sure you'd want to keep that up to date so I went for the extract function.

This is automatically updated in test.pl so there is nothing to maintain.

> As the json output loops over each stanza, it felt weird to me to have the same output in all the stanzas.

Same here, but I can't see anywhere else to put it.

> AssertError: expected return type 'String' already found but also found 'String'

This is because the strNewFmt outside of the MEM_CONTEXT_TEMP block was leaking a string. Fixed that.

pgstef:
> This is because the strNewFmt outside of the MEM_CONTEXT_TEMP block was leaking a string. 

Oh, okay. Good to spot it with the tests then. That's again a good reason to add tests :laughing: 

> Fixed that.

Thanks again for your help on this PR :pray: I definitely learned a lot on this :bowing_man:

pgstef:
> Same here, but I can't see anywhere else to put it.

First of all, I must admit I'm pretty happy with `version --output=num`. I believe it will be very efficient to monitor the version installed on the various hosts used and may help detect potential mismatch earlier and easier.

Although, I've read again the initial [issue](https://github.com/pgbackrest/pgbackrest/issues/2410) and given what @keithf4 describes there and having a look at the pgmonitor [queries](https://github.com/CrunchyData/pgmonitor/blob/main/postgres_exporter/linux/queries_backrest.yml#L23), I feel like they're currently extracting `backrest_repo_version` from the `backup:backrest:version` json key. That's then the version which was used to take the backup, not the version currently installed. As we have the limitation on having same version on PG+repo, "repo version" sounds accurate enough.

I'm now wondering if we shouldn't consider to add the version num to `InfoBackupData` too :thinking:
With that value, we could then compare the backup version with the currently installed version. I would then ask if comparing against `version --output=num` would be easy for them to add or if we should keep the new info json field too.

That's maybe a bigger change but I kinda feel it would worth it :man_shrugging: What do you think?

dwsteele:
> First of all, I must admit I'm pretty happy with version --output=num. I believe it will be very efficient to monitor the version installed on the various hosts used and may help detect potential mismatch earlier and easier.

Same here. As far as I'm concerned that is ready to commit.

I'm a little less happy with the way the version is expressed in info per stanza but I don't see that we have any choice.

> I would then ask if comparing against version --output=num would be easy for them to add or if we should keep the new info json field too.

I pinged @keithf4 about it, so hopefully we'll find out.

> I'm now wondering if we shouldn't consider to add the version num to InfoBackupData too

That's a good idea. I'd actually want to add it to the manifest at backup time so we don't try to go back and translate old version numbers. I worried about a problem there that would cause errors (especially in forks). I think having version_num for backups going forward is fine.

pgstef:
> I'm a little less happy with the way the version is expressed in info per stanza but I don't see that we have any choice.

> I'd actually want to add it to the manifest at backup time so we don't try to go back and translate old version numbers. I worried about a problem there that would cause errors (especially in forks). I think having version_num for backups going forward is fine.

I believe adding the version in the backup manifest for future backups and report that in the per backup json field rather than per stanza would make sense. So you know what version was use to generate the backup and use the version command to get the locally running version to compare both info :thinking:

pgstef:
> I think we need both -- the version when a backup was made _and_ the current version. The latter needs to be per stanza due to limitations of the format.

Attempted to add it in 38cb4cc. Hopefully the test should run fine :crossed_fingers:

pgstef:
> I've been thinking about this and for now I'd rather just go with the --output=num feature for the version command. I'm ready to commit that right now, but still not satisfied with the rest. Not your fault -- I'm just uncomfortable with how this has to be jammed in given the current format.

Same feeling here. I like the `version --output=num` too :+1: 

> Could you update this PR to just have that initial feature? Keep your original branch as well because we may come back to that in time.

Ok, I'll try to do that asap

dwsteele (approved):
> Except for the arch (--platform=linux/ppc64le ubuntu:22.04) test which fails for https://github.com/pgbackrest/pgbackrest/issues/2568 reason

This has been fixed and merged into the commit.

The rest looks good to me and is ready to commit. Thank you!

PR #2528, opened 2025-01-21
ADBM-1970: Check corresponding WAL chain during verify of backup

silent-observer:


dwsteele:
@silent-observer I assume you opened this by accident?

silent-observer:
Yes, sorry

dwsteele:
If you decide to contribute this upstream we'll want to talk about it first -- just open an issue. At a quick glance there are parts that look interesting but I'm not sure we would want the table filtering. As I recall, @x4m cast some doubt on the utility (vs added complexity) of this feature during his talk in Athens last year.

x4m:
Yes, IMO user should specify not what to restore, but what to excluse. Either way you have to do a relfilenode dance, which may be changed during recovery history.

PR #2533, opened 2025-01-22
Do not set recovery_target_timeline=current for PostgreSQL < 12.

dwsteele:
PostgreSQL < 12 defaults recovery_target_timeline to current but if current is explicitly set it behaves as if latest was set. Since current is not handled in the PostgreSQL code it looks as if there should be an error during the integer conversion but that doesn't happen due to incorrect strtoul() usage (not checking endptr).

Handle this by omitting recovery_target_timeline from recovery.conf when it is explicitly set by the user to current.


pgstef (approved):
> Do not set current when PostgreSQL < 12 since this is the default and if current is explicitly set it acts as latest

The comment explain it pretty good imo, so the change LGTM :+1:

PR #2534, opened 2025-01-22
Verify recovery target timeline.

dwsteele:
If the user picks an invalid timeline (or the default is invalid) they will not discover it until after the restore is complete and recovery starts. In that case they'll receive a message like this:
```
FATAL:  requested timeline 2 is not a child of this server's history
DETAIL:  Latest checkpoint is at 0/7000028 on timeline 1, but in the history of the requested timeline, the server forked off from that timeline at 0/600AA20.
```
This message generally causes confusion unless one is familiar with it. In this case 1) a standby was promoted creating a new timeline 2) a new backup was made from the primary 3) the new backup was restored but could not follow the new timeline because the backup was made after the new timeline forked off. Since PostgreSQL 12 following the latest timeline has been the default so this error has become common in split brain situations.

Improve pgBackRest to read the history files and (hopefully) provide better error messages. Now this error is thrown before the restore starts:
```
ERROR: [058]: target timeline 2 forked from backup timeline 1 at 0/600aa20 which is before backup lsn of 0/7000028
       HINT: was the target timeline created by accidentally promoting a standby?
       HINT: was the target timeline created by testing a restore without --archive-mode=off?
       HINT: was the backup made after the target timeline was created?
```

This saves time and (again hopefully) gives more information about what has gone wrong.

If the backup timeline is not an ancestor of the target timeline the error message is:
```
ERROR: [058]: backup timeline 6, lsn 0/4ffffff is not in the history of target timeline B
       HINT: was the target timeline created by promoting from a timeline < latest?
```

This is a situation is rare (I've never see it in the wild) but it can happen during complex recovery scenarios where the user is explicitly setting the target time.

The main benefit of this feature is that we have control over the error messages. I have done my best to make them clear but discussion is certainly welcome. There is also the possibility to auto-prune invalid timelines (and more) but that will be saved for a future commit.


src/command/restore/timeline.c
```diff
@@ -0,0 +1,293 @@
...
+                    // always try to follow the latest timeline this restore will fail since there is no path to the target
+                    // timeline.
+                    if (timelineFound != 0)
+                    {
+                        ASSERT(lsnFound != 0);
+
+                        THROW_FMT(
+                            DbMismatchError,
+                            "target timeline %X forked from backup timeline %X at %s which is before backup lsn of %s\n"
+                            "HINT: was the target timeline created by accidentally promoting a standby?\n"
```
pgstef, line 270: This situation can probably even pretty easily with test restores too. Could we add `or performing restore tests` to the hint maybe?

pgstef (approved):
I have the feeling that we're regularly (more and more frequently even) telling users about `--target-timeline=current`. Kinda wonder if we shouldn't make it our default for restores :-/

> This saves time and (again hopefully) gives more information about what has gone wrong.

Early catching potential issues with timelines (without the user having to look into PG logs after PG has been started) is really nice! Even if we can't catch them all, catching most of them is already a big progress imo. Most users are confusing pgBackRest restore and PG recovery, and even not looking at PG logs. So throwing those errors in the restore command directly will, I believe, be beneficial for a lot of users.

> The main benefit of this feature is that we have control over the error messages. I have done my best to make them clear but discussion is certainly welcome.

The error messages and hints look clear to me at least. Even if it's not that clear for users, it would let us guide them faster I think.

> There is also the possibility to auto-prune invalid timelines (and more) but that will be saved for a future commit.

You mean by removing archives and history files not relevant for the backups in the repository? Would be nice indeed. Probably have to pay attention to the `repo-retention-history` setting for the naming of such feature though.

I've played with the branch and could easily get the error messages:

```bash
[postgres@ro9pg-1 ~]$ pgbackrest --stanza=ro9pg restore --pg1-path=/tmp/restore_test
...
P00  ERROR: [058]: target timeline 3 forked from backup timeline 2 at 0/293f48d0 which is before backup lsn of 0/2a000028
            HINT: was the target timeline created by accidentally promoting a standby?
            HINT: was the backup made after the target timeline was created?
P00   INFO: restore command end: aborted with exception [058]

[postgres@ro9pg-2 ~]$ pgbackrest --stanza=ro9pg restore --pg1-path=/tmp/restore_test --set=20250123-102319F
...
P00  ERROR: [058]: backup timeline 3, lsn 0/2b000028 is not in the history of target timeline 4
            HINT: was the target timeline created by promoting from a timeline < latest?
P00   INFO: restore command end: aborted with exception [058]

[postgres@ro9pg-2 ~]$ pgbackrest --stanza=ro9pg restore --pg1-path=/tmp/restore_test --set=20250123-102319F --target-timeline=current
...
P00   INFO: repo1: restore backup set 20250123-102319F, recovery will start at 2025-01-23 10:23:19
P00   INFO: remap data directory to '/tmp/restore_test'
P00   INFO: write updated /tmp/restore_test/postgresql.auto.conf
P00   INFO: restore global/pg_control (performed last to ensure aborted restores cannot be started)
P00   INFO: restore size = 181.9MB, file total = 1280
P00   INFO: restore command end: completed successfully
```

We regularly see users pollute their repo with restore tests. But it means latest timeline > backup timeline which can happen for very good reasons of course.
I wonder if we couldn't still detect this situation and add some message in the restore command too (like a warning), so users would need to be aware of it, knowing they have to look at it, or at least give us more context on support questions?

Besides the small comments I've made, the feature and PR are looking good to me :+1:

src/command/restore/timeline.c
```diff
@@ -0,0 +1,293 @@
...
+#include "postgres/version.h"
+#include "storage/helper.h"
+
+/***********************************************************************************************************************************
+Parse history file
+***********************************************************************************************************************************/
+typedef struct HistoryItem
+{
+    unsigned int timeline;                                          // parent timeline
+    uint64_t lsn;                                                   // Boundary lsn
```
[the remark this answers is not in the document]
dwsteele: Fixed in a493c049a396.

src/command/restore/timeline.c
dwsteele: Good idea. How about a6ff644071d89e.

dwsteele:
> Most users are confusing pgBackRest restore and PG recovery, and even not looking at PG logs. So throwing those errors in the restore command directly will, I believe, be beneficial for a lot of users.

Or it means this becomes more a pgBackRest issue to the user. But I guess they already seem to think that and at least this gives us more information.

src/command/restore/timeline.c
pgstef: Yep, sounds good indeed, that will remind the users one of our most frequent recommendation :+1:

pgstef:
> Or it means this becomes more a pgBackRest issue to the user. But I guess they already seem to think that and at least this gives us more information.

Indeed. We already have to deal with recovery issues anyway so definitely +1 on this PR :-D 

The suggested hint change looks good to me :+1:

dwsteele:
Commit updated to 922e9f0775c to fix commit message.

PR #2538, opened 2025-01-27
Caveat --tablespace-map-all regarding tablespace creation.

dwsteele:
If a tablespace is created after the backup starts then it cannot be mapped using --tablespace-map-all since there is no record of it in the manifest.

This would be extremely complex to fix but it can be documented.


pgstef (approved):
> This would be extremely complex to fix but it can be documented.

We can't really predict what does the WAL contain unless we scan it, just like for finding out xid for recovery target. So documentation definitely looks good enough to me :+1: 

> We could generate a warning on restore but it would not be trivial. Given that this is the first time we've see this issue I'm not inclined to spend time on it.

I guess in theory we could, a bit like the with history files recently, read the following backup manifest if we're not restoring from the latest backup and look for added tablespaces to emit a warning. But that would add extra files read which could probably slow down the restore process especially on object stores. So again I think you're right and it doesn't worth the extra effort :wink: 

Anyway, the suggested text looks good to me :+1: . Let's see what @Krysztophe thinks about it.

PR #2539, opened 2025-01-28
Fix error reporting for queries with no results.

dwsteele:
If a query that expected no results returned an error then it would incorrectly report that no results were expected because the error was interpreted as a result.

Switch the order of the checks so that an error is reported instead and add a test to prevent regression.


pgstef (approved):
Nice catch :+1:  LGTM ;-)

Issue #2540, opened 2025-01-29
Stable protocol version across minor (and possibly some major) versions?


gclough:
Was there any progress made on giving a stable protocol version across different versions of pgBackRest, as opposed to validating the version of the client & server binaries to ensure they match?

```
    status: error (other)
            [ProtocolError] expected value '2.x' for greeting key 'version' but got '2.y'
            HINT: is the same version of pgBackRest installed on the local and remote host?
```

I note this was covered back in #742, with the idea that it would happen in maybe a year or so.  As others have requested, we would like to upgrade our main pgBackRest servers first and then update clients as and when that's possible, whereas currently we have to upgrade all clients simultaneously.

Alternatively, can we explicitly call out this requirement in the documentation?

https://pgbackrest.org/user-guide.html#upgrading/v2.x

>_Upgrading from v2.x to v2.y is straight-forward. The repository format has not changed, so for most installations it is simply a matter of installing binaries for the new version. It is also possible to downgrade if you have not used new features that are unsupported by the older version._

Can I propose adding:

>_**CAUTION**: The local and remote software versions must match exactly, and so should be upgraded together. If there is a mismatch, then WAL archiving and backups will not function until that is corrected. In such a case, the below error will be reported:_
>
> ```
>     status: error (other)
>             [ProtocolError] expected value '2.x' for greeting key 'version' but got '2.y'
>             HINT: is the same version of pgBackRest installed on the local and remote host?
> ```

gclough:
I have sent a pull request for the documentation update: #2541

dwsteele:
> Was there any progress made on giving a stable protocol version across different versions of pgBackRest, as opposed to validating the version of the client & server binaries to ensure they match?

This has always been the goal but the problem is the majority of our release would require a major version bump anyway (if we used semver) and we haven't decided if we want all our releases to be 3.0.0, 4.0.0, 5.0.0, etc. Even for releases (or especially patch releases) we don't have the testing to back up the idea that we have not changed the protocol. I'm not very comfortable with just eyeballing these changes.

However, a clearly documented caveat sounds like a really good idea. I'm closing this issue since we can continue the discussion on PR #2541.

dwsteele:
> honestly, i think you should do exactly that. right now, 2.55.0 was released and it's declared to be incompatible with 2.54.2... is it really?

[...]

PR #2541, opened 2025-01-29
Clarify requirement for local/remote pgBackRest versions to match.

gclough:
Updates documentation to make to clear that local and remote software versions must match at all times.

Partial fix for #2540


dwsteele (approved):
I made some edits in c9621bb7a8. If those look good to you this is ready to go.

PR #2542, opened 2025-01-30
Change --process-max recommendation for object stores to --repo-bundle.

dwsteele:
While process-max is as useful for object stores as any other storage type, for file creation time in particular file bundling is far more effective since fewer files are created.

Update the recommendation to reflect this.

### DIFF
--- a/doc/xml/release/2025/2.55.0.xml
+++ b/doc/xml/release/2025/2.55.0.xml
@@ -125,6 +125,17 @@
 
                 <p>Caveat <br-setting>--tablespace-map-all</br-setting> regarding tablespace creation.</p>
             </release-item>
+
+            <release-item>
+                <github-pull-request id="2544"/>
+
+                <release-item-contributor-list>
+                    <release-item-contributor id="david.steele"/>
+                    <release-item-reviewer id="stefan.fercot"/>
+                </release-item-contributor-list>
+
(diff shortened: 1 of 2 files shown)


pgstef:
Hi,

> While process-max is as useful for object stores as any other storage type, for file creation time in particular file bundling is far more effective since fewer files are created.

If I understand the suggested change correctly, the idea is to emphasize to use file-bundling on object stores rather than masking the problem using multiple processes, right? Basically said, focus on push/get bigger files 1 at a time rather than multiple files at the same time (small or big)?

dwsteele:
> If I understand the suggested change correctly, the idea is to emphasize to use file-bundling on object stores rather than masking the problem using multiple processes, right?

Correct, process-max is still useful for object store backups (for the same reasons as any backups) but bundling is the best solution to object creation latency. We already have a section for process-max.

> Basically said, focus on push/get bigger files 1 at a time rather than multiple files at the same time (small or big)?

Yes.

pgstef (approved):

pgstef:
> We already have a section for process-max.

Right.

> process-max is still useful for object store backups (for the same reasons as any backups) but bundling is the best solution to object creation latency

Then the suggested change looks good to me :+1:

dwsteele:
Commit updated to aa7210a18308 to fix pull request id.

PR #2550, opened 2025-02-05
Add progress-only detail level for info command output.

hilltracer:
The info command fetches a lot of information from the repository about backups and archives, so this operation can be slow. Because progress data is stored in local lock files, accessing the repository is unnecessary when only progress information is required.

This patch introduces the `--detail-level=[progress|full]` option, with `full` as the default. The `progress` level limits the info command output to progress details without querying the repository. The only remaining operations are scanning the folder structure to list available stanzas and reading lock files.

Note: When `progress` is selected, the info command performs no checks beyond verifying stanza availability.


pgstef:
-1 for me on this. It's quite a big behavior change on the error code returned and I fail to see the point to only displaying this info in text format. I don't get what's the value for the final user. The `--no-wal-range` suggested by @dwsteele would be way better/simpler imho.

hilltracer:
What if the client wants to get backup or restore (I hope we will add this function) progress to show it in real time. In this case, the info will be run many times, why not miss all extra operations (e.g. reading and parsing manifest file)? Skipping reading of wals also changes the error code because we miss this check. We also can implement both features (--no-wal-range and --progress-only). Or maybe, we can return another code instead of `ok` (to show that all most of checks was skipped).

pgstef:
> What if the client wants to get backup or restore (I hope we will add this function) progress to show it in real time. In this case, the info will be run many times, why not miss all extra operations (e.g. reading and parsing manifest file)? 

Why is reading/watching log files so underrated... This remains the easiest and most accurate way of seeing what's going on. The progress report in info command doesn't give you the current step of the backup (WAL archives check, waiting for standby replay, waiting for checkpoint, copying data files,...). If you only need the progress pct, you could also directly look at the lock file content.

The info command is meant imho to display backup info. It doesn't parse manifest unless you provide a backup label with --set. So only backup.info and archive.info are loaded, which doesn't sound to be a problem to me even when spamming the info command. I would even argue that tracking the percentage change every second is too much/too frequent anyway.

> Skipping reading of wals also changes the error code because we miss this check. We also can implement both features (--no-wal-range and --progress-only). Or maybe, we can return another code instead of `ok` (to show that all most of checks was skipped).

Which check would you miss if we skip `archiveDbList` exactly? To me there would be no error code change or integrated check missing. This just reports the oldest and latest WAL archives found in the repositories. This scan could be quite slow and expensive on object stores so I see the interest to have the possibility to skip it. 

Progress report is quite a different story imo. You could do it in a separated command without any access to the repo if you want to only rely on the lock files content, but trying to mess-up with the info status text message looks inappropriate to me :-/

Anyway let's see what David thinks about it.

dimoffon:
> Why is reading/watching log files so underrated... This remains the easiest and most accurate way of seeing what's going on. The progress report in info command doesn't give you the current step of the backup (WAL archives check, waiting for standby replay, waiting for checkpoint, copying data files,...). If you only need the progress pct, you could also directly look at the lock file content.

Tracking progress via direct reading a lock file or parsing log files relys on concrete implementation in pgbackrest. There is no guarantee that it won't be changed after some time (log format, lock path, lock file name, progress storage, etc.). 
On the other hand, parsing the result of `info` command, let's say in json format, allows you to automate progress tracking as you have a data scheme of the output and it doesn't expose you all implementation internals.

> The info command is meant imho to display backup info. It doesn't parse manifest unless you provide a backup label with --set. So only backup.info and archive.info are loaded, which doesn't sound to be a problem to me even when spamming the info command.

But it is a problem in practice - under heavy load, while backup/restore is in progress (simultaneously for many PG clusters that use the same physical storage as repository), it may take more than a minute for a simple `pgbackrest info` command for a concrete stanza. 

> Progress report is quite a different story imo. You could do it in a separated command without any access to the repo if you want to only rely on the lock files content, but trying to mess-up with the info status text message looks inappropriate to me :-/

But backup progress is already implemented in `info` command for some reason. And what the point of developing a separate command that makes sense for running `backup` only. Moreover, how it is supposed to be applied for another command progress? For example, `restore` command looks like another candidate that may report its progress, and adding such functionality with similar flag seems reasonable.

dwsteele:
> But it is a problem in practice - under heavy load, while backup/restore is in progress (simultaneously for many PG clusters that use the same physical storage as repository), it may take more than a minute for a simple pgbackrest info command for a concrete stanza.

But that is mostly because of scanning for the begin-end WAL, which can definitely be expensive. If the scan were optionally removed then we'd just be loading archive.info and backup.info, which I imagine would be pretty fast.

I like this idea better as it is more widely applicable for monitoring. One consistent issue we have seen is that generating being/end WAL is slow for monitoring purposes. If we did this as --no-wal-range it would be useful to more users.

dimoffon:
> I like this idea better as it is more widely applicable for monitoring. One consistent issue we have seen is that generating being/end WAL is slow for monitoring purposes. If we did this as --no-wal-range it would be useful to more users.

I've got your point and I agree that having flag like `--no-wal-range` may be seen as the option in one way or another. But let's consider such aspects as performance and efficiency. Tracking backup progress by accessing local file only is simple, fast and doesn't need network communication. Also this activity suppose pulling model, i.e. periodical call of the `info` command to get the current status of a progress. In case of s3 repo getting progress with `--no-wal-range` option will produce unnecessary requests to the storage, reducing the payload.

pgstef:
Tbh, like David said, I'm still thinking `--no-wal-range` is the way to go for the info command. It would benefit lots of users and let the info command run faster in lots of cases. (That + using --stanza option to avoid scanning the repo for existing stanzas)

I understand you would like to also prevent any access to the remote repository (so skip call to the info files). But that's not what the info command is made for and we shouldn't add so many skip checks which would add quite a burden to maintain imho. It would still be valuable sure, but it doesn't have the same purpose. That's why I mentioned maybe another command dedicated for that would be better, and we could factorize/reuse the progress tracking code of the info command there too. It would make more sense to me.

hilltracer:
We decided to create MVP-patch with --no-wal-range option in our repo. And then make performance testing and compare with --progress-only. After that we will be able to discuss the result together and make decisions. Guys, do you agree with this plan?

dwsteele:
> We decided to create MVP-patch with --no-wal-range option in our repo. And then make performance testing and compare with --progress-only. 

This sounds good to me. Obviously there is a little more overhead with --no-wal-range but it seems to me it would be acceptable and the feature would be more broadly applicable. In my past analysis when users have trouble with info running slowly it is WAL range calculation that is to blame.

hilltracer:
Using name `no-wal-range` causes error
```
gpadmin@gpdb6u:~/src/pgbackrest/src$ pgbackrest --stanza=seg-1 --config=/home/gpadmin/src/arenadata.sh/pgbackrest/pgbackrest.conf --no-wal-range info
ERROR: [031]: invalid option '--no-wal-range'
```
Because parcer understands `no-` as negation. Work around can be using option `wal-range` with negation but for me it can lead to misunderstanding.
And also it contradicts [this comment](https://github.com/arenadata/pgbackrest/blob/6d7d940b1a8ade66e90f08ca2e93a339c8d684eb/src/build/config/config.yaml#L239) in config file:
```
#   All config file boolean options automatically may be negated. This setting is used to maintain backward compatibility with older
#   options and should not be used for new options. For instance --online would be named --offline if it were created now to avoid
#   the need for negation. --no-config is another special case that would now be implemented with a separate boolean option,
#   e.g. --config-ignore.
```
@dwsteele how about `--skip-wal-range` name?

hilltracer:
Or maybe not skip the entire archiveDbList, so just miss the Wal scan there? In this case, the result will be more expected:
```json
[{"archive":[{"database":{"id":1,"repo-key":1},"id":"9.4-1","max":null,"min":null}],"backup":[{...
```

dwsteele:
> Using name no-wal-range causes error

I think what we want here is `wal-range`, which defaults to true. Then the user can specify `no-wal-range` to negate it. I think this is most consistent with what we do now.

> Or maybe not skip the entire archiveDbList, so just miss the Wal scan there?

Yes, let's include all the fields but just null out min and max to provide that maximum amount of information possible.

hilltracer:
See attachments here: [files.zip](https://github.com/user-attachments/files/20232868/files.zip)

Hello! Finally, we finished our performance testing and **compared** timings in 3 cases:  
full `pgbackrest info`, `pgbackrest info --progress-only` (similar to the current patch but in our fork), and `pgbackrest info --no-wal-range` (see `1_no_wal_range.patch`).

**Note:** if a file name ends **with** `_part`, it contains only a fragment of the log. We will provide full data on request—just tell us.

### 1 Common perf testing under load and backup

We tested a GPDB (Greenplum 6) cluster:  
Master (coordinator, CPU 16, RAM 128 GB) — **admaster** (2 segments per host).  
Segments (CPU 24, RAM 256 GB) — **adsegment0** (15 segments per host) and **adsegment1** (15 segments per host).  
DB size 500 GB uncompressed / 380 GB compressed (spread across all machines), backup size 380 GB.  
See info logs for more details about stanzas and archives.

|        | Test\Host                                                                      | admaster (sec)                              | adsegment0 (sec)                            | adsegment1 (sec)                            |
| ------ | ------------------------------------------------------------------------------ | ------------------------------------------- | ------------------------------------------- | ------------------------------------------- |
|        |                                                                                |                                             |                                             |                                             |
| **1**  | **load info** <br>see fragment: 2_load_info_part.txt                           | **'avg: 0.05'**<br><br>**'perc95: 0.061'**  | **'avg: 0.317'**<br><br>**'perc95: 0.454'** | **'avg: 0.219'**<br><br>**'perc95: 0.403'** |
| **2**  | **load info --progress-only** 2_load_info_progress_only_part.txt               | **'avg: 0.009'**<br><br>**'perc95: 0.014'** | **'avg: 0.017'**<br><br>**'perc95: 0.024'** | **'avg: 0.016'**<br><br>**'perc95: 0.023'** |
| **3**  | **load info --no-wal-range** 2_load_info_no_wal_range_part.txt<br>             | **'avg: 0.034'**<br><br>**'perc95: 0.049'** | **'avg: 0.134'**<br><br>**'perc95: 0.217'** | **'avg: 0.116'**<br><br>**'perc95: 0.206'** |
|        |                                                                                |                                             |                                             |                                             |
| **4**  | **load info --stanza** <br>                                                    | **'avg: 0.012'**<br><br>**'perc95: 0.017'** | **'avg: 0.041'**<br><br>**'perc95: 0.073'** | **'avg: 0.04'**<br><br>**'perc95: 0.052'**  |
| **5**  | **load  info --stanza --progress-only**<br>                                    | **'avg: 0.008'**<br><br>**'perc95: 0.014'** | **'avg: 0.015'**<br><br>**'perc95: 0.024'** | **'avg: 0.013'**<br><br>**'perc95: 0.02'**  |
| **6**  | **load info --stanza --no-wal-range**<br>                                      | **'avg: 0.01'**<br><br>**'perc95: 0.014'**  | **'avg: 0.041'**<br><br>**'perc95: 0.096'** | **'avg: 0.063'**<br><br>**'perc95: 0.096'** |
|        |                                                                                |                                             |                                             |                                             |
| **7**  | **load info --stanza --set**<br>                                               | **'avg: 0.261'**<br><br>**'perc95: 0.31'**  | **'avg: 0.694'**<br><br>**'perc95: 0.858'** | **'avg: 0.691'**<br><br>**'perc95: 0.882'** |
| **8**  | **load  info --stanza --set --progress-only**<br>                              | **'avg: 0.007'**<br><br>**'perc95: 0.013'** | **'avg: 0.021'**<br><br>**'perc95: 0.039'** | **'avg: 0.022'**<br><br>**'perc95: 0.05'**  |
| **9**  | **load info --stanza --set --no-wal-range**<br>                                | **'avg: 0.278'**<br><br>**'perc95: 0.327'** | **'avg: 0.465'**<br><br>**'perc95: 0.705'** | **'avg: 0.431'**<br><br>**'perc95: 0.612'** |
|        |                                                                                |                                             |                                             |                                             |
| **10** | **backup+load info** 3_backup_load_info_part.txt<br>                           | **'avg: 0.064'**<br><br>**'perc95: 0.122'** | **'avg: 0.837'**<br><br>**'perc95: 1.978'** | **'avg: 0.92'**<br><br>**'perc95: 2.917**   |
| **11** | **backup+load info --progress-only** 3_backup_load_info_progress_only_part.txt | **'avg: 0.015'**<br><br>**'perc95: 0.057**  | **'avg: 0.041'**<br><br>**'perc95: 0.122'** | **'avg: 0.048'**<br><br>**'perc95: 0.145'** |
| **12** | **backup+load info --no-wal-range** 3_backup_load_info_no_wal_range.txt<br>    | **'avg: 0.04'**<br><br>**'perc95: 0.049'**  | **'avg: 0.207'**<br><br>**'perc95: 0.456'** | **'avg: 0.219'**<br><br>**'perc95: 0.494'** |
|        |                                                                                |                                             |                                             |                                             |
| **13** | **backup+load info --stanza**<br>                                              | **'avg: 0.01'**<br><br>**'perc95: 0.013**   | **'avg: 0.068'**<br><br>**'perc95: 0.13**   | **'avg: 0.062'**<br><br>**'perc95: 0.101**  |
| **14** | **backup+load info --stanza --progress-only**<br>                              | **'avg: 0.014'**<br><br>**'perc95: 0.068'** | **'avg: 0.026'**<br><br>**'perc95: 0.072'** | **'avg: 0.031'**<br><br>**'perc95: 0.076'** |
| **15** | **backup+load info --stanza --no-wal-range**<br>                               | **'avg: 0.01'**<br><br>**'perc95: 0.012'**  | **'avg: 0.045'**<br><br>**'perc95: 0.09'**  | **'avg: 0.062'**<br><br>**'perc95: 0.228'** |
|        |                                                                                |                                             |                                             |                                             |
| **16** | **backup+load info --stanza --set**<br>                                        | **'avg: 0.271'**<br><br>**'perc95: 0.291**  | **'avg: 0.854'**<br><br>**'perc95: 1.146'** | **'avg: 0.651'**<br><br>**'perc95: 0.833'** |
| **17** | **backup+load info --stanza --set --progress-only**<br>                        | **'avg: 0.012'**<br><br>**'perc95: 0.029'** | **'avg: 0.064'**<br><br>**'perc95: 0.14'**  | **'avg: 0.046'**<br><br>**'perc95: 0.189'** |
| **18** | **backup+load info --stanza --set --no-wal-range**                             | **'avg: 0.279'**<br><br>**'perc95: 0.364'** | **'avg: 0.65'**<br><br>**'perc95: 1.12'**   | **'avg: 0.619'**<br><br>**'perc95: 1.057'** |

Experiment description:

- `load` - cluster database under DML load.
- `stanza` - specified `--stanza=seg0`
- `backup` - runs parallel backup
- `set` - specified set option for last backup inside seg0.

avg - average of all data across all experiments, separated by hosts  
perc95 - 95% of all values are below this number

### 2 Valgrind trace
I tested it locally on a standard GPDB 6 demo cluster: 1 master (with standby) and 3 segments (with mirrors). Compilation flags: CFLAGS="-O0 -g3 -fno-omit-frame-pointer" ./configure --disable-optimize --enable-test 

Methodology: 
```sh
valgrind --tool=callgrind --collect-jumps=yes --callgrind-out-file=callgrind_noload_simple.out \
pgbackrest --config=/home/gpadmin/src/arenadata.sh/pgbackrest/pgbackrest.conf \
info
kcachegrind  callgrind_noload_simple.out
```
Exported PDF graphs via File->Export Graph in kcachegrind.

Results: 4_callgrind_noload_json_infoRender_focus.pdf, 4_callgrind_noload_json_simple.pdf, 4_callgrind_noload_json_progress_only.pdf, 4_callgrind_noload_json_nowal.pdf.

image


I can provide any addition info by request.
### 3 Results of perf testing cluster with timers injected inside code
I added printing of timing delays for the `info` function (`5_add_time_trace_perf_tests.patch`).

We tested this code under load in the cluster (unfortunately it was a different cluster, not used in experiment 1).  
Log: `5_backup_load_info_with_trace_part.txt` (fragment only).

#### Conditions
- s1 - master, s2, s3, s4 segment hosts
- Number of stanzas ~= 10 per segment host
- Database size ~= 32 GB (per stanza)
- Wal archive min/max ~= 000000010000000A00000001/00000001000000140000000D (total ~= 175 files)
#### Processing Steps: 
Sum timings for each command across all stanzas within each info command execution. Script: 6_summarize_trace_results.py Output: 6_summarize_results_part.log

Average these totals per experiment, grouped by segment. Script: 7_average_summarized_results.py Output: 7_average_results.log
#### Averaged results per segment:

| The name of the operation                            | s1 (ms)    | s2 (ms)     | s3 (ms)     | s4 (ms)     |
| ---------------------------------------------------- | ---------- | ----------- | ----------- | ----------- |
| list of stanzas in the backup directory              | 0.056      | 3.993       | 4.388       | 4.253       |
| infoBackupLoadFile                                   | 3.272      | 13.152      | 15.936      | 13.221      |
| infoArchiveLoadFile                                  | 2.111      | 11.301      | 13.854      | 11.327      |
| infoUpdateStanzaLock                                 | 0.246      | 0.826       | 0.816       | 0.829       |
| infoPgData                                           | 0.001      | 0.001       | 0.001       | 0.001       |
| infoUpdateStanza                                     | 6.191      | 26.769      | 32.221      | 26.918      |
| format text output                                   | 1.519      | 2.704       | 2.767       | 2.701       |
| infoStanzaList includes:                             | 15.866     | 54.485      | 59.166      | 51.501      |
| *- backupList*                                       | *0.374*    | *0.693*     | *0.720*     | *0.706*     |
| *- archiveDbList includes:*                          | *13.597*   | *50.393*    | *54.956*    | *47.321*    |
| *-- list of all WAL directories in the archive repo* | *1.551*    | *12.939*    | *14.161*    | *11.572*    |
| *-- list of all WAL in this WAL dir*                 | *7.182*    | *19.658*    | *21.194*    | *18.633*    |
| *-- list of all WAL in this WAL dir (reverse)*       | *4.195*    | *16.149*    | *18.000*    | *15.610*    |
| **Total time**                                       | **39.512** | **111.264** | **121.975** | **108.672** |
### Conclusion
- --no-wal-range reduces latency based on WAL count (roughly 50% in our conditions).
- Under heavy load and backup (when there is not enough computing power), significant delays occur in other parts of the code (outside `archiveDbList`) and are not affected by `--no-wal-range`.
- --set incurs expected delays due to manifest scanning.
- --progress-only consistently provides minimal latency.

hilltracer:
Our experiments showed that the `--no-wal-range` option alone is not sufficient for our monitoring system, as there is still some overhead. Therefore, we will continue to support the `--progress-only` option regardless.

We would really like to see this option included upstream as well.
We are open to any modification proposals if this option, in its current form, requires improvement.
@dwsteele How should we proceed?

pgstef:
Let's see what David thinks about it but I'm still strongly against this option.

> That's not what the info command is made for and we shouldn't add so many skip checks which would add quite a burden to maintain imho. It would still be valuable sure, but it doesn't have the same purpose. That's why I mentioned maybe another command dedicated for that would be better, and we could factorize/reuse the progress tracking code of the info command there too.

dwsteele:
@hilltracer I have had a look through this again. Thank you for the detailed analysis! I generally agree with @pgstef because even the worst numbers look pretty reasonable to me but I do I understand that when running at scale every bit counts. I doubt you would go through all this effort if this was not a pain point for you and the patch is simple enough (ignoring whitespace).

I'm inclined to commit this but there are a few things:

1) I don't think combining `--set` and `--progress-only` makes sense. A set is part of a stanza and the stanza is where progress is being made. I think we could add a dependency here to make `--set` invalid when `--progress-only`.

2) In general we try to make boolean options true by default and then let users negate them. There are a few exceptions but they are all legacy. Is there a way to name this option so it is true by default?

hilltracer:
@dwsteele thanks for answer!
1. We agree.
2. Ok. What about `--detailed-output` name? So `--no-detailed-output` will be synonym for `--progress-only`.

Are there any more wishes before I start changing the current patch?

dwsteele:
> What about --detailed-output name? So --no-detailed-output will be synonym for --progress-only.

I think maybe a boolean is not the way to go here. We may have other options we want, such as excluding WAL.

Maybe we'd be better off with an option like `level` with values `progress, default`. This would allow us to add new values later with resorting to more booleans. Thoughts?

hilltracer:
> Maybe we'd be better off with an option like level with values progress, default. This would allow us to add new values later with resorting to more booleans. 

We agree. Let's affirm the names and I will change the code.
```
pgbackrest info --level (or mabe `detail-level`?)=[default (or maybe `full`?) | progress (or maybe `progress-only`?)]
```

dwsteele:
How about `--detail=[progress|full]`. It is also easy to tweak the names before commit if we wish to do so.

pgstef:
> How about `--detail=[progress|full]`. It is also easy to tweak the names before commit if we wish to do so.

FWIW, as we already have `compress-level` and `log-level`, I feel like `--detail-level=[progress|full]` would be consistent :thinking:

dwsteele:
Sounds like two votes for `detail-level`. I'm OK with that.

hilltracer:
I have made changes to the code and description. You can check.
I did not find a way to set dependence on the option --set, in the config.yaml, since there is already a --stanza dependence. So I made it inside infoRender().

dwsteele (approved):
> I did not find a way to set dependence on the option --set

There is a no way to make a dependency on another option *not* being set so what you have done here is fine.

I made some updates in 90103920c976 but otherwise this looks ready to commit. Please pull in the updates and release note from https://github.com/dwsteele/pgbackrest/tree/dev-ADBM-2083-ci.

hilltracer:
> Please pull in the updates and release note from https://github.com/dwsteele/pgbackrest/tree/dev-ADBM-2083-ci.

done

dwsteele:
> I would just be a bit concerned about this:
> 
> $ pgbackrest info --stanza=ro9pg --detail-level=progress
> stanza: ro9pg
>    status: ok

Yeah, I noticed this, too. Within the scope of the checks being performed it is correct but I can see how it might be misleading (but I don't have a better idea). We'll just have to see how users react, I think.

dwsteele:
@hilltracer I made one more minor change and added Stefan to the release notes. Please pull from https://github.com/dwsteele/pgbackrest/tree/dev-ADBM-2083-ci again.

pgstef:
> Within the scope of the checks being performed it is correct but I can see how it might be misleading (but I don't have a better idea). We'll just have to see how users react, I think.

I initially thought we could add an "unchecked" status code but, if I understood correctly, we can hit "ok" or "missing stanza path", so it probably wouldn't be a better solution. So overall, I agree, let's wait and see how users react and use this new option :+1:

hilltracer:
> Please pull from https://github.com/dwsteele/pgbackrest/tree/dev-ADBM-2083-ci again.

done

PR #2553, opened 2025-02-06
Add JSON output to the verify command

dkovalev1:
pgbackrest verify command result code is hard to relay on when it is required to
understand the correctness of the backup data in the repository. It is possible
to get more information by adding options --output text --verbose,

But parsing such output is hard and error prone. This patch implements
json format as a new output format for verify command which transforms the
output into json.


dwsteele:
This is on my radar and I think it is a good change but it will be a few weeks or more before I can get to it. You hit us with a bunch of patches at once (which is great!) but we'll need to work through them one by one.

dkovalev1:
Would it be possible to get an update on whether this PR is still under consideration?

dwsteele (changes requested):
I do think we want to add JSON to the verify command but I don't think this implementation is the right direction.

1) It should operate like the info command, i.e. text and JSON should be output to stdout separate from logging.
2) I don't think we should be constructing JSON manually here. It would be better to use JsonWrite to create the JSON.
3) Getting the format right here is important. I'm not convinced that what you have here is flexible enough, but I don't want to spend time on it unless the first two points are addressed.

dkovalev1:
Hi David,

I've made verify to operate like info and collect all the information into keyValue and then output this keyValue into text or JSON using JsonWrite. The resulting JSON is looks like this:
```
{
    "archives":[
        {
            "archiveId":"9.6-1",
            "checked":1,
            "checksumInvalid":0,
            "fileErrors":{"checksumInvalid":0,"missing":1,"other":0,"sizeInvalid":0},
            "missing":1,
            "other":0,
            "sizeInvalid":0,
            "valid":0
        }
    ],
    "backups":[
        {
            "checked":1,
            "checksumInvalid":0,
            "fileErrors":{"checksumInvalid":0,"missing":1,"other":0,"sizeInvalid":0},
            "label":"test-backup-label",
            "missing":1,
            "other":0,
            "sizeInvalid":0,
            "status":"invalid",
            "valid":0
        }
    ],
    "messages":[
        {"level":5,"message":"archiveId: 9.6-1, wal start: 0, wal stop: 2"}
    ]
}
```

dkovalev1:
Hi David, just wondering if you’ve had a chance to take a look at this. No rush at all—just checking if it’s on your radar.

dwsteele:
I have had a quick look through. There are certainly more changes than I was expecting (e.g. using a KeyValue instead of just outputting using JsonWrite). That's probably OK since it gives us more flexibility in the future.

I'm definitely still not satisfied with the JSON output so I'll probably need to focus on that. My time is pretty limited right now but I should be able to start serious review in September.

NOTE: You'll want to remove strEscape() now that it is unused (https://github.com/pgbackrest/pgbackrest/actions/runs/16561284627/job/48347563881?pr=2553)

dkovalev1:
Thanks, David, will be waiting for your feedback. Meanwhile I've removed strEscape().

dkovalev1:
Hi David, following up gently in case this PR got buried — no rush at all, just checking in. Looking forward to your thoughts when you get a chance!

dwsteele:
I've looked at this a number of times but each time I hold off on comment because I don't think this is what we want -- but I'm not exactly sure what we do want. I'll at least put my thoughts down so you have a response.

1) I don't think the JSON output is flexible enough. Right now verify only works on a single stanza/repo but that may not be true in the future. We probably need something more like the info output but not exactly like that. The current verify output separates archiveIds (10-1, 11-2, etc.) but I'm not sure that's something we need. Likewise I don't think it would make sense to report on repos separately. In theory repos can work together to fix missing data and hopefully verify can be made to understand that at some point. So the question is, how to design the JSON in such as way that it takes all this into account without needing to change the format in the future, which is always painful. Or we mark the feature as beta and iterate on it -- also painful.

2) I'm not sure it makes sense to output messages directly into the JSON. The JSON is meant to be machine readable so these warnings/errors should be as well. They should be organized by archive/backup with filenames and messages separated. Also I think the level should be warn/info/error rather than 3/4/5.

3) Why is some of the summary info repeated in the "fileErrors" key?

Add to this the fact that I'm not entirely happy with verify as it stands right now. 945c7d1 didn't break anything but it is when I noticed that some things seem broken -- for instance gaps in WAL are not reported, only invalid WAL. I'm not sure if this is a bug or if it was intentional. Some gaps at the end of the WAL are to be expected when running async archiving.

At this point I'm not sure I want to layer anything else on top of verify before I can go back and correct some issues with the current behavior.

dkovalev1:
Hi David, 

Thank you for your response. . I want to confirm that I’ve understood the current issues with the patch correctly:

-    The JSON output isn’t flexible enough.
    Solution: add support for multiple stanzas/repos, similar to how info works.

-    Messages are written directly into the top-level JSON.
    Solution: group messages by archive/backup, separating filenames and messages.

-    Message levels are numeric.
    Solution: use string levels (warn / info / error) instead of numbers.

-    Summary information is duplicated under fileErrors.
    Solution: remove the duplication.

-    Existing issues in verify (e.g., gaps in WAL not being reported) should be addressed before extending the JSON output.
    Solution: either wait until this work is done or prepare the patch now and finalize it later.

Regarding your idea about separate archiveIds such as 10-1, 11-2, and reporting on repos separately — could you elaborate on how you see that working?

Below is the JSON produced by the current patch:

<details>
<summary>current variant</summary>
```
{
	"archives":[
		{
			"archiveId":"9.5-1",
			"checked":0,
			"fileErrors":{"checksumInvalid":0,"missing":0,"other":0,"sizeInvalid":0},
			"other":0,
			"sizeInvalid":0,
			"valid":0
		},
		{
			"archiveId":"11-2",
			"checked":4,
			"checksumInvalid":1,
			"fileErrors":{"checksumInvalid":1,"missing":0,"other":0,"sizeInvalid":1},
			"missing":0,
			"other":0,
			"sizeInvalid":1,
			"valid":2
		}
	],
	"backups":[
		{
			"checked":2,
			"checksumInvalid":2,
			"fileErrors":{"checksumInvalid":2,"missing":0,"other":0,"sizeInvalid":0},
			"label":"20181119-152900F",
			"missing":0,
			"other":0,
			"sizeInvalid":0,
			"status":"invalid",
			"valid":0
		},
		{
			"checked":2,
			"checksumInvalid":1,
			"fileErrors":{"checksumInvalid":1,"missing":0,"other":0,"sizeInvalid":0},
			"label":"20181119-152900F_20181119-152909D",
			"missing":0,
			"other":0,
			"sizeInvalid":0,
			"status":"invalid",
			"valid":1
		}
	],
	"messages":[
		{
			"level":5,
			"message":"no archives exist in the repo"
		},
		{
			"level":4,
			"message":"invalid checksum '20181119-152900F/pg_data/PG_VERSION'",
			"pid":1
		},
		{
			"level":4,
			"message":"invalid checksum '20181119-152900F/pg_data/biind.pgbi'",
			"pid":1
		}
	],
	"stanza":"db",
	"status":"error"
},
```
</details>

We can change the output so it contains a list of repos/stanzas, similar to info, and attach messages to the corresponding objects. That would look like:

<details>
<summary>option 1</summary>
```
[
	{
		"archives":[
			{
				"archiveId":"9.5-1",
				"checked":0,
				"checksumInvalid":0,
				"missing":0,
				"other":0,
				"sizeInvalid":0,
				"other":0,
				"sizeInvalid":0,
				"valid":0,
				"messages": {
					"level":"detail",
					"message":"no archives exist in the repo"
				},
			},
			{
				"archiveId":"11-2",
				"checked":4,
				"checksumInvalid":1,
				"missing":0,
				"other":0,
				"sizeInvalid":1,
				"missing":0,
				"valid":2,
				"messages": [
					{
						"level":"info",
						"message":"invalid checksum '20181119-152900F/pg_data/PG_VERSION'",
						"pid":1
					},
					{
						"level":"info",
						"message":"invalid checksum '20181119-152900F/pg_data/biind.pgbi'",
						"pid":1
					}
				]
			}
		],
		"backups":[
			{
				"checked":2,
				"checksumInvalid":2,
				"missing":0,
				"other":0,
				"sizeInvalid":0,
				"label":"20181119-152900F",
				"status":"invalid",
				"valid":0
			},
			{
				"checked":2,
				"checksumInvalid":1,
				"missing":0,
				"other":0,
				"sizeInvalid":0,
				"label":"20181119-152900F_20181119-152909D",
				"status":"invalid",
				"valid":1
			}
		],
		"stanza":"db",
		"status":"error"
	}
]
```
</details>

However, this format is tied closely to the current internal object model in pgBackRest, and it reflects what verify does today. It cannot directly capture your idea of using different repos as sources for the same data set.

An alternative is a more generic structure that lists all discovered objects in a flat form and defines relationships throw fields. This avoids any fixed hierarchy and allows multiple parallel structures:

<details>
<summary>option 2</summary>
```
[
	{
		"kind": "stanza",
		"stanza":"db",
		"status":"error" | "ok",
	},
	{
		"kind": "archive",
		"stanza":"db",
		"archiveId":"9.5-1",
		"checked":0,
		"checksumInvalid":0,
		"missing":0,
		"other":0,
		"sizeInvalid":0,
		"valid":0,
		"messages": [{
			"level":"detail",
			"message":"no archives exist in the repo"
		}],	
	},
	{
		"kind": "archive",
		"archiveId":"11-2",
		"checked":4,
		"checksumInvalid":1,
		"missing":0,
		"other":0,
		"sizeInvalid":1,
		"valid":2,
		"messages": [
			{
				"level":"info",
				"message":"invalid checksum '20181119-152900F/pg_data/PG_VERSION'",
				"pid":1
			},
			{
				"level":"info",
				"message":"invalid checksum '20181119-152900F/pg_data/biind.pgbi'",
				"pid":1
			}
		]
	},
	{
		"kind": "backup",
		"checked":2,
		"checksumInvalid":2,
		"missing":0,
		"other":0,
		"sizeInvalid":0,
		"label":"20181119-152900F",
		"status":"invalid",
		"valid":0
	},
	{
		"kind": "backup",
		"checked":2,
		"checksumInvalid":1,
		"missing":0,
		"other":0,
		"sizeInvalid":0,
		"label":"20181119-152900F_20181119-152909D",
		"status":"invalid",
		"valid":1
	}
]
```
</details>

In this model, each item has a `kind` field plus its own attributes, and relationships are expressed by referencing related objects (for example, archives and backups pointing back to a stanza). This makes the format flexible enough to represent multiple stanzas, multiple repos, or mixed hierarchies withing a single verify report.

We’re ready to continue refining either approach, and if it helps, we can move ahead with a beta version.

dkovalev1:
Hi David! Have you got any thoughts regarding verify and JSON?

dwsteele:
For the time being I don't want to consider any large changes in this module. I have some refactors and improvements planned first.

PR #2554, opened 2025-02-10
Fix expire archive range logging.

dwsteele:
The prior logging only output the last path to be removed since start was overwritten as each path was deleted. This had no affect on expire functionality but was confusing since many more files might be expired than the logs indicated.

Fix logging so the correct start path is logged.

### DIFF
--- a/src/command/expire/expire.c
+++ b/src/command/expire/expire.c
@@ -678,8 +678,11 @@ removeExpiredArchive(const InfoBackup *const infoBackup, const bool timeBasedFul
                                     }
 
                                     archiveExpire.total++;
-                                    archiveExpire.start = strDup(walPath);
                                     archiveExpire.stop = strDup(walPath);
+
+                                    // Update start for the first WAL path removed
+                                    if (archiveExpire.start == NULL)
+                                        archiveExpire.start = strDup(walPath);
                                 }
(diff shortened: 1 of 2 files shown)


pgstef (approved):
The fix looks good to me :+1: 

Regarding OP latest question:

> If it looks like it is behaving properly, I'd guess having in detail log level a progress message every minute or so( what was already deleted...) might be an improvement making life for not enough patient DBAs easier.

I guess from the logs, they could watch `DEBUG:     storage/storage::storagePathRemove` to find out all the archives directory being removed.

It's quite convenient to have an accurate summary in INFO about what was expired, and logging each archive removed at DETAIL level would be too much probably but couldn't we add a DETAIL log after each removed path? It would add least indicate more regularly that something's happening?

dwsteele:
> but couldn't we add a DETAIL log after each removed path? It would add least indicate more regularly that something's happening?

Yeah, that seems like it would be easy enough to do. This doesn't solve the problem of logging during stanza-delete since that is done in one big delete, but I guess it makes sense to fix what we can and expire is much more commonly run.

Were you planning to submit a patch for that?

pgstef:
> Were you planning to submit a patch for that?

It was just an idea but if it sounds interesting, I can have a look and submit a patch sure :+1:

PR #2557, opened 2025-02-16
Add support for GCS requester pays.

cyberdelia:
Add support for [GCS Requester Pays](https://cloud.google.com/storage/docs/using-requester-pays#rest-access-requester-pays) feature which allows to change who is getting billed for associated costs of the requests.

This is probably more a work in progress than a full feature for now, but it does work as is.


dwsteele:
I did some refactoring to make this work more like tags and fit the rest of the code.

One thing I'm noticing, though, this feature is enabled for all ops (read, write, list, info) except remove. If we added it for remove it would simplify the code a lot since and make it more robust -- does that make sense?

On the other hand, if this is intended to be used in hosted environments to allow customers to access their repo, would it make sense to enable this for read, list, info ops only?

cyberdelia:
@dwsteele I’m fairly sure you can send it for any type of requests.

dwsteele:
> I’m fairly sure you can send it for any type of requests.

That's my reading of this as well. That means we can greatly simplify this as done in 9908e5fa. If the user wants to limit requests they can do that via permissions.

dwsteele (approved):
This looks good to me now. We are ready to commit if you are finished testing.

PR #2560, opened 2025-02-18
Add support for S3 requester pays.

cyberdelia:
Add support for [S3 Requester Pays](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RequesterPaysBuckets.html) feature which allows to change who is getting billed for associated costs of the requests, similar to #2557.

Again, it's halfway there, but somewhat further than last PR.


dwsteele (approved):
This looks good to me now. We are ready to commit if you are finished testing.

cyberdelia:
This appears to be good to go as well.

PR #2571, opened 2025-03-07
feat: support upload s3 by http

Wenne:
support upload s3 by http


dwsteele (changes requested):
Thanks for the patch!

I don't think we want to add a new option for this feature. We'd rather have the user specify `http` in the endpoint instead. You can allow this by changing `httpProtocolTypeHttps` to `httpProtocolTypeAny` at https://github.com/pgbackrest/pgbackrest/blob/release/2.54.2/src/storage/s3/helper.c#L34. Then use `httpUrlProtocolType()` to get the protocol.

PR #2578, opened 2025-03-19
Fix test logging.

YegorZ:
I'm new, may be I'm wrong here?


dwsteele (approved):
You're not wrong -- this is missing test logging.

PR #2581, opened 2025-03-21
Fix missing return in FUNCTION_LOG_RETURN_VOID().

YegorZ:
I checked: this macro is used only once at the end of all functions, so the missing `return` did not cause issues


dwsteele (approved):
This macro (and related ones) should always be called last when possible, but of course there may be situations where we can't guarantee that. Thanks for the fix!

PR #2583, opened 2025-03-24
Clarify behavior of --repo-retention-full-type.

dwsteele:
Make some clarifications and add explicit documentation for repo-retention-full-type=count.

### DIFF
--- a/doc/xml/release/2025/2.55.0.xml
+++ b/doc/xml/release/2025/2.55.0.xml
@@ -220,6 +220,19 @@
             </release-item>
 
             <release-item>
+                <github-issue id="2562"/>
+                <github-pull-request id="2583"/>
+
+                <release-item-contributor-list>
+                    <release-item-ideator id="antoine.beaupre"/>
+                    <release-item-contributor id="david.steele"/>
+                    <release-item-reviewer id="antoine.beaupre"/>
+                </release-item-contributor-list>
+
+                <p>Clarify behavior of <br-setting>--repo-retention-full-type</br-setting>.</p>
+            </release-item>
+
+            <release-item>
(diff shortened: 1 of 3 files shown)


anarcat (approved):
LGTM! i would also cross-reference this section from the [expire docs](https://pgbackrest.org/command.html#command-expire) and [--expire-auto](https://pgbackrest.org/command.html#command-backup/category-command/option-expire-auto)...

dwsteele:
> i would also cross-reference this section from the [expire docs](https://pgbackrest.org/command.html#command-expire)

These docs already reference https://pgbackrest.org/user-guide.html#retention/full, which also goes into detail on this subject and names the relevant options (maybe too much overlap, really).

We do have plans to add links for all the option references but that's another thing.

> and [--expire-auto](https://pgbackrest.org/command.html#command-backup/category-command/option-expire-auto)

Not sure how this is relevant?

anarcat:
> Not sure how this is relevant?

well it's where i looked. :)

those are just ideas, this all looks fine to me and is already an improvement!

dwsteele:
> those are just ideas, this all looks fine to me and is already an improvement!

Cool -- and I do appreciate this input! Updating complex docs like this can cost a lot of time in terms of verification that it all makes sense. There's an argument that all the expire docs could use an overhaul but I just can't dedicate time to that right now. Hopefully this will at least help the situation.

Please feel free to submit a PR if you have ideas for further improvement. It is much easier for us to see what you are getting at and review if we have something concrete to start from.

Again, thank you!

anarcat:
> @anarcat Oh, and how would you like your name to appear in the release notes?

anarcat or Antoine Beaupré, as you see fit.

PR #2585, opened 2025-03-26
Fix instances of "of of".

albertchae:
While reading https://pgbackrest.org/user-guide.html#quickstart/performance-tuning, we
noticed a typo with "of of". Grepping for this found another incident in
a comment (as well as 2 release notes generated from commit messages, which I ignored).


dwsteele (approved):
Thanks for this! Amazing how many times you can proofread (and have others do so) and still miss a simple mistake like this.

Actually, I'd rather fix the release note and leave the uncrustify config alone. The latter has been fixed upstream and we'll get it next time we bring in the config.

PR #2588, opened 2025-03-27
Revert "full/incremental backup method".

dwsteele:
This method was introduced in cad595f but on further reflection it does not seem worth the added complexity just to make restore consistency faster without improving the speed of the overall backup. The most common recovery case is PITR and this method produces diminishing returns as the recovery time gets further from the backup end time.

A better solution (not implemented here) is to copy unmodified files from prior backups. This is much faster than recopying and compressing files from the cluster (especially on object stores with a copy command) and can even be done after the backup window to further reduce WAL replay required for consistency. It also reduces load on the host where the backup is made.


pgguru (approved):
Reviewed the diffs between the original feature and the revert; code changes seemed to be largely correct (at least that a quick skim could provide), but there was some difference in the test harness functions, which I assume was related to an internal refactor or additional features touching tests between the original commit and this revert.

Assuming that my assumption here is correct then I think it makes sense, both the mechanics of the revert as well as the reasoning for it.  (It was an admittedly overly complicated feature.)

PR #2592, opened 2025-04-01
Calculate content-md5 on S3 only when required.

dwsteele:
The content-md5 header was generated for all requests with content but it is only required for batch delete requests. It is not clear why this header is required when x-amz-content-sha256 is also provided or why it is required only for this request but the documentation is clear on the matter. However, the content for these requests is relatively small compared to uploading files so omitting content-md5 where possible will save some CPU cycles.

Current AWS S3 and recent Minio don't complain if this header is missing but since it is still required by older versions of Minio and it is specified in the documentation for batch delete it is makes sense to keep it.


pgguru (commented):
The name of this flag seems odd to me; perhaps something indicating an action like `computeMd5`, or more of an attribute about the param like`requiresMd5`.

dwsteele:
> perhaps something indicating an action like computeMd5

Compute md5 for what? We calculate calculate checksums for other parts of the request as well. I think at the least including content is important. I would be OK with `contentMd5Required` or some such.

PR #2594, opened 2025-04-02
Warn when a value for a multi-key option is overwritten.

dwsteele:
Some options can contain multiple key/value pairs. However, if if the key is specified again the value will be silently overwritten. In most cases one value per key is appropriate, but it makes sense to warn the user about the overwrite.


pgguru (approved):
Change makes sense, and def seems to guard against the problem seen in #2595 (at least as far as not overwriting such a value silently).

More for my own understanding, where are the allowed multiple keys defined (i.e., ones that can be validly provided more than once without triggering this error)?

dwsteele:
> More for my own understanding, where are the allowed multiple keys defined (i.e., ones that can be validly provided more than once without triggering this error)?

There aren't any. Overwriting keys was a Perl behavior that the C code inherited. We could change that for tls-server-auth with little chance of breakage, and we might still do so, but for now it seems safer to just document the implemented behavior as we did in #2595.

pgstef:
Hi,

I've played a bit with this change and got confused :-/ 

```bash
$ pgbackrest --stanza=ro9pg restore
P00   WARN: key 'primary_conninfo' value 'host=172.18.0.2 user=repuser port=5432 application_name=ro9pg-3' is overwritten with 'db.mydomain.com' for 'recovery-option' option
2025-04-08 07:06:59.299 P00   INFO: restore command begin 2.55.0dev:
...
```

The WARN doesn't have the timestamp and is only displayed in the console output, not in the restore log file. Was that expected?

```bash
$ pgbackrest --stanza=ro9pg restore --recovery-option="primary_conninfo=host1" --recovery-option="primary_conninfo=host2"
P00   WARN: key 'primary_conninfo' value 'host1' is overwritten with 'host2' for 'recovery-option' option
...
```

Works with multiple recovery-option specified in the command line. But not when multiple options are defined in the config file and one in the command line, which is expected I guess as it would be the way to silently override a configuration setting.

```bash
$ pgbackrest --stanza=ro9pg backup --annotation=source="annot1" --annotation=source="annot2"
P00   WARN: key 'source' value 'annot1' is overwritten with 'annot2' for 'annotation' option
...
```

Seems to work with other configuration settings too, the same way.

hunleyd:
> Overwriting keys was a Perl behavior that the C code inherited. We could change that for tls-server-auth with little chance of breakage, and we might still do so,

Should we open an issue for that because having this behavior changed would *greatly* help our automation deployments

dwsteele:
> The WARN doesn't have the timestamp and is only displayed in the console output, not in the restore log file. Was that expected?

This is always true of warnings before config is fully initialized. It also used to be true of warnings even after initialization but we changed that in 7d54073.

> But not when multiple options are defined in the config file and one in the command line, which is expected I guess as it would be the way to silently override a configuration setting.

This is because a command-line override keeps the options in the config file from being considered at all so there is no warning. This seems fine to me. Certainly we do not want to warn on a command-line override as that is an important and commonly used feature.

pgstef:
> This is because a command-line override keeps the options in the config file from being considered at all so there is no warning. This seems fine to me. Certainly we do not want to warn on a command-line override as that is an important and commonly used feature.

Agreed, pretty common use-case indeed.

> This is always true of warnings before config is fully initialized.

As it's not a new behavior introduced here, it's not something to worry about in this PR of course. But it still feels weird and I was wondering if or how we could improve that separately.

Thanks for your answer anyway, PR looks good to me too :+1:

dwsteele:
> As it's not a new behavior introduced here, it's not something to worry about in this PR of course. But it still feels weird and I was wondering if or how we could improve that separately.

At this point in the parser we don't know which log file to log to or if timestamps are enabled. We could default the timestamps but not much we could do about the log file except to delay outputting the warnings until after the log file has been initialized. Defaulting timestamps to on would break many tests so we'd need a way to keep it disabled for testing.

It think it would be a lot of work for very little gain.

dwsteele:
> but not much we could do about the log file except to delay outputting the warnings until after the log file has been initialized

One more thing -- even if we could get warnings to work this way we couldn't do the same with errors because an error means parsing cannot continue and the log file can never be initialized. Therefore an error must be output immediately using the only output we have -- the console. I think warnings should work the same way.

PR #2595, opened 2025-04-02
Fix documentation for specifying multiple stanzas with tls-server-auth.

dwsteele:
The prior documentation said that multiple stanzas should be specified by repeating the tls-server-auth option. This is incorrect -- in fact a comma-separated list of stanza should be used.

Fix the documentation and add a test to confirm this behavior.

In passing add some const qualifiers that were missing in the relevant code.


pgguru (approved):
Had this ever been valid, or was always wrong?

PR #2596, opened 2025-04-04
Remove extraneous const from cvtIntToZ() prototype.

YegorZ:
The function has signature with "const" in src/common/type/convert.c


dwsteele (approved):
Looks good, thank you!

PR #2601, opened 2025-04-10
Clarify incremental backup expiration.

dwsteele:
The documentation was a bit misleading regarding how incremental backups are expired. Update the misleading part ("Differential and Incremental backups are count-based...") and move the explanation of how incremental expiration works out of differential expiration into the introductory paragraph.

Also add a note about full backups are considered as differential for the purpose of expiration.


src/build/help/help.xml
```diff
@@ -2227,7 +2229,9 @@
...
+                    <p><backrest/> does full backup rotation based on the retention type which can be a count or a time period. When a count is specified, then expiration is not concerned with when the backups were created but with how many must be retained. Differential backups are count-based but will always be expired when the backup they depend on is expired. Incremental backups are not expired by retention independently &amp;mdash; they are always expired with their related full or differential backup. See sections <link page="user-guide" section="/retention/full">Full Backup Retention</link> and <link page="user-guide" section="/retention/diff">Differential Backup Retention</link> for details and examples.</p>
```
pgstef, line 2232: _Differential backups are count-based but will always be expired when the backup they depend on..._

When the _full_ backup they depend on?

doc/xml/user-guide.xml
```diff
@@ -1852,7 +1852,7 @@
                 <backrest-config-option section="global" key="repo1-retention-diff">1</backrest-config-option>
             </backrest-config>
 
-            <p>Backup <br-setting>repo1-retention-diff=1</br-setting> so two differentials will need to be performed before one is expired. An incremental backup is added to demonstrate incremental expiration. Incremental backups cannot be expired independently &amp;mdash; they are always expired with their related full or differential backup.</p>
+            <p>Backup <br-setting>repo1-retention-diff=1</br-setting> so two differentials will need to be performed before one is expired. An incremental backup is added to demonstrate incremental expiration, which depends on the differential expiration.</p>
```
pgstef, line 1855: _which depends on the differential expiration_

Maybe add _which in this case..._ to avoid general assumption?

src/build/help/help.xml
dwsteele: Yes, that makes sense now that the sentence only references differential backups. Fixed in 3abc05420f9a.

PR #2605, opened 2025-04-15
Add FAQ about exporting self-contained cluster.

pgstef:
Based on discussion in issue 2602, while we require the repository to efficiently work with pgBackRest, we can generate a restored pgdata (just like what `pg_basebackup` does) for the users to compress/encrypt/copy wherever they want.

### DIFF
--- a/doc/xml/faq.xml
+++ b/doc/xml/faq.xml
@@ -132,15 +132,15 @@ process-max=1
 
     <!-- ======================================================================================================================= -->
     <section id="self-contained-backup">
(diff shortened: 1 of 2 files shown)


dwsteele (approved):
I went ahead and made edits directly since we are short on time to get this into the release.

Mostly, I remove the last paragraph -- any operation you can do entirely with pgBackRest I think we need to consider supported. The real issue IMV is the lack of timeline switch so I think we need to call that out.

pgstef:
> The real issue IMV is the lack of timeline switch so I think we need to call that out.

Agreed, you're right. Thanks for the edits :+1:

PR #2614, opened 2025-04-28
Add x-amz-checksum-sha256 header to S3 requests.

dwsteele:
The x-amz-checksum-sha256 header is required in place of content-md5 (which was removed in 20bfd14b) for PUT when object lock is enabled. I confused x-amz-content-sha256 (which we already have) with x-amz-checksum-sha256 (which we now need) when writing the patch.

Determining if object lock is enabled may not be compatible across S3 clones, so just add the header to all requests with content. Since SHA-256 is already calculated for authentication this will not add overhead other than encoding the header in base64 (instead of the hex required for authentication).

However, this is not going to help clones that do not accept x-amz-checksum-sha256 and still require content-md5.


dwsteele:
Closing since the original commit will be reverted. This patch may be part of a better solution in a future feature release.

PR #2625, opened 2025-05-08
Fix defaults in command-line help.

dwsteele:
This is another attempt to fix broken defaults in the command-line help. PR #2395 tried to do this by removing current and default values (at least from the list) but this removed much of the value of command-line help being context sensitive. Improvements were made to current values in #2453 but defaults were left as-is, i.e. broken in some cases.

This new approach handles defaults much the same way as current values. If there are multiple defaults for an option then it is displayed in the option list as <multi> and enumerated in the specific option help. For example, if repo1-type=s3 and repo2-type=azure then the defaults for repo1-storage-upload-chunk-size and repo2-storage-upload-chunk-size will differ.

However, getting the above right means we can no longer display defaults for options that are not valid in the current configuration. For instance, if repo1-type=posix then none of the repo1-s3-* options are valid and therefore their defaults are not displayed in help. In this case they don't have defaults and if you tried to set one of these options you would get an error. Another option here would be to not display the option at all (probably confusing) or mark it [inactive] in the list. Marking options inactive would require more work than I've done here due to some behind-the-scenes defaulting we do for programmatic convenience, but it is a possibility.


src/config/config.c
```diff
@@ -410,21 +410,20 @@ cfgOptionIdxTotal(const ConfigOption optionId)
...
 {
     FUNCTION_TEST_BEGIN();
         FUNCTION_TEST_PARAM(ENUM, optionId);
+        FUNCTION_TEST_PARAM(UINT, optionIdx);
     FUNCTION_TEST_END();
 
     ASSERT(cfgInited());
     ASSERT(optionId < CFG_OPTION_TOTAL);
+    ASSERT_DECLARE(const bool group = configLocal->option[optionId].group);
+    ASSERT_DECLARE(const unsigned int indexTotal = configLocal->optionGroup[configLocal->option[optionId].groupId].indexTotal);
```
pgguru, line 423: This is a little hard to follow; not sure if there's anything to do to make it a little more clear, but maybe some sort of macro?

pgguru: (Though at that point, you're just hiding that complexity somewhere else for me to complain about, so maybe this is already as isolated as we can hope for...)

test/src/module/command/helpTest.c
```diff
@@ -639,6 +637,61 @@ testRun(void)
...
+            "Azure. No attempt is made to validate configured chunk sizes so selecting an\n"                                       \
+            "invalid value will lead to errors from the storage service or undefined\n"                                            \
+            "behavior.\n"
+
+        optionHelp = zNewFmt(
+            HELP_OPTION_CHUNK
+            "\n"
+            "default:\n"
+            "  repo1: 5242880\n"
+            "  repo2: 4194304\n",
```
pgguru, line 678: So this is the meat of what we're actually testing in this test; otherwise, the rest of it is just prepping the expected output test so it can do (more-or-less) a strcmp to verify that we get the exact help output for this command given the later option setup, yes?

src/config/config.c
dwsteele: Yeah, that was my thinking. I originally tried to create wrapper macros but there are subtle differences per function so they were impossible to maintain or understand.

fivemeyestore:
Very clean code with good test coverage. This is a great example of best practices.

fivemhostingserver:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

fivemserverhostings:
This is a well-structured codebase. I appreciate the clean separation of concerns.

eyesgodcode:
Interesting solution to this problem! I've been working on similar challenges recently.

eyestorefivem:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

scriptforfivem:
Very clean code with good test coverage. This is a great example of best practices.

myeyestore:
Very clean code with good test coverage. This is a great example of best practices.

fivemservereyestore:
Interesting solution to this problem! I've been working on similar challenges recently.

godofcodeeyes:
Excellent contribution to open source! The community will benefit greatly from this.

eyesaider:
Excellent contribution to open source! The community will benefit greatly from this.

cbandy (approved):
LGTM

src/config/parse.c
```diff
@@ -2672,17 +2672,23 @@ cfgParse(const Storage *const storage, const unsigned int argListSize, const cha
...
+                                configOptionValue->set = true;
+                                configOptionValue->value = optionalRules.defaultValue;
+                                configOptionValue->display = optionalRules.defaultRaw;
+                            }
+
+                            configOptionValue->defaultValue = optionalRules.defaultRaw;
                         }
                         // Else error if option is required and help was not requested
-                        else
+                        else if (!configOptionValue->set)
```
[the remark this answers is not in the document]
pgguru: No strong feelings on that; looks fine.

pgguru (approved):

PR #2636, opened 2025-06-03
Improve the predictability of floating point numbers formatted as strings.

dwsteele:
Some C libraries (e.g.musl) render floating point numbers differently when using printf(). This does not cause any problems for the core code but the unit tests require more predictability to function smoothly without a lot of exceptions.

To accomplish this, wrap the various floating point operations in functions that mostly continue doing math using double but format the output string using integers. This leads to more predictable output at the cost of some complexity.

Rounding could also be accomplished using nearbyint() but this would require linking the math library, which does not seem worth for a fairly simple operation.


dwsteele:
> I assume if we're wanting to support alternate C libs we should include at least one other one in one of the test images

An Alpine (with musl libc) test container is coming in a subsequent patch. These were the core code changes required to make that patch work with all the unit tests. The rest is updates to the tests for varying messages, etc.

PR #2638, opened 2025-06-07
Describe discrete option values in a list where appropriate.

woblerr:
In the documentation and help output the possible values for most options are described as a list. But there are several commands for which formatting is done in a different style.

Formatting has been applied to commands and their options to display option values as a list:
* `info` command, `--type` option;
* `verify` command, `--output` option;
* `version` command, `--output` option;
* `manifest` command, `--output` option.


woblerr:
Looks good to me, thanks.

PR #2646, opened 2025-06-23
Fix issue with adhoc expiration when no backups in a repository.

dwsteele:
If there are no backups in one or more repositories then the following error occurs during adhoc expiration:

ASSERT: [025]: cannot get index 0 from list with 0 value(s)

Fix this by skipping the adhoc logic when there are no backups in a repository.


pgstef (approved):
Hi,

So the error comes in when we use `expire --set` and there's another repository defined without any backup in it.

I could reproduce the error and confirm that this change fixes it.
I also tried the expireTest in the main branch, which leads to `BUT GOT AssertError: cannot get index 0 from list with 0 value(s)`. So the test added is clearly meaningful.

The PR is good for me :+1:

dwsteele:
Thanks, @pgstef!

PR #2652, opened 2025-07-03
Add restore progress to info command output.

hilltracer:
The current implementation of the restore command did not provide any progress information. This patch implements it similarly to the backup command and also updates the info command to display restore progress alongside backup progress.

Restore is a read operation and should not block other commands. The only exception is that multiple restores with the same lock and repo path are not allowed, as each restore must write progress to a separate file. Therefore, no lock is needed for restores with a remote role and restore command should not be terminated when the stop command is executed.

Co-authored-by: Maxim Michkov <m.michkov@arenadata.io>


kathia-barahona:
Just wanted to say I’m looking forward to this change, it’s quite useful. I tested the patch and it works 💯. Thanks for the effort!

hilltracer:
> I think the ok stanzas should look the same

[done](https://github.com/pgbackrest/pgbackrest/pull/2652/commits/520059d8124294e6f60f8da3978884142c942b64) I also refactored status formatting a little bit.

Update: looks like there is problem with CI. And it's not related with current patch.

hilltracer:
> please just pull in my changes

Done.
Thank you for taking our patches!

PR #2658, opened 2025-07-30
S3 Storage Class Support

### LINKED ISSUE #371: add support for IA storage class in S3 buckets
Now that we have support for S3 buckets as a backend (thanks!!) can we get support for IA as well? I use `duplicity` to backup my non-pg bits and it has a `--s3-use-ia` option:
> Store volumes using Standard - Infrequent Access when uploading to Amazon S3. This storage class has a lower storage cost but a higher per-request cost, and the storage cost is calculated against a 30-day storage minimum. According to Amazon, this storage is ideal for long-term file storage, backups, and disaster recovery.

I know that I can accomplish this by setting a Lifecycle on my bucket (and I have) but the files need to be *at least* 30 days old before they will migrate to IA via Lifecycle. It would be great if I could store them in IA from the get-go like `duplicity` allows. Maybe a new option like:
```
repo-s3-storage-class = normal (default)
repo-s3-storage-class = ia
```

Thanks for considering!

### LINKED ISSUE #2671: s3 object storage-class option
Would it be possible to add an option for specifying the storage-class for backup objects in AWS s3?  I have a case where, in an effort to save money on storage costs for large database backups, we want to change the storage class for these objects.  Right now, I just have a script in place to re-copy all files of the last backup, and any WAL with the aws cli and the --storage-class option.

felixbuenemann:
This implements support for setting the AWS S3 storage class for uploaded files when using an S3 repo.

It adds two new options:

* `repo-s3-storage-class`: The storage class to use, eg. `STANDARD_IA` or `GLACIER_IR` (no default)
* `repo-s3-storage-class-threshold`: The size threshold to apply the storage class (`128KiB` by default)

The storage class ensures small objects are written in the cheaper `STANDARD` storage class, since most other classes enforce a minimum billed size of 128K.

This feature is useful to save storage costs, when the retention period matches the minimum billed storage time for the class, eg. 30 days for `STANDARD_ID` or 90 days for `GLACIER_IR`.

To simplify logic for multipart uploads, where the size is not know, the maximum threshold was chosen to match the minimum multipart chunk size of `5MiB` for S3, meaning multipart uploads will always apply the storage class, if configured.

---

The implementation of `repo-s3-storage-class` and `repo-s3-storage-class-threshold` has been split to two commits, since it is possible to only support the storage class without the threshold logic, even though it saves money, especially if bundling is not used. It would also be possible to simplify the threshold logic by using a hardcoded 128K constant.

---

This feature was previously requested in https://github.com/pgbackrest/pgbackrest/issues/371.


felixbuenemann:
Another idea I have is to also exclude certain files that are frequently read like `*.manifest` from storage class, but I'm not sure its worth the effort given pricing of eg. 0.01 USD per GB retrieved for `STANDARD_IA` you can do a lot of manifest reads for 1 USD, given a manifest size of eg. 5MiB.

However I'm not sure how large manifests can get with really large databases.

dwsteele (commented):
Thank you for the well thought out and complete patch!

> The implementation of repo-s3-storage-class and repo-s3-storage-class-threshold has been split to two commits

I think it is fine to commit both of these (and together).

See my separate comment on threshold handling but overall this looks good to me. I do wonder if the tests could be more integrated with the existing tests to avoid starting a new test server. Probably doesn't add that much time, though.

felixbuenemann:
@dwsteele I've done some math on how often the `backup.info` and `backup.manifest` files get read, especially with longer backup chains and it can quickly add up, so I've implemented my idea of excluding these files in a new commit.

Let me now what you think about it. I can also share my calculations in more detail if you'd like.

Since the new code touches some of the places you commented on, I will address feedback in bulk, when I've had feedback on the exclusion code.

felixbuenemann:
Oh and since you know the code much better, are `backup.info` and `backup.manifest` really the only files that are frequently read from S3 when storing new backups?

dwsteele:
> so I've implemented my idea of excluding these files in a new commit.

I'm definitely not in favor of hard-coding these filters into the storage driver. Presumably we'll want to add storage classes to GCS and Azure at some point.

This functionality should be added as a parameter to the storageNewWriteP() macro and pushed down to the storage write driver. Probably something like `bool classDefault` would work so it defaults to the set storage class if any.

In addition, I'm not sure if the .copy files need this exclusion. They are only read when the *.info file is corrupt or missing.

But after going back and reading the original issue I still see some problems for users here. It's pretty normal for full backups to be kept 30 days or more but diff/incr backups are less likely to be kept that long so they will be penalized when those backup types are removed. At the least this should be called out in the docs -- but maybe it means the storage class really needs to be a backup parameter and not a repo parameter.

felixbuenemann:
> In addition, I'm not sure if the .copy files need this exclusion. They are only read when the *.info file is corrupt or missing.

The .copy files are not excluded. 

> but maybe it means the storage class really needs to be a backup parameter and not a repo parameter.

Wouldn’t it already be possible to either omit `repo-s3-storage-class` from pgbackrest.conf and only pass it to `pgbackrest backup --type=full --repo1-s3-storage-class=STANDARD_IA` or override it on a diff/incr with `pgbackrest backup --type=diff --repo1-s3-storage-class=STANDARD`?

dwsteele:
> Wouldn’t it already be possible to either omit repo-s3-storage-class from pgbackrest.conf and only pass it to pgbackrest backup --type=full --repo1-s3-storage-class=STANDARD_IA or override it on a diff/incr with pgbackrest backup --type=diff --repo1-s3-storage-class=STANDARD?

Yes, but you'd also need to *add* it to the `[stanza:archive-push]` configuration or to the archive-push command itself. Workable, sure, but certainly more complicated than we prefer. And it would definitely need to be documented for anyone to have a chance of figuring it out.

dwsteele:
> Yes, but you'd also need to add it to the [stanza:archive-push] configuration or to the archive-push command itself.

Actually, I'm not sure this is a good idea since the WAL is frequently accessed by standby servers in most environments.

Another wrinkle is that when `repo-block` is enabled maps are read from the prior backups in order to build the deltas. It's a relatively small percentage of the prior backup but that might be enough to negate the savings from IA.

felixbuenemann:
@dwsteele I've implemented the requested changes regarding file exclusions, threshold checking and merging the s3 test suites. I'll merge commits after review.

I'm not sure what to do about making storage class a backup option. Wouldn't that cause confusion, because it is no longer called `repo-s3-*` like all other s3 options?

dwsteele:
> I've implemented the requested changes regarding file exclusions, threshold checking and merging the s3 test suites.

On a quick review this looks good!

> I'll merge commits after review.

No need for that -- all commits to integration are automatically squashed.

> I'm not sure what to do about making storage class a backup option. Wouldn't that cause confusion, because it is no longer called repo-s3-* like all other s3 options?

To start with, these may end up as `repo-storage-*` options anyway (see --repo-storage-upload-chunk-size) since other object stores support some sort of class/tiering. Don't worry about that yet -- let's get the important stuff worked out. Naming comes later.

But, it would probably end up as something like `backup-` which is not very good at all. Given that we'd probably want to limit this to full backups it might be something like `backup-full`.

Maybe we only apply the storage class options to full backups (even though they are `repo-*`)? That wouldn't be too hard.

Ultimately we could have an internal mapping for storage tiers vs their minimum retention and use our own retention settings to determine what tier is best. We don't need to do that now but whatever we do should keep that in mind.

> MY COMMENT ABOVE: Another wrinkle is that when repo-block is enabled maps are read from the prior backups in order to build the deltas. It's a relatively small percentage of the prior backup but that might be enough to negate the savings from IA.

I looked at the pricing again and I think that repo-block should not cause too much of a problem. Generally speaking the block incremental maps are .01% of data size (though percent of compressed size varies) so reading them should not blow up costs too badly.

But since WAL would be generally be read in its entirety even one scan of a segment would negate the benefits of IA. Therefore we probably would not want to use IA for WAL.

dwsteele:
Another thing: I do think we should limit the tiers that a user can select. Some of the tiers would prevent pgBackRest from reading data and that seems like a bad idea. Sure, they can use life cycles to do the same thing but it seems worse if they do it through our interface.

So of the available tiers (STANDARD | REDUCED_REDUNDANCY | STANDARD_IA | ONEZONE_IA | INTELLIGENT_TIERING | GLACIER | DEEP_ARCHIVE | OUTPOSTS | GLACIER_IR | SNOW | EXPRESS_ONEZONE | FSX_OPENZFS) I would probably want limit the selection to (STANDARD | REDUCED_REDUNDANCY | STANDARD_IA | ONEZONE_IA | GLACIER_IR | EXPRESS_ONEZONE). In theory INTELLIGENT_TIERING should be in this list but it will ultimately migrate to tiers that we cannot read and anyway I believe it can be configured on the bucket.

felixbuenemann:
> I looked at the pricing again and I think that repo-block should not cause too much of a problem. Generally speaking the block incremental maps are .01% of data size (though percent of compressed size varies) so reading them should not blow up costs too badly.

I find it hard to estimate the I/O patterns given the different combinations of full/diff/incr backups. At least for me (WAL excluded) the `.pgbi` file make up most of the backup storage, so excluding them negates the usefulness of setting the storage class. So the question is, does accessing older `.pgbi` generate enough traffic to exceed the storage savings.

Since all my calculations so far have been flawed in one way or another the best approach I can come up with is to backup a database to two repos, one with `STANDARD` and one with `STANDARD_IA`, setup cost allocation tags and after about a month compare billing.

> But since WAL would be generally be read in its entirety even one scan of a segment would negate the benefits of IA. Therefore we probably would not want to use IA for WAL.

Wouldn't it only be read in restore case? I would be happy to pay extra for the rare case of restoring from backup.

For use cases were eg. slaves are restored from backup regularly, excluding WAL from storage class would be best.

> Another thing: I do think we should limit the tiers that a user can select. Some of the tiers would prevent pgBackRest from reading data and that seems like a bad idea. Sure, they can use life cycles to do the same thing but it seems worse if they do it through our interface.

I initially had implemented just that (with a similar whitelist), but removed it, because the `config.yaml` `allow-list` did not allow underscores and I didn't want to add a translation layer between storage class names.

The list of storage classes is also changing, eg.  `REDUCED_REDUNDANCY` while still available, is now more expensive than `STANDARD`. Also other S3 compatible providers like Minio use other storage class names, so I figured best not to hardcode.

felixbuenemann:
I've setup the before mentioned experiment running backup to two s3 repos in parallel one with `STANDARD` and one with `STANDARD_IA` class. I've also added tags with the bucket name that should allow me to filter billing reports by bucket.

dwsteele:
> At least for me (WAL excluded) the .pgbi file make up most of the backup storage, so excluding them negates the usefulness of setting the storage class. So the question is, does accessing older .pgbi generate enough traffic to exceed the storage savings.

For backups < 1% of the .pgbi files should be read (just the maps). That's probably OK. But a single non-delta restore would negate the benefit of IA and another restore would cost more than standard would have. At least as I read it.

> Wouldn't it only be read in restore case? I would be happy to pay extra for the rare case of restoring from backup.

Also catching up streaming replicas that have fallen behind WAL retention on the primary. This happens to users fairly frequently in my experience and users are often amazed that the replica is up-to-date even though streaming rep has been broken for days.

> For use cases were eg. slaves are restored from backup regularly, excluding WAL from storage class would be best.

In my experience this is a pretty common use case. Especially these days with Patroni clusters automatically rebuilding clusters. In dynamic production environments the most recent backup set is often busy (and all the WAL associated with it). In this case a life cycle rule that transitions files older than the most recent full backup is probably most efficient.

> Also other S3 compatible providers like Minio use other storage class names, so I figured best not to hardcode.

OK, I agree that the way you have done it is probably best.

Overall the commit is looking in pretty good shape but I think the this deserves some documentation in the new option. I'm sure your experiments will help inform that.

dwsteele:
> I would be happy to pay extra for the rare case of restoring from backup.

We also encourage users to test restores *regularly*.

felixbuenemann:
Amended last commit to fix code format error caused by extra newline in `s3Test.c`.

felixbuenemann:
> For backups < 1% of the .pgbi files should be read (just the maps). That's probably OK. But a single non-delta restore would negate the benefit of IA and another restore would cost more than standard would have. At least as I read it.

What do you mean by "just the maps"? Will pgbackrest to a range request to read only part of the `.pgbi` file or are there some `.pgbi` files that only contain maps and others that contain more data?

Regarding the experiment: I am currently at 0.17 USD  for Standard class and 0.19 USD for Standard-IA class after 6 days, which is not great. – I'm backing up a 1.4 GB PGDATA directory with weekly full, daily differential, hourly incremental backup and 1min WAL archive timeout. Backup storage growth is roughly 1 GB per day.

Digging into the costs with cost explorer it seems for now the main difference is in the more costly PUT/COPY/POST/LIST requests for Standard-IA, which is billed at 0.01 USD vs. 0.0054 USD per 1.000 reqs for Standard storage. IA-Retrieval costs over the duration are 700 MB so far, so less than a cent.

I'm not sure these numbers translate well to larger backups, where storage costs become more prominent compared to request costs. The small archive timeout also causes lots of WAL PUT requests, although if nothing happened these will often be under 128K compressed, so Standard class.

Also since the backup storage builds up over a month until objects are expired I don't have meaningful numbers yet.

However is is possible that for frequent smaller backups S3 requests costs eclipse storage cost in which case the higher PUT request pricing will make a difference.

dwsteele:
> What do you mean by "just the maps"? Will pgbackrest to a range request to read only part of the .pgbi file or are there some .pgbi files that only contain maps and others that contain more data?

The block maps are mixed in the with the block data in every .pgbi file. We use HTTP range requests to extract the maps and minimize the data read.

What settings are you using for the repo? Do you have repo-bundle and repo-block enabled? repo-bundle will certainly reduce the number of put operations.

> However is is possible that for frequent smaller backups S3 requests costs eclipse storage cost in which case the higher PUT request pricing will make a difference.

That sounds likely. Also -- my math says that a restore will likely negate any benefits of IA or result in higher payments.

So -- this solution seems to be targeted to large data volumes that will never be read. The first part is fine but the second part gives me pause. I'd hate for the message to be, "you'll save money if you don't test your restores!"

felixbuenemann:
> What settings are you using for the repo? Do you have repo-bundle and repo-block enabled? repo-bundle will certainly reduce the number of put operations.

I'm testing with:

```ini
[global]
repo1-type=s3
repo1-s3-endpoint=s3.eu-central-1.amazonaws.com
repo1-s3-bucket=redacted-postgres-backup
repo1-s3-region=eu-central-1
repo1-path=/pgbackrest
repo1-bundle=y
repo1-block=y
repo1-retention-full=30
repo1-retention-full-type=time
repo2-type=s3
repo2-s3-endpoint=s3.eu-central-1.amazonaws.com
repo2-s3-bucket=redacted-postgres-backup-2
repo2-s3-region=eu-central-1
repo2-s3-storage-class=STANDARD_IA
repo2-path=/pgbackrest
repo2-bundle=y
repo2-block=y
repo2-retention-full=30
repo2-retention-full-type=time
compress-type=zst
delta=y
archive-async=y
process-max=2
start-fast=y
log-level-console=info
log-level-file=off
```

And crontab:

```cron
# Full backup every Sunday at 02:00
0 2 * * 0 pgbackrest-multi backup --type=full

# Differential backup Monday through Saturday at 02:00
0 2 * * 1-6 pgbackrest-multi backup --type=diff

# Incremental backup every hour except at 02:00
0 0-1,3-23 * * * pgbackrest-multi backup --type=incr
```

`pgbackrest-multi` script:

```sh
#!/bin/sh

# pgbackrest wrapper to run backups sequentially across all repos
# Usage: pgbackrest-multi backup --type=full

exit_code=0

for repo in 1 2; do
    echo "Running pgbackrest $@ --repo=$repo"
    pgbackrest "$@" --repo=$repo
    if [ $? -ne 0 ]; then
        echo "ERROR: pgbackrest failed for repo $repo"
        exit_code=1
    fi
done

exit $exit_code
```

I've also just setup a second experiment with a 12 GB uncompressed PGDATA (1.5 GB compressed full backup). It is from a staging environment, so not as busy as I would like it to be for realistic results.

That's still not quite useful, a better test would be a 100GB DB that is in active use, but I don't have one available.

felixbuenemann:
Rebased to resolve conflict in `doc/xml/release/2025/2.57.0.xml`.

felixbuenemann:
@dwsteele After roughly two months of testing it seems that this beneficial for larger databases and not beneficial for smaller.

Given this schedule:
* Full backup every Sunday at 02:00
* Differential backup Monday through Saturday at 02:00
* Incremental backup every hour except at 02:00

- With a 1.3GB database, frequently changing: 16% cost increase
- With a 12.5GB database, infrequently changing: 6% cost decrease

I would expect higher savings the larger the database. But I don't have any large databases to test on.

dwsteele:
And no restores were done, correct? That's what worries me as it appears a single restore would wipe out any possible cost savings and we definitely encourage users to test their restores.

I'm torn here. I appreciate that you have put together such a complete commit but I wonder if it would encourage users to do something that will end up costing them more money. I still think lifecycle rules directly on the bucket may be a better solution.

felixbuenemann:
I tested restore of a 12.5 GB database last month and saw about 15% higher cost for restores. The number is not very exact, because the difference was 1 cent and that is the highest granularity of the AWS Cost Explorer.

If there is no consensus on whether this is useful, I'm in favor of abandoning this PR.

dwsteele:
> If there is no consensus on whether this is useful, I'm in favor of abandoning this PR.

I'd like to let it sit for a while and see if we get other opinions. You've done a great job on this PR so there's really nothing left for you to do if we ultimately decide to move forward on it.

I still believe that a lifecycle rule is the way to go here. I set mine to 30 days but I think you could get away with just older than the last full backup and when all diff/incr backups have been expired. This gives cheaper long term storage without making normal operations like restoring and recovering WAL more expensive in the short term.

alexherington:
This is a really useful feature - please don't abandon it! 🙏 Our use case is that we have an _enormous_ set of data to backup and retain for 30 days, so it makes sense for us to store as `ONEZONE_IA` immediately, rather than use a lifecycle policy to transition the objects to that storage class (which can only happen after a min of 30 days).

The potential increase in restore cost isn't an issue in our use case because the alternative to not using S3 is EBS which is considerably more expensive.

My two cents: Having the option to specify the storage class empowers the users with the flexibility to fit a wider range of use cases, as there are lots of different ways S3 fits into a backup solution. This is awesome software!

dwsteele:
@alexherington Thanks for the input! Feedback from users is incredibly valuable.

I see that @alexherington @felixbuenemann @hunleyd, and @seandano87 are all on board with this so here is what I propose. I have modified the commit to simplify it by removing the size and file filters. Some aspects of the commit were not tested -- e.g. which files in a backup were excluded from the specified storage class (Felix' original version did test this but embedded the logic in an awkward place). Writing tests for that would have been complicated and probably not worth it in the long run. Even overwriting backup.info and archive.info 31 times in IA still only costs $0.0001 per month by my calculations. Storing backup.manifest in IA is fine for the best use case of full backups only since the previous full backup manifest will not be read.

I see the ideal use case as weekly full backups only (no diff/incr), retained for >= 30 days, with <= 2 restores per month. Preferably on ONEZONE_IA and always with repo-bundle enabled.

Thoughts? Nothing here precludes adding size or file filters later but I honestly don't see them as having significant value.

alexherington:
@dwsteele - I'm glad the feedback helps! Not having the size or file filters wouldn't affect our particular use case, and the cost savings from having the overall feature would vastly exceed any IA costs from a few files. It still sounds like a great addition :)

dwsteele:
@felixbuenemann Any thoughts on this version? Would it be sufficient for your needs?

felixbuenemann:
@dwsteele I can understand skipping the file name filters, since they spread outside the storage driver code. I don't think skipping the minimum size filter is a good idea, since you will be billed a minimum of 128 KB for each object and the code is in the storage driver itself.

What is your motivation for excluding those two features?

dwsteele:
> I don't think skipping the minimum size filter is a good idea, since you will be billed a minimum of 128 KB for each object and the code is in the storage driver itself.

@felixbuenemann If file bundling is enabled there are maybe one or two files per backup that would be below 128KiB so it is just not much advantage. File bundling should always be enabled for object stores anyway. Since GCS and Azure do not have this penalty (and likely other S3-based object stores) it becomes an AWS only feature. Seems like not worth the bother of an additional option.

hunleyd:
> File bundling should always be enabled for object stores anyway.

at the risk of hijacking this thread, are there plans to enforce that in code? should i file an issue for the same?

dwsteele:
> at the risk of hijacking this thread, are there plans to enforce that in code? should i file an issue for the same?

@hunleyd Since it would break backward compatibility it is not something we can enable by default. The description for repo-storage-class says to enable repo-bundle so I think that's enough.

dwsteele:
@felixbuenemann So where are you on this? I'm planning to commit it either way since it implements the basic functionality needed and we can always expand on it later. But if you don't approve I won't commit it under your name.

dwsteele:
I'm closing this since I haven't heard from the original author and I don't want to commit it without approval.

I had thought about committing it myself but I still feel like it is too much of an edge case to be generally useful and encourages bad behaviors like not testing restores.

distrimed-github:
Que pena que foi encerrado, essa melhoria ia me ajudar muito.

dwsteele:
I don't think it really helps. In almost all cases you would end up paying more -- infrequent access means exactly that and you pay a real penalty for reading the data.

But pgBackRest repos are designed to be used, for restores, to catch up replicas, etc. Restore testing is also a critical component of any good DR setup.

So, the only way this makes sense is if you make backups but never use or test them. That's just a bad idea. It makes much more sense to use lifecycle rules to automatically transition older data to IA. I've been doing this for many years and I think it strikes the right balance.

PR #2661, opened 2025-08-06
Add HTTP retries for 408 and 429 errors.

dwsteele:
HTTP client errors 408 and 429 are not currently being retried but there seems to be some benefit to doing so.

408 has only been seen once and in that case the server was returning 400 (with Request Timeout in the response body) but it seems worth doing since this could happen during times of high congestion. Requests are not sent until request content is ready so there is not much else to be done to handle this error.

429 has been seen occasionally in the past but now seems to be common on Cloudflare R2. Ideally we would have a large initial back off here but it is not clear that is worth it at this time. The existing Fibonacci back off should be enough to allow operations to proceed if possible during the configured timeout.

### DIFF
--- a/test/src/module/common/ioHttpTest.c
+++ b/test/src/module/common/ioHttpTest.c
@@ -633,7 +633,7 @@ testRun(void)
                 hrnServerScriptAccept(http);
 
                 hrnServerScriptExpectZ(http, "GET / HTTP/1.1\r\n" TEST_USER_AGENT "\r\n");
-                hrnServerScriptReplyZ(http, "HTTP/1.1 429 To Many Requests\r\ncontent-length:0\r\nConnection:close\r\n\r\n");
+                hrnServerScriptReplyZ(http, "HTTP/1.1 429 Too Many Requests\r\ncontent-length:0\r\nConnection:close\r\n\r\n");
 
                 hrnServerScriptClose(http);
                 hrnServerScriptAccept(http);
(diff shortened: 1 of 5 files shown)


test/src/module/common/ioHttpTest.c
```diff
@@ -626,6 +626,18 @@ testRun(void)
...
                 hrnServerScriptAccept(http);
 
+                hrnServerScriptExpectZ(http, "GET / HTTP/1.1\r\n" TEST_USER_AGENT "\r\n");
+                hrnServerScriptReplyZ(http, "HTTP/1.1 408 Request Timeout\r\ncontent-length:0\r\nConnection:close\r\n\r\n");
+
+                hrnServerScriptClose(http);
+                hrnServerScriptAccept(http);
+
+                hrnServerScriptExpectZ(http, "GET / HTTP/1.1\r\n" TEST_USER_AGENT "\r\n");
+                hrnServerScriptReplyZ(http, "HTTP/1.1 429 To Many Requests\r\ncontent-length:0\r\nConnection:close\r\n\r\n");
```
pgguru, line 636: Probably doesn't matter, but should it be `Too Many Requests`?  Assuming the message is mainly human-readable.

pgguru (approved):
A bit tangential, but are all 5xx codes retry-able?  ISTR 501 (not implemented) is permanent, but maybe we haven't hit those in practice.

test/src/module/common/ioHttpTest.c
dwsteele: It doesn't hurt anything because it is a test but it is always better to be tidy. Fixed in c4a481406b1.

dwsteele:
> A bit tangential, but are all 5xx codes retry-able? ISTR 501 (not implemented) is permanent, but maybe we haven't hit those in practice.

In practice we have only seen 500 and 503, which can both be retried. Agreed that some of the others look permanent but not sure it is worth making the distinction.

dwsteele:
> Agreed that some of the others look permanent but not sure it is worth making the distinction.

At least not until we get a complaint!

PR #2674, opened 2025-08-23
Add repo-symlink option to suppress creation of repository symlinks.

dwsteele:
Some otherwise Posix-compatible filesystems do not support symlinks. To support this allow symlinks to be disabled.


dwsteele:
Thank you, @hunleyd!

Since the next release is not until October I'll let this sit a while to see if anyone can test in an immutable env. A long shot, but doesn't hurt to try.

PR #2675, opened 2025-08-25
Fix crash during checksum error warning

dutow:
The STRDEF definition of plural is out of scope in the log message just below it, which leads to either garbage in the log message or a crash with SIGSEGV.

This is particularly visible with pg_tde, which encrypts tables: if pgbackrest is run with checksum enabled, it tries to emit this message for all encrypted files.

### DIFF
--- a/src/command/backup/backup.c
+++ b/src/command/backup/backup.c
@@ -1569,7 +1569,7 @@ backupJobResult(
                                 }
 
                                 // Make message plural when appropriate
-                                const String *const plural = errorTotalMin > 1 ? STRDEF("s") : EMPTY_STR;
+                                const String *const plural = errorTotalMin > 1 ? strNewZ("s") : EMPTY_STR;
 
                                 // ??? Update formatting after migration
                                 LOG_WARN_FMT(


dutow:
Thank you for the quick review!

Yes, I originally wrote the patch the same way, but then I noticed that 75b26319ae changed every occurrence to `strNewZ`, so I followed that instead. Both works.

Ubuntu 24.04 packages definitely show this issue for me. My colleagues also tested it on other platforms and found the same issue. We only tested it with pg_tde, but it should be also reproducible by just manually editing the checksums / data for a few pages.

I don't see the crash itself with pgbackrest built from the source (probably different compilation flags?), but with valgrind the issue is clearly present both in debugoptimized and release build types (when executing `pgbackrest --stanza=db --type=full backup, with checksums enabled in postgres and the database containing tables that will result in multiple checksum errors)

In our case, the script that generates the crash results in this log message:

```WARN: invalid page checksums found in file /home/dutow/pg17inst/data/pgdata/base/16384/16436 at pages 0-1```

Files that only have one checksum error work correctly without the patch, as they don't use the plural form.

dwsteele:
> Yes, I originally wrote the patch the same way, but then I noticed that https://github.com/pgbackrest/pgbackrest/commit/75b26319aea4493d8b9624a61795f8128d50c3b5 changed every occurrence to strNewZ, so I followed that instead. Both works.

Makes sense. In those cases a String is required because of how the var is used/reused later but for this special case we can simplify. I probably should have done that in the first place.

> Ubuntu 24.04 packages definitely show this issue for me.

OK, good to know. I have not added a 24.04 test container yet so that definitely needs to happen.

> the issue is clearly present both in debugoptimized and release build types

I bet this is the key to reproducing. We run valgrind on our unit tests but with -O0 so they can also be used for coverage testing. It looks like we need a valgrind-only unit test run with -O2 enabled to catch these errors.

Thank you for the detailed feedback! Improving the test suite is incredibly valuable.

dutow:
> We run valgrind on our unit tests but with -O0 so they can also be used for coverage testing. It looks like we need a valgrind-only unit test run with -O2 enabled to catch these errors.

Yes, O0/debug build doesn't show the issue even with valgrid.

dwsteele:
Committed!

> This is particularly visible with pg_tde, which encrypts tables: if pgbackrest is run with checksum enabled, it tries to emit this message for all encrypted files.

Bit of a bummer that this doesn't work. There are other TDE implementations that generate the checksum based on the encrypted data (pretty sure, at least) so verification can still be done.

PR #2692, opened 2025-11-06
Remove support for PostgreSQL 9.5.

dwsteele:
Per our policy to support five EOL versions of PostgreSQL, 9.5 is no longer supported by pgBackRest. Remove all logic associated with 9.5 and update the tests.

An effort was made to advance versions as much as possible in the tests while still providing coverage. Hopefully this will reduce churn when future versions expire, though it has created a bit more here.

Tests for 9.4/9.5 are left in the expire/info tests to demonstrate that these commands work with old versions present.


hunleyd (approved):

dwsteele:
Thank you, @hunleyd!

Issue #2697, opened 2025-11-13
Allow cipher suite limitation


bluthg:
Hi!

Many environments require a limitation of allowed TLS cipher suites / levels to e.g. "at least TLS 1.2".
See e.g. the german BSI requirements:
https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-2.pdf?__blob=publicationFile

Ideally, a string like for Postgres' GUC `ssl_ciphers` or Apache's `SSLCipherSuite` should be accepted (which we did with https://github.com/patroni/patroni/pull/1824 for example).

The good news is: organisations that require this want it for each and every connection, so it could and should be set at the lowest possible level :)

I'm struggling to find where and how (`SSL_CTX_set_cipher_list()` & `SSL_CTX_set_ciphersuites()` ?) those options should be set in the code, hence this issue instead of a PR ;-).

What I looked at by now:
- `typedef struct TlsClientNewParam` in `src/common/io/tls/client.h` (would need an additional line `const String *allowedCiphers;` or so (?))
- `tlsContext(void)` in `src/common/io/tls/common.c` (which already disables legacy SSL)

dwsteele:
I think the first thing to do would be to limit the client to TLS >= 1.2. We already do this on the server but we are using a deprecated method to do it, see https://github.com/pgbackrest/pgbackrest/blob/release/2.57.0/src/common/io/tls/server.c#L310C9-L310C29. As of 1.1.0 (which is below our minimum version of 1.1.1) SSL_CTX_set_min_proto_version() should be used instead. This should be true of both the client and server.

As for setting cipher suites, our implementation is based heavily on the Postgres implementation, so it should not be too hard to fit however Postgres does it into our code. I think the suggestion from dosubot looks correct but I don't have time to dig into it right now. A bit annoying if we need different calls for <= 1.2 and > 1.3 as that will make testing more complicated, but again the Postgres code should show the way here.

bluthg:
I _believe_ (could be terribly wrong though!) `SSL_CTX_set_cipher_list()` and `SSL_CTX_set_ciphersuites()` 
a) accept the same string and pick "their relevant part" by themselves
b) effectively render calling `SSL_CTX_set_min_proto_version()` unnecessary (if decent default values are used, i.e. including `:!SSLv1:!SSLv2:!SSLv3:!TLSv1:!TLSv1.1`)

[...]

PR #2700, opened 2025-11-17
Allow configuration of TLS cipher suites.

bluthg:
This is my swing at issue #2697 .

It is about as good as I - as a C, and esp. object-oriented C, layman - can get this.

I was hoping to just set everything in `tlsContext()`, but alas! the configuration cannot be accessed from there.

The tests throw 
```
    ==284== Invalid read of size 8
    ==284==    at 0x199B2E: strZ (string.h:127)
    ==284==    by 0x199B2E: tlsClientNew (client.c:459)
```
errors, which I am pretty clueless about (adding a `MEM_CONTEXT_TEMP_BEGIN()` block didn't fix it).

Also, I'm quite unhappy with the way tests are currently opening TLS connections. I don't see how using the `HRN_CFG_LOAD()` would work...

### DIFF
--- a/doc/xml/release/2020s/2026/2.58.0.xml
+++ b/doc/xml/release/2020s/2026/2.58.0.xml
@@ -58,6 +58,18 @@
             </release-item>
 
             <release-item>
+                <github-issue id="2697"/>
+                <github-pull-request id="2700"/>
+
+                <release-item-contributor-list>
+                    <release-item-contributor id="gunnar.bluth"/>
+                    <release-item-reviewer id="david.steele"/>
+                </release-item-contributor-list>
+
+                <p>Allow configuration of TLS cipher suites.</p>
+            </release-item>
+
+            <release-item>
                 <github-pull-request id="2693"/>
 
                 <release-item-contributor-list>


dwsteele:
This is a pretty good first pass! There are a lot of details to work through, though. For example, I don't know if we want to use the same settings for all servers and clients. For example our internal TLS server can be locked down without needing to configure the client side at all. Also, I think S3/Azure/GCS should have independent configuration in case they need to be different than the internal TLS server, which seems fairly probable to me. Thoughts?

bluthg:
Well, the general idea behind the patch is not to make your CGS et. al. access safer (they almost certainly offer the highest possible levels, so those _will_ be negotiated already), but to satisfy audit requirements that read something like "all communication links _MUST_ be encrypted according to so-and-so recommendations".

Organizations that have this kind of requirement wouldn't allow exceptions anyway... and all the others won't care ;-)

Hence my initial (failed) approach to put this right into `tlsContext()`. Easy to prove: "see, this is where each and every TLS socket comes from, there's no way to forget applying the security setting".

dwsteele:
OK - I see what you mean. Even if we do allow overrides per connection type (GCS, S3, etc.) there should be a global option that affects everything unless overridden. I think I'm OK with that but probably not with your default, which looks pretty aggressive. I might be OK with using the same default PostgreSQL has `HIGH:MEDIUM:+3DES:!aNULL1` pulled from https://www.postgresql.org/docs/current/runtime-config-connection.html#GUC-SSL-CIPHERS. But really I'd rather have no default at all to preserve existing behavior.

My concern is that if we make the default too aggressive we'll break existing installations, something we strive not to do.

bluthg:
Probably force of habit ;-)
I've aligned the default to Postgres'.

dwsteele:
@bluthg OK, I've updated the patch to try to get back to what you were trying to do originally. We can't query the config from directly withing the TLS module but we can add an init function that is run when the config is loaded. We still need some tests to show that the ciphers are being loaded properly, but this is the basic idea.

bluthg:
I LOVE it!
It's still a lot of changed files, but that's just the ironclad style code I'll love to show to auditors :)

I'll try to figure out the test options we have. Not sure about my mileage there, tbth...

dwsteele:
For the tests I think it will be enough to show that the cipher suites have been set appropriately using the getter functions.

But first -- I think the new limit of TLS 1.2 needs to be split out into a separate commit so it can be reviewed on its own -- that's probably the largest change here. You can do that if you like or I'll do it when I have time.

bluthg:
TBTH, I'm pretty sure that I'd make a terrible mess trying... I'm also at best a mediocre git user ;-)

dwsteele:
@bluthg No worries. I'll get to it, just might take a bit. But I still expect this to make the Jan release so all good.

bluthg:
Awesome! :+1:

dwsteele:
@bluthg OK, I broke out the TLS 1.2 requirement into PR #2720. This PR is now based on that PR's branch but there is no way to chain PRs in Github so all the changes from #2720 are included here. But you can diff from that branch locally if you want. As soon as PR #2720 commits this one will reflect the correct diff.

The big change is that I removed the default for tls-cipher-12. I'm just worried that this will cause regressions in object store clients. There are a lot of dodgy custom S3 implementations out there and the default could cause significant breakage.

On the other hand, I think we have pretty good tests to show the cipher suites are used when configured. So what you configure is what you get.

bluthg:
LGTM too! :+1:

PR #2701, opened 2025-11-18
Dynamically size S3/GCS/Azure chunks for large uploads.

dwsteele:
Previously an S3 upload with default repo-storage-upload-chunk-size would only work for files <= 50GiB because of the limited number of chunks allowed. GCS has a smaller chunk size default so it topped out at 40GiB. Azure allows 50,000 chunks so it allowed up to 200GiB.

These are all far larger than files PostgreSQL will create but these days a data directory might also contain files created by plugins that can be much larger.

Since the eventual file size is not known in advance (due to compression) it is hard to pick an appropriate chunk size in advance. Instead, dynamically grow the chunk size over time to reach 5TiB for S3 and GCS (their upper limit). Azure has more parts so it will reach 45TiB, which is smaller than the upper limit of 190TiB, but seems sufficient for now.

The default buffer size is used for the first GiB (plus some) to provide compatibility with any clones that do not support variable block sizes. There is no evidence that this is a problem but better to be safe.

The minimum values for repo-storage-upload-chunk-size have been increased to match vendor minimums and simply the chunk size algorithm.

### DIFF
--- a/doc/xml/release/2020s/2026/2.58.0.xml
+++ b/doc/xml/release/2020s/2026/2.58.0.xml
@@ -19,6 +19,19 @@
 
         <release-improvement-list>
             <release-item>
+                <github-issue id="2679"/>
+                <github-pull-request id="2701"/>
+
+                <release-item-contributor-list>
+                    <release-item-ideator id="timothee.peignier"/>
+                    <release-item-contributor id="david.steele"/>
+                    <release-item-reviewer id="douglas.j.hunley"/>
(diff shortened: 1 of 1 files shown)


dwsteele:
@hunleyd It's definitely tricky to test. Outside of the unit tests you'd need to jack the log level up to trace and check for increasing buffer sizes in files > ~1.5GiB compressed. For sanity testing I added some logging at info level, which is no longer there of course.

Thank you for the review!

PR #2587, opened 2025-03-27
feat: support upload s3 by http

Wenne:
Please read [Submitting a Pull Request](https://github.com/pgbackrest/pgbackrest/blob/main/CONTRIBUTING.md#submitting-a-pull-request) before submitting.

Issue #2340, opened 2024-04-24
About pgbackrest supports the http protocol


booven:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version:2.47

2. PostgreSQL version:12

7. Describe the issue:
   Hello, I see that pgbackrest currently only supports https protocol, is there any relevant option for users to choose, or can users release the restrictions by changing the underlying code? If yes, could you please inform us of the location for modification? Thank you very much.

dwsteele:
> is there any relevant option for users to choose

Not currently.

> can users release the restrictions by changing the underlying code

It's not just one place, though in concept it would not be that hard to do. Basically, wherever HTTP is used you'd need to remove the `tlsClientNewP()` constructor and leave the `sckClientNew()` constructor, see https://github.com/pgbackrest/pgbackrest/blob/main/src/storage/s3/storage.c#L1226 for an example. Might not be that simple, but that's the idea.

You want to tie that into the endpoint for each service so it would behave this way for `http` and keep the current behavior for `http`. If you do that for S3, GCS, and Azure you can submit a PR and we'll likely commit it.

booven:
Hi, I followed your suggestion to remove the tlsClientNewP() from storage.c and keep the sckClientNew() construct, but it still has the following problem: **unexpected eof while reading line**, it seems to have nothing to do with the http protocol, I don't know what is causing it. 

pgbackrest.conf:
```
[global]
repo1-path=/data
compress-type=zst
compress-level=1
repo1-retention-full-type=count
repo1-retention-full=7
process-max=10
log-level-console=detail
log-level-file=info
start-fast=y
delta=y
resume=y
archive-check=y
log-path=./pgbackrest_backup/logs
archive-timeout=300
lock-path=./pgbackrest_backup/lock

repo1-s3-bucket=100007
repo1-s3-endpoint=http://100.124.255.12:80
repo1-s3-key=5DCEFE69034761D9A86106DBEBD521F9D
repo1-s3-key-secret=QHTajSR6LM2vzc3
repo1-s3-region=cn-1
repo1-storage-verify-tls=n
repo1-s3-uri-style=path
repo1-type=s3
repo1-block=y
repo1-bundle=y
backup-standby=n

tls-server-address=*
tls-server-cert-file=./certs/backup-srv.crt
tls-server-key-file=/./certs/backup-srv.key
tls-server-ca-file=./certs/ca.crt

[...]

Issue #1590, opened 2021-12-09
Allow non-TLS connections to S3 servers


ron1:
pgbackrest should allow non-TLS connections to S3 servers for the following reasons:
1. The gold standard Amazon S3 allows both HTTP and HTTPS requests.
2. Kubernetes service meshes already provide mTLS between services so to require additional certificate management in order to connect a pgbackrest client to an S3 server via one-way TLS in a k8s cluster introduces unnecessary complexity.
3. Some Kubernetes clusters encrypt traffic between nodes via IPSec thereby eliminating the need to encrypt the connection between a pgbackrest client and an in-cluster S3 server.

Consider enhancing pgbackrest to support non-TLS connections identified either by specification of an http protocol url scheme or port 80.

dwsteele:
This seems reasonable. Added a project card for this feature at https://github.com/pgbackrest/pgbackrest/projects/2#card-74786045.

How would you like your name to appear in the release notes?

PR #2703, opened 2025-11-24
HTTP support for S3, GCS, and Azure.

wjmorland:
Add support for HTTP endpoints for S3, Azure, and GCS. Allow users to specify HTTP in the endpoint, but default to using HTTPS in all other scenarios to preserve existing behavior.

Extend the URL parser with a `defaultType` parameter to support either:
- explicitly specifying a protocol via `.type` and enforcing that protocol is used in the URL
- allowing protocol to be parsed from URL, but providing default via `.defaultType` if no protocol is found in the URL

Add partial write handling in fdWrite to support non-blocking socket operations. The write loop now handles EAGAIN errors by waiting for the file descriptor to become writable, and continues writing the remaining bytes when write() returns fewer bytes than requested. This is required for HTTP, which may use non-blocking sockets, but doesn't have built in handling like the TLS client we are using for HTTPS. Also wrap the write() call, add a shim, and additional logging for easier unit testing.

The main motivation for this is to allow pgBackRest to work natively with Ceph HTTP frontends. This was previously brought up in #2340 for a Ceph offshoot called MicroCeph, where it looks like they instead fixed the issue for that specific product on the Ceph side by adding HTTPS support [in this PR](https://github.com/canonical/microceph/pull/355). However, I think it would still be beneficial to have HTTP support so the problem is solved from both ways (and this would help my company's use case for backing up to our own private Ceph cluster that doesn't have HTTPS enabled). Also saw that there was some work done on this in #2587, but wasn't sure if that was actively being worked on, so wanted to propose another implementation + add support for Azure/Google Cloud. 

Looks like everything is passing when I run the CI checks on my end (unit tests, code coverage, code-format) and did some manual testing with backing up one of my dev databases, seemed to be working well.

### DIFF
--- a/doc/xml/release/2020s/2026/2.58.0.xml
+++ b/doc/xml/release/2020s/2026/2.58.0.xml
@@ -20,6 +20,18 @@
 
         <release-feature-list>
             <release-item>
+                <github-issue id="2340"/>
+                <github-pull-request id="2703"/>
+
+                <release-item-contributor-list>
+                    <release-item-contributor id="will.morland"/>
+                    <release-item-reviewer id="david.steele"/>
+                </release-item-contributor-list>
+
+                <p>HTTP support for <proper>S3</proper>, <proper>GCS</proper>, and <proper>Azure</proper>.</p>
+            </release-item>
+
+            <release-item>
                 <github-issue id="2666"/>
                 <github-pull-request id="2709"/>
 
--- a/doc/xml/release/contributor.xml
+++ b/doc/xml/release/contributor.xml
@@ -1206,6 +1206,11 @@
     <contributor-name-display>Will M</contributor-name-display>
 </contributor>
 
+<contributor id="will.morland">
+    <contributor-name-display>Will Morland</contributor-name-display>
+    <contributor-id type="github">wjmorland</contributor-id>
+</contributor>
+
 <contributor id="william.cox">
     <contributor-name-display>William Cox</contributor-name-display>
     <contributor-id type="github">mydimension</contributor-id>


dwsteele:
I must say this is a very well thought out and complete patch. Kudos! I see a few minor things I'll want changed but overall I don't see any barriers to getting this committed. Should be no problem for get this in for the January release.

wjmorland:
Thank you! That's great, looking forward to having this included. 

Please let me know if you want me to address any of those minor issues, didn't see any comments on the code yet, so wasn't sure if you were planning on just fixing yourself before merging.

dwsteele:
I'll post a review soon once I have digested the patch a bit.

4n3w:
Curious! Was this ever addressed in https://github.com/pgbackrest/pgbackrest/issues/1590 ? I'm here because we can't get http on a self-hosted minio server to play ball with pgbackrest. The link to the card in the old issue gives me a 404 🤷🏼‍♂️

dwsteele:
@4n3w It wasn't addressed. That's what the current patch is about. There was also an earlier attempt that was abandoned by the author.

> The link to the card in the old issue gives me a 404

At some point GitHub forced a migration to their new project system, which broke all the links.

wjmorland:
Thanks for the review! Makes sense on testing some actual operations with HTTP, added a few basic cases (read, write, error), and happy to add some of the more complex cases too if you want. Also made the helper function static as suggested.

wjmorland:
Sounds good, I see how that makes the S3/GCS/Azure tests much cleaner, thanks for making those changes

dwsteele:
Committed. Thank you!

wjmorland:
Awesome, thanks for including this!

PR #2705, opened 2025-11-24
Optimize S3/GCS/Azure chunk size for small files.

dwsteele:
The prior code allocated the entire chunk buffer when the file was opened. However, in practice many files are smaller than the chunk buffer, especially in the main process.

Instead grow the chunk buffer as data comes in to save memory when smaller files are being processed. This adds some overhead for reallocations but modern processors do this very efficiently so it should not be significant compared to the cost of compressing, encrypting, and transferring files. Even so, the growth is fairly aggressive when the input buffers are full so only one or two reallocation are required to get to the default chunk size.

### DIFF
--- a/doc/xml/release/2020s/2026/2.58.0.xml
+++ b/doc/xml/release/2020s/2026/2.58.0.xml
@@ -32,6 +32,17 @@
             </release-item>
 
             <release-item>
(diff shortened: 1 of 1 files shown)


dwsteele:
@hunleyd Thank you for the review!

PR #2715, opened 2025-12-17
Fix deadlock due to logging in signal handler.

silent-observer:
Previously it was possible to achieve a deadlock in a signal handler, for example when SIGTERM (i.e. sent by `pgbackrest stop --force`) arrives when a lock used in `gmtime_r` is taken. Then the next time logging is done, it will deadlock on `gmtime_r`.

In general, most stdlib functions are not safe to call in signal handlers, only so called "async-signal-safe" ones are. In particular, `snprintf` isn't safe either since it is allowed to internally call `malloc`.  The whole `exitSafe` function isn't safe either due to extensive use of allocations. Because of this, we need to use a simpler logging format in signal handlers, one that only uses plain `strncpy` and `write`.

### DIFF
--- a/doc/xml/release/2020s/2026/2.58.0.xml
+++ b/doc/xml/release/2020s/2026/2.58.0.xml
@@ -4,6 +4,20 @@
             <p><b>IMPORTANT NOTE</b>: The minimum values for the <setting>repo-storage-upload-chunk-size</setting> option have increased. They now represent the minimum allowed by the vendors.</p>
         </text>
 
+        <release-bug-list>
+            <release-item>
+                <github-issue id="2710"/>
+                <github-pull-request id="2715"/>
+
+                <release-item-contributor-list>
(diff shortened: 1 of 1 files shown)


dwsteele:
Committed. Thank you for the fix!

dwsteele:
@silent-observer BTW -- this may not concern you or you may already know, but the email address you use for your commits is not registered with GitHub (I presume) so you don't show up as a contributor to the project on the website. Fine by us but just in case you were not aware...

silent-observer:
That's weird, pretty sure m.michkov@arenadata.io is used on those commits and it is set as primary email on GitHub. Thank you for telling me though!

dwsteele:
Hrm, my mistake. You are there. I did a search earlier and couldn't find you but now I can. Wish I could use lack of coffee as an excuse...

PR #2720, opened 2026-01-06
Require TLS >= 1.2 unless verification is disabled.

dwsteele:
TLS 1.0 and 1.1 have been deprecated since 2021 (see RFC-8996) and TLS 1.2 has been available since 2008. As such it makes sense to require TLS >= 1.2 when verification is enabled. Verification is always enabled for TLS protocol sessions within pgBackRest but can be disabled for object stores (.e.g. S3) to support self-signed certificates on internal servers.

There is a slight change in behavior when verification is disabled. In prior versions SSL 2/3 would be disabled but now they are allowed (as well as TLS 1.0/1.2). With verification disabled it doesn't seem useful to be picky about protocol versions and disabling TLS 1.0/1.1 could easily cause breakage on older TLS servers.

### DIFF
--- a/doc/xml/release/2020s/2026/2.58.0.xml
+++ b/doc/xml/release/2020s/2026/2.58.0.xml
@@ -82,6 +82,18 @@
             </release-item>
 
             <release-item>
+                <github-pull-request id="2720"/>
+
+                <release-item-contributor-list>
+                    <release-item-contributor id="david.steele"/>
+                    <release-item-reviewer id="douglas.j.hunley"/>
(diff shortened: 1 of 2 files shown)


dwsteele:
@turnstep @pgguru @hunleyd @bluthg @pgstef Comments and/or review on this appreciated!

dwsteele:
@hunleyd Yeah, looked like that was a bad package update so was just waiting for it to resolve. All good now.

bluthg:
LGTM too

dwsteele:
@hunleyd @bluthg Thank you for the review!

PR #2722, opened 2026-01-15
pgbackrest: inclusion list for Table Access Methods

kiwicopple:
This RFC proposes making pgBackRest's hardcoded relation file pattern configurable to support alternative PostgreSQL Table Access Methods (TAMs) like OrioleDB.

Currently, pgBackRest uses a hardcoded regex pattern that only recognizes standard PostgreSQL fork types (_fsm, _vm, _init). This prevents proper backup of databases using alternative storage engines that may use different file naming conventions.

The RFC proposes several implementation options, with a recommendation for a configuration-based approach that adds new options like `relation-fork` and `relation-no-checksum` to extend the default pattern without breaking backward compatibility.

### DIFF
--- a/RFC.md
+++ b/RFC.md
@@ -1,0 +1,338 @@
+# RFC: Externalize Relation File Inclusion List for Table Access Methods
+
+## Status
+Draft - Seeking Feedback
+
+## Authors
+- Supabase Team
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


dwsteele:
This level of discussion would be better done as an issue. Maybe that's what you were thinking when you closed the PR?

We'd prefer to use PRs for actual implementation, though of course your RFC would inform eventual documentation and implementation.

PR #2730, opened 2026-02-08
Push read range handling into the storage layer.

dwsteele:
It was previously the responsibility of the calling code to combine adjacent ranges and issue a read for each range. This worked but it added complexity to already complex code, added duplication, and limited opportunities for optimization.

Instead, handle ranges in the storage layer. This reduces duplication (especially of adjacency detection) and simplifies the calling code. More importantly it allows for various optimizations.

For the most part this is a refactor and does not add any optimizations but since ranges are now understood by the storage layer they are pushed to the remote (indeed this is the easiest way to do it) which reduces the number of remote calls.

Most of the changes are rote based on the new data structures. The files to look at closely are: `src/command/restore/blockDelta.c`, `src/command/restore/file.c`, `src/storage/range.c`, and `src/storage/read.c`

Review of this commit benefits from hiding whitespace since there are some big indentation changes.


dwsteele:
Closing this in favor of an upcoming implementation.

PR #2731, opened 2026-02-09
Refactor restore module into included modules.

dwsteele:
The restore module has been large and unwieldy for some time but I have been loathe to split it into separate compilation units because it means maintaining additional header files, updating build files, and losing optimizations from static functions. These functions are only used internally by restore and it seems wasteful to extern them.

We already have a number of cases where C files are included directly into other C files, especially .vendor.c.inc and .auto.c.inc files. We also include C files to add functionality needed for build/doc/test to core objects without having to add that functionality to core. See src/build/common/string.c for an example.

The test/coverage code already supports C includes but I had to update it to recognize the new extension.

I'm curious to know if there are any objections to this pattern. It seems the best of both worlds to me but definitely makes C includes more prominent in the project whereas before they were not too common. The idea would be to move on to other big files like backup.c and manifest.c and give them the same treatment.

### DIFF
--- a/doc/xml/release/2020s/2026/2.59.0.xml
+++ b/doc/xml/release/2020s/2026/2.59.0.xml
@@ -12,5 +12,19 @@
                 <p>Suppress <id>unused parameter</id> errors in <proper>meson</proper> compiler probes.</p>
             </release-item>
         </release-improvement-list>
+
+        <release-development-list>
+            <release-item>
+                <github-issue id="2731"/>
+
+                <release-item-contributor-list>
+                    <release-item-contributor id="david.steele"/>
+                    <release-item-reviewer id="douglas.j.hunley"/>
+                    <release-item-reviewer id="david.christensen"/>
+                </release-item-contributor-list>
+
+                <p>Refactor restore module into included modules.</p>
+            </release-item>
+        </release-development-list>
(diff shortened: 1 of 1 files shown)


hunleyd:
I _like_ the idea!

dwsteele:
> I like the idea!

Glad to hear it. I don't want to stray too far from what is "normal" in C projects but at the same time...

pgguru:
What are the benefits of this approach, basically just smaller conceptual units that are stitched together at compile time?  I guess personally I'm +0 with this approach; seems like code editors should support the navigation issues, but if others find benefits with this approach I'm not against it.

dwsteele:
> What are the benefits of this approach, basically just smaller conceptual units that are stitched together at compile time?

@pgguru Yes -- I find that I'm often working in two sections at the same time (normally a section and what is now restore.c). Sure, I can open two windows for that but it gets a bit confusing which is which since they are both labelled with the same name. Also -- it is hard to find the beginning and end of each section. They are grouped together in the original file but you need to find the boundaries -- I could add some dashes or something to help but a separate file seems easier.

At least you don't hate it! ;)

dwsteele:
Since there were no objections I'm committing this. Thank you both for the review!

PR #2738, opened 2026-02-20
C11 is now the minimum C standard.

dwsteele:
This standard is fifteen years old and the features we are interested in seem well supported on popular compilers.

The main advantage is that static_assert() will now display the specified message on error rather than the ever-cryptic `negative width in bit-field '__error_if_negative'`. Now that we can depend on having static_assert() we can replace our STATIC_ASSERT_STMT() macro.

Replace our ALIGN_OF() macro with alignof().

Replace our FN_NO_RETURN macro with noreturn. Include stdnoreturn.h in build.h to avoid needing to include it in many header files.

Use an anonymous union in common/type/json.c where it simplifies syntax. Other uses of union seem better as they are.


AliNazirKosar:
I reviewed all 13 files in this PR. The changes are consistent across the entire project, both in the main code and in the tests.

The most frequent change is replacing FN_NO_RETURN with noreturn, which appears in almost every file. The most notable change is in meson.build, where the C standard is updated from C99 to C11 and a check is added to ensure the compiler supports it.

STATIC_ASSERT_STMT and ALIGN_OF are also replaced by their standard equivalents static_assert and alignof, and in json.c the field access is simplified using an anonymous union, removing the intermediate .value.

No issues found.

dwsteele:
@AliNazirKosar OK, great, now please approve the PR. "Files Changed" -> "Submit Review" -> "Approve". Thanks!

AliNazirKosar (approved):

AliNazirKosar:
@dwsteele 

If you ever need me to review another PR or help with anything else, just let me know here or on LinkedIn.

PR #2740, opened 2026-02-24
Add per-repo backup progress to info command output.

### LINKED ISSUE #2736: backup progress reported as average when multiple repos running simultaneously
I noticed some interesting behavior related to checking backup progress, when running backups on multiple repos simultaneously, running `pgbackrest info --detail-level=progress` does not report progress for each individual repo, but instead, appears to be doing some sort of averaging. 

Example:
```
[18:38:41] [dba-test@tot-t-db-01 ~]$ pgbackrest info --stanza=test_01 --detail-level=progress
stanza: test_01
    status: ok (backup/expire running - 41.40% complete)
[18:39:29] [dba-test@tot-t-db-01 ~]$
[18:39:30] [dba-test@tot-t-db-01 ~]$
[18:39:30] [dba-test@tot-t-db-01 ~]$
[18:39:30] [dba-test@tot-t-db-01 ~]$ pgbackrest info --stanza=test_01 --detail-level=progress --repo=1
stanza: test_01
    status: ok (backup/expire running - 62.64% complete)
[18:39:41] [dba-test@tot-t-db-01 ~]$
[18:39:42] [dba-test@tot-t-db-01 ~]$
[18:39:42] [dba-test@tot-t-db-01 ~]$
[18:39:42] [dba-test@tot-t-db-01 ~]$ pgbackrest info --stanza=test_01 --detail-level=progress --repo=2
stanza: test_01
    status: ok (backup/expire running - 21.84% complete)
[18:39:43] [dba-test@tot-t-db-01 ~]$
```

I thought this was somewhat interesting, would probably have expected a per-repo progress value, or even using the slowest value. Otherwise, when the backup on the quicker repo finishes, I think the progress would actually go backwards, which feels a bit confusing to me.

Thoughts on updating this to report progress for each individual repo?

wjmorland:
When backups are running on multiple repositories simultaneously, the info command now reports per-repo progress in addition to the existing overall progress. A new repo array is included in JSON output for backup locks. This avoids confusing progress jumps when one repo finishes before another.

Resolves #2736.


dwsteele (approved):
I made a few changes in 7da4121ae7 but overall looks pretty good to me.

I'm not entirely satisfied with the formatting of the text output (grouping with the error statuses) but I can't think of a better way. The good news is we are free to change the text output at any time so it is not too big a deal. The JSON looks fine.

wjmorland:
Great, thanks for making those updates!

Looking back at this, I see what you mean about the grouping with error statuses. Unfortunately can't think of a better way at the moment either

test/src/module/command/infoTest.c
```diff
@@ -1882,6 +1882,8 @@ testRun(void)
                     infoRender(),
                     "stanza: stanza1\n"
                     "    status: ok (backup/expire running - 65.28% complete, restore running - 65.28% complete)\n"
+                    "        repo1 backup: 55.55% complete\n"
```
pgstef, line 1885: We actually have 2 restore running here. 
Repo1: size = 3159000, sizeComplete = 1754830, percentComplete = 5555
Repo2: size = 3159000, sizeComplete = 2369250, percentComplete = 7500

So this change only display the backup progress per repo, not the restore part, right?
Wouldn't that be confusing? (not that multi-restore would happen quite often though)

wjmorland: I was previously assuming that multiple concurrent restores would be disallowed, can definitely add the same progress reporting for restores. I think that even if multi-restore is rare, could be helpful to standardize this so you can always tell from progress output which repo the restore is running off of. I did see that David already approved - please let me know if you want to merge this and I can add restore support in a follow-up, or if I should just update this PR. 

Additionally, for my own understanding, I'm not sure I fully understand the value of the multi-restore scenario. To me, it seems that having multiple restores writing to the same data directory at the same time would be a bad idea/could easily cause issues if there are differences between the backups being restored. I also did a quick test of a multi-restore scenario just to see how it would work, found that it was pretty easy to hit issues. Is there a use case for multi-restores that I'm missing? 

---

Example of doing multiple restores at the same time causing issues
```
[18:30:29] [~]$ pgbackrest restore --stanza=dev_v02_5435 --repo=2 --set=20260324-174907F_20260324-175906I
2026-03-24 18:30:50.981 P00   INFO: restore command begin 2.58.0: --cmd=/usr/bin/pgbackrest --config=/home/dba-test/.config/pgbackrest/5435/pgbackrest.conf --delta --exec-id=861322-afcf4f3c --log-level-console=info --log-level-file=info --log-path=/home/dba-test/pgbackrest/log/5435 --pg1-path=/postgres/pg15/5435/data --process-max=64 --repo=2 --repo2-azure-account=<redacted> --repo2-azure-container=*** --repo2-azure-key=<redacted> --repo2-azure-key-type=sas --repo1-path=/dev/backups/pgbackrest/ --repo2-path=/dev/backups/pgbackrest/ --repo1-s3-bucket=*** --repo1-s3-endpoint=*** --repo1-s3-key=<redacted> --repo1-s3-key-secret=<redacted> --repo1-s3-region=default --repo1-s3-uri-style=path --no-repo1-storage-verify-tls --repo1-type=s3 --repo2-type=azure --set=20260324-174907F_20260324-175906I --spool-path=/postgres/backups/pgbackrest/spool --stanza=dev_v02_5435
2026-03-24 18:30:51.259 P00   INFO: repo2: restore backup set 20260324-174907F_20260324-175906I, recovery will start at 2026-03-24 17:59:06
2026-03-24 18:30:51.261 P00   INFO: remove invalid files/links/paths from '/postgres/pg15/5435/data'
2026-03-24 18:30:59.569 P00   INFO: write updated /postgres/pg15/5435/data/postgresql.auto.conf
2026-03-24 18:30:59.573 P00  ERROR: [073]: unable to sync missing path '/postgres/pg15/5435/data/base/16422'
2026-03-24 18:30:59.573 P00   INFO: restore command end: aborted with exception [073]
```
```
[18:29:27] [~]$ pgbackrest restore --stanza=dev_v02_5435 --repo=1 --set=20260304-142717F
2026-03-24 18:30:53.651 P00   INFO: restore command begin 2.58.0: --cmd=/usr/bin/pgbackrest --config=/home/dba-test/.config/pgbackrest/5435/pgbackrest.conf --delta --exec-id=861431-c4a530f1 --log-level-console=info --log-level-file=info --log-path=/home/dba-test/pgbackrest/log/5435 --pg1-path=/postgres/pg15/5435/data --process-max=64 --repo=1 --repo2-azure-account=<redacted> --repo2-azure-container=*** --repo2-azure-key=<redacted> --repo2-azure-key-type=sas --repo1-path=/dev/backups/pgbackrest/ --repo2-path=/dev/backups/pgbackrest/ --repo1-s3-bucket=*** --repo1-s3-endpoint=*** --repo1-s3-key=<redacted> --repo1-s3-key-secret=<redacted> --repo1-s3-region=default --repo1-s3-uri-style=path --no-repo1-storage-verify-tls --repo1-type=s3 --repo2-type=azure --set=20260304-142717F --spool-path=/postgres/backups/pgbackrest/spool --stanza=dev_v02_5435
2026-03-24 18:30:53.721 P00   INFO: repo1: restore backup set 20260304-142717F, recovery will start at 2026-03-04 14:27:17
2026-03-24 18:30:53.722 P00   INFO: remove invalid files/links/paths from '/postgres/pg15/5435/data'
2026-03-24 18:31:01.813 P00   INFO: write updated /postgres/pg15/5435/data/postgresql.auto.conf
2026-03-24 18:31:01.816 P00   INFO: restore global/pg_control (performed last to ensure aborted restores cannot be started)
2026-03-24 18:31:01.816 P00   INFO: restore size = 1.5GB, file total = 1588
2026-03-24 18:31:01.817 P00   INFO: restore command end: completed successfully (8169ms)
```
Server starts fine, seems like the restore that finished last worked ok

pgstef: It can be useful to restore on a different pgdata so you can validate both backups separately as test restores. 
But definitely restoring from a different backup set into the same pgdata should be avoided imo.
Maybe that's another topic to discuss in a separated issue though.

What I was mentioning here is just a matter of text message standards. It's tracking progress per repo overall on top of the combined view. So I would definitely prefer to see the same message as above but per repo. If that would be doable.

dwsteele: > But definitely restoring from a different backup set into the same pgdata should be avoided imo.

Agreed, but that would require a different kind of locking -- something in pg-path, I think. Even so I don't even think the currently locking is really right -- if you don't specify a repo for restore I think the restore will be reported as repo1 no matter where it actually comes from.

> It's tracking progress per repo overall on top of the combined view. So I would definitely prefer to see the same message as above but per repo. If that would be doable.

I'm not sure what you mean by this. Do you mean:
```
status: ok (backup/expire running - 65.28% complete, restore running - 65.28% complete)
    repo1 backup: 55.55% complete
    repo2 backup: 75.00% complete
    repo1 restore: 65.28% complete 
```
In that case I wonder if we want the averaged output in the text output at all. We need to keep it in the JSON but we are free to update the text output however we like.

pgstef: > I'm not sure what you mean by this. Do you mean:
> 
> ```
> status: ok (backup/expire running - 65.28% complete, restore running - 65.28% complete)
>     repo1 backup: 55.55% complete
>     repo2 backup: 75.00% complete
>     repo1 restore: 65.28% complete 
> ```

Something like that yes.

> In that case I wonder if we want the averaged output in the text output at all. We need to keep it in the JSON but we are free to update the text output however we like.

It feels a bit odd to only display backup progress when a restore is also running, especially if that progress is an aggregation of multiple restores. I do like having all the information available in the JSON output, and a combined status makes sense for a high level view across repositories. However, I have mixed feelings about adding more per repo text output.

From what I have seen, users working with multiple repositories typically do not run backups in parallel anyway because of the IOPS impact. They tend to chain them instead. In the rare case where multiple backups are running and per repo visibility is needed, I would rather suggest using `--repo=X` for more detailed output. I am generally not a fan of increasing verbosity in the global report.

Also, I suspect that running backups and restores simultaneously on the same host is not a very common scenario, so I would prefer not to make this PR overly complex for that edge case. I just wanted to raise the question to confirm this is the intended behavior and to make sure restores were considered too.

I think backup progress is the more useful anyway, so I am totally fine with the current approach in this PR 🙂

dwsteele:
> However, I have mixed feelings about adding more per repo text output.

After some thought I agree. As Stefan pointed out, if you want the per-repo output in text mode you can just add the --repo option.

I think we should remove the additional text output but leave the JSON as is.

test/src/module/command/infoTest.c
wjmorland: > It can be useful to restore on a different pgdata so you can validate both backups separately as test restores.
But definitely restoring from a different backup set into the same pgdata should be avoided imo.

> Agreed, but that would require a different kind of locking -- something in pg-path, I think. Even so I don't even think the currently locking is really right -- if you don't specify a repo for restore I think the restore will be reported as repo1 no matter where it actually comes from.

Makes sense, agree on this. Definitely seems like pg-path locking would be better - this would be super helpful now that I'm thinking about it. Repo-level locking doesn't really do much imo, as it seems like a completely valid scenario to use the same repo to restore two different backups to two different data directories for testing, or if i want to do multiple PiTRs to different times if I don't know the exact timestamp of some data getting dropped, etc. 

However, this is a separate issue from the goal of this PR.

> From what I have seen, users working with multiple repositories typically do not run backups in parallel anyway because of the IOPS impact. They tend to chain them instead.

+1 on this, that is our normal usage. I just happened to notice this discrepancy when doing some one-off manual backups, didn't come up in normal operations. 

> It feels a bit odd to only display backup progress when a restore is also running, especially if that progress is an aggregation of multiple restores.

Agree, think whichever way we go, it should be consistent across both backups and restores. 

> However, I have mixed feelings about adding more per repo text output.... I am generally not a fan of increasing verbosity in the global report.

Makes sense, this is a good point. I would also add that since this is a relatively uncommon scenario, not sure if it  makes sense to make the output more complex all the time just to handle it. 

However, if we leave the per-repo output unchanged, it doesn't address the original issue of the text output having confusing jumps. Totally fine if the answer is to use per-repo output, this is pretty easy to do, mainly wanted to call out that addressing those jumps in the aggregated output was my original goal.

Since the text output seems relatively flexible, I'm wondering if we could do something like the following:
- text output
    - keep existing output if only a single backup/restore is running
    - if multiple concurrent backups/restores are running, switch to showing the repo-level output as we have in this PR
- json output
   - always show repo level progress, and add this output for restores too so it is consistent

That would address the original goal as well as avoid making the text output overly verbose for normal usage. Please lmk your thoughts on that - otherwise, I'm good with removing the text changes entirely and adding the per repo output to both JSON outputs like I think David is suggesting below.

dwsteele: > However, this is a separate issue from the goal of this PR.

I think the way to go with restores is explicit stanza-level locking. Since users rarely specify an explicit repo (by design) that's the way it works in practice now so I think it makes sense to formalize it. But, as you say, that's a separate issue.

> keep existing output if only a single backup/restore is running -- if multiple concurrent backups/restores are running, switch to showing the repo-level output as we have in this PR

I'm OK with this unless Stefan has an objection. I'd rather skip the restore logic for now because I think we should change the locking.

> always show repo level progress, and add this output for restores too so it is consistent

Let's hold off on adding repo-level restore progress for the same reason as above.

pgstef: > I'm OK with this unless Stefan has an objection.

No objection, I'm totally fine with that :+1:

Issue #2727, opened 2026-02-03
Idea : Direct I/O


anayrat:
Hello,
While reading this PR https://github.com/wal-g/wal-g/pull/2003

I wonder if it could be useful for pgbackrest. The idea is to bypass OS cache in order to not affect OS cache used by Postgres.

Maybe there are ideas to explore with io_uring too.

Regards,

anayrat:
Oh, I notice, it is a duplicate of https://github.com/pgbackrest/pgbackrest/issues/1038

dwsteele:
I thought about using `io_uring` after seeing Munro's talk in Montreal last year but came to the conclusion that for our use case it is not very helpful while at the same time adding a ton of complexity. We do only sequential reads and the kernel has no problem figuring this out so there is no reason for us to think about read ahead.

Polluting the Linux cache is absolutely a problem and it seems like `POSIX_FADV_DONTNEED` would be the simplest solution. However, I have not found a way to test that this is actually working. If I simply flush the kernel cache it will happily fill it back up no matter what `fadvise()` says, probably because it is empty and why not?

Any ideas on how this could be tested?

anayrat:
Indeed, io_uring may be overkill for just sequential read. 

I played with several linux tools (vmtouch, dd, advise ...), but I was not able to see how to avoid cache.

I tried to load a file1 in the cache with willneed flag, then read a file2 with dontneed flag. And check if file1 is still loaded (and not file2).
Unfortunately, % of file1 in cache dropped and file2 increased.

dwsteele:
I'll generate a PR based on what I have and see if anyone can figure out how to test that it works.

dwsteele:
Closing since Direct IO is probably not the approach we want to take and there is a PR for fadvise.

PR #2758, opened 2026-03-30
Set fadvise for repo type posix

### LINKED ISSUE #1038: Feature Request: Ability to prevent pgbackrest from impacting heavily OS disk block cache
Something which has been already raised in some issues like https://github.com/pgbackrest/pgbackrest/issues/822

PgBackRest is reading/writing a lot on disk and for large databases, it simply nuke the OS disk cache. However these data are not really useful in the cache.

In the following case: 16GB box, with a PG with ~300GB of data on disk, when running a pgbackrest backup, it happens that the load of the box gets really high because of the `kswapd` process of the kernel starting to move pages around, especially pages potentially being swapped out.

From my point of view, there is no problem in preventing pgbackrest from using the disk cache, to keep the useful data from PG into the OS cache, and use a little more disk. At least I don't take the risk seeing the server load average getting to hundreds.

There are two ways to do this usually:
* Using the flag `O_DIRECT` when opening a file (usually not advised)
* Using the syscall `fadvise` on the read/write blocks: https://linux.die.net/man/2/fadvise

For instance the `dd` command is using `fadvise` when the "nocache" flag is enabled: https://github.com/coreutils/coreutils/blob/master/src/dd.c#L1121

Seeing such configuration option in PgBackRest would be absolutely great.

Thanks a lot

fmbiete:
Some years back we read [Issue #1038](https://github.com/pgbackrest/pgbackrest/issues/1038), we have had some success marking the reads/writes from/to repo type posix as "don't need".

While it still doesn't solve the part of the cache being polluted by the reads from the datadir, it avoids the pollution because of the backup copy...

### DIFF
--- a/src/storage/posix/read.c
+++ b/src/storage/posix/read.c
@@ -89,6 +89,9 @@ storageReadPosixOpen(THIS_VOID)
         // Set free callback to ensure the file descriptor is freed
         memContextCallbackSet(objMemContext(this), storageReadPosixFreeResource, this);
 
+        // advise we don't need to keep on cache
+        posix_fadvise(this->fd, 0, 0, POSIX_FADV_DONTNEED);
+        
         // Seek to offset
         if (this->interface.offset != 0)
         {
--- a/src/storage/posix/write.c
+++ b/src/storage/posix/write.c
@@ -117,6 +117,9 @@ storageWritePosixOpen(THIS_VOID)
     // Set free callback to ensure the file descriptor is freed
     memContextCallbackSet(objMemContext(this), storageWritePosixFreeResource, this);
 
+    // advise we don't need to keep on cache
+    posix_fadvise(this->fd, 0, 0, POSIX_FADV_DONTNEED);
+
     // Update user/group owner
     if (this->interface.user != NULL || this->interface.group != NULL)
     {


dwsteele:
We have recently discussed this in #2727 and I created a branch to test it out https://github.com/pgbackrest/pgbackrest/compare/main...dwsteele:pgbackrest:dws-fadvise-ci. @anayrat and I were not able to verify that this actually works.

I also wonder if we want a couple of exceptions here, delta backup for one (files may be read twice), and block incremental restore (files are checksummed after write).

fbernier:
@dwsteele I just stumbled on the patch you linked and as it is written it's basically a no op. `POSIX_FADV_DONTNEED` tells the kernel that it doesn't need whatever is in cache anymore (so the kernel is free to decide to evict them whenever it wants to). Calling it on a freshly opened fd before any `read()` evicts nothing and doesn't "persist" on subsequent reads. `POSIX_FADV_NOREUSE` is likely what you're looking for, but its implementation changed throughout the kernel history and it was mostly a no-op until 6.3.

I think the right placement for `POSIX_FADV_DONTNEED` would be in `storageReadPosixFreeResource`, although you could also make it more granular in `storageReadPosix`.

fbernier:
Here's a claude generated PR. Let me know if you want me to open it:

https://github.com/pgbackrest/pgbackrest/compare/main...fbernier:pgbackrest:fadvise?expand=1

dwsteele:
@fbernier Thanks for having a look at this! 

The patch look reasonable but I think it should be inverted, i.e. we should be specifying the reads we *do* want cached since no cache seems like the better default. I'm sure Claude can invert that with ease. I don't agree with with what it has decided to cache vs not cache but that's something we can clean up in the PR.

So, if you would, please do the inversion and then submit a PR.

PR #2760, opened 2026-03-30
Add systemd notify integration.

### LINKED ISSUE #2757: Systemd Notify Integration
It would be great if the pgbackrest server has systemd notify integration. This integration would allow systemd to have better understanding of when a pgbackrest process is starting, reloading, ready, etc. See for example postgres[0] and pgbouncer[1] implementations below.

[0] https://www.postgresql.org/message-id/flat/CAFj8pRA4%3DhVj-d%3D8O7PSMjopsFUHPcAftd5tLqFC_xb035hNQA%40mail.gmail.com#e346a6189d8b0ed44c745c4aaaef587f
[1] https://github.com/pgbouncer/pgbouncer/commit/c1d8f4dcace19e8ea87c759654b00d4c764c5c11

AndrewJackson2020:
This PR adds systemd notify integration. This will allow systemd to have a better understanding as to when an pgbackrest has finished "starting" or is stopping. 

I previously opened issue [0] for this but was hoping to take a crack at this issue myself. 

I based this implementation off of the existing implementations in postgres [1] and pgbouncer [2]. Please note that pgbouncer also has an implementation for `notify-reload`. I decided not to implement that here as it is a very recent feature [3] that is unavailable on many linux distributions. For example it is not usable in either Rocky8 or Rocky9 linux.

[0] https://github.com/pgbackrest/pgbackrest/issues/2757
[1]  https://www.postgresql.org/message-id/flat/CAFj8pRA4%3DhVj-d%3D8O7PSMjopsFUHPcAftd5tLqFC_xb035hNQA%40mail.gmail.com#e346a6189d8b0ed44c745c4aaaef587f
[2] https://github.com/pgbouncer/pgbouncer/commit/3816a0073f09944a6f7eaa278d2226ca4942b911
[3] https://github.com/systemd/systemd/blob/fa6d3bffe30064c4d4092b3daa749465f08d35fb/NEWS#L6176


dwsteele (commented):
This looks good overall (I made various changes and added tests). 

One thing -- shouldn't we be considering reload, e.g.:
```
#ifdef HAVE_LIBSYSTEMD
sd_notify(0, "RELOADING=1");
#endif
// ... reload logic ...
#ifdef HAVE_LIBSYSTEMD
sd_notify(0, "READY=1");
#endif
```

AndrewJackson2020:
> This looks good overall (I made various changes and added tests).
> 
> One thing -- shouldn't we be considering reload, e.g.:
> 
> ```
> #ifdef HAVE_LIBSYSTEMD
> sd_notify(0, "RELOADING=1");
> #endif
> // ... reload logic ...
> #ifdef HAVE_LIBSYSTEMD
> sd_notify(0, "READY=1");
> #endif
> ```

Thanks for working on this. I would definitely not known to mock mock sd_notify for testing the way you did.

I left reload out because it is in such a new release of systemd that most LTS linux distros do not support it yet. Rocky8 and 9 for example do not support it. Probably rocky10 does but have not actually tried it.

IMO it ideally should be added. Keep in mind the function is to use is sd_notifyf and not sd_notify. 

Per the systemd.sercice man page [0]

> When initiating the reload process the service is expected to reply with a notification message via [sd_notify(3)](https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html#) that contains the "RELOADING=1" field in combination with "MONOTONIC_USEC=" set to the current monotonic time (i.e. CLOCK_MONOTONIC in [clock_gettime(2)](https://man7.org/linux/man-pages/man2/clock_gettime.2.html)) in μs, formatted as decimal string. Once reloading is complete another notification message must be sent, containing "READY=1".

Pgbouncers codebase has it implemented and can be referenced [1].

[0] https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html?__goaway_challenge=meta-refresh&__goaway_id=d63c05c3b7b5b0b4af0e71819f78dfad&__goaway_referer=https%3A%2F%2Fwww.google.com%2F
[0] https://github.com/pgbouncer/pgbouncer/commit/c1d8f4dcace19e8ea87c759654b00d4c764c5c11

dwsteele:
> Thanks for working on this. I would definitely not known to mock mock sd_notify for testing the way you did.

My pleasure. We use shims for testing in many modules. This one is a bit primitive but I think it does the job.

> I left reload out because it is in such a new release of systemd that most LTS linux distros do not support it yet. Rocky8 and 9 for example do not support it. 

Ugh, yeah, that would definitely cause problems. We could make the code work but it would make documentation and packaging more difficult.

Let me think about this a bit but for now let's leave the patch as it is.

dwsteele (approved):
This is an incremental but worthwhile improvement. Thank you!

AndrewJackson2020:
> This is an incremental but worthwhile improvement. Thank you!

Of course. Thank you for picking up this PR and maintaining pgbackrest as a whole. Very impressive codebase.

PR #2762, opened 2026-04-09
Add support for S3 Outposts.

### LINKED ISSUE #2659: Repo on S3 under AWS Outposts
I'm trying to create a stanza on S3 under AWS Outposts, but I'm getting an `UnsupportedOperation` error. I have multiple repos, and trying to use S3 on repo2. I'm using the full ARN for the --repo2-s3-bucket option, since S3 on Outpost doesn't use traditional S3 bucket names.

Something like `--repo2-s3-bucket=arn:aws:s3-outposts:us-east-1:123456789012:outpost/op-01234567890abcdef/accesspoint/my-pgbackrest-ap`

I was able to connect to the bucket with the aws command line client.  Digging around, the `UnsupportedOperation` error from S3 Outposts typically means pgBackRest is trying to use an S3 API operation that Outposts doesn't support. So I am wondering:

1) if I missing something
2) if there's some incompatibility between backrest and outpost I should be aware of.

This is pgbackrest 2.54.2.

Thanks in advance.

```
$ pgbackrest stanza-create --stanza-db
  2025-07-09 10:02:07.555 P00   INFO: stanza-create command begin 2.53.1: --exec-id=1653-6fafe3 --log-level-console=info   --log-level-file=debug --log-path=/pgbackrest/repo/log --pg1-host=host1 --pg2-host=host2 --pg1-host-ca-file=/etc/pgbackrest/conf.d/~postgres.operator/tls-ca.crt --pg2-host-ca-file=/etc/pgbackrest/conf.d/~postgres.operator/tls-ca.crt --repo-host=host1   --pg1-host-cert-file=/etc/pgbackrest/conf.d/~postgres.operator/client-tls.crt --pg2-host-cert-file=/etc/pgbackrest/conf.d/~postgres.operator/client-tls.crt  --pg1-host-key-file=/etc/pgbackrest/conf.d/~postgres.operator/client-tls.key --pg2-host-key-file=/etc/pgbackrest/conf.d/~postgres.operator/client-tls.key --pg1-host-type=tls  --pg2-host-type=tls --pg1-path=/pgdata/pg16 --pg2-path=/pgdata/pg16 --pg1-port=5432 --pg2-port=5432  --pg1-socket-path=/tmp/postgres --pg2-socket-path=/tmp/postgres --repo1-path=/pgbackrest/repo1 --repo2-path=/pgbackrest/repo2 --repo2-s3-bucket=arn:aws:s3-outposts:us-east-1:<redacted>:outpost/op-<redacted>/accesspoint/<redacted>-op-s3-pgbackrest-ep --repo2-s3-endpoint=s3.outposts.us-east-1.amazonaws.com --repo2-s3-key-type=auto --repo2-s3-region=us-east-1 --repo2-s3-uri-style=path --no-repo2-storage-verify-tls --repo2-type=s3 --stanza=db

INFO: stanza-create for stanza 'db' on repo2
ERROR: [039]: HTTP request failed with 404 (Not Found):
*** Path/Query ***:
GET /arn%3Aaws%3As3-outposts%3Aus-east-1%3A<REDACTED>%3Aoutpost/op-<REDACTED>/accesspoint/<REDACTED>-op-s3-pgbackrest-ep/?delimiter=%2F&list-type=2&prefix=pgbackrest%2Frepo2%2Farchive%2Fdb%2F
*** Request Headers ***: authorization: <redacted>
content-length: 0
host: s3-outposts.us-east-1.amazonaws.com
x-amz-content-sha256: <redacted>
x-amz-date: <redacted> x-amz-security-token: ‹redacted>
*** Response Headers ***: connection: keep-alive content-length: 298
content-type: application/xml date: Wed, 09 Jul 2025 14:26:22 GMT
x-amz-id-2: <redacted>
x-amzn-requestid: <redacted>
*** Response Content ***:
‹?xml version="1.0" encoding="UTF-8"?›
‹ErrorResponse><Error><Code>UnsupportedOperation</Code><Message>Unsupported operation‹/Message></Error><RequestId>redacted</RequestId><HostId>redacted</HostId></ErrorResponse>

```

ShivakumarAmbigiTR:
Base branch: integration

Description:

Auto-detect S3 Outposts endpoints by checking if the endpoint contains s3-outposts and switch the SigV4 signing service name from s3 to s3-outposts in the credential scope, HMAC signing key derivation, and authorization header.

Problem:
When using pgBackRest with an S3 Outposts bucket, SigV4 authentication fails because the signing service in the credential scope, HMAC key derivation, and Authorization header is hardcoded to s3. AWS S3 Outposts requires the signing service to be s3-outposts.

Solution:
Added a signingService field to the StorageS3 struct that is auto-detected in storageS3New() based on whether the endpoint contains s3-outposts. This value is then used in storageS3Auth() for:

The credential scope in the string-to-sign
The HMAC signing key derivation
The Authorization header
Changes:

storage.c— Added S3_OUTPOSTS constant, STRING_STATIC definitions for signing service names, signingService struct field with auto-detection from endpoint, and updated all three signing locations in storageS3Auth().
s3Test.c— Added unit tests for S3 Outposts signing service detection, signing key regeneration on date change, and standard endpoint fallback, all with exact SigV4 signature verification.
Testing:

Unit tests pass with 100% coverage (test.pl --vm-out --module=storage --test=s3 --vm=none)
Code formatting verified (--code-format)
Auto-generation verified (--gen-only)
Documentation generated (doc.pl --out=html --no-exe)


src/storage/s3/storage.c
```diff
@@ -1287,6 +1290,11 @@ storageS3New(
             .interface = storageInterfaceS3,
             .bucket = strDup(bucket),
             .region = strDup(region),
+
+            // Use "s3-outposts" signing service when endpoint contains "s3-outposts", otherwise use "s3"
+            .signingService =
+                strstr(strZ(endPoint), S3_OUTPOSTS) != NULL ? S3_OUTPOSTS_SIGNING_SERVICE_STR : S3_SIGNING_SERVICE_STR,
```
dwsteele: This seems a bit fragile and I worry that you could get an unintended match. I know that seems unlikely but my experience is that changes that *can* affect existing installations often will.

I think we need an option, e.g. `repo-s3-service`, that defaults to `s3` and also accepts `s3-outposts`.

ShivakumarAmbigiTR: I have made the suggested changes.. Please review..


 Also I would want this to work on top of 2.53.1 too.. As i see integration branch has 2.59 .. would it mean we will have to upgrade?

src/build/config/config.yaml
```diff
@@ -2383,6 +2383,10 @@ option:
       list:
         - s3
 
+  repo-s3-service:
+    inherit: repo-s3-bucket
+    default: s3
```
dwsteele, line 2388: I was thinking of an `allow-list` here but maybe this is better. If there are any additional services in the future we don't need to make changes for users to utilize them.

src/storage/s3/storage.c
```diff
@@ -1287,6 +1291,7 @@ storageS3New(
             .interface = storageInterfaceS3,
             .bucket = strDup(bucket),
             .region = strDup(region),
+            .signingService = strEqZ(service, S3_OUTPOSTS) ? S3_OUTPOSTS_SIGNING_SERVICE_STR : S3_SIGNING_SERVICE_STR,
```
dwsteele: This is a bug. If I specify `repo1-s3-service=s3-valid-service` then the service will be set as `s3`. This will be surprising to the user. If the `repo-s3-service` option is free-form then the input should be respected.

It is also wasteful. `strDup(service)` is more efficient and allows us to get rid of some constants.

dwsteele (changes requested):
See my comments. In addition you'll need to add option help in /src/build/help/help.xml.

dwsteele (approved):
OK, this looks good to me now.

@roberto-mello Any chance you would be able to test this in your environment?

roberto-mello:
Hi @dwsteele ,

I'm in contact with the OP (Shiva) about the issue that led to the patch. I'll work with them to test it, but they've deployed it internally. Since one can't just spin up a temporary outposts instance to test, I'll try to get a temporary bucket on an outposts S3 to test.

Roberto

dwsteele:
@roberto-mello @ShivakumarAmbigiTR This patch is ready to commit but I want to get confirmation from somebody that it has been tested in a real environment. If this has been deployed internally (as indicated above) that is definitely sufficient testing.

roberto-mello:
@dwsteele I did go through runs of pgbr patched with this patch, and we were able to use Outposts as S3 repository. @ShivakumarAmbigiTR was the one running it but I did walk things over with him and we did basic usage testing.

dwsteele:
@roberto-mello @ShivakumarAmbigiTR Thank you for testing this patch. It has now been merged.

PR #2764, opened 2026-04-09
[do not merge] Restart POSIX directory listing on cookie reset

bandetto:
Restart POSIX directory listing on cookie reset

> Our team recently encountered a bug rooted in some implementations of NFS, such
> as [UNFS3](https://github.com/unfs3/unfs3) which may produce duplicated file
> entries when using `storagePosixList()` (which is unchanged from 2.54.2, this
> bug was reproduced on).
> 
> By design, UNFS3 holds a global variable called `rcookie` that is used
> globally for all NFS directories. After it gets invalidated when a directory
> is changed, UNFS3 returns 0 instead of `NFS3ERR_BAD_COOKIE`--which exists to
> signify the client that it needs to restart the scan via `readdir()` and
> invalidate existing entries--to any ongoing `readdir()` calls.
> 
> As the cookie gets sent to the NFS server on the next `readdir()` call,
> server will list entries from the beginning, even the previous call was in
> the middle of the directory.

Introduce mechanism to restart directory listing when offset cookie spuriously
appears to be 0 in the middle reading.


bandetto:
Oops, wrong repository.

PR #2770, opened 2026-05-21
POC: Mount a backup as a filesystem.

### LINKED ISSUE #2747: Idea : Mount a backup and start Postgres above it
Hello,
I read this article : https://www.plakar.io/posts/2026-01-11/researching-a-postgresql-viewer-for-plakar/

TL;DR : Plakar is a backup tool and allows mounting the backup. In this article, they mount a pg_basebackup backup and with a little sorcery (qcow2), they are able to start Postgres on this mount point. **Without restoring all the data**.

It is a common complaint : "If I want to restore my 1MB table from my 5TB database, I have to restore the whole database? - Yes"

If pgbackrest was able to mount a backup, then, we could perform a PITR without restoring all the data. For sure, it will be slower than having all data locally, but it may be acceptable to have a slow dump of my 1MB table than restoring my 5TB database.

dwsteele:
This was suggested by @anayrat and it sounded interesting so I put together a POC.  I'm wondering whether people think this is broadly useful and should be developed into a working feature.

It could be really handy for cases where you need to restore a very large database just to dump a few rows or a small table. However, recovery is still required so it might end up needing to fetch a lot of files from the repo depending on write patterns. Probably a lot less than the full size of the cluster, though.

Is this something you do a lot? Do you have alternative ways to do it (.e.g. snapshot) that you'd still prefer if this feature existed?

Currently the entire file is fetched from the repo the first time it is requested. This could be made smarter in the future but is probably fine for a first pass. Only uncompressed/unbundled backups are supported in the POC.

Basic usage is do a backup and then add some data with pgbench. Calculate a checksum of the data.
```
docker exec -it -u postgres doc-pg-primary pgbackrest --stanza=demo backup --start-fast --no-resume --compress-type=none --type=full --log-timestamp --repo1-retention-full=1 --log-level-console=info
docker exec -it -u postgres doc-pg-primary pgbench -i -s 100
docker exec -it -u postgres doc-pg-primary bash -c 'pg_dump -a --restrict-key=x postgres | sha1sum'
```

A spool path is required (this is where the backing data is stored)
```
docker exec -it -u postgres doc-pg-primary rm -rf /var/spool/pgbackrest
docker exec -it doc-pg-primary mkdir -p 700 /var/spool/pgbackrest/
docker exec -it doc-pg-primary chown postgres:postgres /var/spool/pgbackrest
```

Stop the cluster, remove PGDATA, and mount the backup:
```
docker exec -it -u postgres doc-pg-primary pg_ctlcluster 17 demo stop
docker exec -it -u postgres doc-pg-primary pgbackrest --stanza=demo --mount=fuse --type=standby restore
```

Start the cluster, wait for recovery to finish, and recheck the checksum (it matches!)
```
docker exec -it -u postgres doc-pg-primary pg_ctlcluster 17 demo start
docker exec -it -u postgres doc-pg-primary tail -f /var/log/postgresql/postgresql-17-demo.log
docker exec -it -u postgres doc-pg-primary bash -c 'pg_dump -a --restrict-key=x postgres | sha1sum'
```

There's so much missing here. Only uncompressed backups are supported, no tests, would probably be better to have a separate command rather than just forking restore, sparse files in backing storage, etc., etc., etc.


AliNazirKosar:
@dwsteele i was reading the thread in issue #2747 where @anayrat raised the problem and I think you've hit the nail on the head with this POC.

To answer your questions directly yes, this is something I do. The most common scenario for me is falling back to a backup when I need to recover a table, either restoring it under a different name on the same server or on a different server altogether. Snapshots are something I use occasionally, but not often.

The problem is always the same you have to download everything, restore the whole backup and then export only what you need. With this approach you would only fetch the files you actually touch, saving time, disk space and transfer costs. Especially useful when the backup is stored in the cloud and you're dealing with large databases.

I also see value in cases where the table already exists in production with current data and you only need to recover specific rows from a previous point in time. You mount the backup, query exactly what you need and carefully merge just those rows back into production without touching anything else.

That said, I have some concerns around FUSE latency could be a real issue especially with cloud-based repositories where every read means a network round trip. And if the backup is compressed or encrypted, handling decompression and decryption on the fly adds significant complexity. How is that being considered for a more complete version?

Either way, I think this is worth developing into a full feature. Is there an open issue where one could help out?

anayrat:
Thanks for working on this. I got severely injured. I won't be able to work on this. I will try to follow this pr. Thanks for working on this. Regards

dwsteele:
@anayrat I'm really sorry to hear that! Just focus on getting better and don't worry too much about the PR.

dwsteele:
@AliNazirKosar Thanks for having a look at this.

> I have some concerns around FUSE latency could be a real issue especially with cloud-based repositories where every read means a network round trip.

This version fetches the entire file on first access so it does not have to do a round trip for every read. But there are plenty of ways this could be made more efficient in the future. For instance, we could prefetch by scanning the WAL or WAL summaries. Since we are fetching the WAL in the background that might be a good place to do the scan.

> And if the backup is compressed or encrypted, handling decompression and decryption on the fly adds significant complexity. How is that being considered for a more complete version?

It does add complexity and would require the restore code to be refactored so it could be used from both places. It might be enough just to tweak `restoreFile()` a bit but I have not looked at it closely.

> Is there an open issue where one could help out?

For the moment no. There is some parallel work going on that will change restore a lot so I'd rather finish that first to avoid conflicts.

For now I'm just trying to get an idea of how useful this feature would be.

marco44:
This seems a super interesting feature !

PR #2771, opened 2026-05-22
expire: fix inconsistent dry-run guard in expireAdhocBackup

jodevsa:
expire: fix inconsistent dry-run guard in expireAdhocBackup

All other dry-run checks in this file use:

    if (!cfgOptionValid(cfgOptDryRun) || !cfgOptionBool(cfgOptDryRun))

to guard against calling cfgOptionBool() when the option is not valid in the current command context. The guard in expireAdhocBackup skips the cfgOptionValid() check, meaning backupLinkLatest() could be called during a dry-run if cfgOptDryRun is not a valid option and cfgOptionBool() returns an unexpected default.


dwsteele:
Adding this check causes coverage testing to fail. `expireAdhocBackup()` is already gated at line 1010 by an option that is specific to the expire command, so there is no need to test `--dry-run` in `expireAdhocBackup()`. The comment above this line calls this out.

We appreciate the PR but this is not something we can commit.

PR #2773, opened 2026-05-27
Add user/group caching for faster manifest build.

GLFNSE:
Background:

We have a special situation with our databases where we have millions of files (6-8 millions) in our databases, so we are using the bundling feature. But the building of the manifest still takes a very long time. I'm not an expert C programmer so I let Claude AI have a go at finding what is taking so long when building the manifest. It turned out to be different kinds of "overhead" so this is the first patch to help reduce the impact of the overhead.

This overhead is not noticeable with few files, but certainly is for large amount of files.

This is the most important patch since it reduced the running time the most (more patches with noticeable gains will also be submitted). It made manifest build go from several hours to minutes when tested.

What it adds:

A small cache to cache nss lookup results to avoid repeated lookups since it gives a large overhead when doing many lookups.

--------------

Cache uid/gid → name lookups during manifest build

Summary

Adds a small in-process cache to userNameFromId() and groupNameFromId() in src/common/user.c so that each unique uid/gid triggers at most one libc NSS lookup for the lifetime of a backup, instead of one per file.

Motivation

storagePosixInfo() calls userNameFromId() and groupNameFromId() for every file at storageInfoLevelDetail, which is the level used during manifestNewBuild(). The current implementations of those two functions in common/user.c call getpwuid() / getgrgid() directly with no caching.

On systems where NSS is configured to consult a remote name service (sssd, systemd-userdbd, LDAP via nss-pam-ldapd, etc.) every lookup is a Unix-socket round-trip to the resolver, plus its own syscall and audit overhead. PostgreSQL data files almost always share a single owner (postgres:postgres), so this work is essentially repeated millions of times for an answer that doesn't change.

CPU profiling on a 6.5-million-file cluster running with sssd (Active Directory backend) showed roughly 25–30% of pgbackrest CPU time during the manifest walk being spent in:

- libnss_systemd.so.2 and libnss_sss.so.2 (NSS module work)
- unix_stream_* / __libc_connect / __sys_sendto (socket round-trips to sssd)
- _nss_files_parse_pwent / _nss_files_parse_grent (libc fallback parsing of /etc/passwd//etc/group)
- The __audit_filter_op and syscall_enter_from_user_mode kernel paths driven by those syscalls

This isn't visible in top as obvious blocked time because each lookup is microsecond-fast individually — it only becomes obvious when you sample CPU stacks at scale.

Change

src/common/user.c gains two file-scope static arrays, 16 entries each:

static UserNameCacheEntry userNameCache[USER_NAME_CACHE_SIZE]; static GroupNameCacheEntry groupNameCache[USER_NAME_CACHE_SIZE];

userNameFromId() and groupNameFromId() now do a linear scan over their cache first. On hit, they return strDup() of the cached name (or NULL if the id was previously unresolvable). On miss, they call getpwuid() / getgrgid() as before and store the answer (positive or negative) into the next free slot. Cached names are allocated in memContextTop() so they survive the caller's context.

Linear scan is intentional — the working set on a PG host is essentially always 1–3 unique owners. Beyond 16 unique ids the cache simply stops growing and we fall back to per-call NSS, which is correct (just slower for that pathological case).

Performance impact

On a 6.5-million-file PostgreSQL cluster with sssd (XFS over LVM on a network volume):

┌───────────────────────┬────────────────────────┬─────────────────────┐
│         Phase         │         Before         │        After        │
├───────────────────────┼────────────────────────┼─────────────────────┤
│ manifestNewBuild walk │ 13,121 s (~3 h 39 min) │ 224 s (~3 min 44 s) │
├───────────────────────┼────────────────────────┼─────────────────────┤
│ Full backup wall time │          several hours │           1 h 4 min │
└───────────────────────┴────────────────────────┴─────────────────────┘

The 58× speedup on the walk is larger than the "remove 25% of CPU" suggests because the NSS overhead was preventing the kernel XFS inode cache from staying warm — once per-file work dropped from ~2 ms to ~30 µs, cache hits started compounding.

Profile after the change: NSS-related symbols (libnss_systemd*, libnss_sss*, _nss_files_*, unix_stream_*) drop to zero. userNameFromId and groupNameFromId themselves register at ~0.05% and ~0.17% of CPU — confirming the cache is doing its job.

Tradeoffs and risks

- Negative results are cached. If a uid is unresolvable at the start of a backup, it stays unresolvable for the rest of that process. For backup workloads this is the desired behavior (don't retry NSS in a tight loop).
- No invalidation. The cache is process-local and lives for the process's lifetime. pgbackrest is invoked as a one-shot CLI, so there's no daemon scenario to worry about.
- Process forks inherit a copy. Each pgbackrest local/remote worker spawned via ProtocolParallel warms its own cache lazily after fork. Working as intended.
- Cardinality assumption. The 16-entry fixed cache is sized for the realistic case. If anyone has a PG host where the data directory has more than 16 unique owners, they get the existing uncached behavior for the excess, which is no worse than today.

Tests

Existing unit tests in test/src/module/common/userTest.c continue to pass. Behavioral equivalence: cache returns a strDup of identical content to what the uncached path would have returned, in the caller's current memory context.

Notes

This is a real bug-class issue on systems with remote NSS, not just an optimization for our environment. We suspect any user running pgbackrest against a PG host with sssd/LDAP/AD-backed NSS sees a meaningful share of this overhead; it's just invisible without profiling.

### DIFF
--- a/doc/xml/release/2020s/2026/2.59.0.xml
+++ b/doc/xml/release/2020s/2026/2.59.0.xml
@@ -2,6 +2,17 @@
     <release-core-list>
         <release-improvement-list>
             <release-item>
+                <github-pull-request id="2773"/>
+
+                <release-item-contributor-list>
+                    <release-item-contributor id="gunnar.lindholm"/>
+                    <release-item-reviewer id="david.steele"/>
+                </release-item-contributor-list>
+
+                <p>Add user/group caching for faster manifest build.</p>
+            </release-item>
+
+            <release-item>
                 <github-issue id="2736"/>
                 <github-pull-request id="2740"/>
 
--- a/doc/xml/release/contributor.xml
+++ b/doc/xml/release/contributor.xml
@@ -458,6 +458,11 @@
     <contributor-id type="github">bluthg</contributor-id>
 </contributor>
 
+<contributor id="gunnar.lindholm">
+    <contributor-name-display>Gunnar Lindholm</contributor-name-display>
+    <contributor-id type="github">GLFNSE</contributor-id>
+</contributor>
+
 <contributor id="gurucubano">
     <contributor-name-display>Matthias</contributor-name-display>
     <contributor-id type="github">gurucubano</contributor-id>
--- a/src/common/user.c
+++ b/src/common/user.c
@@ -12,6 +12,34 @@ System User/Group Management
 #include "common/user.h"
 
 /***********************************************************************************************************************************
+Per-process uid->name and gid->name caches.
+
+On systems where uid/gid lookups are routed to a remote name service (sssd, systemd-userdbd, LDAP via nss-pam-ldapd, etc.), every
+call costs a Unix-socket round-trip plus its own audit-and-syscall overhead. Profiling shows this can dominate the manifest-build
+phase for clusters with millions of files even though the data files almost always share a single owner (typically "postgres").
+
+The cache is a small fixed-size array because the working set of unique ids on a PG host is tiny (usually 1-3). Linear scan is
+faster than a hash table at this size and avoids any allocation in the hot path. The cached String lives in memContextTop() so it
+survives the caller's context.
+
+Cache misses past USER_NAME_CACHE_SIZE fall through to an uncached lookup (correct but slow); in practice we never exceed the table
+for a backup of a normally-configured cluster.
+***********************************************************************************************************************************/
+#define USER_NAME_CACHE_SIZE                                        16
+
+typedef struct UserNameCache
+{
+    uid_t id;                                                       // The uid that was looked up
+    const String *name;                                             // User name returned from lookup, NULL if unresolvable
+} UserNameCache;
+
+typedef struct GroupNameCache
+{
+    gid_t id;                                                       // The gid that was looked up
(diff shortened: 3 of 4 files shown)


GLFNSE:
Sorry, I forgot to change that this should go into integration, not main. Not sure how to change that.

dwsteele (approved):
Thank you for the patch!

This looks like a win to me. The added code is minimal and isolated in a rarely-updated module.

We've occasionally seen manifest builds that seemed unreasonably long but generally put it down to slow storage. This provides an alternate explanation that may end up making people's lives easier.

I made a bunch of updates to make the code conform to our standards, cleaned up comments, and fixed some usage that Claude missed, e.g. strDup() is NULL-tolerant. The changes should be superficial but you'll want to review and retest.

GLFNSE:
I've built and tested your fixes to my patch. It worked perfectly.  
Huge performance gains. Hard to tell how much any of the other patches would do in comparison.

dwsteele:
Great, thanks for testing it! I would be surprised if any of the other patches made as big a difference since we regularly test with millions of files and have found the performance acceptable (though this patch proves there are exceptions). If manifest processing time is a small enough percentage of overall time then we consider some overhead acceptable vs. code complexity. This is always a balancing act.

PR #2774, opened 2026-05-27
Fewer binary searches when building manifest.

GLFNSE:
Background:

We have a special situation with our databases where we have millions of files (6-8 millions) in our databases, so we are using the bundling feature. But the building of the manifest still takes a very long time. I'm not an expert C programmer so I let Claude AI have a go at finding what is taking so long when building the manifest. It turned out to be different kinds of "overhead" and this is the second patch to help reduce the impact of the overhead.

This overhead is not noticeable with few files, but is for large amount of files, although the gain is modest.

This is not the most important patch since it reduces only the number of lookups, but it was still noticeable during our tests.

What it adds:

It replaces 3 searches with only 1 search.

--------------

Reduce per-file binary searches in manifestBuildIncr

  Summary

  The loop in manifestBuildIncr performs three name-based binary searches per file when one would suffice. This PR adds two small helper functions and rewrites the loop to use them, eliminating the redundancy.

  Motivation

  For every file in the new manifest, manifestBuildIncr's main loop currently does:

  if (manifestFileExists(manifestPrior, file.name))           // bsearch #1
  {
      const ManifestFile filePrior =
          manifestFileFind(manifestPrior, file.name);         // bsearch #2 (same key as above)
      ...
      manifestFileUpdate(this, &file);                        // bsearch #3 (in current manifest, by name)
  }

  Search #2 is fully redundant with #1 — they have the same key in the same list. Search #3 happens even though the enclosing loop already knows the file's index in the current manifest's file list.

  For a 6.5-million-file PostgreSQL cluster this means ~15 million binary searches when ~5 million would suffice. Each search involves string comparisons against the names embedded in ManifestFilePack entries.

  Change

  Two new helper functions in src/info/manifest/manifest.h and src/info/manifest/store.c.inc:

  - manifestFilePackFindDefault(this, name) — a find-or-NULL variant of the existing manifestFilePackFind(). Single binary search; returns NULL on miss instead of throwing.
  - manifestFileUpdateByIdx(this, fileIdx, file) — equivalent to manifestFileUpdate() but skips the by-name search since the caller already knows the index. Internally uses lstGet() on the file list and the existing static manifestFilePackUpdate().

  The loop in manifestBuildIncr is rewritten to use both:

  const ManifestFilePack *const filePackPrior =
      file.copy ? manifestFilePackFindDefault(manifestPrior, file.name) : NULL;

  if (filePackPrior != NULL)
  {
      const ManifestFile filePrior = manifestFileUnpack(manifestPrior, filePackPrior);
      ...
      manifestFileUpdateByIdx(this, fileIdx, &file);
  }

  Performance impact

  On a 6.5-million-file incremental backup, manifestBuildIncr dropped from approximately 80 s (estimated) to 21.8 s. Modest in absolute terms but a clear constant-factor improvement on the only phase of an incremental that scales linearly with file
   count after the manifest walk itself.

  What this does NOT do

  It does not eliminate the per-file repack inside manifestFilePackUpdate(). That remains the dominant cost for unchanged-but-referenced files in an incremental backup. A larger refactor that consults the prior manifest during manifestNewBuild and
  packs files in their final state directly would eliminate the repack entirely, but is a much bigger change with broader behavioral implications. Not pursued here — left as a follow-up if/when the repack cost becomes the next visible bottleneck.

  Tradeoffs and risks

  - Behavior is functionally identical to the existing implementation.
  - Two new public API functions, both with tightly-scoped semantics.
  - manifestFileUpdateByIdx asserts the index is in range.
  - The existing manifestFileUpdate remains for callers that don't have an index.

  Tests

  Existing unit tests covering manifestFileUpdate and the manifestBuildIncr path continue to pass.


dwsteele:
This looks good overall but I simplified it a bit. Not quite ready for a final review but I think this is pretty much what the patch will look like.

dwsteele:
Oh, as for the extra tests, I just didn't feel like they were adding much. We already have coverage for the "not found" condition and I think that is sufficient.

GLFNSE:
I'm afraid that I'd rather not test this patch just yet, since it takes so long to run without the nss-cache patch. But feel free to notify me when you are getting closer to a release and I'll re-run the test with the latest version.

dwsteele:
Oh, yeah, that sounds really painful. Better to wait. I'll commit the nss-cache patch tomorrow and merge to make this easier.

BTW -- you may have noticed I scrubbed mentions of NSS from the patch and I'll do the same for the commit message. As far as I know that's Linux-specific and we try to be generally *nix compatible.

dwsteele:
> I'll commit the nss-cache patch tomorrow and merge to make this easier.

This has been committed and merged. That should make your testing less painful.

dwsteele (approved):

GLFNSE:
The backup worked fine with this version (will later try to do a restore also to verify). It took 19 seconds longer (out of over 1 h of total backup time), so I guess there are some variation in load to the system. And no much gained due to the previous patch that was applied.

dwsteele:
< .5% is definitely not enough to prove anything either way. It might be improving the performance of manifest generation but that is still a small percentage of the total backup.

dwsteele:
@GLFNSE I'm planning to commit this unless you have an objection.

PR #2775, opened 2026-05-27
Replace pack/unpack with less costly alternative

GLFNSE:
Background:

We have a special situation with our databases where we have millions of files (6-8 millions) in our databases, so we are using the bundling feature. But the building of the manifest still takes a very long time. I'm not an expert C programmer so I let Claude AI have a go at finding what is taking so long when building the manifest. It turned out to be different kinds of "overhead" and this is the third patch to help reduce the impact of the overhead.

This overhead is quite noticeable with a large amount of files, up to 1 hour on some systems.

What it adds:

It removes the 2 manifestFileUnpack calls when comparing files, and instead use a more direct datastructure for comparison.

It also changes where the packaging is calculated.

--------------

Scope: src/command/backup/backup.c only. Pure refactor of how files flow from the sorted manifest into dispatched backup jobs. Bundle composition is byte-identical to upstream given the same input.

  Two related problems it fixes

  1. The sort comparator calls manifestFileUnpack() twice per compare. The current backupProcessQueueComparator unpacks each file's varint-encoded pack header (to read size/timestamp/name) on every comparison. For an N-file sort that's roughly 2 × N log N unpack operations — at multiple million files, this is hundreds of millions of unpacks just to determine ordering.
  2. The dispatcher (backupJobCallback) re-walks the queue every time it builds a job. It pulls files from the queue head and tries to fit them into a bundle up to repo-bundle-size, with a forward-skip pass so a smaller later file can fill remaining capacity. This walk runs once per dispatched job, with another manifestFileUnpack inside the inner loop. The main process is essentially doing bin-packing work on every job hand-off to a worker.

  What it does instead

  Splits the work into two clean phases:

  1. Sort phase (in backupProcessQueue) — uses a new local struct that caches the sort keys: typedef struct BackupQueueItem {
      uint64_t size;                          // size sort key
      time_t timestamp;                       // timestamp sort key
      const String *name;                     // tiebreak
      const ManifestFilePack *filePack;       // unpack later when actually building the job
  } BackupQueueItem;
  1. The per-target list is now List<BackupQueueItem> and the comparator reads scalars directly. No unpack in the hot path.
  2. Bin phase (new function backupProcessQueueBin) — runs immediately after the sort, walks each per-target list once, and produces a List<BackupJobDescriptor>:
  typedef struct BackupJobDescriptor
  {
      uint64_t bundleId;          // 0 = standalone single-file job
      List *fileList;             // 1+ pack pointers, in dispatch order
  } BackupJobDescriptor;
  2. Each descriptor is exactly one future protocol job. The forward-skip greedy fit logic that used to live in the dispatcher now lives here, mirroring the old algorithm exactly so bundle composition is byte-identical.
  3. Dispatcher (backupJobCallback) — becomes a thin consumer: peek the first descriptor in a queue, unpack each pack to build the job param, pop the descriptor, return one ProtocolParallelJob. No fit-checking, no forward-skip, no lstRemoveIdx
  shuffles.

  Net effect

  - Sort comparator: 2 × N log N unpacks → 0 unpacks (scalar reads)
  - Bin-packing: runs once per backup → still runs once, but at queue-build time, in a single forward pass
  - Dispatcher per-job work: drops from "scan queue, fit-check, possibly skip multiple times" to "pop next descriptor, build job from its files"
  - Memory: each BackupQueueItem is ~40 bytes temporary during sort/bin (discarded immediately after binning). Descriptor list holds per-job entries each containing a List of pack pointers.

  What does NOT change

  - Wire format (protocol pack sent to workers) — identical
  - On-disk bundle layout — identical
  - Bundle id assignment — still monotonic via jobData->bundleId++
  - Standalone-vs-bundled decision — same file.size > bundleLimit rule
  - File order within bundles — same as upstream given the same sorted input
  - Per-queue dispatch round-robin in backupJobQueueNext — unchanged
  - Standby-only-files-go-to-queue-0 rule — unchanged
  - pg_control must-exist check — unchanged

  Test coverage caveat

  The existing backup tests pass and exercise the new code paths, but there's still a coverage gap (~7 lines, ~16 branches) on the descriptor-list edge cases. None of the gap is in the hot path; mostly defensive branches like "queue is empty,
  advance to next" and the like.

### DIFF
--- a/src/command/backup/backup.c
+++ b/src/command/backup/backup.c
@@ -1717,8 +1717,27 @@ backupProcessFilePrimary(RegExp *const standbyExp, const String *const name)
         BOOL, strEqZ(name, MANIFEST_TARGET_PGDATA "/" PG_PATH_GLOBAL "/" PG_FILE_PGCONTROL) || !regExpMatch(standbyExp, name));
 }
 
-// Comparator to order ManifestFile objects by size, date, and name
-static const Manifest *backupProcessQueueComparatorManifest = NULL;
+// Per-file queue item carrying the keys needed for sorting and binning so the comparator (called O(N log N) times) does not have
+// to call manifestFileUnpack(), which deserializes the pack header. Each item also remembers its source file pack so the dispatcher
+// can still produce the full job param from it.
+typedef struct BackupQueueItem
+{
+    uint64_t size;                                                  // file.size (sort key, bundle accumulator)
+    time_t timestamp;                                               // file.timestamp (sort key for small files when bundling)
+    const String *name;                                             // file.name (tie-break + dispatch)
+    const ManifestFilePack *filePack;                               // Source pack, unpacked once at dispatch time
+} BackupQueueItem;
+
+// One dispatchable unit produced by the pre-bin pass. For a non-bundle job (bundleId == 0) the descriptor holds exactly one file
+// pack; for a bundle it holds the full set of file packs that share a bundleId, in dispatch order. Built once during
+// backupProcessQueue so the dispatcher does no fit-checking or bundle-id bookkeeping.
+typedef struct BackupJobDescriptor
+{
+    uint64_t bundleId;                                              // 0 = standalone (one-file job), >0 = bundled
+    List *fileList;                                                 // List of const ManifestFilePack * (>=1 entry)
+} BackupJobDescriptor;
+
+// Comparator over BackupQueueItem. Reads scalars directly -- no manifestFileUnpack() in the hot path.
 static bool backupProcessQueueComparatorBundle;
 static uint64_t backupProcessQueueComparatorBundleLimit;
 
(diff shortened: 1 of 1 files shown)


dwsteele:
I'm not convinced that this is worth the additional complexity -- I'd need to see substantial performance improvements to offset the additional memory cost and I just don't think it's there.

However, WRT to scanning for a file that fits, I have an optimization in another branch that just takes the next file that is <= repo-bundle-limit. This technically violates the repo-bundle-size contract but would simplify the code, improve performance, could be documented, and is likely negligible after compression is applied.

dwsteele:
> However, WRT to scanning for a file that fits, I have an optimization in another branch that just takes the next file that is <= repo-bundle-limit. This technically violates the repo-bundle-size contract but would simplify the code, improve performance, could be documented, and is likely negligible after compression is applied.

This was committed at 253dbc3f

However, regarding the rest of this commit I don't think this is something we want. It trades memory for CPU and I think that is the wrong direction -- for very large manifests memory is the scarcer resource. In addition, a future commit adds a number of fields to the sort (to help with data locality) which would blow up the memory requirement even more.

Given that the user/group caching seems to have solved the vast majority of your performance issue I think it makes sense to close this.

PR #1775, opened 2022-06-08
Fix a compilation error in the performance module test

andr-sokolov:
If DEBUG macro is not defined then ASSERT macro expands to nothing. In this case the timeBegin variable is never used and a compilation error occurs.

```
$ ./test/test.pl --test-path=/tmp/test --tz=GMT --no-coverage --module=performance --vm=rh7
2022-06-08 18:12:29.910 P00   INFO: test begin on x86_64 - log level info
2022-06-08 18:12:29.969 P00   INFO: autogenerate configure
2022-06-08 18:12:29.971 P00   INFO:     autogenerated version in configure.ac script: no changes
2022-06-08 18:12:29.975 P00   INFO:     autogenerated configure script: no changes
2022-06-08 18:12:30.132 P00   INFO: clean autogenerate code
2022-06-08 18:12:31.796 P00   INFO: cleanup old data and containers
2022-06-08 18:12:31.953 P00   INFO: builds required: bin, bin host
2022-06-08 18:12:31.989 P00   INFO:     bin build for rh7 (/tmp/test/bin/rh7)
2022-06-08 18:12:32.192 P00   INFO:         bin dependencies have changed, rebuilding
2022-06-08 18:12:36.461 P00   INFO:     clean bin build for none (/tmp/test/bin/none)
2022-06-08 18:12:40.384 P00   INFO: 3 tests selected
                                        
2022-06-08 18:12:40.991 P00   INFO: P1-T1/3 - vm=rh7, module=performance, test=archive-perl (0.61s)
2022-06-08 18:12:45.623 P00  ERROR: [125]: P1-T2/3 - vm=rh7, module=performance, test=type (err2-4.28s):
    
    STDOUT:
    /tmp/test/repo/src/command/archive/common.c: In function 'archiveAsyncExec':
    /tmp/test/repo/src/command/archive/common.c:274:14: error: unused variable 'timeBegin' [-Werror=unused-variable]
         TimeMSec timeBegin = timeMSec();
                  ^
    compilation terminated due to -Wfatal-errors.
    cc1: all warnings being treated as errors
    make: *** [.build/command/archive/common.o] Error 1
    make: *** Waiting for unfinished jobs....
    
2022-06-08 18:12:50.209 P00  ERROR: [125]: P1-T3/3 - vm=rh7, module=performance, test=storage (err2-4.23s):
    
    STDOUT:
    /tmp/test/repo/src/command/archive/common.c: In function 'archiveAsyncExec':
[...]

PR #2776, opened 2026-05-27
Skip older files faster

GLFNSE:
Skip sorting for very old files when creating bundles.
    
    Notice:
    
    This builds on the manifest-unpack branch!
    
    Background:
    
    We have a special situation with our databases where we have
    millions of files (6-8 millions) in our databases, so we are using the
    bundling feature. But the building of the manifest still takes a very
    long time. I'm not an expert C programmer so I let Claude AI have a go
    at finding what is taking so long when building the manifest.
    It turned out to be different kinds of "overhead" and this is the fourth
    patch to help reduce the impact of the overhead.
    
    What it adds:
    
    It skips sorting very old files and just send them straight to bundle packing,
    since there is not much point in sorting them (hopefully).
    Togehter with the manifest-unpack patch this helps shaving of time from backups using the bundle feature.
    
    --------------
    
    — Add repo-bundle-stable-age option to skip mtime sort for stable files
    
      Summary
    
      Adds a new repo-bundle-stable-age configuration option (integer, days, default 90) that lets files older than the threshold skip the timestamp sort during bundle assignment. Recent files still sort by timestamp-desc; old files end up in directory
       order via the natural traversal order from the manifest walk.
    
      Motivation
    
      The current bundle assignment sorts every small file by timestamp descending so that "newer files cluster together, older files cluster together," which the original bundle commit describes as the mechanism for keeping bundles dense across
      incrementals.
    
      For very large PostgreSQL clusters — particularly OLTP-style workloads where the bulk of data is historical and only a small fraction of files have recent mtimes — sorting the entire set by timestamp is wasted work. A 90-day-old file and a
      365-day-old file are both equally "stable" from the perspective of bundle density across incrementals. The mtime sort within the "stable" bucket adds no fragmentation-resistance benefit while still costing CPU and memory locality.
    
      Change
    
      A new option in src/build/config/config.yaml:
    
        repo-bundle-stable-age:
          section: global
          group: repo
          type: integer
          default: 90
          allow-range: [0, 1825]
          command:
            backup: {}
          depend:
            option: repo-bundle
            list:
              - true
    
      Type is integer (days) rather than time because pgbackrest's time-option storage is unsigned int milliseconds, which caps at ~49 days; we want the range to comfortably cover months/years.
    
      In backupProcessQueue, when bundling is enabled and the threshold is > 0:
    
      1. Two parallel sub-lists per target: a sortable one (recent small files + all large files) and a stable one (old small files).
      2. The sortable list goes through the comparator as before (size-desc for large, mtime-desc for recent small).
      3. The stable list is appended to the sortable list in insertion order before being fed to backupProcessQueueBin.
    
      When the threshold is 0 (or bundling is off), the partition is bypassed and behavior is identical to today.
    
      The "insertion order" for the stable bucket happens to be directory order, because manifestBuildInfo walks directories with sortOrder=sortOrderAsc. This is arguably better for restore locality than mtime-desc within the stable bucket — files from
       the same relation/directory cluster together in the same bundle.
    
      Help text
    
      Updated src/build/help/help.xml with a new <config-key> for repo-bundle-stable-age describing the semantics and recommending 0 to disable.
    
      Performance impact
    
      For a 6.5 M-file PG cluster where ~95% of files are months old, the partition reduces the sort input from 6.5 M to ~325 K. With the existing pre-bin patch the per-compare cost is already low; with this partition the absolute count of
      compares drops by roughly 20×. On our workload manifestNewBuild's sort= sub-phase is currently 21 s; with the partition we expect that to drop to single-digit seconds. Modest absolute savings, mostly relevant for very large clusters.
    
      Semantic effects on bundle composition
    
      - Large files: unchanged (still size-desc sorted, each gets its own job)
      - Recent small files: unchanged (still mtime-desc + name)
      - Old small files: file order within these bundles becomes directory order instead of mtime-desc. The restore-fragmentation argument from the original bundle commit is preserved at the bucket boundary: "old" files as a group still stay densely
      referenced across incrementals.
    
      Tradeoffs and risks
    
      - Default behavior changes when repo-bundle=y: today, all small files sort by timestamp; with this default 90-day threshold, files older than 90 days skip the timestamp sort. Setting repo-bundle-stable-age=0 restores the previous behavior.
      - Adds one configuration knob. The default is a guess based on typical OLTP workloads. Could be lower or higher depending on operator preference.
    
      Tests
    
      Existing bundle-related tests continue to pass. We did not add specific tests for the partition itself; happy to add some if maintainers prefer (a unit test that walks a known set of timestamps and verifies the resulting bundle composition would
      be straightforward).
    
      Alternative considered
    
      The threshold could be derived from the prior manifest's backup-timestamp-start automatically (a file unchanged since the previous backup is "stable" by definition). We didn't pursue this for incrementals specifically because it requires plumbing
       the prior manifest into backupProcessQueue — small but more invasive than a flat integer option. Mentioned here in case it's of interest as a future refinement.

### DIFF
--- a/src/build/config/config.yaml
+++ b/src/build/config/config.yaml
@@ -1872,6 +1872,21 @@ option:
     default: 2MiB
     allow-range: [8KiB, 1PiB]
 
+  repo-bundle-stable-age:
+    section: global
+    group: repo
+    type: integer
+    default: 0
+    allow-range: [0, 1825]
+    command:
+      backup: {}
+    command-role:
+      main: {}
+    depend:
+      option: repo-bundle
+      list:
+        - true
+
   repo-block:
     section: global
     group: repo
--- a/src/build/help/help.xml
+++ b/src/build/help/help.xml
@@ -699,6 +699,20 @@
                         <example>10MiB</example>
                     </config-key>
 
+                    <config-key id="repo-bundle-stable-age" name="Repository Bundle Stable Age">
+                        <summary>Age in days above which files are considered stable for bundling.</summary>
+
+                        <text>
+                            <p>When sorting files for bundle assignment, files whose modification time is at least this many days in the past are treated as stable and are not sorted by timestamp. They are still placed in bundles, but in the order they were discovered during the manifest walk (which is directory order). Files newer than this threshold are still sorted by timestamp descending so that "recent" files cluster together in the same bundles -- preserving the restore-time read locality benefit of the original sort.</p>
+
+                            <p>This optimization is significant when the cluster has millions of small files of which most are old/stable and only a small fraction were recently modified. Sorting then runs only over the recent fraction instead of the whole manifest.</p>
+
+                            <p>The default is <setting>0</setting>, which disables the partition: all files are sorted by timestamp as in the original behavior. Set to a positive value, e.g. <setting>90</setting>, to enable the partition for files whose modification time is older than that many days.</p>
+                        </text>
+
+                        <example>90</example>
+                    </config-key>
+
                     <config-key id="repo-gcs-bucket" name="GCS Repository Bucket">
                         <summary>GCS repository bucket.</summary>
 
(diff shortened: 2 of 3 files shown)


dwsteele:
Given that this a follow-on to #1775 and we are not proceeding with that PR, I'm closing this as well.

PR #2778, opened 2026-06-04
Add S3 process authentication.

dwsteele:
This allows S3 credentials to be provided by an arbitrary executed process. The credentials must be returned in a specific JSON format, i.e.:
```
{
    "Version": 1,
    "AccessKeyId": "an AWS access key",
    "SecretAccessKey": "your AWS secret access key",
    "SessionToken": "the AWS session token for temporary credentials", 
    "Expiration": "RFC3339 timestamp for when the credentials expire"
} 
```

When `repo-s3-key-type=process` the `repo-s3-process-cmd` option can be used to set the command. If there are parameters just repeat `repo-s3-process-cmd`, e.g.:
```
repo1-s3-process-cmd=/usr/local/bin/get-credentials
repo1-s3-process-cmd=--role
repo1-s3-process-cmd=my-role
```


pgstef:
This sounds like a way to fetch credentials from external systems like a kms or something :thinking: ?

dwsteele:
> This sounds like a way to fetch credentials from external systems like a kms or something 

You can fetch credentials in any way you like -- this is a general purpose interface for doing that follows the API defined at https://docs.aws.amazon.com/sdkref/latest/guide/feature-process-credentials.html (documented in a comment).

The specific use case is to work around the role limit on AWS. Cloud providers working at scale can run into these limits and this allows another way to provide credentials.

andrewcharlton (approved):
Thank you for adding this @dwsteele.

I've tested this exetensively in our cloud environment, and it works beautifully.  All of the changes look entirely reasonable, its some lovely C code!

I checked a bunch of failure cases as well, and all were handled as expected:
- [x] Process script doesn't exist
- [x] Process exits with an error code
- [x] Process output is invalid json
- [x] Expired credentials returned
- [x] Missing AccessKeyId
- [x] Missing SecretAccessKey
- [x] Missing SessionToken
- [x] Missing Expiration
- [x] Expiration is not a valid timestamp

In all cases, pgBackRest was always able to retry and recover once the provider was working as expected again.

dwsteele:
@andrewcharlton Committed, thank you!

PR #2779, opened 2026-06-05
PostgreSQL 19beta1 support.

dwsteele:
This update is fairly straightforward except for the new values in ControlFileData.data_checksum_version (PG_DATA_CHECKSUM_INPROGRESS_OFF, PG_DATA_CHECKSUM_INPROGRESS_ON). The new values represent states where checksums are incomplete so we need to treat the cluster as if it has no checksums since we have no easy way of knowing which checksums are completed.


pgstef (approved):
> The new values represent states where checksums are incomplete so we need to treat the cluster as if it has no checksums since we have no easy way of knowing which checksums are completed.

The code change looks good to me. 
I tried to compile this branch and use it with PG 19beta1 (and with PG 18 to cross check), I could take backups and restore them successfully.
So, approved the PR 👍

dwsteele:
> The code change looks good to me.

Thank you for looking at it. I'm going to let this one sit for a bit to think about the page checksum changes. Probably fine, but best to be careful.

flopez777-ops (commented):
Greatness

PR #2783, opened 2026-06-10
Add StorageReadMulti for prefetched multi-file/range reads from object stores.

dwsteele:
StorageReadMulti submits HTTP requests ahead of time to reduce latency when reading multiple files or ranges from S3, Azure, and GCS. Adjacent reads within a configurable gap are merged into a single request, avoiding the cost of a new connection for small gaps ("read over").

Add async open to S3, Azure, and GCS read drivers so requests are submitted immediately and responses are deferred until the first read, enabling prefetch without blocking.

Two new options control the behavior:
* repo-storage-prefetch (default 3): number of reads to queue ahead.
* repo-storage-read-over (default 64KiB): max gap to read over rather than issuing a new request.

Simplify the restore file copy loop to use StorageReadMulti instead of manually tracking offsets and limits across bundled files.

---

Restore performance sees significant improvement with block incremental disabled. With block incremental enabled performance improvement is only about 10% but that will be addressed in a future PR.

For performance testing 10K small tables were created in a fresh cluster. A full backup was made, half of the tables were modified, then a diff backup was made. Every other table was modified to make the worst case scenario for restore.

Then three restores were done with various settings:
```
BOTH OFF: --repo1-storage-prefetch=0 --repo1-storage-read-over=0
PREFETCH ONLY: --repo1-storage-read-over=0
READ OVER ONLY: --repo1-storage-prefetch=0
BOTH ON:
```
|     OPERATION           | BOTH OFF | PREFETCH ONLY   | READ OVER ONLY | BOTH ON |
| ----------------------- | -------: | --------: | ------: | ------: |
| DELTA RESTORE FROM FULL |   299.1s |     91.3s |    9.6s |    8.6s |
| DELTA RESTORE FROM DIFF |     7.6s |      7.5s |    6.9s |    6.9s |
| FRESH RESTORE FROM DIFF |   327.7s |    107.3s |   25.1s |   23.3s |


src/storage/read.c
```diff
@@ -118,10 +141,10 @@ storageRead(THIS_VOID, Buffer *const buffer, const bool block)
                             .offset = this->pub.offset + this->bytesRead,
                             .limit = this->pub.limit != NULL ? varNewUInt64(varUInt64(this->pub.limit) - this->bytesRead) : NULL,
                             .versionId = this->pub.versionId);
+                        storageReadDriverInterface(this->driver)->open(this->driver);
```
dAdAbird: Non-async storage would ignore `open()`'s error should an error happen.

src/storage/storage.c
dAdAbird, line 146: The assumption `(ioReadOpen == true) == file_exist` isn't true anymore for some (async) drivers.  Hence, `ioWriteOpen` may create a destination file for a non-existent source.

dAdAbird, line 214: The same as in `storageCopy()`, `ioReadOpen()` always returns true for asynchronous drivers. I don't think anything could happen here, as the subsequent read will throw an error after openAsync(). But the comment and the code structure, for that matter, became misleading.

src/storage/readMulti.c
```diff
@@ -0,0 +1,560 @@
...
+        FUNCTION_LOG_PARAM(STORAGE_READ_MULTI_DEFAULT, this);
+        FUNCTION_LOG_PARAM(STRING, file);
+        FUNCTION_LOG_PARAM(BOOL, param.compressible);
+        FUNCTION_LOG_PARAM(UINT64, param.offset);
+        FUNCTION_LOG_PARAM(VARIANT, param.limit);
+        FUNCTION_LOG_PARAM(STRING, param.versionId);
+    FUNCTION_LOG_END();
+
+    ASSERT(this != NULL);
+    ASSERT(file != NULL);
```
dAdAbird, line 286: Does it make sense to add something like
```
#ifdef DEBUG
bool opened;
#endif
```
to `struct StorageReadMultiDefault`, set it in `storageReadMultiDefaultOpen()` and ASSERT here just to be sure?

To my understanding, in `storageReadMultiRemote`, for example, any adds after open would be just ignored. But here we may modify the offset and limit of an already opened (in `storageReadMultiDefaultQueue()`) last request, which may lead to silently reading wrong data.

src/build/help/help.xml
```diff
@@ -1304,6 +1267,18 @@
                         <example>/etc/pki/tls/certs</example>
                     </config-key>
 
+                    <config-key id="repo-storage-prefetch" name="Repository Storage Prefetch">
+                        <summary>Repository storage prefetch.</summary>
+
+                        <text>
+                            <p>Determines the number of requests to prefetch when read from object stores, e.g. <proper>S3</proper>. This is not true concurrency but it reduces the latency involved in submitting requests.</p>
```
dAdAbird: "... when **_reading_** from ..."?

src/storage/readMulti.c
```diff
@@ -0,0 +1,560 @@
...
+    FUNCTION_LOG_END();
+
+    ASSERT(this != NULL);
+    ASSERT(!prelim || !lstEmpty(this->requestList));
+
+    // If the read queue is less than max then queue more reads
+    if (lstSize(this->queue) < this->queueMax)
+    {
+        // Look for request to queue but skip the last request during prelim queueing since it might not be complete
+        for (unsigned int requestIdx = lstSize(this->queue); requestIdx < lstSize(this->requestList) - prelim; requestIdx++)
```
dAdAbird: maybe `requestIdx + prelim < lstSize(this->requestList);` to avoid an unsigned overflow when `this->requestList` is empty and `prelim == 1`. I see that there is an ASSERT to prevent this, and according to the current code, `storageReadMultiDefaultQueue` with `prelim == true` is called only after adding a request to the list. But just in case...

src/storage/readMulti.c
```diff
@@ -0,0 +1,560 @@
...
+    ASSERT(param.limit == NULL || varType(param.limit) == varTypeUInt64);
+
+    MEM_CONTEXT_TEMP_BEGIN()
+    {
+        const String *const path = storagePathP(this->storage, fileExp);
+
+        // If targeting by time and the version has not already been resolved then look up the version id
+        if (param.versionId == NULL && storageTargetTime(this->storage) != 0)
+        {
+            param.versionId = storageInfoP(this->storage, fileExp, .ignoreMissing = true).versionId;
```
dAdAbird: Looks like the version will be resolved for every individual file before they are merged into one request. Maybe add some version "cache" either to storageReadMulti or somewhere in `command/restore`

dAdAbird (commented):
Looks good in general.
There are a couple of issues though. The main is a potential ignoring of `open()` errors for non-async drivers. It also retoroatievly broke contract for the existsing code if few places (again for non-async drivers). See comments to the code, plus due to asyncOpen, [this code](https://github.com/dwsteele/pgbackrest/blob/55f9406769aef79d35641773f35fa1f9753e73f7/src/command/repo/get.c#L145-L149) isn't correct anymore. It may create a destination file for the absent source. 
And in all cases it's not only about non-existent file. ioReadOpen() may return false due to other errors which might be ignored in the asynchronous case.

The rest is nitpicking and potential issues. Except for the version resolving, which may lead to tone redundant S3/Azure/GCP requests if I understand code correctly (again see the code comment).

Cheers

src/storage/read.c
dwsteele: Do mean a file missing error? Currently this can't happen because there are no drivers that have retry without async, but of course this could change in the future. I added a CHECK() in fe1654211 to guard against this. We can't put a regular error in because since the condition is currently not possible so it would not be easy to cover.

src/storage/storage.c
dwsteele: Yeah, very good point. I puzzled over this one for a bit but finally decided that async should only be enabled when it will be useful (done in d8cd01af122), e.g. from StorageReadMulti, rather than all the time. Who knows what other weird side effects could be lurking?

src/build/help/help.xml
dwsteele: Fixed in d8cd01af122.

src/storage/readMulti.c
dwsteele: This is a good idea. Done in 6f2832097c.

src/storage/readMulti.c
dwsteele: Make sense. Done in 23bb66964a0.

src/storage/readMulti.c
dwsteele: Huh, I though I have done this. Must have been something I intended to get back to. In any case, this would be a big hit to performance. The versionId lookups are not cheap. Fixed in d7625b416.

I chose to cache only the last file since in correct/optimal usage file reads should be grouped together with offsets ascending. However, it will still work if a file is not contiguous, it just won't be quite as efficient.

src/storage/storage.c
dAdAbird: Yeah, that's a great solution

dAdAbird (approved):
LGTM

dwsteele:
@dAdAbird Committed, thank you very much!

PR #2794, opened 2026-06-23
Bundle backup files in order rather than scanning for a better fit.

dwsteele:
Previously a file that would push a bundle over repo-bundle-size was skipped in the hope that a smaller file later in the queue would fit, which kept bundles under the target size at the cost of additional file unpacks.

Now files are added to the bundle in queue order and the bundle is closed once it reaches repo-bundle-size. As a result a bundle may exceed the target size by up to repo-bundle-limit, i.e. the size of the file that crosses the threshold, which is now noted in the help for repo-bundle-size.

In addition to the immediate benefits, this is important for a future commit where preserving bundle order from the prior backup is important.


pgstef (approved):
The help message looks clear enough to me, the code change is quite simple and the test pretty straightforward.

LGTM 👍

dwsteele:
@pgstef Committed. Thank you for the review!

PR #2799, opened 2026-07-02
Fix archive-push-queue-max not enforced when a WAL file errors.

### LINKED ISSUE #2629: archive-push-queue-max does not work when archive-push fails
Hello, i use pgBackrest 2.54.2.
And I'm a little confused by the behavior of the archive-push-queue-max with async archiving.
We kept getting an error on one WAL file. 
```
WARN: could not push WAL file '00000001000015920000005B' to the archive (will be retried): [45] raised from local-1 protocol: WAL file '00000001000015920000005B' already exists in the repo1 archive with a different checksum
                                        [ArchiveDuplicateError] on retry at 138ms: WAL file '00000001000015920000005B' already exists in the repo1 archive with a different checksum
```
I thought the queue would be cleared when archive-push-queue-max was exceeded. But we got a filesystem overflow.
It turned out that while we were getting the error, subsequent WALs were being archived successfully and the queue did not grow. But the status about successful archiving of these files was not being passed to PostgreSQL.

Is this normal behavior? Is there a way to avoid this problem?

dwsteele:
In asynchronous mode archive-push-queue-max measured the queue against the process list (WAL that still needs to be pushed) rather than the ready list (all WAL that PostgreSQL is waiting to archive). When a WAL file errored repeatedly the async look-ahead kept pushing the later files successfully and writing their ok status files, but PostgreSQL retries the failed file in order and never acknowledges the later ones, so those files dropped out of the process list while their WAL remained in pg_wal. The queue therefore appeared tiny and was never dropped, letting pg_wal fill the disk — exactly the outcome the option exists to prevent.

Measure the queue against the ready list, as synchronous mode already does. The files that get dropped are unchanged, so writing the ok status for the failed file unblocks PostgreSQL, which then acknowledges the already-pushed files from their existing status files and drains the queue.

### DIFF
--- a/doc/xml/release/2020s/2026/2.59.0.xml
+++ b/doc/xml/release/2020s/2026/2.59.0.xml
@@ -15,6 +15,19 @@
             </release-item>
 
             <release-item>
+                <github-issue id="2629"/>
+                <github-pull-request id="2799"/>
+
+                <release-item-contributor-list>
+                    <release-item-ideator id="vut12"/>
+                    <release-item-contributor id="david.steele"/>
+                    <release-item-reviewer id="stefan.fercot"/>
+                </release-item-contributor-list>
+
+                <p>Fix <setting>archive-push-queue-max</setting> not enforced when a WAL file errors.</p>
+            </release-item>
+
+            <release-item>
                 <release-item-contributor-list>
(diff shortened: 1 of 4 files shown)


pgstef (approved):
LGTM 👍

PR #2801, opened 2026-07-02
Add archive-expire-before option to clean up WAL archive.

pgstef:
Add a new command-line option `--archive-expire-before` to the `expire` command.
It removes WAL segments earlier than a given segment from the archive, while never removing WAL that a retained backup still requires.

A typical use is to drive archive clean-up from a standby's restartpoint (using PostgreSQL  `archive_cleanup_command`) without having to fall back to `pg_archivecleanup` and lose pgBackRest's retention safeguards.


src/command/expire/expire.c
```diff
@@ -767,6 +785,22 @@ removeExpiredArchive(const InfoBackup *const infoBackup, const bool timeBasedFul
                                 }
                             }
 
+                            // If --archive-expire-before is set, preserve everything from the requested segment up to the retention
+                            // backup start, so that only WAL strictly before the segment is eligible for removal. The condition
+                            // walExpireBefore < archiveExpireMax avoids an inverted range and skips the case where the segment is
+                            // newer than the retention backup start (nothing extra to preserve there). archiveExpireMax is always
+                            // set here: the retention backup is in archiveIdBackupList and has an archive start (checked above).
+                            if (walExpireBefore != NULL && strCmp(walExpireBefore, archiveExpireMax) < 0)
```
dwsteele, line 793: This comparison won't work correctly if the specified WAL segment is a .partial (which walIsSegment() allows). I doubt it makes sense to specify a .partial so probably best to filter that. A proposed fix is in fa3be2d858.

dwsteele (commented):
I moved the function refactor into a separate branch to make the PR more readable.

Overall it seems to work, but it does sort of work against the existing expiration. The best use seems to be on an archive-only repo with no backups and that should probably be reflected in the docs.

I'll do some more testing.

pgstef:
Thanks for having a look at this :-)

> Overall it seems to work, but it does sort of work against the existing expiration. The best use seems to be on an archive-only repo with no backups and that should probably be reflected in the docs.

Indeed, it works best when there are no backups in the repository, which was my initial target use case: allowing WAL archives to be sent to a separate repository for each standby, with each standby cleaning up its own repository using PostgreSQL's `archive_cleanup_command`.

Another scenario is when a user creates a repository but does not take a backup immediately. This option allows them to reclaim WAL archive space before the first backup is taken.

That said, I think it can also be useful when the repository does contain backups. It provides a way to free some space without having to adjust the retention settings, which can sometimes be a bit tricky.

I've suggested a documentation improvement in aeb556c.

src/command/expire/expire.c
pgstef: Oh yeah, definitely, we don't want to allow `.partial` here :+1:

dwsteele:
> That said, I think it can also be useful when the repository does contain backups. It provides a way to free some space without having to adjust the retention settings, which can sometimes be a bit tricky.

Only if you mean before the first full backup. I guess it is handy to not have to change the full retention to remove those WAL.

Somehow I had the idea it was going to allow you to expire WAL between backups but that doesn't work unless you change repo-retention-archive, which does the expiration itself.

The uses cases are only 1) repo with no backups, 2) WAL before the oldest backup (before retention has been met). That's what you added to the documentation so I assume this is all you intended?

pgstef:
> Only if you mean before the first full backup. I guess it is handy to not have to change the full retention to remove those WAL.

Yes, only before the 1st full backup indeed.
 
> The uses cases are only 1) repo with no backups, 2) WAL before the oldest backup (before retention has been met). That's what you added to the documentation so I assume this is all you intended?

Yes, correct :-)

> Somehow I had the idea it was going to allow you to expire WAL between backups but that doesn't work unless you change repo-retention-archive, which does the expiration itself.

Now that you've said that, is it something we should think about? Like killing PITR but only keeping WAL archives for the backup consistency before WAL segment number X?

dwsteele:
We should probably update this then. I implies there are three uses but there are actually two. The first sentence and the first clause of the second sentence are essentially the same case.
```
This option is primarily intended for archive-only repositories, where no backups are stored. It can also be used to reclaim WAL archive space before the first backup is taken, or to remove obsolete WAL without modifying the configured retention policy.
```
Maybe something like:
```
This option is primarily intended for archive-only repositories where no backups are stored. It can also be used to reclaim WAL archive space before the first backup when full retention has not yet been met to trigger archive expiration.
```

pgstef:
> Maybe something like:
> 
> ```
> This option is primarily intended for archive-only repositories where no backups are stored. It can also be used to reclaim WAL archive space before the first backup when full retention has not yet been met to trigger archive expiration.
> ```

I liked the `without modifying the configured retention policy` reasoning but `when full retention has not yet been met to trigger archive expiration` makes sense too :+1:

dwsteele:
@pgstef BTW, I'm assuming that you are going to make the doc change.

pgstef:
> @pgstef BTW, I'm assuming that you are going to make the doc change.

Just pushed then :+1:

PR #2802, opened 2026-07-02
Exit async archive-push on first error.

### LINKED ISSUE #1848: Archive Queue Max doesn't apply when archiving failed
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version: 2.39 

2. PostgreSQL version: 10.21

3. Operating system/version : Debian GNU/Linux 11 (bullseye)

4. Did you install pgBackRest from source or from a package? Debian PGDG packages

    - `pgbackrest.conf` file(s) [pgbackrest.txt](https://github.com/pgbackrest/pgbackrest/files/9403254/pgbackrest.txt)
~~~
archive_mode = on
archive_command = 'pgbackrest --stanza=lsi archive-push %p'
~~~
 
6. Describe the issue:

Using pgbackrest to backup and archive PostgreSQL 10 to S3 (at scaleway), I setup `archive-push-queue-max=200GiB` to limit the queue, because of the large amount of transaction every morning, which are slow to archive because of S3, which can fill the PG local storage. It works well, sometime I got that in  : 

~~~
2022-08-21 08:41:08.380 P00   WARN: dropped WAL file '000000030000AF05000000FC' because archive queue exceeded 200GB
~~~

But yesterday, there was a problem is S3, which return `500 Internal error` : 

~~~
2022-08-22 11:59:09.521 P02   WARN: could not push WAL file '000000030000AF980000009A' to the archive (will be retried): [104] raised from local-2 protocol: archive-push command encountered error(s):
                                    repo1: [ServiceError] [500] Internal Server Error
                                    [CommandError] on retry after 0ms
~~~

Problem : I thought that the queue will be full, then WAL file will be drop, but no, this morning, the disk was full and PG stop. 

Now, the S3 error is fixed, and it works, but I think it's not OK that pgbackrest does not apply archive-push-queue-max when there is a problem with the archive system (S3 here) 

It is normal  ? do I miss something ?

dwsteele:
When the repository was unavailable the async archive-push process kept trying every ready WAL segment, retrying each failure and writing an error file for it. Because the archive queue is only checked at the start of each run, a run could take so long that the WAL volume filled and PostgreSQL stopped before archive-push-queue-max was rechecked.

Exit the run as soon as any job errors so the async process is respawned by the next archive-push, which rechecks the queue and drops WAL once it exceeds archive-push-queue-max. Continuing past an error cannot relieve disk pressure because PostgreSQL cannot recycle any WAL past the oldest unarchived segment, so the segments pushed ahead free nothing; the only relief is the queue-max drop. job-retry already absorbs transient failures before they are reported here, so a reported error is a real failure and there is little benefit to continuing.

In-flight jobs are allowed to finish so their status is recorded and no segment is pushed twice.

### DIFF
--- a/src/command/archive/push/push.c
+++ b/src/command/archive/push/push.c
@@ -40,13 +40,6 @@ Ready file extension constants
 #define STATUS_EXT_READY_SIZE                                       (sizeof(STATUS_EXT_READY) - 1)
 
 /***********************************************************************************************************************************
-Stop async processing after this many errors. The async process will be spawned again by the next archive-push, which rechecks the
-queue and drops WAL if it now exceeds queue-max. This prevents a failing repo from filling the WAL volume while retrying every ready
-segment and creating an error file for each one.
-***********************************************************************************************************************************/
-#define ARCHIVE_PUSH_ERROR_MAX                                      3
-
-/***********************************************************************************************************************************
 Format the warning when a file is dropped
 ***********************************************************************************************************************************/
 static String *
@@ -447,7 +440,7 @@ typedef struct ArchivePushAsyncData
     CompressType compressType;                                      // Type of compression for WAL segments
     int compressLevel;                                              // Compression level for wal files
     ArchivePushCheckResult archiveInfo;                             // Archive info
-    unsigned int errorTotal;                                        // Number of errored jobs so far
+    bool errorFound;                                                // Has a job errored? If so, stop scheduling new jobs
 } ArchivePushAsyncData;
 
 static ProtocolParallelJob *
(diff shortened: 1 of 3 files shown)


src/command/archive/push/push.c
```diff
@@ -457,10 +458,12 @@ archivePushAsyncCallback(void *const data, const unsigned int clientIdx)
         // No special logic based on the client, we'll just get the next job
         (void)clientIdx;
 
-        // Get a new job if there are any left
+        // Get a new job if there are any left. Stop scheduling new jobs as soon as one errors so the async process exits and the
+        // next run rechecks the queue. Continuing past an error does not relieve disk pressure since PostgreSQL cannot recycle any
+        // WAL past the oldest unarchived segment, so the sooner the queue is rechecked the sooner WAL can be dropped if needed.
```
pgstef, line 463: I do like this comment 👍

pgstef (approved):
LGTM 👍

dwsteele:
The disadvantage here is that we are not "working ahead" so when the error is cleared there is more WAL to deal with. On the other hand, there's a good chance that WAL will either a) not be useful because there is a unrecoverable error, e.g. duplicate WAL, leading to a GAP or b) the issue is a general error that will cause all WAL segments to error out.

pgstef:
> The disadvantage here is that we are not "working ahead" so when the error is cleared there is more WAL to deal with. On the other hand, there's a good chance that WAL will either a) not be useful because there is a unrecoverable error, e.g. duplicate WAL, leading to a GAP or b) the issue is a general error that will cause all WAL segments to error out.

In practice, I also believe that the chances of only a single WAL segment failing are lower than the chances of encountering a general error (e.g. a repository permission or availability issue). So having to deal with a bit more work on the next run doesn't seem like a significant overhead compared to protecting against pg_wal exhaustion sooner 🤔

dwsteele:
Yeah, this was my thinking as well.

PR #2803, opened 2026-07-07
Add archive-push-batch-size to limit WAL pushed per asynchronous run.

dwsteele:
In asynchronous mode archive-push pushes every ready WAL segment in a single run and archive-push-queue-max is only checked at the start of the run, so a run that works through a large backlog can let the queue grow well past the limit before it is rechecked.

Add the archive-push-batch-size option (default 16GiB) to cap the amount of WAL pushed per run. The list of WAL to be pushed is trimmed to a whole number of segments using the size of the first ready segment, always keeping at least one, so the process exits and is respawned by the next archive-push, which rechecks the queue. Lower the value to recheck the queue more often at the cost of spawning the asynchronous process more frequently.


pgstef (approved):
LGTM 👍

PR #2805, opened 2026-07-07
Document that info reads encryption settings from global section.

### LINKED ISSUE #1407: Document issue regarding info and encryption per stanza
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version:
2.27

2. PostgreSQL version:
9.6.21

RPM from postgresql.org.

    - `pgbackrest.conf` file(s)
    - errors in the postgresql log file before or during the time you experienced the issue
    - log file in `/var/log/pgbackrest` for the commands run (e.g. `/var/log/pgbackrest/mystanza_backup.log`)

7. Describe the issue:
Please mention in the Encryption sections of the Configuration and User Guides that "pgbackrest info" only works when the cypher type and cypher-pass are in [global].  It's obvious **after the fact**, but not so beforehand.

dwsteele:
The info command reads encryption settings only from the global section when run without the stanza option, so encryption settings configured per stanza require the stanza option to read an encrypted stanza. Note this in the user guide encryption section and the repo-cipher-pass config reference.

### DIFF
--- a/doc/xml/release/2020s/2026/2.59.0.xml
+++ b/doc/xml/release/2020s/2026/2.59.0.xml
@@ -307,6 +307,23 @@
         </release-development-list>
     </release-core-list>
 
+    <release-doc-list>
+        <release-improvement-list>
+            <release-item>
+                <github-issue id="1407"/>
+                <github-pull-request id="2805"/>
+
+                <release-item-contributor-list>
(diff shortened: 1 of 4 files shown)


pgstef (approved):
Makes sense. The text looks good to me 👍

PR #2807, opened 2026-07-07
Document that expire-auto uses the backup command configuration.

### LINKED ISSUE #1874: inconsistency between expire parameter and backup command
1. pgBackRest version: 2.40

With a specific setting for the `expire` command, the setting is not used when the `expire` command is called by the `backup` command.

by example : 

~~~
[<stanza>:expire]
io-timeout=360
~~~

is not used by `expire` inside the `backup` command. 

To me, that's an inconsistency in the behavior, which should be fixed

best regards, 
Sébastien

dwsteele:
When expire runs automatically after a successful backup it uses the configuration loaded for the backup command, so options set only in an expire command section are not applied. Note this behavior in the expire-auto option help and point users to disabling it and running expire separately when expire-specific configuration is needed.

### DIFF
--- a/doc/xml/release/2020s/2026/2.59.0.xml
+++ b/doc/xml/release/2020s/2026/2.59.0.xml
@@ -335,6 +335,19 @@
 
                 <p>Document that info reads encryption settings from global section.</p>
             </release-item>
+
+            <release-item>
+                <github-issue id="1874"/>
+                <github-pull-request id="2807"/>
+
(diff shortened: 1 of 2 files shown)


pgstef (approved):
The suggested text looks good to me 👍 

I guess we don't have a way to reload the expire settings, or even simply detect it somehow to warn/info about it. So at least documenting the behavior is a good thing to do.

PR #2812, opened 2026-07-08
Error when running as root unless allow-root is enabled.

### LINKED ISSUE #1693: Keep file owner when running expire as root
1. pgBackRest version: 2.36
2. PostgreSQL version: 14.2
3. Operating system/version: Ubuntu 18.04
4. Did you install pgBackRest from source or from a package? Package

I have pgBackRest normally running as user "postgres" (in cron jobs). Sometimes I manually run `pgbackrest expire` as root (forgetting to sudo to user postgres). This changes the ownership of the backup.info and backup.info.copy files to root, so user postgres can no longer access them and the regular backups fail. This is a "gotcha" that got me a few times now. It would nice if pgBackRest kept the current owner of those files when running as root. (Perhaps commands other than expire are affected, too.)

dwsteele:
Running as the root user for commands other than restore can create files (e.g. backup.info in the repository) owned by root that the PostgreSQL user cannot access, causing later commands to fail.

Add the allow-root option to gate this. It defaults to false for all commands except restore, which is designed to run as root and manages file ownership carefully. The check applies to all command roles so a non-root command cannot run as root on a remote, and is skipped for help/version.


src/config/load.c
```diff
@@ -465,6 +466,26 @@ cfgLoad(const unsigned int argListSize, const char *argList[])
...
+        {
+            userInit();
+
+            if (userRoot())
+            {
+                THROW_FMT(
+                    OptionInvalidError,
+                    "the '%s' command must not be run as the root user\n"
+                    "HINT: running as root can create files that the regular user cannot access, causing later commands to fail.\n"
+                    "HINT: set the '" CFGOPT_ALLOW_ROOT "' option to run this command as root.",
```
hunleyd:
```suggestion
                    "HINT: set the '" CFGOPT_ALLOW_ROOT "' option to allow running this command as root.",
```

hunleyd (approved):

src/config/load.c
dwsteele: Good clarification. The prior version made it sound like the command would be forced to run as root when allow-root is set. Fixed in 4f92799f4de.

PR #2815, opened 2026-07-20
Skip unexpected output before protocol greeting.

nateships:
A login shell that prints output for non-interactive SSH sessions (an unguarded banner such as neofetch or motd in a shell rc) corrupts the protocol stream, and every remote operation fails with an opaque error deep in the first command, e.g.:

```
ERROR: [099]: invalid type at: [?25l[?7l    /#####   #####.
```

Since the greeting is always a JSON object on a single line, `protocolClientNew()` now searches for the greeting by skipping leading lines that do not begin as a JSON object, bounded at 64KiB so a stream that never produces a greeting still errors cleanly. Per review discussion, unexpected output is treated as a hard error rather than a warning: output that precedes the greeting is likely to continue during the protocol session, where it cannot be detected and will cause remote operations to fail unpredictably. The greeting is still searched for first so the error can confirm that the correct remote was reached, and the error includes the number of bytes skipped, a sample of the output to help identify the source, and hints:

```
ERROR: [103]: greeting found after 17 byte(s) of unexpected output beginning with '.[2Jlogin banner'
       HINT: is the remote shell printing output for non-interactive sessions?
       HINT: unexpected output may interrupt the protocol at any time so the connection cannot be used until the output is disabled.
```

Also per review:

* The `JsonFormatError` catch is narrowed to the check that the line begins as a JSON object. Errors from parsing or validating the rest of the greeting (e.g. a version skew or protocol change) now propagate as real failures instead of being treated as skippable noise.
* The `JsonRead` object is freed on the skip path.
* The skipped output sample is sanitized to printable characters so remote output cannot inject terminal escapes or forge log lines in error messages.
* The greeting-not-found errors are consolidated into a single helper.

Only the client read direction is affected; the reverse direction never passes through a login shell.

Tested with the protocol module unit tests (noise before a valid greeting, search bound exceeded, eof after noise, sanitized sample, trailing greeting key, existing bogus-greeting cases) at full coverage.

### DIFF
--- a/src/protocol/client.c
+++ b/src/protocol/client.c
@@ -21,6 +21,8 @@ output sampled for error reporting, and the hint added to errors caused by unexp
 #define PROTOCOL_GREETING_SKIP_MAX                                  65536
 #define PROTOCOL_GREETING_SKIP_SAMPLE_MAX                           64
 
+#define PROTOCOL_GREETING_UNEXPECTED_OUTPUT_FMT                                                                                    \
+    "after %zu byte(s) of unexpected output beginning with '%s'"
 #define PROTOCOL_GREETING_UNEXPECTED_OUTPUT_HINT                                                                                   \
     "HINT: is the remote shell printing output for non-interactive sessions?"
 
@@ -341,8 +343,8 @@ protocolClientGreetingNotFound(const char *const reason, const size_t skipSize, 
 
     THROW_FMT(
         ProtocolError,
-        "greeting not found (%s) after %zu byte(s) of unexpected output beginning with '%s'\n"
-        PROTOCOL_GREETING_UNEXPECTED_OUTPUT_HINT, reason, skipSize, strZ(skipSample));
+        "greeting not found (%s) " PROTOCOL_GREETING_UNEXPECTED_OUTPUT_FMT "\n" PROTOCOL_GREETING_UNEXPECTED_OUTPUT_HINT,
+        reason, skipSize, strZ(skipSample));
 }
 
 /**********************************************************************************************************************************/
@@ -376,20 +378,21 @@ protocolClientNew(const String *const name, const String *const service, IoRead 
             .sessionList = lstNewP(sizeof(ProtocolClientSession)),
         };
 
-        // Read, parse, and check the protocol greeting. The greeting is a JSON object on a single line so any line that does not
-        // begin as one is unexpected output (e.g. a login banner printed by the remote shell for a non-interactive SSH session).
-        // Unexpected output is an error since more of it is likely to arrive in the middle of the protocol session, but search for
-        // the greeting first so the error can confirm that the correct remote was reached. The search is bounded so a stream that
-        // never produces a greeting errors cleanly and errors include a sample of the output to help identify the source.
+        // Read, parse, and check the protocol greeting. The greeting is a machine-generated JSON object on a single line so it
+        // always begins with { and any line that does not is unexpected output (e.g. a login banner printed by the remote shell
+        // for a non-interactive SSH session). Unexpected output is an error since more of it is likely to arrive in the middle of
+        // the protocol session, but search for the greeting first so the error can confirm that the correct remote was reached.
+        // The search is bounded so a stream that never produces a greeting errors cleanly and errors include a sample of the
+        // output to help identify the source.
         MEM_CONTEXT_TEMP_BEGIN()
         {
             String *const skipSample = strNew();
-            volatile size_t skipSize = 0;
-            volatile bool greetingFound = false;
+            size_t skipSize = 0;
+            bool greetingFound = false;
 
             do
             {
-                String *volatile greetingLine = NULL;
+                String *greetingLine = NULL;
 
                 TRY_BEGIN()
                 {
@@ -402,18 +405,7 @@ protocolClientNew(const String *const name, const String *const service, IoRead 
                 }
                 TRY_END();
 
-                JsonRead *volatile greeting = NULL;
-                volatile bool greetingIsJson = false;
-
-                TRY_BEGIN()
-                {
-                    greeting = jsonReadNew(greetingLine);
-
-                    jsonReadObjectBegin(greeting);
-
-                    greetingIsJson = true;
-                }
-                CATCH(JsonFormatError)
+                if (!strBeginsWithZ(greetingLine, "{"))
                 {
                     // The line does not begin as a JSON object so it is unexpected output rather than a malformed greeting. Skip
                     // it, keeping a sample of the skipped output (sanitized to printable characters) for error reporting.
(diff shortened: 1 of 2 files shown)


src/protocol/client.c
```diff
@@ -340,45 +350,110 @@ protocolClientNew(const String *const name, const String *const service, IoRead
...
+                    {
+                        THROW_FMT(
+                            ProtocolError,
+                            "greeting not found after %zu byte(s) of unexpected output beginning with '%s'\n"
+                            PROTOCOL_GREETING_NOT_FOUND_HINT, skipSize, strZ(skipSample));
+                    }
+
+                    strFree(greetingLine);
                 }
+                TRY_END();
```
hunleyd: Note: this suggestion uses `isprint()`, which needs `#include <ctype.h>` added near the top of the file alongside the other includes (that line falls outside this diff so it can't be suggested inline here).

The `CATCH(JsonFormatError)` here wraps the entire greeting parse, including `jsonReadObjectEnd()` and the key/value validation loop, not just the initial "is this even JSON" check. That means a real greeting that fails validation for a genuine reason (e.g. a newer server adds a trailing field and `jsonReadObjectEnd()` throws `JsonFormatError`) gets silently treated as skippable noise instead of surfacing a clear error, and the client eventually just reports a generic "greeting not found" after exhausting the skip budget — masking what should be an obvious version/protocol mismatch.

Separately, on the skip path the `JsonRead` object allocated by `jsonReadNew()` is never freed (only `greetingLine` is), and the skip-sample bytes are copied verbatim from remote-controlled output into the error message with no sanitization of non-printable/control characters — which is a channel for terminal escape injection or forged-looking log lines if the remote is compromised or misconfigured.

Suggested fix narrows the catch to just the "does this line begin as a JSON object" check, lets any error past that point propagate as real, frees the `JsonRead` on the skip path, and sanitizes the sample to printable characters only. This also resolves the question of whether a line that parses as JSON but fails key validation should be skipped or hard-fail — with this change that's now an intentional, documented hard-fail rather than an inconsistency.

```suggestion
                JsonRead *volatile greeting = NULL;
                volatile bool greetingIsJson = false;

                TRY_BEGIN()
                {
                    greeting = jsonReadNew(greetingLine);

                    jsonReadObjectBegin(greeting);

                    greetingIsJson = true;
                }
                CATCH(JsonFormatError)
                {
                    // The line does not even begin as a JSON object, so it is unexpected output rather than a malformed greeting --
                    // skip it, keeping a sanitized sample of the beginning of the skipped output for error reporting
                    skipSize += strSize(greetingLine) + 1;

                    if (strEmpty(skipSample))
                    {
                        for (size_t sampleIdx = 0; sampleIdx < strSize(greetingLine) && sampleIdx < PROTOCOL_GREETING_SKIP_SAMPLE_MAX;
                             sampleIdx++)
                        {
                            const char sampleChar = strZ(greetingLine)[sampleIdx];
                            strCatChr(skipSample, isprint((unsigned char)sampleChar) ? sampleChar : '.');
                        }

                        if (strSize(greetingLine) > PROTOCOL_GREETING_SKIP_SAMPLE_MAX)
                            strCatZ(skipSample, "...");
                    }

                    if (skipSize > PROTOCOL_GREETING_SKIP_MAX)
                    {
                        THROW_FMT(
                            ProtocolError,
                            "greeting not found after %zu byte(s) of unexpected output beginning with '%s'\n"
                            PROTOCOL_GREETING_NOT_FOUND_HINT, skipSize, strZ(skipSample));
                    }

                    strFree(greetingLine);

                    if (greeting != NULL)
                        jsonReadFree(greeting);
                }
                TRY_END();

                // Once the line is confirmed to begin as a JSON object, any further format or validation error is a real greeting
                // mismatch (e.g. a version skew or protocol change), not unexpected output -- let it propagate instead of being
                // treated as skippable noise
                if (greetingIsJson)
                {
                    const struct
                    {
                        const StringId key;
                        const char *const value;
                    } expected[] =
                    {
                        {.key = PROTOCOL_GREETING_NAME, .value = PROJECT_NAME},
                        {.key = PROTOCOL_GREETING_SERVICE, .value = strZ(service)},
                        {.key = PROTOCOL_GREETING_VERSION, .value = PROJECT_VERSION},
                    };

                    for (unsigned int expectedIdx = 0; expectedIdx < LENGTH_OF(expected); expectedIdx++)
                    {
                        if (!jsonReadKeyExpectStrId(greeting, expected[expectedIdx].key))
                            THROW_FMT(ProtocolError, "unable to find greeting key '%s'", zNewStrId(expected[expectedIdx].key));

                        if (jsonReadTypeNext(greeting) != jsonTypeString)
                            THROW_FMT(ProtocolError, "greeting key '%s' must be string type", zNewStrId(expected[expectedIdx].key));

                        const String *const actualValue = jsonReadStr(greeting);

                        if (!strEqZ(actualValue, expected[expectedIdx].value))
                        {
                            THROW_FMT(
                                ProtocolError,
                                "expected value '%s' for greeting key '%s' but got '%s'\n"
                                "HINT: is the same version of " PROJECT_NAME " installed on the local and remote host?",
                                expected[expectedIdx].value, zNewStrId(expected[expectedIdx].key), strZ(actualValue));
                        }
                    }

                    jsonReadObjectEnd(greeting);

                    greetingFound = true;
                }
```

src/protocol/client.c
```diff
@@ -340,45 +350,110 @@ protocolClientNew(const String *const name, const String *const service, IoRead
...
+                String *volatile greetingLine = NULL;
 
-                const String *const actualValue = jsonReadStr(greeting);
-
-                if (!strEqZ(actualValue, expected[expectedIdx].value))
+                TRY_BEGIN()
+                {
+                    greetingLine = ioReadLine(this->pub.read);
+                }
+                CATCH(FileReadError)
```
hunleyd, line 401: Minor: the "greeting not found ... HINT: is the remote shell printing output for non-interactive sessions?" message is constructed independently at three call sites in this function (this one, and two more further down), two of them nearly identical. Might be worth a small static helper so the format/hint text has a single source instead of three copies to keep in sync. Not proposing a literal diff here since it would overlap with the other suggested change in this file — just flagging for whichever way you'd rather fold it in.

hunleyd (changes requested):
A few findings from reviewing the greeting-skip logic, with suggested fixes where the change is concrete enough to propose directly.

dwsteele:
@nateships As you suggested in a separate conversation, I think it makes sense to treat this as an error. So, search the input for the greeting (refactoring with @hunleyd's suggestions) but error if there is noise before the greeting with hints. My experience is that if an SSH connection has noise before the greeting it will often have it later as well and there won't be any way for us to detect that later noise. If the user gets a good error message, i.e. yes you are connecting to the right place but it's not likely to work, then they can configure SSH to suppress extraneous output.

PR #2816, opened 2026-07-24
Combine prior block incremental map reads during backup.

dwsteele:
During a block incremental backup, each file based on a prior backup reads that backup's block map to determine which blocks have changed. Previously each map was read with its own open and read, which is slow against an object store where every read is a separate round trip.

Order the files so the prior maps are laid out in read order and read them all through a single StorageReadMulti, combining many small reads into one sequential pass. Adjacent maps can be read over so they are fetched together.

Files are ordered at two levels. The process queue orders bundled files to mirror the prior backup they reference (by reference, then bundle id, then offset). Within a backup job the files are ordered block incremental first (by prior map file, then map offset) and then whole files by size, which keeps the maps together and in order and places the smallest whole files nearest them. Keeping files in the same relative order from one backup to the next also lets the same read pattern benefit future backup and restore.

The backupFileComparator() test shim places pg_control at the end of the bundle because its contents vary by architecture and would otherwise shift bundle offsets, making test output non-reproducible across architectures.

NOTE: An alternative to the above would be to always exclude pg_control from bundling. The advantage is it would make pg_control easier to get at for debugging, the disadvantage would be another small file in the repo. On the other hand the shim is test-only, it is just a little awkward the way it modifies the test output.

---

For performance testing 10K small tables were created in a fresh cluster against real S3. A full backup was made with block incremental enabled, half of the tables were modified, then a diff backup was made. Every other table was modified to make the worst case scenario for diff backup. Below is the average of three runs.

|     OPERATION           |  MAIN  | BRANCH | BRANCH RO OFF | BRANCH RO OFF PF 16 |
| ----------------------- | -----: | -----: | ------------: | ------------------: |
| FULL BACKUP             |   9.5s |   9.2s |             - |                   - |
| DIFF BACKUP             | 288.1s |   7.3s |         87.8s |               30.2s |

Diff backup performance for block incremental is significantly improved. The HTTP statistics from the local process show why:
```
MAIN statistics: {"http.client":{"total":2},"http.close":{"total":50},"http.request":{"total":5004},"http.retry":{"total":3},"http.session":{"total":54}}
BRANCH statistics: {"http.client":{"total":2},"http.request":{"total":12},"http.session":{"total":3}}
```

Even with read over disabled it does pretty well and even better when prefetch is increased.

The full backup is unchanged and is not actually faster in this branch. Since this was run against real S3 there was some variation. I did other runs were main was slightly faster.


dAdAbird (approved):
LGTM!

src/command/backup/file.c.inc
hunleyd, line 1: the loop that adds prior block-map reads to biMapRead and the loop that later drains them in FIFO order via ioCopyP each re-derive the same three-part eligibility condition independently. Nothing ties them together, so an edit to one condition without the other would silently desync which bytes go to which file, with no error until a checksum mismatch downstream.
Maybe something like:
```
+// Is a prior block incremental map needed for this file? Shared by the loop that adds maps to biMapRead and the loop that later
  +// reads them back out of it, so the two loops cannot drift out of the lockstep order biMapRead depends on.
  +static bool
  +backupFileBlockIncrMapNeeded(const BackupFileResult *const fileResult, const BackupFile *const file)
  +{
  +    FUNCTION_TEST_BEGIN();
  +        FUNCTION_TEST_PARAM(BOOL, fileResult->backupCopyResult == backupCopyResultCopy);
  +        FUNCTION_TEST_PARAM(SIZE, file->blockIncrSize);
  +        FUNCTION_TEST_PARAM(STRING, file->blockIncrMapPriorFile);
  +    FUNCTION_TEST_END();
  +
  +    FUNCTION_TEST_RETURN(
  +        BOOL,
  +        fileResult->backupCopyResult == backupCopyResultCopy && file->blockIncrSize != 0 && file->blockIncrMapPriorFile != NULL);
  +}
  +
   static List *
   backupFile(
   ...
                   // If block incremental file will be copied add block map to the read
  -                if (fileResult->backupCopyResult == backupCopyResultCopy && file->blockIncrSize != 0 &&
  -                    file->blockIncrMapPriorFile != NULL)
  +                if (backupFileBlockIncrMapNeeded(fileResult, file))
                   {
   ...
                           // Read prior block map
                           Buffer *blockMap = NULL;

  -                        if (file->blockIncrMapPriorFile != NULL)
  +                        if (backupFileBlockIncrMapNeeded(fileResult, file))
                           {
```

src/command/backup/process.c.inc
hunleyd, line 1: backupFileComparator and backupProcessQueueComparator have the same "preserve prior-backup locality" policy on two different structs but don't reference each other. Add a comment to both?
```
--- a/src/command/backup/protocol.c
  +++ b/src/command/backup/protocol.c
  @@ header comment for backupFileComparator
   The rationale for each ordering step is in the inline comments below.
  +
  +This is the within-job counterpart to backupProcessQueueComparator() in process.c.inc, which orders files across queues by the
  same
  +prior-backup locality (reference, then bundle id, then bundle offset) before jobs are assembled. The two comparators sort
  different
  +structs (BackupFile here vs ManifestFile there) at different stages and cannot share code, but both exist to preserve the same
  +read-locality property end to end, so a change to one ordering rule should be checked against the other.
   **********************************************************************************************************************************
  */
```
```
--- a/src/command/backup/process.c.inc
  +++ b/src/command/backup/process.c.inc
  @@ header comment for backupProcessQueueComparator
   // lets block incremental map reads (and reads of whole files in prior bundles) run in sequence rather than seeking back and
  forth.
  +//
  +// This is the across-queue counterpart to backupFileComparator() in protocol.c, which orders files within a single job by the
  same
  +// prior-backup locality once they reach the remote process. See that comparator's header comment for the full picture.
   static const Manifest *backupProcessQueueComparatorManifest = NULL;
```

hunleyd (approved):
approving with comments

src/command/backup/file.c.inc
dwsteele: We can't rid of the first two parts of the second comparison because they are required to route the non-BI path. So it really just replaces file->blockIncrMapPriorFile != NULL with the function, including two conditions that we already know to be true. I prefer it as is for keeping the conditions in the path of execution where I can see them

src/command/backup/process.c.inc
dwsteele: Good idea, done in e4536787b8e0d1.

PR #2817, opened 2026-07-28
Reuse large freed allocations from a memory context free list.

dwsteele:
Commands that process many files, e.g. backup, cycle through large allocations (buffers, etc.) in short-lived contexts. When these are freed the C library may return the memory to the kernel and fault it back in for the next file. Minor page faults are cheap on some hosts but expensive on others, e.g. x86 with kernel page-table isolation, where the faults have been observed to more than double backup time. Whether the memory is returned depends on heap layout, which varies from run to run with address randomization, so performance can also vary significantly between otherwise identical runs.

Retain large freed allocations on a size-keyed free list in the memory context manager and hand them back for allocations of the same size. This keeps the pages resident and makes reuse deterministic, independent of the C library and heap layout. Only allocations above a threshold are recycled since smaller allocations are served well by the allocator's own arena and do not cause the churn. Retained memory is bounded by a per-size block limit and an overall byte limit so the cost stays a few MiB per process at the default buffer size.

The first free list bucket is permanently reserved so it is never assigned to an arbitrary size. main() reserves it for ioBufferSize() after the configuration is loaded, which guarantees the size that churns most always has a bucket even when many other large sizes are freed first, and that it is checked first.

Each bucket tracks how many times it has been reused, up to a limit. Once all buckets are in use a new size evicts the least reused bucket and the reuse of every remaining bucket is decremented. A new bucket starts with a reuse of one so it is not evicted before it has had a chance to be reused, which would leave two sizes trading the same slot and never settling. Specialized one-off sizes, e.g. from config parsing, never get reused so they erode immediately, while a bucket that was reused a lot but is no longer being reused erodes over time and is eventually evicted. The reserved bucket is never evicted.

Because retained memory is not returned to the allocator, Valgrind can no longer detect use-after-free or uninitialized reads on a recycled block. Add memcheck client requests that poison a block while it is on the free list and mark it undefined when it is reused, which restores both. The requests are compiled into debug builds when the Valgrind header is available, which includes the unit tests, and are no-ops when not run under Valgrind.

---

I tested this in the original environment where I was seeing the regressions (while testing #2816) and they are resolved. It does not work *quite* as well as something brute force like `MALLOC_TRIM_THRESHOLD_=536870912 MALLOC_MMAP_THRESHOLD_=67108864` but is definitely more portable and at least the test timings are now consistent.


dAdAbird (approved):
LGTM.

Some thoughts outloud:
1. Does it make sense to have more persitent buckets like for ioBuffer? But then BLOB_BLOCK_SIZE, for example, fall in the same bucket if ioBuffer. Besides frequently allocated size would stay in the list longer anyway. So it doesn't make sense to do anything here w/o some stats or or other evideence.
2. Should we do something with memReAllocInternal()? I mean to do memAllocReuse() + memcpy() + memFreeReuse() instead of calling realloc() if certain contiditions are met. But this needs benchmrking.

dwsteele:
Thank you for the review!

> So it doesn't make sense to do anything here w/o some stats or or other evidence.

Agreed. I/O buffer churn is what led to this feature so that is the main optimization, rather than try to reuse I/O buffers throughout the code. Since chunk buffers grow as they are used they probably won't hit this cache as much.

> Should we do something with memReAllocInternal()?

It's an interesting idea but I doubt it would lead to much in the way of performance. What we have here is already *really* close to `MALLOC_TRIM_THRESHOLD_=536870912 MALLOC_MMAP_THRESHOLD_=67108864` even under stress so I don't think more optimization would be worth it.

dwsteele:
Sorry for the commit noise -- I thought localizing the valgrind defines would help with ccache churn but really I just need to commit to integration so the ccache can be rebuilt in the main repo. PRs can't do that.

PR #2819, opened 2026-07-29
Fix hang when the chunk buffer is smaller than the input buffer.

### LINKED ISSUE #2818: Full backup stalls near completion with CPU pegged in storageWriteGcs/bufCatSub when writing many small files to GCS repo with bundling enabled (2.58.0 and 2.59.0)
**Environment:**
- pgBackRest 2.58.0 (original occurrence), reproduced again after upgrading to 2.59.0
- Repo type: GCS (`repo1-type=gcs`), with `repo1-block=y` and `repo1-bundle=y` enabled
- OS: Rocky Linux 8.10, PostgreSQL 15
- Backup source: standby (`backup-standby=y`), remote `pg1-host`/`pg2-host` topology (primary + standby, backup run from a separate repo host)
- Database: multi-TB scale, very high relation/file count (large number of small/zero-byte data files relative to total data volume)

**Config (anonymized):**
```ini
[global]
repo1-type=gcs
repo1-path=/example_stanza/repository
repo1-gcs-bucket=<redacted-gcs-bucket>
repo1-gcs-key=/var/lib/pgsql/backup-service-key.json
repo1-block=y
repo1-bundle=y
log-level-console=detail
log-level-file=detail
log-path=/var/log/pgbackrest
start-fast=y
log-subprocess=y
delta=y
process-max=4
archive-async=y
spool-path=/var/spool/pgbackrest
archive-timeout=120
buffer-size=16MB

[global:backup]
process-max=10
[global:restore]
process-max=8
[global:archive-push]
process-max=8
[global:archive-get]
process-max=8

[example_stanza]
pg1-host=primary-host
pg1-path=/data/pgdata15
pg2-host=standby-host
pg2-path=/data/pgdata15
backup-standby=y
repo1-retention-full=4
```

**Symptom:**
Full backup (`--type=full --repo1-retention-full=999`) proceeds normally, then stalls near completion — `pgbackrest info` reports ~99.66% complete and stops advancing. The `backup:local` worker process remains alive and pegged at 100% CPU for 2+ hours with no further log output (log level `detail`), while:
- `/proc/<pid>/syscall` repeatedly shows `running` (not blocked in any syscall)
- `wchan` is `0`
- No sockets attributable to the process at the OS level during the stall

Immediately before the stall, the log shows a rapid burst of entries for many small/0-byte files being backed up from the standby host, suggesting the stall correlates with the tail of the backup where remaining files are numerous and tiny.

**Profiling:**
`perf record -p <pid> -g --call-graph=dwarf -- sleep 30` during the stall shows:
```
74.88%  storageWriteGcs
  74.16% storageWriteGcs
    61.82% storageWrite → ioWrite → ioCopy → backupFile → backupFileProtocol
    12.34% storageWriteGcs → storageWrite → ioWrite → ioCopy → backupFile → backupFileProtocol   (self-recursion)
  0.72%  bufRemains (inlined) → storageWriteGcs → ...

25.11%  bufCatSub
  25.08% bufCatSub → storageWriteGcs → storageWrite → ioWrite → ioCopy → backupFile → backupFileProtocol
```
Essentially all CPU time is self-time inside `storageWriteGcs`/`bufCatSub`; kernel-side network transmission functions appear in the raw sample set but at ~0% overhead, indicating actual data transmission is a negligible fraction of the work being done.

**Hypothesis:** With `repo1-bundle=y` enabled, many small files get packed together into a shared bundle object before being written to the repo. If bundling accumulates each file's content into a shared buffer via repeated `bufCatSub` calls without O(1) amortized append (e.g. re-copying already-accumulated buffer content on each append rather than growing geometrically), a workload with an unusually high count of small/zero-byte files could produce a stall that scales quadratically with file count rather than linearly with data volume — matching what we observed.

**Question for maintainers:** Is this a known interaction between `repo1-bundle`/`repo1-block` and the GCS write path's buffer handling? Would disabling bundling (`repo1-bundle=n`) be expected to avoid this, and if so, is that recommended for very-high-file-count databases, or is there a bundle-size tuning knob that would help?

dwsteele:
The chunk buffer used for S3/GCS/Azure multi-part upload is only written when it contains exactly chunk size bytes, so when the chunk buffer is allocated smaller than chunk size and fills up, no more data can be copied into it and no write is triggered, which results in an infinite loop.

This could happen when chunk size grew after a chunk was written, since the chunk buffer was resized only once before the copy loop, or when the chunk buffer had been sized exactly to fit a prior partial write and doubling was not enough to hold the next input buffer.

Resize the chunk buffer in the copy loop and make sure it is always large enough to hold the entire input buffer, unless max chunk size has been reached. Also grow the chunk buffer aggressively when it is not empty, since in that case more data is likely no matter how full the input buffer is.


dAdAbird (approved):
LGTM!

src/storage/write.c
hunleyd, line 1: storageWriteChunkBufferResize() was moved inside the copy loop but still takes the caller's whole input buffer and since bufUsed(input) reports total size on iteration 2+ the required-size calculation overcounts by bytesTotal, i think. it appears bounded by chunkSizeMax so it's just wasted allocation/copy.

hunleyd (changes requested):

src/storage/write.c
dwsteele: Good catch! The extra allocation was bounded and zero for common cases, but no reason to leave bytes on the floor due to bad math. Fixed in 61f3ddf4129.

hunleyd (changes requested):
fair warning, claude helped with this:
`src/storage/write.c:273-295`
```diff
       ASSERT(input != NULL);
       ASSERT(inputOffset <= bufUsed(input));
       ASSERT(chunk != NULL);
       ASSERT(chunkSizeMax > 0);
  +    ASSERT(bufSize(chunk) <= chunkSizeMax);

       // Resize chunk buffer if it cannot hold the remaining input and is less than max chunk size
       if (chunkSizeRequired > bufSize(chunk) && bufSize(chunk) < chunkSizeMax)
       {
  -        size_t chunkSize;
  +        // Chunk size must be at least large enough to hold the remaining input
  +        size_t chunkSize = chunkSizeRequired;

  -        // If the input buffer is full or the chunk buffer is not empty there is very likely more data so increase size of the
  chunk
  -        // buffer aggressively
  +        // If the input buffer is full or the chunk buffer is not empty there is very likely more data coming, so increase the
  chunk
  +        // buffer size aggressively rather than resizing repeatedly as small amounts of data arrive
           if (bufFull(input) || !bufEmpty(chunk))
           {
  -            // If this is the first write set chunk size equal to double the input buffer size
  -            if (bufSize(chunk) == 0)
  -            {
  -                chunkSize = bufSize(input) * 2;
  -            }
  -            // Else double chunk size so as not to resize the chunk buffer too many times since prior data must be copied
  -            else
  -                chunkSize = bufSize(chunk) * 2;
  -
  -            // Chunk size must be large enough to hold the remaining input
  -            if (chunkSize < chunkSizeRequired)
  -                chunkSize = chunkSizeRequired;
  +            // If this is the first write set chunk size equal to double the input buffer size, else double the current chunk size
  +            const size_t chunkSizeDouble = bufSize(chunk) == 0 ? bufSize(input) * 2 : bufSize(chunk) * 2;
  +
  +            if (chunkSizeDouble > chunkSize)
  +                chunkSize = chunkSizeDouble;
           }
  -        // Else no more writes are expected so allocate only what is needed
  -        else
  -            chunkSize = chunkSizeRequired;
  ```
test/src/module/storage/posixTest.c:1013
```diff
+        Buffer *const chunkInputLarge = bufNew(18);
  +        Buffer *const chunkBufferRequired = bufNew(2);
  +
  +        bufUsedSet(chunkInputLarge, 18);
  +        TEST_RESULT_VOID(
  +            storageWriteChunkBufferResize(chunkInputLarge, 0, chunkBufferRequired, 40),
  +            "resize when required size exceeds doubled size");
  +        TEST_RESULT_UINT(bufSize(chunkBufferRequired), 18, "chunk buffer sized for required, not doubled");
  +
  +        Buffer *const chunkInputHuge = bufNew(20);
  +        Buffer *const chunkBufferMax = bufNew(4);
  +
  +        bufUsedSet(chunkInputHuge, 20);
  +        TEST_RESULT_VOID(
  +            storageWriteChunkBufferResize(chunkInputHuge, 0, chunkBufferMax, 10), "resize when required size exceeds chunk size
  max");
  +        TEST_RESULT_UINT(bufSize(chunkBufferMax), 10, "chunk buffer capped at chunk size max");
```
doc/xml/release/2020s/2026/2.60.0.xml:2
```diff
<release-core-list>
  +        <release-bug-list>
  +            <release-item>
  +                <github-pull-request id="2819"/>
  +
  +                <release-item-contributor-list>
  +                    <release-item-contributor id="david.steele"/>
  +                </release-item-contributor-list>
  +
  +                <p>Fix hang on S3/GCS/Azure multi-part upload when the chunk buffer is too small to hold the next chunk.</p>
  +            </release-item>
  +        </release-bug-list>
  +
```

probably not for this PR but for a follow-up, in `src/storage/s3` (and azure/gcs), the copy loop is byte-for-byte identical across all three drivers except field names, and this bug had to be patched three times because of it. could you not hoist the loop into src/storage/write.c behind a driver callback (flush + return next chunk size), collapsing each driver's write function to one call and making storageWriteChunkBufferResize a static internal with no inputOffset parameter?

dwsteele:
> fair warning, claude helped with this:

This is really the same algorithm, just restated, and I don't prefer it. The new ASSERT() breaks existing tests that are designed to check the robustness of the code. True in production code, but a problem for our testing.

I did like part of the new comments better, though, so applied that in b1e7f97b9e647a8.

> probably not for this PR but for a follow-up, in src/storage/s3 (and azure/gcs), the copy loop is byte-for-byte identical across all three drivers except field names

They aren't really because each driver's *Async() function (e.g. storageWriteS3PartAsync()) has a different signature. These could be unified but would require changing the interface, adding a new struct to handle the different handling in the GCS driver, etc. This may be worthwhile but not something I want to take on right now.

If you feels strongly about it feel free to submit a patch once this one is in.

hunleyd (approved):

PR #2822, opened 2026-08-06
Remove the repo-put command and the cipher-pass option.

dwsteele:
Both are internal and in the past the integration tests used them. They were left in place for possible administration, investigation, and testing but in fact those uses have not materialized. The repo format has moved on so repo-put is now actively dangerous (it always was, really).

Without cipher-pass, repo-get always determines the passphrase from the repository: the repo passphrase for the info files, and the passphrase each of those carries for the files below it. That was already what happened whenever the option was not given, so what is left is a single path rather than two.

The tests now write the repository directly with the storage harness, which is all the get tests needed from repo-put.


pgstef (approved):
I completely agree that `repo-put` is potentially dangerous, so I'm in favor of removing it.

For `cipher-pass`, I also think it's better to rely on `repo-cipher-pass`, so removing this option makes sense as well. Having a single code path is simpler and easier to maintain.

Overall, the code changes look good to me. So 👍 for the PR

PR #2823, opened 2026-08-07
Replace deprecated OpenSSL APIs.

dwsteele:
The minimum required OpenSSL version is 1.1.1, which is new enough to provide replacements for nearly all the deprecated APIs in use.

Hardcoded 2048-bit DH parameters are replaced by SSL_CTX_set_dh_auto(), which selects parameters that match the strength of the certificate key. This is at least as strong as the parameters that were hardcoded and removes the need to read them from a PEM buffer with PEM_read_bio_DHparams().

The ECDH curve is now set with SSL_CTX_set1_groups_list() rather than being constructed as an EC_KEY and passed to SSL_CTX_set_tmp_ecdh(). Since 1.1.0 the latter simply extracts the curve from the key and sets it as the only supported group, so this is equivalent. The macro casts the curve list to char *, which drops const because the build enables -Wwrite-strings, so the resulting -Wcast-qual error is suppressed. GCC does not report this but clang does.

Together these remove all the -Wdeprecated-declarations pragmas from the TLS server. SSL_OP_SINGLE_DH_USE and SSL_OP_SINGLE_ECDH_USE are also removed since they have been defined as no-ops since 1.1.0.

Also replace SSLv23_method() with TLS_method(), ERR_load_crypto_strings()/OpenSSL_add_all_algorithms() with the OPENSSL_init_crypto() call they expand to, and EVP_MD_CTX_create()/EVP_MD_CTX_destroy() with EVP_MD_CTX_new()/EVP_MD_CTX_free().

SSL_get_peer_certificate() is left as is since SSL_get1_peer_certificate() is not available in 1.1.1 and adding a compatibility macro is not worth it while the deprecated name still works.


hunleyd (approved):
always love to see code simplifications

PR #2824, opened 2026-08-07
Compare a new backup label only against its own full backup set.

### LINKED ISSUE #2792: Error -  "new backup label is not later than latest backup"
1. pgBackRest version: 2.58.0

2. PostgreSQL version: 17.10

AlmaLinux 9.8
Master-Replica (patroni), Minio S3 Storage with Object locking (Governance mode)

package from pgdg

```
[main]
pg1-path=/var/lib/pgsql/17/data

[global]
repo1-path=/
repo1-type=s3
repo1-storage-port=9000
repo1-s3-endpoint=https://s3-cluster
repo1-s3-bucket=my-bucket
repo1-s3-uri-style=path
repo1-s3-key=<redacted>
repo1-s3-key-secret=<redacted>
repo1-s3-region=eu-west-3

repo1-retention-full-type=time
repo1-retention-full=14

archive-timeout=600
io-timeout=600

repo1-storage-upload-chunk-size=20MiB
process-max=1
start-fast=y
repo1-block=y
repo1-bundle=y
compress-type=zst
archive-async=y
spool-path=/var/lib/pgsql/spool

log-level-console=info
log-level-file=info

[global:archive-push]
compress-type=zst
process-max=2
```

6. Describe the issue:

Hi.
We started using object locking (GOVERNANCE) mode in minio.
After that, we started encountering an occasional problem, in the form of an error:
```
2026-06-22 00:00:23.894 P00   INFO: last backup label = 20260615-103753F, version = 2.58.0
2026-06-22 00:00:23.894 P00   INFO: execute backup start: backup begins after the requested immediate checkpoint completes
2026-06-22 00:00:23.966 P00   INFO: backup start archive = 00000036000008C500000028, lsn = 8C5/A0000028
2026-06-22 00:00:23.967 P00   INFO: check archive for prior segment 00000036000008C500000027
2026-06-22 00:00:25.002 P00   WARN: a timeline switch has occurred since the 20260615-103753F backup, enabling delta checksum
                                    HINT: this is normal after restoring from backup or promoting a standby.
2026-06-22 00:00:29.578 P00   INFO: backup '20260620-000009F' cannot be resumed: resume only valid for full backup
2026-06-22 00:00:38.989 P00  ERROR: [106]: new backup label '20260615-103753F_20260622-000024D' is not later than latest backup label '20260620-000009F'
                                    HINT: has the timezone changed?
                                    HINT: is there clock skew?
2026-06-22 00:00:38.989 P00   INFO: backup command end: aborted with exception [106]
```

My guess is that this happens when a full backup is interrupted, and the next one deletes the incorrect contents, but due to protection, the empty directory 20260620-000009F remains (since it's protected), and pgbackrest thinks something is wrong.

Output from mcli
```
$ ./mcli ls minio_prod/backup/main/
[2026-06-19 00:29:12] 5.2KiB STANDARD backup.info
[2026-06-19 00:29:12] 5.2KiB STANDARD backup.info.copy
[2026-06-22 11:15:13]     0B 20260612-000003F/
[2026-06-22 11:15:13]     0B 20260613-000003F/
[2026-06-22 11:15:13]     0B 20260615-103753F/
[2026-06-22 11:15:13]     0B 20260615-103753F_20260616-000007D/
[2026-06-22 11:15:13]     0B 20260615-103753F_20260617-000011D/
[2026-06-22 11:15:13]     0B 20260615-103753F_20260618-000021D/
[2026-06-22 11:15:13]     0B 20260615-103753F_20260619-000013D/
[2026-06-22 11:15:13]     0B 20260620-000009F/
[2026-06-22 11:15:13]     0B backup.history/
```

dwsteele:
A diff or incr label is generated from the prior backup label, so it begins with the full backup part of that label and can only collide with, or need to sort after, a label in the same full backup set. The history lookup has always known this and restricts itself to one set, but the repository listing beside it compared against every backup in the repository, including full backups from other sets, which a diff or incr label sorts before whenever the full backup it belongs to is older.

That costs nothing while the listing holds only backups that exist. On object stores that keep a prefix alive after everything under it has been deleted, such as MinIO with object locking or SeaweedFS, a backup that pgBackRest removed goes on being returned by the listing. An aborted full backup cleaned up by resume therefore keeps sorting after the prior full backup, and every diff and incr taken afterwards fails with "new backup label ... is not later than latest backup label ...", naming a backup that is no longer there.

Build the expression that restricts a comparison to a single full backup set once and filter both the backup listing and the history listing with it. For a full backup the expression matches everything, as it did before.

Build the expression from the prior backup label rather than from the newest label in the repository listing, since that is the value the listing gets wrong and it is also the label the new label is generated from. Where the last full backup has been removed and a diff or incr is generated from the last remaining full backup, which is the case the history restriction was written for, the two labels are the same.

### DIFF
--- a/doc/xml/release/2020s/2026/2.60.0.xml
+++ b/doc/xml/release/2020s/2026/2.60.0.xml
@@ -23,6 +23,19 @@
 
                 <p>Combine prior block incremental map reads during backup.</p>
             </release-item>
+
+            <release-item>
+                <github-issue id="2792"/>
+                <github-pull-request id="2824"/>
+
+                <release-item-contributor-list>
+                    <release-item-ideator id="anton.glushakov"/>
+                    <release-item-contributor id="david.steele"/>
+                    <release-item-reviewer id="douglas.j.hunley"/>
+                </release-item-contributor-list>
+
+                <p>Compare a new backup label only against its own full backup set.</p>
+            </release-item>
         </release-improvement-list>
 
         <release-development-list>
--- a/src/command/backup/label.c.inc
+++ b/src/command/backup/label.c.inc
@@ -19,11 +19,20 @@ backupLabelCreate(const BackupType type, const String *const backupLabelPrior, c
     {
         const String *backupLabelLatest = NULL;
 
+        // For full backup compare against all backups. For other backup types compare only against backups in the same full backup
+        // set, i.e. those whose name begins with the full backup part of backupLabelPrior. Since the new label is generated from
+        // backupLabelPrior it can never conflict with a label in another full backup set. This prevents label generation from
+        // failing when the last full backup is removed and then a diff/incr is generated from the last remaining full backup, or
+        // when a removed full backup is still returned by the repository listing, which can happen on object stores that retain
+        // empty prefixes.
+        const String *const labelRegExp =
+            type == backupTypeFull ?
+                backupRegExpP(.full = true, .differential = true, .incremental = true, .noAnchorEnd = true) :
+                strNewFmt("^%.*sF\\_" DATE_TIME_REGEX "(D|I)", DATE_TIME_LEN, strZ(backupLabelPrior));
+
         // Get the newest backup
         const StringList *const backupList = strLstSort(
-            storageListP(
-                storageRepo(), STRDEF(STORAGE_REPO_BACKUP),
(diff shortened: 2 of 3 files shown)


hunleyd (approved):
lgtm

PR #2826, opened 2026-08-10
Pass a cipher spec rather than a type and passphrase.

dwsteele:
The cipher type and passphrase were separate parameters in every function that reads or writes an encrypted file, every info file load and save, and every protocol message that hands a file to a local process. Anything added alongside them, e.g. the digest the key derives with, meant changing all of it again.

A cipher spec keeps them together. The pass is a buffer rather than a string so an absent pass is simply NULL, and it is copied into the object so the caller is free to release whatever it was read from. The spec is packed and unpacked as a unit for the protocol.

The specs form a chain. cfgCipherSpec() builds one from the options and caches it per repository, the info files are read with that, and each info file carries the spec for the files that depend on it, so nothing below the info files consults the options again.

A spec is named for what it opens: cipherSpecMain from the configuration, cipherSpecArchive from archive.info, cipherSpecManifest from backup.info, and cipherSpecBackup from the manifest. That turned up names that said the opposite of what they held, since what backup called the backup passphrase is what opens manifests, and restore documented the passphrase it loads the manifest with as being for backup files.

The test harness takes a spec as well, so the storage and backup harnesses no longer accept a type and pass pair, and the rule that a missing pass meant the main test pass is now spelled out in TEST_CIPHER_SPEC.


src/common/crypto/spec.c
```diff
@@ -0,0 +1,137 @@
...
+
+/**********************************************************************************************************************************/
+FN_EXTERN void
+cipherSpecToLog(const CipherSpec *const this, StringStatic *const debugLog)
+{
+    char typeZ[STRID_MAX + 1];
+    strIdToZ(cipherSpecType(this), typeZ);
+
+    // There is no digest when there is no cipher. The pass is never logged.
+    if (cipherSpecType(this) == cipherTypeNone)
```
hunleyd: if i'm reading this right, cipherSpecDup() calls cipherSpecNewP(cipherSpecType(this), cipherSpecPass(this)) without forwarding the source spec's digest, so the duplicate silently falls back to the default (SHA1) digest?

src/command/archive/get/get.c
```diff
@@ -95,7 +95,7 @@ typedef struct ArchiveGetFindCacheRepo
 {
     unsigned int repoIdx;
     CipherType cipherType;                                          // Repo cipher type
```
hunleyd:
```suggestion
```
isn't this stale?

test/src/module/common/cryptoTest.c
hunleyd, line 1: missing a test that covers the non-default digest in the dup test?

hunleyd (changes requested):

src/common/crypto/spec.c
dwsteele: I debated even keeping this because the old code never used another digest. But, the next commit does so I decided to leave it and then missed the dup. Oops! Fixed in db1a35dde9a4d36.

src/command/archive/get/get.c
dwsteele: Yes, dead. This inspired me to search for others but none were related to this commit so I'll fix them separately. Fixed in 9e6285b60.

test/src/module/common/cryptoTest.c
dwsteele: Yes, added along with the code change in db1a35dde9.

src/command/backup/process.c.inc
```diff
@@ -334,7 +334,7 @@ typedef struct BackupJobData
     const bool backupStandby;                                       // Backup from standby
     RegExp *standbyExp;                                             // Identify files that may be copied from the standby
     const CipherType cipherType;                                    // Cipher type
```
hunleyd: BackupJobData.cipherType (const CipherType cipherType;) is still declared and set (process.c.inc:801, .cipherType = cfgOptionStrId(cfgOptRepoCipherType)), but nothing in the file reads it anymore since cipherSpecBackup/cipherSpecType() now carries the same info

test/src/module/common/cryptoTest.c
hunleyd, line 1: Maybe I'm missing it but i don't see anything calling cipherSpecPack/cipherSpecNewPack directly with a real type+pass+digest and asserting the unpacked fields match. i _think_ it's being tested indirectly (cipherBlockNewPack), but I don't know if that would catch a field-order bug?

hunleyd (changes requested):

src/command/backup/process.c.inc
dwsteele: Yep, there's another one. This is trickier because it was not completely referenced. Fixed in 8083149ea2c27cc.

test/src/module/common/cryptoTest.c
dwsteele: > but I don't know if that would catch a field-order bug

Yes, because the field types differ, Buffer vs String.

But there is a real coverage gap in the serialization because digest again is not really be exercised. Added a test in 4ea507a16c.

This whole digest thing is a bit of a mess because the ability to specify the digest existed in the prior code but it was never used or properly exercised. I decided to keep it to reduce code churn but now I'm regretting it.

I think we are at the point where we either accept it as is (it is exercised a *lot* in the next commit) or just go ahead and rip it all out and accept the diff in the next commit (which I don't think would be too bad since it should be localized to CipherSpec and the tests). I'm good either way.

dwsteele: I'm leaning a bit towards rip it out. I dislike code that is not being used and it's always possible that we'll go a different way in the next commit and make this code useless. I rather doubt it, though, since I think upgrading the digest to sha256 is the right way to go.

dwsteele: Not hearing anything to the contrary, I pulled out the alternate digest support in 17e7ebd02d76. It is now hard-coded to SHA1, which reflects what we currently do.

I'll reintroduce it when we use it, proper exercised and with better test coverage than was demonstrated here.

src/command/repo/get.c
```diff
@@ -121,19 +121,22 @@ storageGetProcess(IoWrite *const destination)
...
+                            // Else use the manifest passphrase
+                            else
+                                cipherSpec = cipherSpecManifest;
                         }
                     }
                 }
 
                 // Error when unable to determine cipher passphrase
-                if (cipherPass == NULL)
+                if (cipherSpec == NULL)
```
hunleyd, line 135:
```suggestion
                if (cipherSpec == NULL || cipherSpecType(cipherSpec) == cipherTypeNone)
```

src/command/verify/verify.c
```diff
@@ -993,8 +987,8 @@ verifyBackup(VerifyJobData *const jobData)
                     // Initialize the jobData
                     MEM_CONTEXT_BEGIN(jobData->memContext)
                     {
-                        // Get the cipher subpass used to decrypt files in the backup and initialize the file list index
-                        jobData->backupCipherPass = strDup(manifestCipherSubPass(jobData->manifest));
+                        // Get the cipher spec used to decrypt files in the backup and initialize the file list index
+                        jobData->cipherSpecBackup = manifestCipherSpec(jobData->manifest);
```
hunleyd:
```suggestion
                        jobData->cipherSpecBackup = cipherSpecDup(manifestCipherSpec(jobData->manifest));
```

src/command/verify/verify.c
```diff
@@ -168,11 +168,11 @@ verifyInvalidFileAdd(List *const invalidFileList, const VerifyResult reason, con
 Load a file into memory
 ***********************************************************************************************************************************/
 static StorageRead *
-verifyFileLoad(const String *const pathFileName, const String *const cipherPass)
+verifyFileLoad(const String *const pathFileName, const CipherSpec *const cipherSpec)
 {
     FUNCTION_TEST_BEGIN();
         FUNCTION_TEST_PARAM(STRING, pathFileName);                  // Fully qualified path/file name
-        FUNCTION_TEST_PARAM(STRING, cipherPass);                    // Password to open file if encrypted
+        FUNCTION_LOG_PARAM(CIPHER_SPEC, cipherSpec);                // Cipher spec to open file if encrypted
```
hunleyd, line 174:
```suggestion
        FUNCTION_TEST_PARAM(CIPHER_SPEC, cipherSpec);               // Cipher spec to open file if encrypted
```

src/info/infoPg.c
hunleyd, line 166:
```suggestion
        FUNCTION_LOG_PARAM(FUNCTIONP, callbackFunction);
        FUNCTION_LOG_PARAM(CIPHER_SPEC, cipherSpec);
```

src/common/crypto/spec.c
```diff
@@ -0,0 +1,115 @@
...
+    FUNCTION_TEST_END();
+
+    ASSERT(this != NULL);
+
+    CipherSpec *result;
+
+    if (cipherSpecType(this) == cipherTypeNone)
+        result = cipherSpecNewNone();
+    else
+        result = cipherSpecNew(cipherSpecType(this), cipherSpecPass(this));
```
hunleyd:
```suggestion
    CipherSpec *const result = cipherSpecNew(cipherSpecType(this), cipherSpecPass(this));
```

src/config/config.c
```diff
@@ -1078,3 +1078,37 @@ cfgOptionInvalidate(const ConfigOption optionId)
...
     FUNCTION_TEST_RETURN_VOID();
 }
+
+/**********************************************************************************************************************************/
+FN_EXTERN const CipherSpec *
+cfgCipherSpecMainIdx(const unsigned int repoIdx)
+{
+    FUNCTION_TEST_BEGIN();
+        FUNCTION_TEST_PARAM(UINT, repoIdx);
+    FUNCTION_TEST_END();
```
hunleyd, line 1088:
```suggestion
    ASSERT(repoIdx < cfgOptionGroupIdxTotal(cfgOptGrpRepo));
    FUNCTION_TEST_END();
```

src/common/crypto/cipherBlock.h
```diff
@@ -18,21 +18,19 @@ Constructors
...
 FN_EXTERN IoFilter *cipherBlockNewPack(const Pack *paramList);
 
 /***********************************************************************************************************************************
 Helper functions
 ***********************************************************************************************************************************/
-// Add a block cipher to an io object
-FN_EXTERN IoFilterGroup *cipherBlockFilterGroupAdd(
-    IoFilterGroup *filterGroup, CipherType type, CipherMode mode, const String *pass);
+// Add a block cipher to an io object. Nothing is added when the repository is not encrypted.
+FN_EXTERN IoFilterGroup *cipherBlockFilterGroupAdd(IoFilterGroup *filterGroup, CipherMode mode, const CipherSpec *cipherSpec);
```
hunleyd, line 34: 9 hand-rolled if (type != none) cipherBlockNewP(...) blocks could collapse to 7 unconditional calls to the canonical cipherBlockFilterGroupAdd() by adding a .raw param to it.

src/info/info.c
```diff
@@ -213,7 +216,9 @@ infoNewLoad(IoRead *const read, InfoLoadNewCallback *const callbackFunction, voi
                             {
                                 MEM_CONTEXT_OBJ_BEGIN(this)
                                 {
-                                    this->pub.cipherPass = varStr(jsonToVar(value->value));
+                                    // The dependent files are encrypted with the same cipher type as this one
+                                    this->pub.cipherSpec = cipherSpecNew(
```
hunleyd, line 220: infoNewLoad()'s cipherSpec param is never actually used to decrypt (only its type is read), and the NULL/none dual-encoding for "unencrypted" is handled three different ways across the codebase. Pass CipherType not CipherSpec ?

hunleyd (changes requested):

src/command/verify/verify.c
dwsteele: This is good as-is. CipherSpec is smart enough to not log the passphrase so we don't need to do this everywhere.

src/common/crypto/cipherBlock.h
dwsteele: This is a bigger question which I would prefer to consider separately. My instinct is that we should *remove* this helper rather than propagate it. It is almost certainly responsible for the missing coverage I added in cb50a5d4fcecbb.

I originally thought that this would help keep the tests simpler and that it was not necessary to test encryption on every path because the encryption is "transparent". Cynthia disagreed and she was right. Encryption is not quite transparent and this helper masks that to the detriment of our test coverage.

src/info/info.c
dwsteele: We could, but we'd need to change it back in the next commit. I'm OK with leaving this as-is because CipherSpec is an existing type here (not something we invent for this usage) and it is just a question of pushing the cipherType() call around, which causes plenty churn down the road.

> and the NULL/none dual-encoding for "unencrypted" is handled three different ways across the codebase

That's not related to this field, as far as I can tell, since it is always non-NULL.

There is still some NULL vs none handling but that is in places where we are setting up (infoNew()/infoCipherSpecSet(), cipherSpecType(), cfgCipherSpecMainIdx()) and rarely touched in day to day programming, so I think "across the codebase" is not an accurate description.

I had a look and I don't see ready ways to improve it without shuffling complexity out into the general code. For example infoNew() could require cipherSpec to be non-NULL but then all the call sites that currently pass NULL need to pass cipherSpecNewNone() instead. That's not quite right, though, because at that point the eventual encryption is unknown rather than none. I think NULL represents that better and lets us know if we try to use the info cipherPass before it gets set.

In addition, these few call sites are exactly the ones that a going to change a *lot* in subsequent commits so a refactor here would likely be wasted (anyway I'm fine with them as-is). The goal of this patch is to isolate the places where we have complex logic outside of CipherSpec to make these subsequent commits more readable.

src/command/repo/get.c
dwsteele: This condition should never happen so it would cause a loss of coverage. However, it can be asserted, as in 85c4b6bf8a3d06629.

src/command/verify/verify.c
dwsteele: No dup is required here since jobData->manifest lives in the same mem context. However, that makes the MEM_CONTEXT_BEGIN() block useless. And it turns out jobData->cipherSpecBackup is only used once, so easier just to replace it with manifestCipherSpec(jobData->manifest). Done that way in 1ab6972f6268af.

Note that cipherSpecManifest and cipherSpecArchive cannot get the same treatment because infoBackup and infoArchive are not present in jobData.

src/info/infoPg.c
dwsteele: Added in 2f14f03573c.

src/common/crypto/spec.c
dwsteele: That makes sense and better yet make in inline as in 14a399d0ddb062e.

src/config/config.c
dwsteele: Added in 20978fb207d9e.

src/common/crypto/cipherBlock.h
dwsteele: Interesting historical note here -- encryption was once contained entirely within the storage layer so it really was transparent. Unfortunately, that meant the storage had to understand rules about the repo format and that was ultimately not sustainable, especially with features like bundling and block incremental which setup multiple encryption blocks in the same file.

hunleyd: want an issue for it then? just to not lose track of it

src/command/verify/verify.c
hunleyd: ah, that makes sense. thx for the details!

src/common/crypto/cipherBlock.h
dwsteele: I have a note for it.

src/command/annotate/annotate.c
```diff
@@ -53,11 +53,8 @@ cmdAnnotate(void)
             TRY_BEGIN()
             {
                 // Attempt to load the backup info file
```
hunleyd, line 58:
```suggestion
{
	// Get the repo storage in case it is remote and encryption settings need to be pulled down
	const Storage *const storageRepo = storageRepoIdx(repoIdx);
  
	// Attempt to load the backup info file
```

src/command/manifest/manifest.c
```diff
@@ -439,16 +438,12 @@ cmdManifestRender(void)
...
-        const InfoBackup *const infoBackup = infoBackupLoadFile(
-            storageRepo(), INFO_BACKUP_PATH_FILE_STR, cfgOptionStrId(cfgOptRepoCipherType),
-            cfgOptionStrNull(cfgOptRepoCipherPass));
-        const String *const cipherPass = infoPgCipherPass(infoBackupPg(infoBackup));
-        const CipherType cipherType = cipherPass == NULL ? cipherTypeNone : cipherTypeAes256Cbc;
+        const InfoBackup *const infoBackup = infoBackupLoadFile(storageRepo(), INFO_BACKUP_PATH_FILE_STR, cfgCipherSpecMain());
 
         // Load manifest
         const Manifest *const manifest = manifestLoadFile(
             storageRepo(), strNewFmt(STORAGE_REPO_BACKUP "/%s/" BACKUP_MANIFEST_FILE, strZ(cfgOptionStr(cfgOptSet))),
```
hunleyd, line 448:
```
@@ -437,12 +437,15 @@ cmdManifestRender(void)
       MEM_CONTEXT_TEMP_BEGIN()
       {
  +        // Get the repo storage in case it is remote and encryption settings need to be pulled down
  +        const Storage *const storage = storageRepo();
  +
           // Load backup.info and cipher
  -        const InfoBackup *const infoBackup = infoBackupLoadFile(storageRepo(), INFO_BACKUP_PATH_FILE_STR, cfgCipherSpecMain());
  +        const InfoBackup *const infoBackup = infoBackupLoadFile(storage, INFO_BACKUP_PATH_FILE_STR, cfgCipherSpecMain());
           const Manifest *const manifest = manifestLoadFile(
  -            storageRepo(), strNewFmt(...), infoBackupCipherSpec(infoBackup));
  +            storage, strNewFmt(...), infoBackupCipherSpec(infoBackup));
```

hunleyd (changes requested):

src/command/annotate/annotate.c
dwsteele: This exact code would throw an error on an unused variable, but done in 062c17ca1.

This is a bug fix unrelated to this patch so I'll commit it separately.

src/command/manifest/manifest.c
dwsteele: Done in 0e4e78e1ce. This is a bug fix unrelated to this patch so I'll commit it separately.

src/common/crypto/spec.c
```diff
@@ -0,0 +1,95 @@
...
+/**********************************************************************************************************************************/
+FN_EXTERN CipherSpec *
+cipherSpecNew(const CipherType type, const Buffer *const pass)
+{
+    FUNCTION_TEST_BEGIN();
+        FUNCTION_TEST_PARAM(STRING_ID, type);
+        FUNCTION_TEST_PARAM(BUFFER, pass);
+    FUNCTION_TEST_END();
+
+    ASSERT(type == cipherTypeNone || (pass != NULL && !bufEmpty(pass)));
```
hunleyd: cipherSpecNew()'s invariant assert is one-sided: it catches "encrypted but no pass" but not "unencrypted with a stray pass," which then gets silently dropped. 
```suggestion
    ASSERT((type == cipherTypeNone) == (pass == NULL));
    ASSERT(pass == NULL || !bufEmpty(pass));```

hunleyd (changes requested):
cipherSpecGen() is a CipherSpec factory living in a command module; backup.c reaches into stanza's header for it. Better moving to common/crypto/spec.c?

dwsteele:
> cipherSpecGen() is a CipherSpec factory living in a command module; backup.c reaches into stanza's header for it. Better moving to common/crypto/spec.c?

I'd rather not. The length of the passphrase is application defined and in the future there will be format specific logic which includes which digest to use, etc.

src/common/crypto/spec.c
dwsteele: Fixed in 5645f9034a/e31d84704f/fb908d5f6ec0. The assert turned up some issues in the test data but nothing that would have affected production.

hunleyd (approved):

dwsteele:
@hunleyd Whew, that was a big one. Thank you for your patience!

Issue #2829, opened 2026-08-14
Backup fails with [042] unexpected eof near 99.9% with repository bundling/hardlink enabled


JPCOSTA78:
1. pgBackRest version:

pgBackRest 2.59.0

The issue was originally observed multiple times with pgBackRest 2.56.0 and was reproduced after upgrading Production and DR to pgBackRest 2.59.0.

2. PostgreSQL version:

PostgreSQL 17

3. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each:

Production database server:
- Red Hat Enterprise Linux 9
- PostgreSQL 17
- pgBackRest 2.59.0
- Local POSIX backup repository
- Database size approximately 7.7-8.3 TB

DR standby:
- Red Hat Enterprise Linux 9
- PostgreSQL 17
- pgBackRest 2.59.0

DEV:
- Red Hat Enterprise Linux 9
- PostgreSQL 17
- pgBackRest 2.59.0

4. Did you install pgBackRest from source or from a package?

Package.

pgBackRest 2.59.0 was installed using the official PGDG RHEL 9 RPM:

pgbackrest-2.59.0-1PGDG.rhel9.8.x86_64

5. Please attach the following as applicable:

Relevant pgBackRest configuration used when the failures occurred:

```ini
[global]
repo1-path=/backup/<prod-stanza>
repo1-retention-full=1
process-max=8
delta=y
hardlink=y
repo1-bundle=y
repo1-block=y
checksum-page=y
start-fast=y
log-level-console=info
log-level-file=detail
compress-level=5
compress-type=zst
protocol-timeout=600

archive-async=y
archive-timeout=60
spool-path=/var/lib/pgbackrest/spool
archive-push-queue-max=2GB

[<prod-stanza>]
pg1-path=/data/postgres/17
```

Relevant PostgreSQL settings:

```text
archive_command = pgbackrest --stanza=<prod-stanza> archive-push %p >> /log/17/pgbackrest_archive.log 2>&1
archive_mode = on
listen_addresses = *
max_wal_senders = 10
wal_level = replica
port = 5432
```

WAL archiving itself is working normally.

I can also provide/attach:

- pgBackRest DEBUG log from the failed pgBackRest 2.59.0 backup
- Relevant PostgreSQL log entries around the failure
- Additional TRACE logging if required

At the timestamps of the failures, we found no corresponding:

- PostgreSQL error indicating corruption
- OOM event
[...]

Issue #2489, opened 2024-11-19
Incremental backups failing since enabling repo-block


pinaraf:
Please provide the following information when submitting an issue (feature requests or general comments can skip this):

1. pgBackRest version: 2.54.0

2. PostgreSQL version: 13.16

3. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each: debian 12 (bookworm)

4. Did you install pgBackRest from source or from a package? package from pgdg

5. Please attach the following as applicable:
    - `pgbackrest.conf` file(s)
[pgbackrest.conf.txt](https://github.com/user-attachments/files/17818097/pgbackrest.conf.txt)

    - `postgresql.conf` settings applicable to pgBackRest (`archive_command`, `archive_mode`, `listen_addresses`, `max_wal_senders`, `wal_level`, `port

```
       name       |                         setting                         
------------------+---------------------------------------------------------
 archive_command  | pgbackrest --stanza=saas.prod-13-publik archive-push %p
 archive_mode     | on
 listen_addresses | *
 max_wal_senders  | 10
 port             | 5432
 wal_level        | replica
(6 lignes)
```

    - errors in the postgresql log file before or during the time you experienced the issue
no error  in postgresql log.
    - log file in `/var/log/pgbackrest` for the commands run (e.g. `/var/log/pgbackrest/mystanza_backup.log`)
No log on postgresql side. On backup server side, see attached file.
[pgbackrest.log](https://github.com/user-attachments/files/17818085/pgbackrest.log)

7. Describe the issue:

Since enabling the repo-block feature, the backup full and the first incremental backup succeed, but each following incremental backup will fail, as can be seen in the attached log, with an "unexpected eof" error.
We are using pgbackrest in server/TLS mode. None of our other databases show this behaviour while the configuration is strictly identical (deployed using an automation tool).
Here are the info on our last two backups:

```
        full backup: 20241117-230001F
            timestamp start/stop: 2024-11-17 23:00:01+01 / 2024-11-18 00:16:56+01
            wal start/stop: 000000020000C71200000055 / 000000020000C713000000E8
            database size: 813GB, database backup size: 813GB
            repo1: backup size: 192.9GB

        incr backup: 20241117-230001F_20241118-025720I
            timestamp start/stop: 2024-11-18 02:57:20+01 / 2024-11-18 03:43:01+01
            wal start/stop: 000000020000C7180000001F / 000000020000C718000000B7
[...]

PR #2830, opened 2026-08-18
Add context to block map read errors during backup.

dwsteele:
A block incremental file stores its block map at the end of its data and the location of the map is recorded in the manifest, so when the data at that location is not a complete map the error comes from the map parser, which knows nothing about where the map came from. In #2829 that leaves the user with "unexpected eof" and no indication of which file failed, where the map was expected to be, or whether the data is short or simply not a map. The same signature in #2489 could not be diagnosed for the same reason.

Rethrow errors from the read and the parse with the manifest file, repository path, offset and map size. Also measure the bytes read from the repository, before decryption, and error when they are fewer than the manifest expects, since a short read would otherwise surface as a parse error that gives no hint that data is missing. This does not address whatever causes the mismatch, but it makes the next report actionable.

Move the read into backupFileBlockIncr() to keep the copy loop readable. The map is read in its own memory context with the filter created in the prior context, so the map buffer is freed once the filter has parsed it rather than living until the file is done.

### DIFF
--- a/doc/xml/release/2020s/2026/2.60.0.xml
+++ b/doc/xml/release/2020s/2026/2.60.0.xml
@@ -49,6 +49,18 @@
 
                 <p>Reduce the manifest checks made when reconstructing <file>backup.info</file>.</p>
             </release-item>
+
+            <release-item>
+                <github-pull-request id="2830"/>
+
+                <release-item-contributor-list>
+                    <release-item-contributor id="david.steele"/>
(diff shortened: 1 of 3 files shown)


pgstef (approved):
The code change looks good to me and the added output would indeed be helpful.

So 👍 for the PR

hunleyd (approved):

dwsteele:
@pgstef @hunleyd Thank you for the reviews!

PR #2831, opened 2026-08-18
General framework for new repository formats.

dwsteele:
The repository format has been fixed at 5 since 1.00. It was a compile-time constant written into every info file and manifest, and any mismatch at load was a hard error. A version could only read the format it wrote, so introducing a format meant a flag day where every repository had to be recreated. 1.00 was that flag day and this branch is so the next format does not need one.

A file is readable when its format is between REPOSITORY_FORMAT_MIN and REPOSITORY_FORMAT_MAX. A version can read a format it does not write and a repository can hold both while older backups and archives expire.

The format is requested with repo-format, which is repo indexed and valid only for stanza-create and stanza-upgrade. It is command line only so a format left in the configuration cannot migrate a stanza as a side effect of an upgrade run for another reason, such as a PostgreSQL version upgrade. The default stays 5.

stanza-upgrade --repo1-format=6 rewrites the two info files and nothing else, so a repository of any size migrates in the time it takes to write them. A downgrade is refused, since the info files are what stop a version from reading a format it does not support. The two files are saved separately, so an upgrade interrupted between them leaves a mismatch that running the upgrade again repairs and that every command reading both files now reports.

A backup set adopts a new format only at a full backup. A prior backup is a candidate for diff or incr only if it is at the format new backups are written with, and a resumable backup at another format is discarded, so a backup set is never mixed and its format can be read once rather than per file. Migrating therefore costs one full backup per stanza.

Once the info files are at format 6 a version that supports only format 5 cannot read the stanza at all, including the backups still at format 5. Such a version fails at info load with "expected format 5 but found 6" before touching any data.

test/spec-repofmt.md records the design and the reasoning for each rule. It will be removed before the final commit, with anything durable moving to the user documentation.


dwsteele:
Closed since I was attempting to stack PRs but it turns out "Stacked pull requests require all branches to be in the same repository. Cross-fork stacks are not supported.", which makes them functionally useless for us.

src/command/stanza/upgrade.c
```diff
@@ -58,6 +58,30 @@ cmdStanzaUpgrade(void)
...
             InfoPgData backupInfo = infoPgData(infoBackupPg(infoBackup), infoPgDataCurrentId(infoBackupPg(infoBackup)));
 
+            // Determine if the repository format should be updated. The format is only updated when it is explicitly requested,
+            // since the option default would otherwise downgrade a repository that has already been upgraded.
+            bool formatUpgrade = false;
+
+            if (cfgOptionIdxSource(cfgOptRepoFormat, repoIdx) != cfgSourceDefault)
+            {
+                const unsigned int format = cfgOptionIdxUInt(cfgOptRepoFormat, repoIdx);
+                const unsigned int formatRepo = infoArchiveFormat(infoArchive);
```
hunleyd:
```diff
      -                const unsigned int formatRepo = infoArchiveFormat(infoArchive);
      +                const unsigned int formatArchive = infoArchiveFormat(infoArchive);
      +                const unsigned int formatBackup = infoBackupFormat(infoBackup);
      +
      +                // Use the higher of the two info file formats so an interrupted upgrade that left them apart cannot be used
  to
      +                // sneak a downgrade of whichever file is still behind past this guard.
      +                const unsigned int formatRepo = formatArchive > formatBackup ? formatArchive : formatBackup;
      @@
      -                formatUpgrade = format != formatRepo || format != infoBackupFormat(infoBackup);
      +                formatUpgrade = format != formatArchive || format != formatBackup;
```

src/command/backup/backup.c
hunleyd, line 216:
```diff
               BackupData *const backupData = backupInit(infoBackup);
      +
      +        // Error if the info files do not agree on repository format. Archive info is only loaded when archive-check is
  enabled.
      +        if (backupData->archiveInfo != NULL)
      +            checkStanzaInfo(infoArchivePg(backupData->archiveInfo), infoBackupPg(infoBackup));
```

src/info/info.c
hunleyd, line 152:
```diff
               const String *checksumExpected = NULL;
      +        bool formatFound = false;                                          // Was the format key found in ini file
      @@
                               this->pub.format = (unsigned int)format;
      +                        formatFound = true;
      @@
      +            if (!formatFound)
      +                THROW_FMT(FormatError, "no repository format found\nHINT: is this a valid repository info file?");
      +
                   if (checksumExpected == NULL)
```

hunleyd (changes requested):
src/config/parse.c:2025 — The PR body states repo-format is "command line only so a format left in the configuration cannot migrate a stanza as a side effect" of an unrelated upgrade run. The config-FILE parser rejects a cfgSectionCommandLine option with a WARN (parse.c:2143-2148), but the ENVIRONMENT-variable parser (parse.c:1982-2050) has no equivalent check — it accepts any option valid for the command and sets `source = cfgSourceConfig`. A `PGBACKREST_REPO1_FORMAT=6` left in a cron/systemd env would silently trigger the migration on every later stanza-upgrade run for an unrelated reason — exactly the failure mode the design claims to prevent.
```diff
                           // Warn if reset option found in env
                           else if (option.reset) { ... continue; }
      +
      +                    // Warn if this option should be command-line only
      +                    if (parseRuleOption[option.id].section == cfgSectionCommandLine)
      +                    {
      +                        LOG_WARN_FMT("environment contains command-line only option '%s'", strZ(key));
      +                        continue;
      +                    }
```

dwsteele:
> The PR body states repo-format is "command line only so a format left in the configuration cannot migrate a stanza as a side effect" of an unrelated upgrade run.

Yeah, turns out this is not really true. At first I thought the only thing to do was make sure it was properly documented, done at 750de978b.

Then I realized we could at least have a warning so did that in 7571afff30. I'm a bit conflicted, though, since this warn will come up in normal operation if you choose to drive the upgrade process through env vars. That seems like a violation of the env var contract. Maybe we should stop at just documenting it.

What do you think?

src/command/stanza/upgrade.c
dwsteele: This scenario implies some manual shenanigans on the part of the user. Since archive.info is always written first is should not happen in a regular interrupted upgrade.

But given that user shenanigans are a bit part of what we protect against, done at 0337209da9 with minor changes.

src/command/backup/backup.c
dwsteele: Since backup is the command they are most likely to run after upgrade this check makes a lot of sense. Backup is the only command that handled both files but did not to the check. Done in 46f091b34ac7f8 but moved into backupInit() and used our existing checkStanzaInfo() function for the error.

src/info/info.c
dwsteele: This didn't matter before because the format was fixed and would be rewritten, but it definitely matters now. Done in fbdd5a1164ed.

hunleyd:
> > The PR body states repo-format is "command line only so a format left in the configuration cannot migrate a stanza as a side effect" of an unrelated upgrade run.
> 
> Yeah, turns out this is not really true. At first I thought the only thing to do was make sure it was properly documented, done at [750de97](https://github.com/pgbackrest/pgbackrest/commit/750de978be2bb0a3e77613f9a3879decc5c7625d).
> 
> Then I realized we could at least have a warning so did that in [7571aff](https://github.com/pgbackrest/pgbackrest/commit/7571afff30dfe9a2d829bc1c0857adc9ded7d7b8). I'm a bit conflicted, though, since this warn will come up in normal operation if you choose to drive the upgrade process through env vars. That seems like a violation of the env var contract. Maybe we should stop at just documenting it.
> 
> What do you think?

unless i;'m reading things wrong, the LOG_WARN sits inside if (formatUpgrade), which is only true when the requested format differs from what's already on disk. Once a stanza migrates to format 6, formatUpgrade is false on every later run so the warning fires exactly once, at the actual migration event, not on every invocation?

src/info/info.c
```diff
@@ -239,6 +261,11 @@ infoNewLoad(
             }
             TRY_END();
 
+            // The format defines how everything else is read, so a file without one is not an info file this version can use. The
```
hunleyd:
```diff --git i/src/info/info.c w/src/info/info.c
index 2c8f982d3..974027602 100644
--- i/src/info/info.c
+++ w/src/info/info.c
@@ -261,14 +261,10 @@ infoNewLoad(
             }
             TRY_END();
 
-            // The format defines how everything else is read, so a file without one is not an info file this version can use. The
-            // format is zero until the key is found and the value stored, so if we got here then the key was not found.
-            if (infoFormat(this) == 0)
-                THROW(FormatError, "repository format not found\nHINT: is this a valid " PROJECT_NAME " info file?");
-
             INFO_CHECKSUM_END(checksumActualFilter);
 
-            // Verify the checksum
+            // Verify the checksum first, so a truncated or corrupted file is reported as a checksum failure rather than as a
+            // missing format
             const String *const checksumActual = strNewEncode(
                 encodingHex, pckReadBinP(pckReadNew(ioFilterResult(checksumActualFilter))));
 
@@ -280,6 +276,11 @@ infoNewLoad(
                     ChecksumError, "invalid checksum, actual '%s' but expected '%s'", strZ(checksumActual),
                     strZ(checksumExpected));
             }
+
+            // The format defines how everything else is read, so a file without one is not an info file this version can use. The
+            // format is zero until the key is found and the value stored, so if we got here then the key was not found.
+            if (infoFormat(this) == 0)
+                THROW(FormatError, "repository format not found\nHINT: is this a valid " PROJECT_NAME " info file?");
         }
         MEM_CONTEXT_TEMP_END();```

File: src/command/backup/backup.c
Comment (hunleyd) [inline src/command/backup/backup.c:166]: the new archive/backup info-file consistency check only runs when archive-check is on, so --no-archive-check backups skip it entirely. Fixing it means always loading archive.info during backup, which changes behavior for anyone using --no-archive-check specifically because archive storage isn't reachable. Related: the pg-data-mismatch branch of the same check gives no hint pointing at stanza-upgrade, and reusing the check here also adds a pg-data cross-check backup never did before; worth calling out as a behavior change in the PR description either way.

File: src/command/check/common.c [hunk 3]
```diff
@@ -95,27 +95,46 @@ checkDbConfig(const unsigned int pgVersion, const unsigned int pgIdx, const Db *
...
+    {
+        const unsigned int formatArchive = infoPgFormat(archiveInfoPg);
+        const unsigned int formatBackup = infoPgFormat(backupInfoPg);
+
+        THROW_FMT(
+            FileInvalidError,
+            "backup info file and archive info file are at different repository formats\n"
+            "archive: format = %u\n"
+            "backup : format = %u\n"
+            "HINT: run " CFGCMD_STANZA_UPGRADE " with --repo-format=%u to complete an interrupted upgrade.",
```
hunleyd: the interrupted-upgrade repair hint prints --repo-format=N with no repo index; on a multi-repo config, following it migrates the wrong, untouched repo.

src/version.h
```diff
@@ -25,10 +25,19 @@ Config include path name. The parent path will vary based on configuration.
...
+REPOSITORY_FORMAT_MIN/MAX are the range that can be read. REPOSITORY_FORMAT_DEFAULT is the format used for new repositories when
+repo-format is not specified. The allow list for repo-format in build/config.yaml must be kept in sync with MIN/MAX.
 ***********************************************************************************************************************************/
-#define REPOSITORY_FORMAT                                           5
+#define REPOSITORY_FORMAT_5                                         5
+#define REPOSITORY_FORMAT_6                                         6
+
+#define REPOSITORY_FORMAT_MIN                                       REPOSITORY_FORMAT_5
+#define REPOSITORY_FORMAT_MAX                                       REPOSITORY_FORMAT_6
+#define REPOSITORY_FORMAT_DEFAULT                                   REPOSITORY_FORMAT_5
```
hunleyd: am i missing it, or is this unread anywhere?

hunleyd (changes requested):

dwsteele:
> unless i;'m reading things wrong, the LOG_WARN sits inside if (formatUpgrade), which is only true when the requested format differs from what's already on disk

You are reading it right. The WARN will only log on an upgrade, but the question is should we be warning at all when env vars are used in the documented way? Maybe we just need a note in the docs to note the danger.

src/command/backup/backup.c
dwsteele: > so --no-archive-check backups skip it entirely

My perspective is if anyone is using this as more than a debug flag -- then tough. :) But seriously, the user said `--no-archive-check` so we would not be honoring the flag if we went ahead and did an archive check.

> reusing the check here also adds a pg-data cross-check backup never did before; worth calling out as a behavior change in the PR description either way.

That was the intention and I think this is a reasonable additional check to introduce (I'm planning to commit it separately to call it out). When the check was originally written there was no stanza-create so archive.info was created on first archive and backup.info was created on first backup so the chance of a mismatch was very real. These days it is a lot less so but the check is still a good thing to do, IMO.

src/version.h
dwsteele: These are all used in the core code and plenty of tests.

src/info/info.c
dwsteele: That ordering makes more sense so done that way in 83859a284e.

src/command/check/common.c
dwsteele: > on a multi-repo config, following it migrates the wrong, untouched repo.

You'd get an error, actually, but the message should still be updated. Done in 03d9151319ec. This turned out to be a big change because we have never output indexing for an option not valid for the current command before, but I was able to commit the config changes separately in 3aa53ba24c7d4 (with a real use case) to keep that out of this patch.

hunleyd:
> > unless i;'m reading things wrong, the LOG_WARN sits inside if (formatUpgrade), which is only true when the requested format differs from what's already on disk
> 
> You are reading it right. The WARN will only log on an upgrade, but the question is should we be warning at all when env vars are used in the documented way? Maybe we just need a note in the docs to note the danger.

ISTM if I'm following the docs and have set up the vars correctly, and then I get a warning, that's something 'wrong' with the code. So I guess I'm leaning towards doc only.

Maybe @pgstef has an opinion?

src/command/stanza/upgrade.c
```diff
@@ -58,6 +58,35 @@ cmdStanzaUpgrade(void)
...
             InfoPgData backupInfo = infoPgData(infoBackupPg(infoBackup), infoPgDataCurrentId(infoBackupPg(infoBackup)));
 
+            // Determine if the repository format should be updated. The format is only updated when it is explicitly requested,
+            // since the option default would otherwise downgrade a repository that has already been upgraded.
+            bool formatUpgrade = false;
+
+            if (cfgOptionIdxSource(cfgOptRepoFormat, repoIdx) != cfgSourceDefault)
+            {
+                const unsigned int format = cfgOptionIdxUInt(cfgOptRepoFormat, repoIdx);
+                const unsigned int formatArchive = infoArchiveFormat(infoArchive);
```
hunleyd: A half-migrated stanza (archive.info at 6, backup.info at 5 after an interrupted upgrade) is only repaired if the operator re-supplies --repoN-format=6. A plain re-run of stanza-upgrade throws at the checkStanzaInfo() guard before saving anything, leaving the stanza wedged until someone manually re-adds the flag.

Since downgrade is already refused, the lagging file has exactly one legal target (max(formatArchive, formatBackup), which this code already computes as formatRepo) so why not auto-heal unconditionally, reserving the option purely for requesting a new target format?

src/version.h
hunleyd: REPOSITORY_FORMAT_DEFAULT is referenced nowhere in src/ (only its own #define and ~60 test call sites) ?

hunleyd (approved):

pgstef:
> > You are reading it right. The WARN will only log on an upgrade, but the question is should we be warning at all when env vars are used in the documented way? Maybe we just need a note in the docs to note the danger.
> 
> ISTM if I'm following the docs and have set up the vars correctly, and then I get a warning, that's something 'wrong' with the code. So I guess I'm leaning towards doc only.
> 
> Maybe @pgstef has an opinion?

I've not been through the entire PR yet, so I might miss something. But as we can't downgrade, I'm wondering if allowing to use the env vars is even a good idea. It feels quite dangerous to me at least. And experience shows that people tend to not read the docs or the release notes very carefully.

If you both feel it's not that dangerous, maybe a LOG_INFO might be enough? It could be handy for the user to figure out what happened, and it could even be a reminder to clean-up the env vars afterwards :thinking:

dwsteele:
> I'm wondering if allowing to use the env vars is even a good idea.

This would be a big change and carve out an exception for one option, where env vars are supposed to work for any option, including command-line options. So this would hit the container people who like to use env vars.

> And experience shows that people tend to not read the docs or the release notes very carefully.

True enough but there is only so much we can do.

> maybe a LOG_INFO might be enough

I doubt anyone would ever notice it.

I think we should just document it and see how it goes in the field. Some things are problems and some aren't...

dwsteele:
> I think we should just document it and see how it goes in the field.

Here's what that looks like with an admonition, 4db5ae1c42e. I'd rather not bikeshed this too much right now, it is easy enough to change later if we have a better idea.

dwsteele:
> REPOSITORY_FORMAT_DEFAULT is referenced nowhere in src/ (only its own #define and ~60 test call sites) ?

Whoops, yeah I saw info.c and did not notice it was in the test harness dir. Moved to the test harness in cabc5a3e35a.

This is harmless in theory but I don't want anyone thinking they can use it. The default format should always be pulled from the option default in the core code.

src/version.h
dwsteele: Whoops, yeah I saw info.c and did not notice it was in the test harness dir. Moved to the test harness in https://github.com/pgbackrest/pgbackrest/commit/cabc5a3e35a318cae13090f48baa6f0746a0c786.

This is harmless in theory but I don't want anyone thinking they can use it. The default format should always be pulled from the option default in the core code.

src/command/stanza/upgrade.c
dwsteele: This makes good sense so done that way in f3eb33c3981c. I left --repo1-format in the check hint since this is the proper way to run the command, but we'll still deal with it if they forget the option.

Either way they'll get a warning that the process was interrupted and we are repairing it.

PR #2833, opened 2026-08-19
Add checksum-page-error option to error on invalid page checksums.

### LINKED ISSUE #2687: Possible detection or actions on invalid checksums
Currently, checksum failures simply throw a warning in the output:

```
2025-10-10 16:15:23.532 P00   WARN: invalid page checksum found in file /home/greg/pg/17/data/base/16388/2151861 at page 0
```

The backup ends successfully, however. After that, the only way to see the problem is in the last line of the pgbackrest info output for that backup:

```
...
            repo1: backup set size: 75.9MB, backup size: 75.9MB
            error(s) detected during backup
```

(or as an error field in the json output). Is this information available anywhere else? Maybe pgbackrest should exit with a non-zero code if a checksum failed? Or a feature flag to have the entire backup fail if corruption is detected? I would think the bad checksum should advertise itself a little stronger, so I'm searching for ways to do that. Other quick ideas:

* Have a final line in the backup output that mentions the failure. That WARN line gets buried.
* Flag to add to backup command that works like -Werror  So any warnings result in the backup failing (or at least cause a non-zero exit, per above)
* Separate log file to track WARN level items

I suppose for now the best solution is to parse the info output after the backup completes and check for errors, but I would like to know if there are any other ideas, or objections to some of the bullets above.

dAdAbird:
The new checksum-page-error option determines whether invalid page checksums are reported as a warning or an error. It is available for the backup, restore, and expire commands.

For backup the option defaults to false, which preserves the prior behavior of emitting a warning for each file that contains invalid page checksums. A warning with the total number of checksum warnings is now emitted at the end of the backup so they are less likely to be missed in the log. When the option is enabled the backup terminates with an error as soon as an invalid page checksum is found. Page misalignment is handled the same way since it prevents checksum validation.

For restore the option defaults to true. When the option is enabled the restore terminates with an error as soon as a file with invalid page checksums is restored. Since this happens before pg_control is renamed, the cluster cannot be started. When the option is disabled a warning is emitted for each file that contains invalid page checksums and the total is reported at the end of the restore. Previously the restore command did not report invalid page checksums at all.

For expire the option defaults to true. If the backup that would become the oldest retained backup contains invalid page checksums then expire terminates with an error before removing anything. When the option is disabled expire proceeds and emits a warning if the oldest retained backup contains invalid page checksums. An adhoc expire proceeds in either case but still emits the warning.


src/command/expire/expire.c
```diff
@@ -89,6 +89,45 @@ expireBackup(InfoBackup *const infoBackup, const String *const backupLabel, cons
...
+
+    ASSERT(infoBackup != NULL);
+
+    if (infoBackupDataTotal(infoBackup) > 0)
+    {
+        const InfoBackupData *const backupData = infoBackupDataByLabel(infoBackup, backupLabel);
+
+        if (backupData->backupError != NULL && varBool(backupData->backupError))
+        {
+            ASSERT(backupData->optionChecksumPage);
```
dAdAbird: Is this enough? Or is it better to also check files in the manifest for checksum errors?

src/command/restore/process.c.inc
```diff
@@ -232,6 +233,15 @@ restoreJobResult(
 
                 String *const log = strCatZ(strNew(), "restore");
 
+                if (file.checksumPage && file.checksumPageError)
+                {
+                    *checksumPageError = true;
```
dAdAbird: Should it be more explicit, e.g. a StringList of files/errors?

src/command/restore/process.c.inc
```diff
@@ -232,6 +233,15 @@ restoreJobResult(
...
+                    
+                    LOG_WARN_FMT(
+                        "invalid page checksum found in file %s: %s",
+                        strZ(restoreFilePgPath(manifest, file.name)), strZ(file.checksumPageErrorList));
```
dAdAbird: Should the description be more explicit?

src/command/expire/expire.c
dwsteele: This is enough. You can use `info --set` to get  details about errors in a backup (https://github.com/pgbackrest/pgbackrest/blob/release/2.59.1/src/command/info/info.c#L1203). That would be good to add as a hint.

src/command/restore/process.c.inc
dwsteele: This is fine since the warnings are in the log. We just need to copy the language from other places like expire on how we log this: https://github.com/pgbackrest/pgbackrest/blob/release/2.59.1/src/command/expire/expire.c#L1202.

src/command/restore/process.c.inc
dwsteele: As long as this matches the message in backup then it's fine.

doc/xml/reference.xml
```diff
@@ -2239,6 +2239,16 @@
                 </text>
 
                 <option-list>
+                    <option id="checksum-page-error" name="Error on Invalid Page Checksums">
+                        <summary>Error on Invalid Page Checksums.</summary>
+
+                        <text>
+                            <p>Disabled by default. If this option is enabled with the <br-option>checksum-page</br-option>, a first failure in checksum validation will abort a backup.</p>
```
dwsteele: No need to put defaults in the description. They are automatically output in the help, e.g. https://pgbackrest.org/configuration.html#section-backup/option-archive-copy.

doc/xml/reference.xml
```diff
@@ -2337,6 +2347,18 @@
                 </text>
 
                 <option-list>
+                    <option id="checksum-page-error" name="Error on Invalid Page Checksums">
+                        <summary>Error on Invalid Page Checksums.</summary>
+
+                        <text>
+                            <p>Enabled by default. And the <cmd>expire</cmd> will stop and issue an error before any changes if the oldest retained backup would contain checksum errors. If this option is disabled, the <cmd>expire</cmd> will proceed and issue a warning if there are checksum errors in the oldest retained backup.</p>
```
dwsteele: Same here.

src/command/backup/backup.c
```diff
@@ -252,10 +257,10 @@ cmdBackup(void)
         }
 
         // Save the manifest before processing starts
-        backupManifestSaveCopy(manifest, cipherSpecManifest, false);
+        backupManifestSaveCopy(manifest, cipherPassBackup, false);
 
         // Process the backup manifest
-        backupProcess(backupData, manifest, cipherSpecManifest);
+        StringList *const warnList = backupProcess(backupData, manifest, cipherPassBackup);
```
dwsteele: I think my phrasing in the spec was not good here. I would like just *one* warning at the end to indicate that there are warnings in the log. The existing warnings should say inline. That last warning should follow the existing format as in https://github.com/pgbackrest/pgbackrest/blob/release/2.59.1/src/command/expire/expire.c#L1202.

src/command/backup/process.c.inc
```diff
@@ -277,10 +278,16 @@ backupJobResult(
...
+                                if (cfgOptionBool(cfgOptChecksumPageError))
+                                {
+                                    THROW_FMT(
+                                        ChecksumError,
+                                        "invalid page checksum%s found in file %s at page%s %s", plural, strZ(fileLog), plural,
+                                        strZ(error));
+                                }
+                                strLstAddFmt(
+                                    warnList, "invalid page checksum%s found in file %s at page%s %s", plural, strZ(fileLog),
+                                    plural, strZ(error));
```
dwsteele: Probably better to format the string once then output as needed.

dwsteele (changes requested):
This is the overall shape we are looking for. See comments for details.

src/command/restore/restore.c
hunleyd, line 1: 161,267 use wasChecksumPageErrors — never declared in this file. Fix: declare bool wasChecksumPageErrors = false; before the MEM_CONTEXT_TEMP_RESET_BEGIN() block that passes its address.

src/command/backup/process.c.inc
hunleyd, line 241: strZ(file.checksumPageErrorList) called unconditionally whenever file.checksumPage && file.checksumPageError. src/command/backup/process.c.inc:234-238 legitimately sets checksumPageError=true with checksumPageErrorList=NULL for the page-misalignment case, and strZ() (string.h:129) dereferences with no NULL check (strZNull() exists specifically as the safe alternative. Restoring any backup containing a misaligned file segfaults/asserts. Fix: use strZNull(), or branch the misalignment case separately.

src/command/expire/expire.c
```diff
@@ -253,6 +292,16 @@ expireFullBackup(InfoBackup *const infoBackup, const unsigned int repoIdx)
             // If there are more full backups then the number to retain, then expire the oldest ones
             if (strLstSize(currentBackupList) > fullRetention)
             {
+                // If checksum-page-error is set, check the would-be-oldest retained backup and abort with an error if it has
+                // checksum errors.
+                if (cfgOptionBool(cfgOptChecksumPageError))
```
hunleyd, line 299: pre-check is wired only into expireFullBackup; repo1-retention-full-type=time dispatches to expireTimeBasedBackup (expire.c:330), which has no equivalent check — contradicts reference.xml's "stop before any changes" text. (spec + maintainability + correctness converge)

src/command/restore/process.c.inc
```diff
@@ -232,6 +233,15 @@ restoreJobResult(
 
                 String *const log = strCatZ(strNew(), "restore");
 
+                if (file.checksumPage && file.checksumPageError)
```
hunleyd: checks file.checksumPage/checksumPageError before the zeroed (selective --db-include) and restoreResultPreserve handling. A selective or delta restore can hard-fail on a file that was zeroed/skipped, never actually restored.

src/command/backup/backup.c
```diff
@@ -299,6 +304,10 @@ cmdBackup(void)
             "%s backup size = %s, file total = %u", zNewStrId(manifestData(manifest)->backupType),
             strZ(strSizeFormat(infoBackupDataByLabel(infoBackup, manifestData(manifest)->backupLabel)->backupInfoSizeDelta)),
             manifestFileTotal(manifest));
+
+        // Print warnings if any
+        for (unsigned int warnIdx = 0; warnIdx < strLstSize(warnList); warnIdx++)
```
hunleyd: now buffers all page-checksum warnings into warnList and only emits them after backupDbPing/backupComplete/archive-check — any error in that stretch (dead standby, archive timeout, repo write failure — all realistic) discards every already-detected corruption warning that used to log immediately.

src/command/backup/process.c.inc
```diff
@@ -230,10 +232,9 @@ backupJobResult(
                             if (!pckReadBoolP(checksumPageResult))
                             {
                                 checksumPageErrorList = NULL;
```
hunleyd, line 232: The page-misalignment branch never checks checksum-page-error — only the invalid-checksum branch does (line 283). A misaligned file always just warns even with the option enabled, yet still sets backup-error=true in backup.info, which then trips my comment on expire.c line 297

src/command/expire/expire.c
```diff
@@ -253,6 +292,16 @@ expireFullBackup(InfoBackup *const infoBackup, const unsigned int repoIdx)
...
+                {
+                    const String *const oldestRetainedBackupLabel =
+                        strLstGet(currentBackupList, strLstSize(currentBackupList) - fullRetention);
+                    
+                    checkChecksumErrors(infoBackup, oldestRetainedBackupLabel, repoIdx, true);
```
hunleyd, line 304: pre-check runs before the dry-run guard, so expire --dry-run on a repo whose oldest-retained backup has checksum errors throws instead of reporting what it would do — defeats the point of --dry-run.

build/config.yaml
```diff
@@ -1161,6 +1161,18 @@ option:
     command-role:
       main: {}
 
+  checksum-page-error:
```
hunleyd, line 1164: checksum-page-error has no section: (confirmed — sibling options checksum-page/page-header-check both have section: global), so it's CommandLine-only. Combined with defaulting true for restore/expire, anyone whose commands are issued by orchestration (Patroni, CloudNativePG, PGO, cron) they don't directly control has no reachable opt-out.

hunleyd (changes requested):
cfgCommandSet() (src/config/config.c:65-79) only sets configLocal->command/commandRole — it never touches configLocal->option[...], and cfgOptionIdxDefaultValue() reads a defaultValue cached per-option at parse time (config.c:427-440). So main.c:157 swapping to cfgCmdExpire after backup does NOT re-resolve checksum-page-error's default — auto-expire (the common path, expire-auto defaults on) still sees backup's default (false), while a standalone pgbackrest expire sees true. Same corruption, opposite outcome depending on invocation path.

src/command/expire/expire.c
dAdAbird: Changed in https://github.com/pgbackrest/pgbackrest/pull/2833/changes/7631fce05750c3d6bc127838a9e1ccb0289dd6f3

src/command/restore/process.c.inc
dAdAbird: Changed in https://github.com/pgbackrest/pgbackrest/pull/2833/changes/9c83dda3374bb216e0f2b9233cdd6e8fa3ff5215

src/command/restore/process.c.inc
dAdAbird: changed in https://github.com/pgbackrest/pgbackrest/pull/2833/changes/9c83dda3374bb216e0f2b9233cdd6e8fa3ff5215

doc/xml/reference.xml
dAdAbird: Fixed in https://github.com/pgbackrest/pgbackrest/pull/2833/changes/6ee0d1ea44e88a45857c69192adfee4958d8692d

doc/xml/reference.xml
dAdAbird: Fixed in https://github.com/pgbackrest/pgbackrest/pull/2833/changes/6ee0d1ea44e88a45857c69192adfee4958d8692d

src/command/backup/backup.c
dAdAbird: Changed in https://github.com/pgbackrest/pgbackrest/pull/2833/changes/f2864e3218671f6c7424396a9414d78efe040aa9

src/command/backup/process.c.inc
dAdAbird: Yeah. Now that code is gone with https://github.com/pgbackrest/pgbackrest/commit/f2864e3218671f6c7424396a9414d78efe040aa9
Changed the same for restore here https://github.com/pgbackrest/pgbackrest/pull/2833/changes/7631fce05750c3d6bc127838a9e1ccb0289dd6f3

src/command/restore/restore.c
dAdAbird: Fixed in https://github.com/pgbackrest/pgbackrest/pull/2833/changes/7c923ad39aee9972828b961c558706599042e9d4
But then logic a bit changed here https://github.com/pgbackrest/pgbackrest/pull/2833/changes/9c83dda3374bb216e0f2b9233cdd6e8fa3ff5215

src/command/backup/process.c.inc
dAdAbird: Gone with changes in https://github.com/pgbackrest/pgbackrest/pull/2833/changes/f2864e3218671f6c7424396a9414d78efe040aa9

src/command/restore/process.c.inc
dAdAbird: Fixed in https://github.com/pgbackrest/pgbackrest/pull/2833/changes/1368cf7ce947c14d00fe5b5b64dd5eed34e32e8b

src/command/backup/backup.c
dAdAbird: Changed in https://github.com/pgbackrest/pgbackrest/pull/2833/changes/f2864e3218671f6c7424396a9414d78efe040aa9

src/command/backup/process.c.inc
dAdAbird: Should be fixed with https://github.com/pgbackrest/pgbackrest/pull/2833/changes/f2864e3218671f6c7424396a9414d78efe040aa9

src/command/expire/expire.c
dAdAbird: Fixed in https://github.com/pgbackrest/pgbackrest/pull/2833/changes/156460634a55219e0dec70afda5a5573fd8418f4

src/command/expire/expire.c
dAdAbird: My logic there was that the oldest full backup in other cases remains the same, hence no need in warining/errors.

But now I think maybe it needs to be checked one more time? What do you think @dwsteele? Or leave it as is and just clarify this in the reference?

build/config.yaml
dAdAbird: I don't quite understand the concern. You mean the situation where someone has access to pgbackrest.conf but cannot change how commands run? Then how do they set those commands in the first place?
Or have I misunderstood the issue?

dAdAbird:
> cfgCommandSet() (src/config/config.c:65-79) only sets configLocal->command/commandRole — it never touches configLocal->option[...], and cfgOptionIdxDefaultValue() reads a defaultValue cached per-option at parse time (config.c:427-440). So main.c:157 swapping to cfgCmdExpire after backup does NOT re-resolve checksum-page-error's default — auto-expire (the common path, expire-auto defaults on) still sees backup's default (false), while a standalone pgbackrest expire sees true. Same corruption, opposite outcome depending on invocation path.

Yeah, I forgot to raise this question and got used to this behaviour. Thanks for catching it @hunleyd 
So currently, if you run a backup with the default `--checksum-page-error` (which is `n`), the expire cmd that runs automatically after the backup inherits that `--checksum-page-error=n` and would show on WARN, although the "default" for the expire is to throw an error. I left it intentionally like that so expire after could clean up after the backup and not leave the storage overflowing. But, yeah, maybe expire has to error in this case as well. @dwsteele, what's your take?

dAdAbird:
Thanks for your reviews @dwsteele and @hunleyd. I hope I addressed/replied to all your feedback

src/command/backup/process.c.inc
```diff
@@ -276,10 +280,19 @@ backupJobResult(
...
+                                        ChecksumError,
+                                        "invalid page checksum%s found in file %s at page%s %s", plural, strZ(fileLog), plural,
+                                        strZ(error));
+                                }
                                 // ??? Update formatting after migration
                                 LOG_WARN_FMT(
                                     "invalid page checksum%s found in file %s at page%s %s", plural, strZ(fileLog), plural,
                                     strZ(error));
+
+                                (*warningTotal)++;
```
dAdAbird, line 288: Now looking at this, I'm not sure if we need to track anything else but the checksum error warnings. Since other warnings could happen before out of the reach of this func and the total number in the end might be misleading...

src/command/restore/restore.c
```diff
@@ -261,6 +263,18 @@ cmdRestore(void)
...
         LOG_INFO_FMT(
             "restore size = %s, file total = %u", strZ(strSizeFormat(sizeRestored)), manifestFileTotal(jobData.manifest));
+
+        if(checksumErrorFileTotal > 0)
+        {
+            const String *message = strNewFmt(
+                CFGCMD_RESTORE " command encountered page checksum error(s) in %u file(s), check the log file for details",
+                    checksumErrorFileTotal);
+
+            if (cfgOptionBool(checksumErrorFileTotal))
```
dwsteele: cfgOptionBool() should not be called here --  just checksumErrorFileTotal checked

src/command/backup/process.c.inc
```diff
@@ -209,6 +210,8 @@ backupJobResult(
                             " corrupted.\n"
                             "NOTE: this does not indicate a problem with the PostgreSQL page checksums.",
                             strZ(file.name), strZ(strNewEncode(encodingHex, BUF(file.checksumSha1, HASH_TYPE_SHA1_SIZE))));
+                        
+                        (*warningTotal)++;
```
dwsteele: This is a different kind of checksum (file checksum) and should not be included here since the issue will be corrected on resume. The warning is just to let the user know that something is funny with the storage.

src/command/backup/process.c.inc
```diff
@@ -276,10 +280,19 @@ backupJobResult(
                                 // Make message plural when appropriate
                                 const char *const plural = errorTotalMin > 1 ? "s" : "";
 
+                                if (cfgOptionBool(cfgOptChecksumPageError))
+                                {
+                                    THROW_FMT(
+                                        ChecksumError,
+                                        "invalid page checksum%s found in file %s at page%s %s", plural, strZ(fileLog), plural,
```
dwsteele: Let's generate the error message once and use it for both the error and warning.

src/command/backup/process.c.inc
dwsteele: No need to track the file checksum (as noted above). I'm not sure about the alignment error -- in theory this is something that should be fixed on replay, but in that case I don't think it should be possible to see it. Let's leave the alignment error. I've never seen one in the field so I don't think this is a big danger.

dwsteele (changes requested):
Looking pretty good but I found a few more minor issues.

src/command/restore/restore.c
dAdAbird: It's a typo; I meant to check cfgOptChecksumPageError here. Fixed in https://github.com/pgbackrest/pgbackrest/commit/f6b216aa8e20d51010b243f10fbf3ac2b36dd117

src/command/backup/process.c.inc
dAdAbird: I removed file checksums from tracking. And I decided to remove the alignment as well. You were not sure about them, and their potential presence made the summary message clunky. I think if we were to track them, it should be done separately.
Changes are here: https://github.com/pgbackrest/pgbackrest/commit/45112203441be53e251929fc3b3404d63b875e9f

src/command/backup/process.c.inc
dAdAbird: Fixed in https://github.com/pgbackrest/pgbackrest/commit/60271098e7cb3c31059a458e1441f9a3ae48ef14

src/command/backup/process.c.inc
dAdAbird: Removed in https://github.com/pgbackrest/pgbackrest/commit/45112203441be53e251929fc3b3404d63b875e9f

src/command/restore/restore.c
hunleyd: In 9c83dda, the second check regressed:

  if (cfgOptionBool(checksumErrorFileTotal))

  cfgOptionBool() takes a ConfigOption id, not a count — this passes the file count where cfgOptChecksumPageError used to be. Looks
  like a search/replace slip during the bool→count refactor. Should stay:

  if (cfgOptionBool(cfgOptChecksumPageError))

src/command/backup/process.c.inc
hunleyd: Still open. f2864e3 changed the delivery mechanism (buffered warnList → immediate LOG_WARN+count), but didn't add the missing option check to this branch. Current state:
```
  if (!pckReadBoolP(checksumPageResult))
  {
      checksumPageErrorList = NULL;
      LOG_WARN_FMT("page misalignment in file %s: ...", ...);
  }
```
No cfgOptionBool(cfgOptChecksumPageError) check, no THROW, unlike the invalid-checksum branch right below it (if (cfgOptionBool(cfgOptChecksumPageError)) THROW(...)). A misaligned file still just warns regardless of the option, while file.checksumPageError still gets set true for it so it still trips the expire.c backup-error handling.

Also: this branch doesn't increment warningTotal, so a misalignment warning won't even show up in the final "encountered N warning(s)" summary.

src/command/expire/expire.c
hunleyd: This stops the throw under --dry-run, but the guard wraps the whole function body, not just the THROW:
```
  if (infoBackupDataTotal(infoBackup) > 0 && (!cfgOptionValid(cfgOptDryRun) || !cfgOptionBool(cfgOptDryRun)))
  {
      ...
      if (shouldFail)
          THROW(ChecksumError, strZ(message));

      LOG_WARN_FMT("%s: %s", ...);
  }
```
Under dry-run this condition is false, so detection and LOG_WARN_FMT both get skipped too. expire --dry-run on a repo whose oldest-retained backup has checksum errors now reports nothing, no crash, but no preview of what the real run would do either, which was the actual point raised (report what it would do, not throw). Same pattern repeats at the adhoc-expire call site (expire.c:1240).

Suggest gating only the throw: `if (shouldFail && !(cfgOptionValid(cfgOptDryRun) && cfgOptionBool(cfgOptDryRun)))`, and leave LOG_WARN_FMT unconditional so dry-run still surfaces the warning.

build/config.yaml
hunleyd: Not about conf-file access vs. command-control being separately restricted for the same person it's that they're two different surfaces, and checksum-page-error is only reachable from one of them.

`build/lib/config/parse.py` confirms it: omitting section: defaults the option to command-line, and the parser explicitly rejects setting a command-line-section option in pgbackrest.conf. Sibling options checksum-page/page-header-check both declare `section: global`, so they're settable in the conf file. checksum-page-error isn't.

Concretely: someone running Patroni, CloudNativePG, PGO, or a cron job someone else wrote controls the box pgbackrest.conf lives on, but not the literal argv the orchestrator execs; that command line is built internally by the orchestrator/operator code, not passed through from the user. So they can freely edit pgbackrest.conf, but have no path to add --checksum-page-error=n to the actual invocation. Combined with restore/expire defaulting this to true (hard-fail), that's a real behavior change with no reachable opt-out for that whole class of user; needs `section: global` (or per-stanza) added here, same as its siblings.

hunleyd (changes requested):
see additional comments. looking better, but still a few changes needed

src/command/restore/restore.c
dAdAbird: Yep, I've that as well; fixed in https://github.com/pgbackrest/pgbackrest/commit/7e18f77bf174242025dd2cde55e75dbb3bf361ca

src/command/backup/process.c.inc
dAdAbird: I didn't intend to throw errors on misalignment, tbh. Just keep showing a warning, and additionally count for the later log. But then, after  I've decided to keep misalignment out of changes at all (do not count warnings as well) see. https://github.com/pgbackrest/pgbackrest/pull/2833#discussion_r3832154460. I don't think it's worth throwing an error for misalignment. But then, this path sets `checksumPageError = true` in the manifest, so maybe it should be counted... Let's keep the discussion there: https://github.com/pgbackrest/pgbackrest/pull/2833#discussion_r3832154460

src/command/backup/process.c.inc
dAdAbird: On the other hand, after @hunleyd's comment, I'm in doubt. The misalignment error path sets `checksumPageError = true` in the manifest. So it makes sense to count them towards the total number of checksum warnings in the end. But I think we still should not throw an error for misalignment even if `cfgOptChecksumPageError` is on. Or should we? @dwsteele , @hunleyd, what do you think?

src/command/expire/expire.c
dAdAbird: Ok, for some reason I thought warnings shouldn't be there as well... Yes, makes sense. Fixed in 8e3e85c7c
_btw, the suggested fix would lead to throwing a warning twice in dry run _

doc/xml/reference.xml
hunleyd: So the linked commit (6ee0d1ea) fixed 2 of the 3 command sections but missed restore's own entry.

src/command/expire/expire.c
hunleyd: The guard moved. It did not go away.

Before this commit, the check for dry run was inside checkChecksumErrors. Now it wraps the whole call at both sites:
```
  // line 299 (pre-flight, shouldFail = true)
  if (cfgOptionBool(cfgOptChecksumPageError) && (!cfgOptionValid(cfgOptDryRun) || !cfgOptionBool(cfgOptDryRun)))
  {
      ...
      checkChecksumErrors(infoBackup, oldestRetainedBackupLabel, repoIdx, true);
  }

  // line 1227 (post-hoc, shouldFail = false)
  if (!cfgOptionValid(cfgOptDryRun) || !cfgOptionBool(cfgOptDryRun))
  {
      ...
      checkChecksumErrors(infoBackup, infoBackupData(infoBackup, 0).backupLabel, repoIdx, false);
  }
```
In --dry-run, both conditions are false. Neither call happens. So expire --dry-run on a repo with checksum errors in the oldest retained backup gives no THROW and no LOG_WARN_FMT. This is the same result as before this commit — the check just moved one level up.

The original ask was to gate only the throw, and let the warning run every time. This code still hides the warning during a dry run.

On the double-warning point: checkChecksumErrors calls LOG_WARN_FMT once, and each site calls checkChecksumErrors once. There is no second warning to collide with.

Suggested fix — gate the throw only, inside checkChecksumErrors:
```
  if (shouldFail && !(cfgOptionValid(cfgOptDryRun) && cfgOptionBool(cfgOptDryRun)))
      THROW(ChecksumError, strZ(message));

  LOG_WARN_FMT("%s: %s", cfgOptionGroupName(cfgOptGrpRepo, repoIdx), strZ(message));
```
Then drop the dry-run condition from both call sites at lines 299 and 1227, so the function always runs and only the throw is conditional.

dAdAbird: Both statements are wrong.
> ...   
// line 1227 (post-hoc, shouldFail = false)
  if (!cfgOptionValid(cfgOptDryRun) || !cfgOptionBool(cfgOptDryRun))
  {
...
In --dry-run, both conditions are false. Neither call happens. So expire --dry-run on a repo with checksum errors in the oldest retained backup gives no THROW and no LOG_WARN_FMT. This is the same result as before this commit — the check just moved one level up.

No, the check in line 1227 wraps only the infoBackupSaveFile() call. The checkChecksumErrors() call in L1240 is conditioned only on infoBackup size. Hence, WARN will be shown

> On the double-warning point: checkChecksumErrors calls LOG_WARN_FMT once, and each site calls checkChecksumErrors once. There is no second warning to collide with.

Yes, `LOG_WARN_FMT` is called only once in `checkChecksumErrors()`. But the `checkChecksumErrors()` is called twice - once by `expireFullBackup()` and then closer to the end of cmdExpire().

------
Anyway, giving it a second thought, I don't agree with the initial statement. I believe we should throw a checksum error in dry-run (the same BackupSetInvalidError, for example). Otherwise, the dry-run is misleading, showing it'll remove backups which it won't do. So I removed dry-run-guardrails from checksum errors here 2a9f53730

And now the expire cmd with or without --dry-run shows exactly the same output. The dry-run just doesn't do any modifications. And this is how it should be, in my opinion.

src/command/expire/expire.c
```diff
@@ -1183,6 +1234,10 @@ cmdExpire(void)
                 removeExpiredBackup(infoBackup, adhocBackupLabel, repoIdx);
                 removeExpiredArchive(infoBackup, timeBasedFullRetention, repoIdx);
                 removeExpiredHistory(infoBackup, repoIdx);
+
+                // Check the oldest retained backup for page checksum errors and issue a warning if any
+                if (infoBackupDataTotal(infoBackup) > 0)
+                    checkChecksumErrors(infoBackup, infoBackupData(infoBackup, 0).backupLabel, repoIdx, false);
```
dAdAbird, line 1240: This will show a WARN even if the expire did nothing, but the oldest backup has a checksum error. 
For example:
1. We have 4 backups, the oldest with a checksum error; the `repo1-retention-full=4`. We run expire and it shows:
```
...
2026-08-25 15:21:26.165 P00   INFO: repo1: 18-6 no archive to remove
2026-08-25 15:21:26.165 P00   WARN: repo1: oldest retained backup 20260825-143231F contains invalid page checksum(s)
                                    HINT: use info --set command to get details about errors in the backup.
...
```

2. The same options, but we run an incremental backup. The auto-expire afterwards will show the same:
```
2026-08-25 15:25:10.307 P00   INFO: repo1: 18-6 no archive to remove
2026-08-25 15:25:10.307 P00   WARN: repo1: oldest retained backup 20260825-143231F contains invalid page checksum(s)
                                    HINT: use info --set command to get details about errors in the backup.
```

In both cases, the expire hasn't expired anything but warns (legitimately) anyway. I think we should leave it. But eager to hear other opinions.

build/config.yaml
dAdAbird: Agreed, added to the global section: 1c701292e

doc/xml/reference.xml
dAdAbird: Fixed: de8f97314

doc/xml/reference.xml
hunleyd, line 1: Error on invalid page checksums. Help generator only lowercases the first char, so Title Case renders as error on Invalid Page Checksums (visible in helpTest.c:224's own expectation). Every other option summary is sentence case.

build/config.yaml
hunleyd, line 1: add to the backup: block of checksum-page-error:
```
      backup:
        default: false
        depend:
          option: checksum-page
          list:
            - true
```

hunleyd (changes requested):
--no-checksum-page --checksum-page-error is currently accepted and silently inert (nothing computes page checksums to trip the throw). reference.xml already states this dependency in prose; other options in this repo enforce it structurally the same way (online, stanza).

Restore's checksumErrorFileTotal only counts files with result == restoreResultCopy — a delta/--force restore where corrupt files are preserved-in-place exits clean with no warning, even though the corrupt pages are on disk. Also: restore's new default-true gate applies retroactively to backups made with older pgbackrest versions, with no escape path named in the thrown message.

One option name, three different semantics/timings/defaults (backup: fail-fast mid-copy, default off; restore: fail after full restore completes, default on; expire: fail before any change, default on); identical doc summary text across all three papers over the difference.

warningTotal/checksumErrorFileTotal out-params threaded  through backupJobResult→backupProcess→cmdBackup (and restore's equivalent) are derivable from state already sitting in the  manifest/backupError at the point of use — could delete the plumbing entirely.
  
Backup's THROW fires before manifestFileUpdate() persists the triggering file's error state, so the one file that caused the abort is the one file missing from the manifest a resumed backup would read.

src/command/expire/expire.c
hunleyd: Yes, it needs a check — but not by copy-pasting the current one.

The existing pre-check (in expireFullBackup) has real problems: it computes "oldest retained backup" differently than the tail check at expire.c:1240 (can disagree), it blocks all expiration for the repo — archive and history cleanup too, not just the flagged backup — and it throws even under --dry-run with no escape hatch named in the error. Mirroring that same function into expireTimeBasedBackup would just double the blast radius of a design that's already too aggressive, not close a gap cleanly.

expireDiffBackup has the identical gap — even count-based retention skips the check when a differential (not full) backup is what's expiring. So this isn't a "time-based retention is special" issue, it's "the check lives in one of four sibling functions instead of at their shared reconvergence point."

Fix: one check in cmdExpire, after retention decisions are applied to the in-memory infoBackup (works identically for full/diff/time-based/adhoc), before infoBackupSaveFile/removeExpired* run. That covers every retention type in one place, deletes the divergent oldest-backup computation, and is the natural answer to "should we check here too" — yes, once, upstream of all four paths, not four times.

build/config.yaml
dAdAbird: Although `checksum-page` is `n` by default, it can be automatically set to `y` by pgBackRest based on pg_cnotrol. Hence it's kinda on by default. And propsed changes make it reuired to explicitly specify `checksum-page=y` even if it would be on implictly. Which is not an issue but to me looks a bit confusing.

I don't think it's an issue that both `--no-checksum-page` and `--checksum-page-error` can be set at the same time. `--checksum-page-error` in my reading doesn't say "there are has to be page checksums", it's more "IF there are page checksums and errors related to them"

doc/xml/reference.xml
dAdAbird: Fixed a804a0916

src/command/expire/expire.c
dwsteele: Agreed, let's leave it. We should be warning often and noisily about this condition. We can always remove the warning later if it proves to be too much.

build/config.yaml
dwsteele: This one is a bit tricky since checksum-page actually has *no* value before it is set by the backup code -- because there is a dependency outside of the config rules. There's not much we can do about this so I think the way Andrew has it set up is correct.

When we introduce options that act on new repo formats we'll need a way to do deferred option resolution and at the point we should be able to fix this.

test/src/module/command/backupTest.c
```diff
@@ -3238,6 +3240,63 @@ testRun(void)
             HRN_STORAGE_PATH_REMOVE(storageRepoWrite(), STORAGE_REPO_BACKUP "/20191027-181320F_20191028-220000I", .recurse = true);
         }
 
+        // -------------------------------------------------------------------------------------------------------------------------
+        TEST_TITLE("incr backup fails immediately on invalid page checksum when checksum-page-error=y");
```
dwsteele: Seems like we could do this test more cheaply by putting it at line 3207, keep `base/1/3` at line 3204 from the prior test, and remove it after the new test.

test/src/module/command/restoreTest.c
```diff
@@ -3246,6 +3247,224 @@ testRun(void)
...
+            " checksum dd71038f3463f511ee7403dbcbc87195302d891c\n"
+            "P00   INFO: write " TEST_PATH "/pg-checksum-warn/recovery.conf\n"
+            "P00 DETAIL: sync path '" TEST_PATH "/pg-checksum-warn'\n"
+            "P00 DETAIL: sync path '" TEST_PATH "/pg-checksum-warn/pg_tblspc'\n"
+            "P00   WARN: backup does not contain 'global/pg_control' -- cluster will not start\n"
+            "P00 DETAIL: sync path '" TEST_PATH "/pg-checksum-warn/global'\n"
+            "P00   INFO: restore size = [SIZE], file total = 2");
+
+        // -------------------------------------------------------------------------------------------------------------------------
+        TEST_TITLE("error on invalid page checksum when checksum-page-error=y");
```
dwsteele, line 3378: Do we need an entirely new test here? Can't we just piggyback on the warn test at 3251?

src/command/backup/process.c.inc
dwsteele: Yeah, given that it sets the error flag in backup.info we should keep alignment as a tracked warning and error when it is set. In practice I have never seen one of these errors in the field so I'm not too worried about it.

dwsteele (changes requested):
I made a few naming/formatting edits and left some new comments.

src/command/backup/process.c.inc
dAdAbird: done in 1a3561aaa

test/src/module/command/backupTest.c
dAdAbird: Changed in a407cf2c2

test/src/module/command/restoreTest.c
dAdAbird: Changed in 7dabd9e20

src/command/expire/expire.c
dAdAbird: I've added a check for expireTimeBasedBackup()  in 1ca79d9a9

dAdAbird:
Also added some more tests for `expire`. Before, it tested only the expireChecksumErrorCheck() function, which was not enough.

turnstep:
Looks good to me!

dwsteele (approved):
Looks all good now. We just need @hunleyd to check his final open item and approve.

dwsteele:
@turnstep Thank you!

hunleyd (approved):

dAdAbird:
@hunleyd, since you've approved the PR, can we resolve [this conversation](https://github.com/pgbackrest/pgbackrest/pull/2833#discussion_r3822391903) so the merge is unblocked?
cheers!

dwsteele:
@turnstep @hunleyd @dAdAbird While writing up the release note and the note about the behavioral change I realized the restore behavior may not be what users expect (though it was implemented as specified). The restore proceeds until the end and errors if there are invalid page checksums but otherwise leaves a valid backup.

First, an error should always indicate a failed command, but in this case the cluster is valid and can be started. At the very least the error should be thrown before pg_control is renamed so the cluster can't be started.

Second, do we really want to wait until the end of the restore to error? This potentially wastes a lot of the user's time. If we do error earlier, I'd prefer to just stop on the first error. They can use the info command to get a comprehensive list.

Or, am I just crazy? I know this seemed reasonable to me at some point but now I'm having qualms.

turnstep:
Yes, that makes sense, and is more inline with my original thinking. I think stopping on the first error is a good thing, assuming this defaults to off.

dwsteele:
> I think stopping on the first error is a good thing, assuming this defaults to off.

Off, you think? I was thinking the point here was to improve the behavior so users don't restore backups with invalid page checksums accidentally. We error by default now, we just do it all the way at the end -- and leave a cluster that can be started.

dAdAbird:
Before reading the spec, I assumed a bechaviour for the restore would be to fail on the first error. But then got into implementation weeds and forgot about my considerations. So yes, I agree; we should fail with an error as soon as we encounter the first checksum issue. And I vote to make it the default. I'd prefer a failed restore to a chance of issues after that in a running cluster

turnstep:
You are right: default on, and fail immediately, for a restore. Sorry for the confusion.

pgstef:
Hey,

Sorry to jump in. I haven't followed the code changes very closely, but I did try to summarize the behavior very briefly.

- Backup: 
  * checksum-page-error is off by default
  * when enabled, a failure in page checksum validation will terminate the backup
  * -> fail the backup fast
- Expire:
  * checksum-page-error is on by default
  * would error before any changes if the oldest retained backup would contain checksum errors
  * -> prevents to remove the oldest backup, hoping that this one doesn't contain the checksum errors (= an extra layer of protection)
- Restore:
  * checksum-page-error is on by default
  * if any file contains checksum errors, an error is emitted at the end, warning if the option is disabled
  * -> fail the restore late

From a user's perspective, I think it would make sense to fail the restore fast as well, imho.
Especially since `--delta --checksum-page-error=n` would still provide a way to explicitly force the restore to continue when the user knows there is a checksum issue. To me, that fits well with the whole point of adding this extra protection feature :-)

So, like @turnstep, I'm +1 on:

> default on, and fail immediately, for a restore

dwsteele:
Sounds like we are all on the same page here. Restore defaults to checksum-page-error=true, fails on first error.

@dAdAbird Please go ahead and do that -- sorry for the late change.

dAdAbird:
Changes in restore: cff24f5c0

Now Restore:
- checksum-page-error is on by default
- the command emits an error and stops as soon as it encounters a checksum error
- if checksum-page-error is off, a warning is emitted for every file with checksum error(s) along the way, and a summary warning at the end.

dwsteele:
Committed, thank you!

PR #2836, opened 2026-08-21
Derive the cipher key with SHA-256 at repository format 6.

dwsteele:
Format 6 derives encryption keys from the passphrase with SHA-256 rather than SHA-1. Older repositories keep SHA-1, since the migration does not rewrite the files they already encrypted. The digest is picked when a passphrase is generated rather than at each use, so it is stored with the passphrase in CipherSpec and defaults to SHA-256.

A passphrase outlives the format of the file it is stored in, so the format cannot decide how the key is derived. Info files at format 6 store cipher-digest beside cipher-pass, and a passphrase stored without one derives with SHA-1. stanza-upgrade writes back the digest the passphrase already had, so archives and backups from before the migration still load. Only a new passphrase gets SHA-256: the one for a stanza created at format 6, or for the manifest of a full backup written at it.

The info file has the same problem in reverse. Its format sits inside the content that the passphrase decrypts, so a reader would have to decrypt the file to find out how to decrypt it. Format 6 puts a plaintext header in front instead: PGBR, the three-digit format, and a spare byte for later. It goes where the OpenSSL Salted__ magic used to go and is the same eight bytes, so the salt still starts in the same place. A file that starts with the magic came from format 5, the only format older than the header. The header format is checked before anything is decrypted, then compared against the format found in the content, so a file stitched together from two others is rejected rather than half read.

The format constants move out of version.h into a new common/format, along with repoFormatValidate(), repoFormatDigest(), and CipherBlockFormat. That filter reads the header, builds a block cipher with the digest the format calls for, and streams the rest of the file through it. The block cipher no longer knows anything about repositories: it takes the digest from the spec and takes what sits in front of the content as a parameter.

Verify checksums info files as stored rather than as decrypted, since the header has to come off before decryption can start. A file and its copy are written from the same bytes, so comparing the two still works. In an encrypted repository created by a Perl version the file and its copy are separate encryptions, so their checksums differ and verify reports it at detail level. A log line is better than a separate code path for repositories with seven year old backups.


src/info/info.c
```diff
@@ -406,6 +429,17 @@ infoSave(Info *const this, IoWrite *const write, InfoSaveCallback *const callbac
         if (cipherSpecType(infoCipherSpec(this)) != cipherTypeNone)
         {
             callbackFunction(callbackData, STRDEF(INFO_SECTION_CIPHER), &data);
+
+            // Store the digest the pass derives with so that a pass outlives the format of the file it is stored in. A pass in a
+            // file written before this could be stored derives with SHA-1, which is what a reader assumes when it finds no digest.
+            if (infoFormat(this) >= REPOSITORY_FORMAT_6)
```
hunleyd, line 439: add ASSERT(cipherSpecDigest(sub)==hashTypeSha1) when infoFormat<6, since the digest silently isn't stored for format<6 and a reader always assumes SHA-1. Two real instances of this exact latent bug in the PR's own tests (test/src/module/info/infoArchiveTest.c:90 "create with cipher", test/src/module/command/restoreTest.c:2313); both fixed by adding an explicit .digest = hashTypeSha1. 

NOT exhaustively audited: ~40 total cipherSpecNewP() call sites exist; only the 2 that trip the new ASSERT via a clearly reachable infoSave()/manifestSave() path were checked.

src/common/crypto/cipherBlock.c
```diff
@@ -11,28 +12,80 @@ Block Cipher
...
+The header is not part of the file content. This filter adds it on encrypt and consumes it on decrypt, so nothing on either side
+sees anything but the content.
+***********************************************************************************************************************************/
+#define CIPHER_BLOCK_HEADER_MAGIC                                   "PGBR"
+#define CIPHER_BLOCK_HEADER_MAGIC_SIZE                              (sizeof(CIPHER_BLOCK_HEADER_MAGIC) - 1)
+#define CIPHER_BLOCK_HEADER_RESERVED                                '_'
+#define CIPHER_BLOCK_HEADER_FORMAT_SIZE                             3
+
 // Total length of cipher header
 #define CIPHER_BLOCK_HEADER_SIZE                                    (CIPHER_BLOCK_MAGIC_SIZE + PKCS5_SALT_LEN)
```
hunleyd: (also line 199) add a static_assert tying CIPHER_BLOCK_HEADER_MAGIC_SIZE + CIPHER_BLOCK_HEADER_FORMAT_SIZE + 1 == CIPHER_BLOCK_MAGIC_SIZE (the whole "header replaces magic" design depends on this coincidence, which is unenforced), and an ASSERT(format <= 999) before the 3-digit header write (else you get silent truncation for format >= 1000). Pure hardening, no behavior change for valid inputs.

test/src/harness/info.c
```diff
@@ -116,6 +120,83 @@ harnessInfoChecksum(const String *const info)
...
+            bufCat(encrypted, BUFSTR(strNewFmt("PGBR%03u_", param.format)));
+
+        IoWrite *const write = ioBufferWriteNew(encrypted);
+        ioFilterGroupAdd(
+            ioWriteFilterGroup(write),
+            cipherBlockNewP(
+                cipherModeEncrypt,
+                cipherSpecNewP(
+                    cipherSpecType(param.cipherSpec), cipherSpecPass(param.cipherSpec),
+                    .digest = param.format >= REPOSITORY_FORMAT_6 ? hashTypeSha256 : hashTypeSha1),
```
hunleyd: hrnInfoPut() should call repoFormatDigest(param.format) instead of hand-deriving the same threshold inline; the two would silently diverge if a format 7 ever adds a third digest.

test/src/harness/info.c
hunleyd, line 1: test/src/harness/info.{c,h}, test/src/module/info/{infoTest.c,manifestTest.c}, test/src/module/command/restoreTest.c — you can merge testInfoEncrypt() (infoTest.c-local) into harnessInfoEncrypt() by giving the latter a format param;

doc/xml/reference.xml
```diff
@@ -2929,7 +2929,7 @@
 
                             <allow-list caption="formats">
                                 <allow-item id="5">Format supported by all versions since <proper>1.00</proper></allow-item>
-                                <allow-item id="6">Adds no features yet and is reserved for features in development</allow-item>
+                                <allow-item id="6">Derives repository encryption keys with <id>SHA-256</id> rather than <id>SHA-1</id></allow-item>
```
hunleyd: the format-6 allow-list text sas encryption keys derive with SHA-256 unconditionally; traced cipherSpecGen()'s 3 call sites and confirmed a migrated stanza's existing archive/manifest sub-passes keep SHA-1 until the next full backup or a new stanza, so maybe reword

src/common/crypto/cipherBlock.h
hunleyd, line 1: raw/header/format is a 3-way parameter overload on cipherBlockNewP()/cipherBlockFilterGroupAddP() whose legal combinations differ by encrypt vs decrypt mode, enforced only by 2 runtime ASSERTs. Proposed fix: single CipherBlockHeader enum (magic/none/format)

src/command/verify/verify.c
hunleyd, line 1: verifyFileLoad()'s decrypt bool param is unused on one of its 2 callers, and the checksum-over-ciphertext behavior now depends on an undocumented invariant (main and copy must be byte-identical, not just decrypt-equal) that holds today but has no assertion protecting it. Split the function or just document/assert the invariant?

hunleyd (changes requested):

src/info/info.c
dwsteele: There were two additional tests that could trip the assert. All added in 2c1e1d1429573.

src/common/crypto/cipherBlock.c
dwsteele: Love me some static asserts! Added in 725b3e0cf78905 along with the runtime assert.

test/src/harness/info.c
dwsteele: Done in d4c322806ea6.

test/src/harness/info.c
dwsteele: Done in fc275ed290a8c7.

doc/xml/reference.xml
dwsteele: Clarified in 84586552548.

src/common/crypto/cipherBlock.h
dwsteele: Done as an enum in b85e18a9c683.

src/command/verify/verify.c
dwsteele: This is really two special cases sharing one function so I removed the function and implemented each special case inline in 7d981527c6.

test/src/harness/info.c
```diff
@@ -116,6 +120,84 @@ harnessInfoChecksum(const String *const info)
...
+
+    ASSERT(info != NULL);
+
+    // Default to the format a new repository is created at
+    if (param.format == 0)
+        param.format = REPOSITORY_FORMAT_DEFAULT;
+
+    const Buffer *content = harnessInfoChecksumFormat(param.format, STR(info));
+
+    // Encrypt the way the format stores the file. A file that contains a header gets it in place of the magic the cipher writes,
```
hunleyd, line 146:
```diff
  -    // Encrypt the way the format stores the file. A file that contains a header gets it in place of the magic the cipher writes,
  -    // and the pass derives with the digest the format calls for.
  +    // Encrypt the way the format stores the file, the same path production writers use, so a header (or lack of one) matches
  +    // exactly what harnessInfoEncrypt() below and the real writers produce.
       if (param.cipherSpec != NULL && cipherSpecType(param.cipherSpec) != cipherTypeNone)
       {
  -        const bool header = param.header && param.format >= REPOSITORY_FORMAT_6;
           Buffer *const encrypted = bufNew(0);
  -
  -        if (header)
  -            bufCat(encrypted, BUFSTR(strNewFmt("PGBR%03u_", param.format)));
  -
           IoWrite *const write = ioBufferWriteNew(encrypted);
           ioFilterGroupAdd(
               ioWriteFilterGroup(write),
  -            cipherBlockNewP(
  -                cipherModeEncrypt,
  -                cipherSpecNewP(
  -                    cipherSpecType(param.cipherSpec), cipherSpecPass(param.cipherSpec),
  -                    .digest = repoFormatDigest(param.format)),
  -                .header = header ? cipherBlockHeaderNone : cipherBlockHeaderMagic));
  +            cipherBlockNewP(cipherModeEncrypt, param.cipherSpec, .format = param.header ? param.format : 0));```

hunleyd (changes requested):
Cipher filter layering inversion (src/common/crypto/cipherBlock.c now #includes common/format.h, calls repoFormatValidate()/repoFormatDigest(), references REPOSITORY_FORMAT_5/6 directly). A generic block-cipher filter now hard-codes repository format policy, and the module dependency is close to circular (common/format.h pulls common/crypto/common.h for HashType). Suggested fix: push validation/digest-mapping one layer up into the info layer, let the filter just report the raw header value

Mixed-version fleet gets a misleading error (src/command/stanza/upgrade.c:129): an older pgbackrest binary reading a migrated repo gets cipher header invalid / HINT: is or was the repo encrypted? instead of the clearer "repository format 6 requires a newer version" message it gets for an unencrypted repo doing the same migration.

test/src/harness/info.c
dwsteele: > an older pgbackrest binary reading a migrated repo gets cipher header invalid / HINT: is or was the repo encrypted?

I don't think there is much to be done about this other than document it as done in a208e01555. Older versions just won't recognize the new header format or versioning.

dwsteele: > Suggested fix: push validation/digest-mapping one layer up into the info layer, let the filter just report the raw header value

This won't work well, especially when we start storing the key id in the header (needed for retrieving WAL keys).

This has been bugging me as well but I knew the solution would be complex so I was reluctant to do it. I've now done the "ideal" solution in 959d051c13. This does add some code, but should be easier to maintain in the future. It adds an extra buffer copy for the encrypted files with format headers, but currently that only applies to info files so not that big a deal. However, it will also be required for WAL files when key rotation is implemented since they will also need format headers.

In the grand scheme of things, though, buffer copies are pretty cheap.

hunleyd (commented):
Left as a plain comment on this fork PR, not as a requested review. These points cover cases the earlier review rounds on this PR did not raise.

### 1. Use-after-free risk in `cipherBlockFormat.c`

`cipherBlockFormatContentWriteNew()` (around line 159) creates `this->content` and `this->contentWrite`. The code does not wrap this creation in `MEM_CONTEXT_OBJ_BEGIN(this)`. `ioFilterProcessInOut()` does not switch the memory context, so these buffers exist in whatever context is active when the filter runs its process callback.

`iniValueNext()` in `ini.c` parses every info file. It runs its read loop inside a temporary context. That context resets periodically, through `MEM_CONTEXT_TEMP_RESET(1000)` at `ini.c:196`.

An encrypted `archive.info` or `backup.info` file larger than `buffer-size` (16KiB minimum) needs more than one call to `process()` to finish decryption. If a context reset happens between those calls, `this->content` points to freed memory. A later read through `bufUsed()` or `bufPtrConst()` then touches that freed memory.

This risk is not limited to format 6. `infoPgNewLoad()` always passes `.header = true`, so every encrypted info file goes through this filter, regardless of its format.

`blockIncr.c:136-143` shows the fix this codebase already uses for the same pattern. It wraps a lazy `ioBufferWriteNew()` call in `MEM_CONTEXT_OBJ_BEGIN(this)` and `MEM_CONTEXT_OBJ_END()`. The same wrap here removes the risk.

### 2. `verify` now checksums encrypted files differently

In `verifyInfoFile()` (`verify.c`, around line 190), the code adds the SHA-1 hash filter before `infoNewLoad()` adds its decrypt filter. Filter groups run in the order you add them, so the checksum now covers the encrypted bytes, not the decrypted content. Before this PR, `verifyFileLoad()` decrypted the file first, then hashed it.

Separately, the `else ioReadDrain(infoRead)` branch, used for `.copy` files when `keepFile` is false, never calls `infoNewLoad()`. This means the code never tries to decrypt the copy file at all. A copy file that fails to decrypt used to produce a clear decrypt error. Now it only shows up as an ordinary checksum mismatch, with no distinct cause.

This causes no harm for repositories that C pgbackrest wrote, because the main file and its copy hold the same encrypted bytes, written twice. A repository whose main file and copy were encrypted separately, for example an old Perl-era repository, gets a false "copy does not match main" error under the new code.

This is a real change in what `verify` checks. The PR description does not mention it.

### 3. Digest lookup duplicated a third time

`cipherBlock.c` adds a third copy of the digest-by-name lookup (`strIdToZ` into a stack buffer, then `EVP_get_digestbyname`). `hash.c` already has this same lookup twice, in `cryptoHashNew` and `cryptoHmacOne`. Move this lookup into one shared function in `common/crypto/common.c`.

### 4. Debug-only assert on a critical invariant

The assert in `info.c` (around line 1496) checks that a passphrase for format less than 6 always uses SHA-1. This assert guards against one failure that is silent, and once it happens, permanent: a stored passphrase that no reader can derive again. `ASSERT` compiles out of release builds, so this check does not run in production.

No current code path triggers this failure. Given the consequence, `CHECK` keeps the same protection in production builds too.

### 5. A cipher spec with a default digest that is wrong for existing repositories

`cfgCipherSpecMainIdx()` in `config.c` (around line 1120) is the only production `CipherSpec` built with no explicit `.digest` value. It takes the new default from `cipherSpecNew()`, which is SHA-256. SHA-256 is the wrong digest for every repository before format 6.

Today this causes no harm. Every real caller, through `cipherBlockFormatFilterGroupReadAdd()` or `cipherBlockFormatFilterGroupWriteAdd()`, rebuilds the spec from `repoFormatDigest(format)` before use. But the silent default in `cipherSpecNew()` is a hazard for a future caller that does not rebuild the spec. Make the digest a required parameter with no default, or set it to SHA-1 explicitly here.

### 6. A possible simplification, not a bug

`infoNewLoad()` takes a new `InfoNewLoadParam.header` flag. The three callers, `infoArchiveLoadFileCallback`, `infoBackupLoadFileCallback`, and `manifestLoadFileCallback`, can instead call `cipherBlockFormatFilterGroupReadAdd()` directly, at the same call site where they called `cipherBlockFilterGroupAdd()` before this PR. That removes the flag.

This is a suggestion, not a request for a change. An open review thread on this PR already gives a forward-looking reason, storage of a key ID for WAL keys, that may make the current shape intentional.

### 7. No size limit on a buffered file

The read filter in `cipherBlockFormat.c` holds an entire decrypted info file in memory, with no size limit, before it passes any content on. Info files stay small today, but nothing in the code enforces that. The exported function, `cipherBlockFormatFilterGroupReadAdd()`, takes no parameter to set a limit.

hunleyd (commented):
A status update on my earlier review round on this PR. I checked the four commits named in the replies to my prior comments, and each one matches the stated fix:

- `2c1e1d1429573` adds the assert that requires SHA-1 for format 5 saves.
- `fc275ed290a8c7` merges `testInfoEncrypt()` into `harnessInfoEncryptP()`.
- `b85e18a9c683` collapses the raw/header params in `cipherBlockNewP()` to a header enum.
- `7d981527c6` splits `verifyFileLoad()` into two special cases.

All four of my prior concerns are resolved. The one remaining open thread stays open on my side, since I do not have a further response to your stated reasons there.

dwsteele:
@hunleyd,

> 1. Use-after-free risk in cipherBlockFormat.c

Fixed in 6672549e32d.

> 2. verify now checksums encrypted files differently

Added a decrypt filter for copies so they get a valid error in 5eef5970f3. On repos with 7+ year old backups created with Perl this code results in a warning because under Perl the checksums could differ, but I think this is acceptable. It is a warn not a hard failure, and the main copy remains usable.

> 3. Digest lookup duplicated a third time

Deduped in 665ac2f914.

> 4. Debug-only assert on a critical invariant

Probably overkill given our coverage, but can't hurt. Done in 0dc57496f1da.

> 5. A cipher spec with a default digest that is wrong for existing repositories

Use SHA1 as the default here (694c4a7a56e) since SHA256 repos will rewrite the CipherSpec.

> 6. A possible simplification, not a bug

Looks good, and it is simpler so done in cec269f0a. Could be we'd need to bring back var params in the future but I can't see an immediate need for it.

> 7. No size limit on a buffered file

Yeah, I thought this could pass because info files are small (and it mirrors how we handle subfilters in BlockIncr) but we will need to deal with larger files in the future so may as well fix this now.

Reworked in 3abe001b282 to stream using the internal API, which avoids additional buffering.

I made a few more misc changes based on my own review in d1237c5fcd15ac. I need to give this a additional full once over but need to give my brain a break for now.

test/src/harness/info.c
dwsteele: I believe I've addressed all the new issues here, up to you to decide.

hunleyd (approved):

test/src/harness/info.c
hunleyd: yes, i'm happy with the explanations and edits. thx!

dwsteele:
> yes, i'm happy with the explanations and edits. thx!

Great! Now I can start on the actual change we are trying to make. :) I didn't want to get too far ahead without getting this all settled.

PR #2837, opened 2026-08-23
Add human readable table output with available backups

deric:
Implement a table-like output for listing backups. Rather than rewriting the `info` command which people are already used to I opted for adding a new command.

usage:
```bash
$ pgbackrest list
STANZA 'psql01'
--------------------------------------------------------------------------------------------------------------------------------------------------------------
 Repo   Version  ID                Recovery Time             Type  Cipher  TLI  Time     DB size   Backup size  Zratio  Start LSN       Stop LSN        Status
--------------------------------------------------------------------------------------------------------------------------------------------------------------
 repo1  13       20260821-010003F  2026-08-21 20:20:37+0000  full  none    4/0  19h:20m  2001.1GB      507.4GB    3.94  552A5/562EAFC8  5534B/288825D0  OK
 repo1  -        -                 -                         -     none    -          -  2087.7GB            -       -  -               -               85.50%
```


hunleyd:
i'll defer to @dwsteele , but i think he's gonna say "fold this into `info` as a third `--output=table` option". i know i would :)

also, this need an open issue to land against.

pgstef:
> I think he's gonna say "fold this into `info` as a third `--output=table` option". i know i would :)

My thoughts too tbh. The question here would be what information this new command would add compared to what's available in the info json output. Then maybe it might be added in the json and be useful for more users.

I'm not really convinced by the table output format, but if that can be interesting to others, why not :shrug:

dwsteele:
I think my concern here is that this patch would likely spawn a bunch of natural followups: what fields to list, data/size formatting, ordering, etc. The idea of the JSON output is that people can adapt it to their needs and we don't need to support custom output types.

For instance, I asked Claude to produce a jq script that would mimic what this patch does and it came up with the attached (should be list.jq but Github won't accept that extension).

[list.txt](https://github.com/user-attachments/files/31803608/list.txt)

So, my question is -- is this something that should be built-in or is it better to let users script what they need?

dwsteele:
I would be open to adding a simplified version of this script to https://pgbackrest.org/user-guide.html#monitor/jq as an example.

deric:
@dwsteele Sure, you're right, however the `jq` script is rather complex - I've already tried writing something like this though my `jq` capabilities are very limited. Anyway, I understand that maintenance of formatting code is not very interesting, but it can heavily simplify daily operations.

I guess I could distribute a bunch of formatting scripts along with `pgbackrest`, but why not include it within `pgbackrest`?  

The current `pgbackrest info` output is ok if you have a few backups, but with multiple stanzas and many backups it fills many screens while it could simply fit into one screen.

I was thinking about adding a `--output / -o` flag that could work with columns:
```
$ pgbackrest list -o id,type,size
STANZA 'psql01'
------------------------------------------------------------
 ID                                  Type       Size
-----------------------------------------------------------
20260821-010003F   full           2001.1GB    
```
something like `lsblk` [does](https://man7.org/linux/man-pages/man8/lsblk.8.html) - it provides JSON output `lsblk -J` but also advanced formatting capabilities which makes it great tool for daily sysadmin usage.

dwsteele:
Hmm, well, I'm on the fence about this myself since it seems like it would require a lot of tweaks to make it flexible enough to be generally useful. If we can get a consensus and somebody else is willing to take on review and maintenance then I won't argue. Personally, I don't have the time to take it on right now.

One area where I definitely agree with Doug and Stefan: this would need to be an output type to the info command. I think a separate command introduces too much code duplication.

deric:
@dwsteele  Ok, I'll try to rewrite it into `info` command.

If the `jq` script is a one-liner, it's a good workaround. But if you need 120 lines of `jq`, the product is clearly missing a feature.

dwsteele:
> But if you need 120 lines of jq, the product is clearly missing a feature.

I'm not sure I agree -- there are plenty of monitoring and other solutions based on the JSON and I don't think it would be appropriate to build them all into pgbackrest. This may be the exception but we'll see what people think.

deric:
Closing in favor of #2850

PR #2839, opened 2026-08-24
feat: implement web-id key type for GCS storage

### LINKED ISSUE #2838: proposal: implement GCS web-id key type
## Problem

Outside of GKE, pgBackRest has no keyless way to authenticate to GCS using explicit pod identities,
On managed or self-managed Kubernetes clusters without Google's pod-aware metadata server, the
existing key type mechanisms fall short:

- `token` is meant for local testing. Once the supplied bearer token expires,
  subsequent GCS requests fail because the driver cannot renew it
- `service` requires using a long-lived service account private-key file available to each
  pgBackRest pod, creating static secrets and a rotation burden.
- `auto` relies on the GCP metadata endpoint. GKE with workload identity provides this mechanism,
  but on plain GCE the metadata server returns the node VM's service account, so every pod on
  a node shares one identity. Preserving per-pod privilege isolation would require a pod-aware
  metadata proxy or emulator.

## Precedent in pgBackRest

This problem was solved in 2.36.0 for s3: `repo-s3-key-type=web-id`, which reads the projected
Kubernetes ServiceAccount token from a file and calls `AssumeRoleWithWebIdentity`, refreshing
credentials as they expire, with the token file re-read on each renewal since 2.54.0, so rotated
projected tokens keep long operations alive.
Google's equivalent flow is Workload Identity Federation (WIF).

## Proposal

Add `repo-gcs-key-type=web-id`:

1. Read a JWT from a configurable file that Kubernetes projects into the pod and rotates. Its
   audience must be accepted by the WIF provider.
2. POST an `application/x-www-form-urlencoded` token-exchange request to
   `https://sts.googleapis.com/v1/token`, with `grant_type=urn:ietf:params:oauth:grant-type:token-exchange`,
   the configured provider resource name as `audience`, 
   `requested_token_type=urn:ietf:params:oauth:token-type:access_token`, the file
   contents as `subject_token`, `subject_token_type=urn:ietf:params:oauth:token-type:jwt`,
   and the same read-only/read-write Cloud Storage scope selection the `service` type
   already uses. The response carries `access_token`/`token_type`/`expires_in`, the exact
   fields the existing GCS token parser consumes.
3. Cache the token; before each GCS request, renew when the driver's adjusted expiration
   threshold is reached (existing renewal machinery), re-reading the token file on each
   renewal as S3 `web-id` does. The subject token must be deleted from logs/errors, as S3
   implementation does with its web-identity token.

Three new options will be added. `repo-gcs-web-id-token-file` is the path to the projected JWT,
`repo-gcs-web-id-audience` is the workload identity provider's full resource name sent as the STS
audience, both are required when the new key type is used and neither carries a secret.
`repo-gcs-sts-host` defaults to `sts.googleapis.com` and allows an alternate STS host. Explicit options
rather than Google's standard external-account credential JSON are intentional: parsing it would
mean supporting impersonation, multiple token-source types, and more. This proposal focuses on
direct access instead, which is the preferred option in GCP docs, service account impersonation will not be
supported for now.

pgBackRest authenticates as the federated principal, which must hold the Cloud Storage
permissions.

## Why upstream

With this in place, non-GKE Kubernetes clusters can use GCS authentication
without static keys or metadata server emulators.
The pod mounts a projected Kubernetes ServiceAccount token, and the cluster's OIDC issuer is
registered as a Workload Identity Federation provider. GCP validates tokens signed by the cluster
and maps their claims to federated principals. IAM bindings can then grant bucket access to a
specific Kubernetes ServiceAccount or a broader set of identities, such as all ServiceAccounts
in a namespace.

```text
[Projected Kubernetes JWT]
             |
             v
[POST to GCP STS]
             |
             |  GCP federated access token
             |  (principal://iam.googleapis.com/...)
             v
[Call the GCS API directly]
```

The implementation can reuse the existing HTTP, JSON, token-parsing, expiry, and file-reading
facilities; the main new code is the STS request path and configuration wiring, with the
S3 `web-id` code as the template.

adolsalamanca:
Adds repo-gcs-key-type=web-id: authenticate to GCS via Workload Identity Federation using a Google external-account credential file passed in repo-gcs-key, which is the file `gcloud iam workload-identity-pools create-cred-config` generates, consumed in the same way service already consumes Google's service-key JSON.

The system already parses audience, token_url, and credential_source.file from the credential file, though service-account impersonation and non-text formats are unsupported. 
It exchanges the referenced JWT at the STS token endpoint, and re-reads the token file on each renewal so rotated Kubernetes projected tokens keep long operations alive. This mirrors s3 behavior and gives keyless per-pod auth on clusters without a pod-aware metadata endpoint.

Note for existing key types: the shared auth-token parser now reports non-JSON error responses via the HTTP error instead of a JSON parse failure, and type-guards the error field, reporting nested error objects (e.g. STS API disabled) with the full response, which also changes error output for service/auto in those cases.

Validated on GKE end to end: STS exchange, WAL archiving, full/diff backups, restore, negative paths, and a 2h+ backup spanning multiple STS renewals and token-file rotations.

Solves this [issue](https://github.com/pgbackrest/pgbackrest/issues/2838)


src/storage/gcs/storage.c
hunleyd: STS response parsed as JSON with no status check
```
  -        const KeyValue *const kvResponse = varKv(jsonToVar(strNewBuf(httpResponseContent(httpRequestResponse(request, true)))));
  +        HttpResponse *const httpResponse = httpRequestResponse(request, true);
  +
  +        // Error if the request was not successful, e.g. the STS host is misconfigured. Without this the error would be a
  +        // confusing report that the response is not valid JSON rather than the actual error returned by the server.
  +        if (!httpResponseCodeOk(httpResponse))
  +            httpRequestError(request, httpResponse);
  +
  +        const KeyValue *const kvResponse = varKv(jsonToVar(strNewBuf(httpResponseContent(httpResponse))));
```
A misconfigured repo-gcs-sts-host (typo, proxy, wrong port) that returns an HTML error page currently surfaces as a confusing JsonFormatError instead of the real HTTP status/body. S3's web-id path already guards this exact case (storage/s3/storage.c:413-416)

test/src/module/storage/gcsTest.c
hunleyd, line 1: bespoke testWebIdRequest helper duplicates the existing testRequest helper 20 lines below it. Delete the 16-line helper, replace all 7 call sites (testWebIdRequest(sts, X)) with the canonical hrnServerScriptAccept(sts);
  testRequestP(sts, HTTP_VERB_POST, .path = "/v1/token", .noAuth = true, .contentType = "application/x-www-form-urlencoded", .content
  = X);
testRequest already supports .path/.noAuth/.contentType/.content and emits headers in the same order the bespoke helper hand-wrote.
```diff
diff --git i/test/src/module/storage/gcsTest.c w/test/src/module/storage/gcsTest.c
index ff2bb119e..d48e099c1 100644
--- i/test/src/module/storage/gcsTest.c
+++ w/test/src/module/storage/gcsTest.c
@@ -89,24 +89,6 @@ testAuthRequest(const Storage *const storage)
     return result;
 }
 
-static void
-testWebIdRequest(IoWrite *const write, const char *const content)
-{
-    hrnServerScriptAccept(write);
-    hrnServerScriptExpectZ(
-        write,
-        strZ(
-            strNewFmt(
-                "POST /v1/token HTTP/1.1\r\n"
-                "user-agent:" PROJECT_NAME "/" PROJECT_VERSION "\r\n"
-                "content-length:%zu\r\n"
-                "content-type:application/x-www-form-urlencoded\r\n"
-                "host:%s\r\n"
-                "\r\n"
-                "%s",
-                strlen(content), strZ(hrnServerHost()), content)));
-}
-
 /***********************************************************************************************************************************
 Helper to build test requests
 ***********************************************************************************************************************************/
@@ -441,7 +423,10 @@ testRun(void)
                     "subject_token_type=urn%3Aietf%3Aparams%3Aoauth%3Atoken-type%3Ajwt&"
                     "scope=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fdevstorage.read_only";
 
-                testWebIdRequest(sts, content);
+                hrnServerScriptAccept(sts);
+                testRequestP(
+                    sts, HTTP_VERB_POST, .path = "/v1/token", .noAuth = true, .contentType = "application/x-www-form-urlencoded",
+                    .content = content);
                 testResponseP(sts, .content = "{\"access_token\":\"federated\",\"token_type\":\"Bearer\",\"expires_in\":1}");
                 hrnServerScriptClose(sts);
 
@@ -464,7 +449,10 @@ testRun(void)
                     "subject_token=rotated%20jwt&"
                     "subject_token_type=urn%3Aietf%3Aparams%3Aoauth%3Atoken-type%3Ajwt&"
                     "scope=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fdevstorage.read_only";
-                testWebIdRequest(sts, rotatedContent);
+                hrnServerScriptAccept(sts);
+                testRequestP(
+                    sts, HTTP_VERB_POST, .path = "/v1/token", .noAuth = true, .contentType = "application/x-www-form-urlencoded",
+                    .content = rotatedContent);
                 testResponseP(sts, .content = "{\"access_token\":\"renewed\",\"token_type\":\"Bearer\",\"expires_in\":3600}");
                 hrnServerScriptClose(sts);
 
@@ -476,7 +464,10 @@ testRun(void)
 
                 // Non-200 STS errors use the standard GCS auth error fields
                 ((StorageGcs *)storageDriver(storage))->tokenTimeExpire = 0;
-                testWebIdRequest(sts, rotatedContent);
+                hrnServerScriptAccept(sts);
+                testRequestP(
+                    sts, HTTP_VERB_POST, .path = "/v1/token", .noAuth = true, .contentType = "application/x-www-form-urlencoded",
+                    .content = rotatedContent);
                 testResponseP(
                     sts, .code = 400,
                     .content = "{\"error\":\"invalid_grant\",\"error_description\":\"subject token rejected\"}");
@@ -486,14 +477,20 @@ testRun(void)
                     "unable to get authentication token: [invalid_grant] subject token rejected");
 
                 // Missing required token response fields fail cleanly
-                testWebIdRequest(sts, rotatedContent);
+                hrnServerScriptAccept(sts);
+                testRequestP(
+                    sts, HTTP_VERB_POST, .path = "/v1/token", .noAuth = true, .contentType = "application/x-www-form-urlencoded",
+                    .content = rotatedContent);
                 testResponseP(sts, .content = "{\"token_type\":\"Bearer\",\"expires_in\":3600}");
                 hrnServerScriptClose(sts);
                 TEST_ERROR(
                     storageGetP(storageNewReadP(storage, STRDEF("missing4"), .ignoreMissing = true)), FormatError,
                     "access token missing");
 
-                testWebIdRequest(sts, rotatedContent);
+                hrnServerScriptAccept(sts);
+                testRequestP(
+                    sts, HTTP_VERB_POST, .path = "/v1/token", .noAuth = true, .contentType = "application/x-www-form-urlencoded",
+                    .content = rotatedContent);
                 testResponseP(sts, .content = "{\"access_token\":\"token\",\"token_type\":\"Bearer\"}");
                 hrnServerScriptClose(sts);
                 TEST_ERROR(
@@ -501,7 +498,10 @@ testRun(void)
                     "expiry missing");
 
                 // The subject token is redacted even if the STS error echoes it
-                testWebIdRequest(sts, rotatedContent);
+                hrnServerScriptAccept(sts);
+                testRequestP(
+                    sts, HTTP_VERB_POST, .path = "/v1/token", .noAuth = true, .contentType = "application/x-www-form-urlencoded",
+                    .content = rotatedContent);
                 testResponseP(
                     sts, .code = 400,
                     .content =
@@ -536,7 +536,10 @@ testRun(void)
                     "subject_token_type=urn%3Aietf%3Aparams%3Aoauth%3Atoken-type%3Ajwt&"
                     "scope=https%3A%2F%2Fwww.googleapis.com%2Fauth%2Fdevstorage.read_write";
 
-                testWebIdRequest(sts, writeContent);
+                hrnServerScriptAccept(sts);
+                testRequestP(
+                    sts, HTTP_VERB_POST, .path = "/v1/token", .noAuth = true, .contentType = "application/x-www-form-urlencoded",
+                    .content = writeContent);
                 testResponseP(sts, .content = "{\"access_token\":\"writer\",\"token_type\":\"Bearer\",\"expires_in\":3600}");
                 hrnServerScriptClose(sts);```

**Review by hunleyd — CHANGES_REQUESTED** [review]
test/src/harness/host.c:764 - build-breaking
```
  -                            STRDEF(HRN_HOST_GCS_KEY), 4 * 1024 * 1024, NULL,
  +                            STRDEF(HRN_HOST_GCS_KEY), NULL, NULL, NULL, 4 * 1024 * 1024, NULL,
```
storageGcsNew() grew from 17 to 20 params (added webIdTokenFile, webIdAudience, stsHost). The diff fixed both gcsTest.c call sites but missed this integration-test-harness call, which still passes the old 17-arg form; a hard "too few arguments" compile error under this repo's -Werror build. 

storageGcsAuthWebId's subject-token redaction is a strstr(errorMessage(), strZ(subjectToken)) exact-byte match against the already-formatted error text. Any provider that truncates, prefixes, or re-encodes the echoed token (the common case, not the literal-echo case this was tested against) defeats it silently; the raw JWT fragment then propagates unredacted through RETHROW().
Separately, the branch that does redact uses THROWP() instead of RETHROW(), discarding the original throw site/stack trace — so a caught-and-redacted error is diagnostically worse than an unredacted one. This driver already has headerRedactList/queryRedactList fields wired on StorageGcs (storage.c:93-94), the same S3-style mechanism (httpQueryNewP(.redactList = ...)) that lets S3's web-id path (storage/s3/storage.c:401-408) redact WebIdentityToken at the point it enters a loggable structure, with no partial-match failure mode. Moving redaction into the shared storageGcsAuthToken() via that existing mechanism would delete the whole TRY/CATCH_ANY/replace-loop, fix the partial-echo gap, and fix the stack-trace loss in one pass.

**Comment by adolsalamanca** [comment]
Thanks for your review @hunleyd, 🙏🏽 I've addressed your comments, please let me know if I missed something.

**Review by hunleyd — CHANGES_REQUESTED** [review]
1 — biggest one: GCP already has a standard file format for exactly this setup — an "external_account" credential file (what gcloud iam workload-identity-pools create-cred-config generates, normally pointed to by GOOGLE_APPLICATION_CREDENTIALS). That JSON file already contains audience, token_url, and credential_source.file — the exact same 3 values this PR asks the operator to retype into 3 separate pgBackRest options. If the WIF pool/provider ever rotates, the operator has to remember to update pgBackRest's 3 scattered options instead of just re-pointing at the regenerated file.

2 — real bug: varStr() on the STS error field has no type guard in release builds (ASSERT compiles to nothing outside DEBUG). Google's STS endpoint returns a nested {"error":{...}} object on exactly the two most likely first-run misconfigs (STS API disabled, missing WIF pool) — I confirmed VariantString/VariantKeyValue share identical struct layout, so that's a type-confused pointer deref in a release build, not a crash in test builds only. Existing auto/service paths share the same unguarded call but never hit it because oauth2.googleapis.com only returns flat errors — this PR is what makes the vulnerable branch reachable.

3 — stsHost gets string-concatenated with /v1/token before being parsed as a URL. A trailing slash or a pasted full URL silently produces a wrong path or a confusing FormatError instead of a clean config error. S3 parses the host first, keeps path separate.

4 - The new httpResponseCodeOk guard landed in the shared storageGcsAuthToken(), so it also changes error behavior for the existing service/auto key types — undisclosed in the PR body.

**Comment by adolsalamanca** [comment]
Thanks again for the review @hunleyd, about your comments:

1: adopted. web-id now takes the external-account credential file via repo-gcs-key, parsed the same way service parses its key JSON. Now the token_url arrives as a full URL, so number 3 is gone too.

2: fixed in the shared parser, string errors keep the description format, anything else reports via httpRequestError with the full response. Tests cover nested objects and non-string descriptions.

On the redaction: I removed the scrub given unlike S3, the GCS subject token only ever exists in the request content, which httpRequestToLog/httpRequestError never emit. That also fixes the partial-match and stack-trace issues you flagged. Apart from that, repo-gcs-key being secure means the credential path itself renders as <redacted> in command logs.

4: disclosed in the updated PR description. The harness call in host.c is fixed too, thanks for the catch!

File: src/storage/gcs/storage.c [hunk 1]
```diff
@@ -1293,6 +1386,56 @@ storageGcsNew(
...
+                CHECK(
+                    FormatError, kvGet(kvKey, GCS_JSON_SA_IMPERSONATION_URL_VAR) == NULL,
+                    "service account impersonation is not supported");
+
+                const String *const audience = varStr(kvGet(kvKey, GCS_JSON_AUDIENCE_VAR));
+                CHECK(FormatError, audience != NULL, "audience missing");
+                const String *const tokenUrl = varStr(kvGet(kvKey, GCS_JSON_TOKEN_URL_VAR));
+                CHECK(FormatError, tokenUrl != NULL, "token url missing");
+
+                const KeyValue *const credentialSource = varKv(kvGet(kvKey, GCS_JSON_CREDENTIAL_SOURCE_VAR));
```
hunleyd: here and line 1422:
varKv() is called on credential_source and credential_source.format with only a NULL check, no type check. varKv()'s own guard is ASSERT, which compiles to nothing outside debug builds. A credential file with "credential_source":"some-string" (a plausible hand-edit) causes undefined behavior/crash in a release build instead of a clean FormatError.
  
Notably: this exact hazard class is the one the PR's own storageGcsAuthToken fix already guards against elsewhere in the same diff (varType(error) == varTypeString) — just not applied here.
```diff
  -                const KeyValue *const credentialSource = varKv(kvGet(kvKey, GCS_JSON_CREDENTIAL_SOURCE_VAR));
  -                CHECK(FormatError, credentialSource != NULL, "credential source missing");
  +                const Variant *const credentialSourceVar = kvGet(kvKey, GCS_JSON_CREDENTIAL_SOURCE_VAR);
  +                CHECK(FormatError, credentialSourceVar != NULL && varType(credentialSourceVar) == varTypeKeyValue,
  +                    "credential source missing");
  +                const KeyValue *const credentialSource = varKv(credentialSourceVar);
                   const String *const tokenFile = varStr(kvGet(credentialSource, GCS_JSON_FILE_VAR));
                   CHECK(FormatError, tokenFile != NULL, "token file missing");

                   if (format != NULL)
                   {
  +                    CHECK(FormatError, varType(format) == varTypeKeyValue, "credential source format is not supported");
  +
                       const String *const formatType = varStr(kvGet(varKv(format), GCS_JSON_TYPE_VAR));```

hunleyd (changes requested):
- storage.c:453 — key-type dispatch is if/else not switch; both storageGcsNew's own constructor and the sibling S3 driver use switch for the same enum, so a future key type silently falls into the wrong auth path instead of failing to build

- storage.c:353 — STS body hand-built via strNewFmt + httpUriEncode() instead of the existing HttpQuery machinery already used elsewhere in this file and in s3/azure. Using HttpQuery would also let the subject token ride on an enforced redactList (as S3's AssumeRoleWithWebIdentity does) instead of relying on an informal "never logged" comment as the only guard 

- doc/xml/reference.xml — PR body states impersonation and non-text credential_source.format are unsupported, but that disclosure never made it into the shipped reference docs, only the PR body

adolsalamanca:
Thanks for the feedback @hunleyd, addressed your comments 🙏🏽

hunleyd (approved):

adolsalamanca:
Pushed an additional test as the CI was failing due to lack of coverage.
cc @hunleyd @dwsteele

hunleyd (approved):

dwsteele (approved):
I made a number of small changes: fix mem context handling (a237b884c5b2), verify response format (d64510eab), improve error message and doc (d0817bf14), add test titles (f962984), and simplify a test (44d58f59).

The big change (dd7240e0) was to remove the response redaction. From my perspective, it does not make a lot of sense to redact responses from just one auth type given that the format is the same. If we want to do that (and I don't think we do) that should definitely be the subject of a separate commit with all auth types included.

I also removed an unused redaction in 442c05eea6. Since it was only used in a request body, which is never logged, it was a noop.

PR #2846, opened 2026-09-02
Disable read retry when filters are passed to the driver.

### LINKED ISSUE #2845: Segfault in ioFilterGroupParamAll via blockDeltaNext during block-incremental restore from a remote repo
### Summary

A `restore:local` worker reliably segfaults on a null pointer during restore of a
block-incremental backup from a remote (`repo1-host`) S3 repository. The parent process only
notices when it next writes to the dead worker, so the user-visible error is
`[039]: client state is 'response' but expected 'idle'` — which reads like a transport
failure but is not one.

This looks like the same defect as #2748, which was closed without diagnosis. That report had
only a corrupted-pointer address; this one has a clean backtrace from a **non-LTO build**, which
exposes two frames that link-time optimisation had inlined away.

### Backtrace

Core from pgBackRest 2.59.1 built with LTO explicitly disabled:

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  ioFilterGroupParamAll ()
#1  storageReadRemoteOpen ()
#2  storageRead ()
#3  ioReadInternal ()
#4  ioRead ()
#5  ioLimitRead ()
#6  ioReadInternal ()
#7  ioRead ()
#8  blockDeltaNext ()
#9  restoreFile ()
#10 restoreFileProtocol ()
#11 protocolServerProcess ()
#12 cmdLocal ()
#13 main ()

rdi 0x17   rsi 0x14   r13 0x0   r14 0x7
=> 0x...  <ioFilterGroupParamAll+158>:  mov 0x0(%r13),%rdi
```

`r13` is NULL and is dereferenced directly. The worker was invoked as:

```
pgbackrest --exec-id=... --log-level-console=off --log-level-file=off \
  --log-level-stderr=error --pg1-path=... --process=1 --remote-type=repo \
  --repo=1 --stanza=... restore:local
```

The `blockDeltaNext` frame puts this squarely in the block-incremental read path.

### Environment

- pgBackRest **2.59.1** (also reproduced on 2.59.0)
- openSUSE Leap 15.5, 15.6 and **16.0** — reproduced on all three
- PostgreSQL 15
- Repository: S3, reached indirectly via a dedicated repo host (`repo1-host`, ssh)

Relevant configuration:

```
repo1-type=s3
repo1-bundle=y
repo1-block=y
compress-type=zst
process-max=4
archive-async=y
```

Restore is `--type=time --set=<full>_<incr> --delta`, of a ~4.9 TB cluster
(incremental backup size ~571 GB) whose full backup is around three weeks older than the
incremental being restored.

### Frequency and pattern

Crashes on most restores, but **not at a fixed point** — observed at 13.1%, 36.98%, 37.17% and
80.74% of the same restore across separate runs. It always crashes in the same function.

Because the fault occurs only for blocks actually fetched, a `--delta` retry always makes
forward progress: files that verify clean are skipped and not re-read through the block-delta
path. Running the restore repeatedly with `--delta` eventually completes it.

### Ruled out

- **Not our packaging or toolchain.** Reproduced on distribution-toolchain builds for three
  different OS releases, and on a build with LTO disabled.
- **Not the network or storage layer.** The `[039]`/`[042]` errors are consequences of the worker
  dying, arriving up to 74 s after the segfault appears in `dmesg`.
- **Not a corrupt backup.** The same backup set restores successfully when the run is retried
  enough times with `--delta`.

### Workaround

Retry with `--delta` until it completes. #2748 reports that `--process-max=1` avoids it, which is
consistent with a concurrency-related lifetime bug, though it makes a large restore impractically
slow.

dwsteele:
When a read is retried the driver is freed and recreated but the filter group is not set on the new driver. For drivers that require the filter group (i.e. remote) this causes a segfault when the read is opened because the filter group is NULL.

This was introduced in a953aa74b, which changed retry to recreate the driver rather than reopen it, added the filter group callback that is only set when the read is created, and replaced the driver retry flag with the storageFeatureReadRetry feature, which remote storage inherits from the repository host. Before that commit retry was never enabled for remote reads.

Even with the filter group set the retry could not work because the filters are passed to the remote when the read is opened so they cannot be run again, and bytes read are counted after the filters have been applied so the offset to restart the read cannot be determined.

Read retry is only enabled for object stores so this happened when the repository was on an object store accessed via a repository host. The repository host still retries the object store read, where the filters are applied above the driver and the retry works correctly, so the only retry lost is a redundant second attempt by the client.


src/storage/read.c
hunleyd, line 145: The bug this PR fixes was a driver-recreate path (storageRead()'s CATCH_ANY retry block) silently losing an invariant (filter group must be NULL for retry to be safe). The fix makes that path unreachable today, but nothing pins the invariant there; a future driver combining retry + filterGroup reintroduces the identical segfault. Add an assert at the recreate site:
    ```diff
                            this->driver = storageInterfaceNewReadP(
                                storageDriver(this->storage), this->pub.name, .compressible = this->compressible,
                                .offset = this->pub.offset + this->bytesRead,
                                .limit = this->pub.limit != NULL ? varNewUInt64(varUInt64(this->pub.limit) - this->bytesRead) : NULL,
                                .versionId = this->pub.versionId);

    +                       // The filter group is never reset on the recreated driver, so retry must already be disabled for any
    +                       // driver that requires one (see storageReadNew())
    +                       ASSERT(storageReadDriverInterface(this->driver)->filterGroup == NULL);
    +
                            // Open should always return true either because a) the file still exists or b) the driver is async and```

test/src/module/storage/remoteTest.c
```diff
@@ -416,6 +416,30 @@ testRun(void)
...
+        // -------------------------------------------------------------------------------------------------------------------------
+        TEST_TITLE("read error is not retried when filters are passed to the remote");
+
+        // Write an encrypted file larger than the buffer size so the error will happen after the first chunk has been read
+        HRN_STORAGE_PUT(
+            storageTest, TEST_PATH "/repo128/test.bin", contentBuf,
+            .cipherSpec = cipherSpecNew(cipherTypeAes256Cbc, BUFSTRDEF("x")));
+
+        // Enable read retry to show that it is disabled when the filters are passed to the remote
+        storageRepo->pub.interface.feature |= 1 << storageFeatureReadRetry;
```
hunleyd, line 429: |= to set the retry bit, ^= to restore it are not true inverses. Remote storage inherits its whole feature mask from the host it proxies, so a remote-over-object-store fixture genuinely can already carry this bit; ^= would then clear a bit that should stay set, corrupting state for later tests in the same block. Matches an existing bug class the repo's own posixTest.c avoids by using ^= both ways:
```diff
          // Enable read retry to show that it is disabled when the filters are passed to the remote
  -       storageRepo->pub.interface.feature |= 1 << storageFeatureReadRetry;
  +       storageRepo->pub.interface.feature ^= 1 << storageFeatureReadRetry;```

hunleyd (changes requested):

src/storage/read.c
dwsteele: Good plan, done in 41fccde9d6cf0.

test/src/module/storage/remoteTest.c
dwsteele: I think it makes more sense to preserve the prior feature value rather than depending on retry being unset. That's true right now but the Posix driver could get a retry in the future and it would break this test. Done that way in 41fccde9d6cf0.

hunleyd (approved):

PR #2849, opened 2026-09-04
Retry Azure batch sub-requests missing from the response.

### LINKED ISSUE #2840: expire command does not delete all files in azure repo after introduction of batch delete in version 2.59.0
1. pgBackRest version:

2.59.1

2. PostgreSQL version:

17.11

`Linux 5.15.0-179-generic #189-Ubuntu SMP Tue May 5 18:20:56 UTC 2026 x86_64 x86_64 x86_64 GNU/Linux`

Via apt package `pgbackrest=2.59.1-1.pgdg22.04+1` from `https://apt.postgresql.org/pub/repos/apt jammy-pgdg/main amd64 Packages`

    - `pgbackrest.conf` file(s)
    - errors in the postgresql log file before or during the time you experienced the issue
    - log file in `/var/log/pgbackrest` for the commands run (e.g. `/var/log/pgbackrest/mystanza_backup.log`)

[expire-1.log](https://github.com/user-attachments/files/31375172/expire-1.log)
[expire-detail.log](https://github.com/user-attachments/files/31375171/expire-detail.log)
[pgbackrest.conf.txt](https://github.com/user-attachments/files/31375199/pgbackrest.conf.txt)
[postgresql.conf.txt](https://github.com/user-attachments/files/31375200/postgresql.conf.txt)

6. Describe the issue:

* Repo type is `azure`.
* We left the relevant flags of the azure sas token in the supplied `pgbackrest.conf`.
* We updated pgbackrest from 2.57.0 to 2.59.1.
* After the last full backup and expire runs monitoring via check_pgbackrest (`--extended-check`) failed with "check min WAL is not the oldest archive".
* `pgbackrest info` no longer listed the expired backup and WAL files.
* The expired backups and wal files (see `expire-1.log`) indeed where not completely deleted from azure.
* Note that `expire` reported success.
* Repeated `pgbackrest expire` calls only removed a few files in each run.
* We downgraded to 2.58.0 and ran `pgbackrest expire` again which deleted the files from azure.
* We suspect the new batch delete feature to be the problem. In the statistics JSON of the expire call supplied in `expire-detail.log` `azure.rm.batch` and `azure.rm.batch.retry` have the same total.
* We ran `expire` with trace log files, the output included:

```
2026-08-24 12:30:30.295 P00  TRACE:         common/io/http/response::httpResponseMultiNext: => {code: 400, reason: One of the request inputs is not valid., contentChunked: false, contentSize: 221, contentRemaining: 221, closeOnContentEof: false, contentExists: false, contentEof: false, contentCached: true}, header: {content-type: 'application/http, application/xml', x-ms-error-code: 'InvalidInput', x-ms-request-id: '86f0947c-e01e-00fd-15b3-33f9e9000000', x-ms-version: '2020-08-04', content-length: '221', server: 'Windows-Azure-Blob/1.0'}}
```

* We can supply the complete trace file(s) if need via a private channel.

dwsteele:
Response parts were mapped to sub-requests by iterating the parts, so when the service returned fewer parts than were requested the remaining sub-requests were silently treated as complete. Azure fails the entire batch when it is unable to parse the request and reports the failure as a single response part, so only one file per batch was removed while the command reported success.

Iterate the sub-requests instead and retry any that have no response part, just as parts that report an error are retried. More response parts than sub-requests is now an error.


src/storage/azure/storage.c
```diff
@@ -809,21 +809,20 @@ storageAzurePathRemoveInternal(StorageAzurePathRemoveData *const data)
...
 
-            // Loop through all response parts. Parts are mapped to the original request by ordinal position since the service
-            // returns one response part per sub-request in request order. The echoed content-id is intentionally not used because
-            // Azure omits it on some error responses, which would otherwise make a failed part impossible to retry.
-            unsigned int partIdx = 0;
+            // Loop through all sub-requests. Response parts are mapped to the original request by ordinal position since the
+            // service returns one response part per sub-request in request order. The echoed content-id is intentionally not used
+            // because Azure omits it on some error responses, which would otherwise make a failed part impossible to retry.
             HttpResponse *responsePart = httpResponseMultiNext(responseMulti);
             CHECK(FormatError, responsePart != NULL, "at least one response part is required");
```
hunleyd, line 819: The retained `CHECK(FormatError, responsePart != NULL, "at least one response part is required")` still hard-fails when Azure returns ZERO response parts. A truncated/empty multipart body aborts the whole remove instead of retrying every sub-request.
Either drop the check (0 parts → retry all, consistent with the PR's stated principle) or keep it and document why zero is different from one-of-N missing?

test/src/module/storage/azureTest.c
hunleyd, line 1: No test exercises the PR's own motivating scenario: one failure/error response part mapped across N>1 sub-requests (index 0 retries via the error branch, indexes 1..N-1 retry via the missing branch, in the same batch). Current coverage only has a single missing part, never a real error part plus multiple missing parts together?

test/src/module/storage/azureTest.c
```diff
@@ -761,7 +761,7 @@ testRun(void)
                 TEST_RESULT_BOOL(storageInfoP(storage, NULL, .ignoreMissing = true).exists, false, "info for /");
 
                 // -----------------------------------------------------------------------------------------------------------------
-                TEST_TITLE("remove files with shared key batched one per request");
+                TEST_TITLE("remove files with shared key batched one per request and extra response part");
```
hunleyd: The only shared-key-signed batch test had its terminal success assertion (`TEST_RESULT_VOID`) converted into the new excess-parts error test. No test now confirms a clean shared-key batched remove completes end to end (the sibling success test at line ~1186 runs under SAS auth only).

hunleyd (changes requested):
src/common/io/http/response.c:544 (httpResponseMultiNext): `boundaryLast` is unconditionally advanced before the NULL check, with no eof latch. Calling `httpResponseMultiNext()` again after it has already returned NULL pushes `begin` past the buffer and trips `bufFindP`'s bounds ASSERT. This diff avoids the hazard by guarding every call site with `if (responsePart != NULL)` rather than fixing the iterator's own contract. the next caller (e.g. GCS, if it gets an equivalent fix) will rediscover this the hard way. Suggested fix: latch `boundaryLast = NULL` on eof and short-circuit at the top of the function.

src/storage/gcs/storage.c:1003 The GCS driver has the identical bug this PR fixes for Azure: `do { ... } while (responsePart != NULL)` iterates received parts only, so a sub-request with no matching response part (by content-id) is silently treated as complete. Same silent-data-loss failure mode, different driver. Open an issue?

src/storage/azure/storage.c
dwsteele: It does not seem to be possible to get zero subparts here based on the documentation and testing, documented the reasoning in 14c3ff13b67.

test/src/module/storage/azureTest.c
dwsteele: Done in 598c6ae9e.

test/src/module/storage/azureTest.c
dwsteele: Done in 652c5efe82.

dwsteele:
> This diff avoids the hazard by guarding every call site with if (responsePart != NULL) rather than fixing the iterator's own contract.

Good point, improved in b358b82d10a518.

> The GCS driver has the identical bug this PR fixes for Azure

I looked into this while fixing the bug and it doesn't really. On GCS the parts are keyed with the content-id, unlike in Azure, so that gives us something to correlate to that is not positional. In any case the GCS version has been working for a while with no known issues so I'm not too excited about making speculative changes there.

hunleyd (approved):

PR #2855, opened 2026-09-11
Fix pagination of S3 version listing.

### LINKED ISSUE #2854: Error when restoring from S3 with --repo-target-time set
1. pgBackRest version: 2.59

2. PostgreSQL version: 18.6

3. Operating system/version - if you have more than one server (for example, a database server, a repository host server, one or more standbys), please specify each: Oracle Linux 9.8

4. Did you install pgBackRest from source or from a package? package

    - `pgbackrest.conf` file(s)

repo1-type=s3
repo1-s3-bucket=<redacted>
repo1-s3-endpoint=<redacted, Ceph RGW>
repo1-s3-uri-style=path
repo1-s3-key-type=... (whatever applies)

(Backend is Ceph RADOS Gateway, but the root cause below is not backend-specific — see analysis.)

Error:

ERROR: [029]: raised from local-1 protocol: unable to find child 'NextContinuationToken':0 in node 'ListVersionsResult'
       [FormatError] on 2 retries from 126-15195ms: unable to find child 'NextContinuationToken':0 in node 'ListVersionsResult'

6. Describe the issue:

Running a restore using  --repo-target-time  fails once the S3 version listing for the target prefix needs to be paginated (i.e. more than 1000 keys/versions under the prefix — the default S3 page size):

pgbackrest restore --repo=1 --repo-target-time="2026-09-10 09:59:00+02" --type=time "--target=2026-09-10 09:53:00.780558+02" --process-max=8

ERROR: [029]: raised from local-1 protocol: unable to find child 'NextContinuationToken':0 in node 'ListVersionsResult'
       [FormatError] on 2 retries from 126-15195ms: unable to find child 'NextContinuationToken':0 in node 'ListVersionsResult'

At the time of the error, the relevant backup prefix contained roughly 3000+ object versions (WAL + backup files accumulated across several test backups/expirations), which is enough to cross S3's default 1000-key page size and trigger a truncated (paginated) response.

The next part is generated by AI, so please handle with care:

7. Root cause (found via source inspection,  src/storage/s3/storage.c , confirmed identical in both the  release/2.59.1  tag and the current  main  branch):

pgBackRest issues two different S3 list requests depending on context ( storageS3ListInternal  /  storageListP , around line 796-798 in 2.59.1):

• Normal listing →  ?list-type=2  ( ListObjectsV2 ) — paginated via the  NextContinuationToken  element.
• Version listing, used specifically when  targetTime != 0  (i.e.  --repo-target-time ) →  ?versions  ( ListObjectVersions ) — per the AWS API spec, this response type is paginated via  NextKeyMarker  /  NextVersionIdMarker , and its XML root ( ListVersionsResult ) never contains a  NextContinuationToken  element, truncated or not.

The pagination-continuation logic (~line 838-844) is shared between both code paths and unconditionally looks for  NextContinuationToken  whenever  IsTruncated=true :

if (strEq(xmlNodeContent(xmlNodeChild(xmlRoot, S3_XML_TAG_IS_TRUNCATED_STR, true)), TRUE_STR))
{
    const String *const nextContinuationToken = xmlNodeContent(
        xmlNodeChild(xmlRoot, S3_XML_TAG_NEXT_CONTINUATION_TOKEN_STR, true));
    CHECK(FormatError, !strEmpty(nextContinuationToken), S3_XML_TAG_NEXT_CONTINUATION_TOKEN " may not be empty");
    ...
}

Since a genuine  ListVersionsResult  response never has this element,  xmlNodeChild(..., true)  (the  true  presumably meaning "required") throws the  FormatError  seen above, as soon as the version listing needs a second page.

This reproduces against any S3-compatible backend that correctly implements the  ListObjectVersions  API per spec (confirmed here against Ceph RADOS Gateway, but the AWS API documentation confirms real AWS S3 behaves the same way) — it is not a Ceph-specific issue. It simply requires the target prefix to have accumulated enough object versions to exceed one page (1000 by default).

8. Expected behavior:  --repo-target-time  should paginate version listings using  NextKeyMarker / NextVersionIdMarker  (per the  ListObjectVersions  API), separately from the  NextContinuationToken  logic used for normal  ListObjectsV2  listings.

9. Workaround currently in use: avoid  --repo-target-time  once affected; instead manually identify and remove the specific S3 delete-marker version(s) hiding the desired backup ( aws s3api list-object-versions  +  aws s3api delete-object --version-id <deletemarker-id> ), then restore normally without  --repo-target-time .

dwsteele:
Listing with a repository target time uses the ListObjectVersions API but pagination read NextContinuationToken from the response, which is only returned by ListObjectsV2. A version listing truncated at the maximum key count failed with "unable to find child 'NextContinuationToken':0 in node 'ListVersionsResult'", so a repository with more than a thousand versions under a prefix could not be read at a target time.

Paginate the version listing with NextKeyMarker and NextVersionIdMarker instead. NextVersionIdMarker is missing or empty when the listing is truncated at a common prefix, in which case the next page starts at the key after the key marker, which is what is needed.

Build the query for each page from the base query so a marker set for one page is not reused for the next.

The mock response used to test version listing pagination returned a continuation token rather than markers, so the error was not caught.

---

I reproduced this issue on both AWS and SeaweedFS and the fix works both places. I also checked Azure and GCS to see if I had make a similar mistake but they both look correct.


guruguruguru:
@dwsteele works perfectly fine now, thanks!

dwsteele:
@guruguruguru Great, thanks! The patch release will be on the 28th this month, which is a week later than usual. I'm doing it later to make sure we get the PG19beta4 changes. It has been an unusually unstable beta period for pg_control/WAL.

src/storage/s3/storage.c
```diff
@@ -839,16 +844,42 @@ storageS3ListInternal(
...
+                    {
+                        const String *const nextKeyMarker = xmlNodeContent(
+                            xmlNodeChild(xmlRoot, S3_XML_TAG_NEXT_KEY_MARKER_STR, true));
+                        CHECK(FormatError, !strEmpty(nextKeyMarker), S3_XML_TAG_NEXT_KEY_MARKER " may not be empty");
+
+                        httpQueryAdd(queryNext, S3_QUERY_KEY_MARKER_STR, nextKeyMarker);
+
+                        // The version id marker may be missing or empty when the next result is a common prefix. In that case the
+                        // next page starts at the key after the key marker, which is what is needed.
+                        const XmlNode *const nextVersionIdMarker = xmlNodeChild(
```
hunleyd, line 861: `xmlNodeContent(nextVersionIdMarker)` called twice (condition + `httpQueryAdd` arg), each call is a fresh libxml2 alloc + copy. Bind once:
  ```diff
  -                        const XmlNode *const nextVersionIdMarker = xmlNodeChild(
  -                            xmlRoot, S3_XML_TAG_NEXT_VERSION_ID_MARKER_STR, false);
  +                        const String *const nextVersionIdMarker = xmlNodeContent(
  +                            xmlNodeChild(xmlRoot, S3_XML_TAG_NEXT_VERSION_ID_MARKER_STR, false));

  -                        if (nextVersionIdMarker != NULL && !strEmpty(xmlNodeContent(nextVersionIdMarker)))
  -                            httpQueryAdd(queryNext, S3_QUERY_VERSION_ID_MARKER_STR,
  xmlNodeContent(nextVersionIdMarker));
  +                        if (nextVersionIdMarker != NULL && !strEmpty(nextVersionIdMarker))
  +                            httpQueryAdd(queryNext, S3_QUERY_VERSION_ID_MARKER_STR, nextVersionIdMarker);```

src/storage/s3/storage.c
hunleyd, line 802: (6 sites: 802, 816, 851, 916, 945, 976) — introduce `const bool versionList = targetTime != 0` and use at all sites?

hunleyd (changes requested):

src/storage/s3/storage.c
dwsteele: Done in 14046b60d.

src/storage/s3/storage.c
dwsteele: I'd rather not since it is already a simple comparison. This comparison is elsewhere, too, and I think it is easier to read if they are all the same.

hunleyd: fair enough

hunleyd (approved):

PR #2857, opened 2026-09-15
Fix async archive-push stalling on a zero-size backup history file.

### LINKED ISSUE #2856: Asynchronous archive-push stops on a zero-size backup history file
1. pgBackRest version:

2.59.0. I also reproduced the problem on main at d7f54b224.

2. PostgreSQL version:

17 and 18

Debian 12 (bookworm) containers on Kubernetes, managed by the CloudNativePG operator. Each cluster has one PostgreSQL instance.

Package from apt.postgresql.org (2.59.0-1.pgdg12+1).

pgbackrest.conf: archive-async=y. The problem occurs only in asynchronous mode.

postgresql.conf: archive_mode=on. archive_command runs pgbackrest archive-push through the CloudNativePG instance manager.

PostgreSQL log. This error repeats on every archive attempt:

```
ERROR: [029]: size of WAL segment '000000010000000000000003.00000028.backup' is 0
```

6. Describe the issue:

Asynchronous archive-push fails on every run when the first ready file is a zero-size backup history file. No WAL is archived after that file, and pg_wal grows until the disk is full.

We saw this on clusters that start from a storage snapshot of another cluster's volume. PostgreSQL does not fsync the backup history file that pg_backup_stop() writes. When a backup ends just before the snapshot, the copy can contain that file with zero size, together with its .ready file.

Commit 34a035f4d (2.59.0) added archive-push-batch-size. To calculate the batch, cmdArchivePushAsync() reads the size of the first file in the ready list. That list also contains files that are not WAL segments, such as backup history files. When the first file has zero size, the size check fails and the async process pushes nothing. PostgreSQL retries the oldest ready file first, so the same error occurs on every run.

Before 2.59.0, pgBackRest pushed the zero-size history file and archiving continued.

To reproduce:

1. Set archive-async=y.
2. In pg_wal, create an empty file 000000010000000100000018.00000028.backup.
3. Create archive_status/000000010000000100000018.00000028.backup.ready.
4. Run archive-push. The async process fails with the error above.

I have a fix with a unit test on my fork:
https://github.com/andrewkroh/pgbackrest/tree/fix/archive-push-zero-size-backup-history-cig

The fix uses the size of the first ready WAL segment. If no WAL segment is ready, it does not apply the batch size. A zero-size WAL segment still causes an error. CI passes on my fork, and I can open a pull request against integration.

archivePushDrop() makes the same first-file assumption for archive-push-queue-max. I did not change it in this fix.

andrewkroh:
34a035f4d added archive-push-batch-size, which converts the batch size to a file count with the size of the first ready file. The ready list also contains files that are not WAL segments, such as backup history files. PostgreSQL does not sync a backup history file to disk, so a crash-consistent copy of the cluster, e.g. a storage snapshot taken just after pg_backup_stop(), can contain the file with zero size.

In that case every asynchronous run errors with "size of WAL segment '000000010000000000000003.00000028.backup' is 0". PostgreSQL always retries the oldest ready file first, so no WAL after that file is archived and pg_wal grows without limit.

Use the size of the first ready WAL segment instead. If no WAL segment is ready then the batch size is not applied, since the other files are small. A zero-size WAL segment is still an error.

Relates: #2856


src/command/archive/push/push.c
```diff
@@ -554,17 +554,27 @@ cmdArchivePushAsync(void)
...
-                while (strLstSize(jobData.walFileList) > 1 && strLstSize(jobData.walFileList) > batchMax)
-                    strLstRemoveIdx(jobData.walFileList, strLstSize(jobData.walFileList) - 1);
+                if (segmentIdx < strLstSize(jobData.walFileList))
+                {
+                    const String *const walSegment = strLstGet(jobData.walFileList, segmentIdx);
+                    const uint64_t segmentSize = storageInfoP(
+                        storagePg(), strNewFmt("%s/%s", strZ(jobData.walPath), strZ(walSegment))).size;
+                    CHECK_FMT(FormatError, segmentSize != 0, "size of WAL segment '%s' is 0", strZ(walSegment));
+
+                    const uint64_t batchMax = cfgOptionUInt64(cfgOptArchivePushBatchSize) / segmentSize;
```
hunleyd, line 583: batchMax is derived from the segment's size, but the trim loop compares it against the count of the whole ready list. So if the ready list is: [zero-size .backup, walSegment1, walSegment2, ...], archive-push-batch-size is sized so batchMax == 1. Trim removes from the tail down to 1 file  and since the non-segment file sits at index 0, it survives while walSegment1 gets trimmed away. That run pushes only the .backup file, zero real WAL. It's self-correcting next run, but contradicts the PR's own comment ("always push at least one segment"). If I'm reading this right.

hunleyd (changes requested):

src/command/archive/push/push.c
dwsteele: True, but the user would need to set `archive-push-batch-size` to more than 100x smaller than the default for this to even be visible -- and it continues work no matter the setting, just with possibly smaller batch sizes than expected (rarely). Improving this is more complicated that it is worth in my opinion.

Also, I'd prefer to base this on the WAL segment size in pg_control, which would get rid of the scan entirely. The "issue" would remain but the code would be far simpler.

dwsteele: it is worth a note, though, so I added that in a0a41f5c70.

hunleyd (approved):

dwsteele:
Committed, thank you!

PR #2859, opened 2026-09-17
Allow test user names that adduser on Debian rejects.

### LINKED ISSUE #2858: vm-build fails on Debian-based VMs when host username contains a dot
## Description

`test.py vm-build` fails when the host user running the tests has a username containing a character not allowed by Debian's `adduser` `NAME_REGEX` (e.g. a dot, as in `first.last` corporate naming conventions). The test container is built with the same username as the host user (`doc/lib/common/user.py::user_name()`) so that files created inside map back to the host UID/name — but Debian's `adduser` package validates that name and aborts:

```
err: Please enter a username matching the regular expression
    configured via the NAME_REGEX configuration variable. Use the
    `--allow-bad-names' option to relax this check or reconfigure
    NAME_REGEX in configuration.
```

## Proposed fix

Add `--allow-bad-names` for the Debian branch only:

```python
def _user_create(os_base, name, id, group):
    if os_base == VM_OS_BASE_RHEL:
        return "adduser -g%s -u%u -N %s" % (group, id, name)

    if os_base == VM_OS_BASE_DEBIAN:
        return 'adduser --uid=%u --ingroup=%s --disabled-password --gecos "" --allow-bad-names %s' % (id, group, name)

    return 'adduser --uid=%u --ingroup=%s --disabled-password --gecos "" %s' % (id, group, name)
```

The corresponding fixture in `test/src/module/test/command/vm/build_test.py` (`SCRIPT_TEST_U24`) would also need the flag added to keep the unit test's exact-match assertion passing.

Should I open a PR for this change?

pgstef:
Debian's adduser rejects usernames that don't match NAME_REGEX (e.g. containing a dot).

Fix #2858


dwsteele (approved):
Looks good, thank you!

pgstef:
> Looks good, thank you!

Thanks for having a look at it! :-)

PR #2860, opened 2026-09-17
Add skip option to backup-standby

### LINKED ISSUE #1550: Have pgbackrest exit silently when run against a standby node

pgstef:
Adds a new `skip` value to `backup-standby`, alongside the existing `n`, `prefer`, and `y` values.

The goal is to allow a scheduled `backup` command running on a standby to exit successfully instead of failing or unexpectedly taking the backup from another cluster.

A `WARN` is logged when the backup is skipped, with wording that distinguishes whether a primary was reachable:

* `standby cluster found, skipping backup since backup-standby=skip`
* `unable to find primary cluster but standby cluster found, skipping backup since backup-standby=skip`

After reading the feature request in https://github.com/pgbackrest/pgbackrest/issues/1550, my initial interpretation was that the main use case was an HA setup where the same backup command may run on either node, but only the locally running cluster is defined in the configuration.

In that case, when the command happens to run on a standby, we want to avoid the `unable to find primary cluster` error and simply skip the backup.

So the initial behavior was essentially:

> Skip the backup if only a standby cluster can be found.

However, there is another case to consider: what should `backup-standby=skip` do when both the primary and standby are configured and reachable?

Today:

* `backup-standby=n` takes the backup from the primary.
* `backup-standby=y` takes the backup from the standby.
* `backup-standby=prefer` takes the backup from the standby when available.

For `backup-standby=skip`, I think the least surprising behavior for users may be to skip the backup whenever the command detects that it is running against a standby, regardless of whether a primary can also be found.

In other words:

> If a standby is detected and `backup-standby=skip`, skip the backup.

This is still open for discussion, of course, but it seems more consistent with what users are likely to expect from the `skip` value.

Current behavior in this PR:

* primary + standby found + `backup-standby=n` -> backup taken from primary
* primary + standby found + `backup-standby=y` -> backup taken from standby
* primary + standby found + `backup-standby=skip` -> backup skipped with:
  `WARN: standby cluster found, skipping backup since backup-standby=skip`
* primary not found + standby found + `backup-standby=skip` -> backup skipped with:
  `WARN: unable to find primary cluster but standby cluster found, skipping backup since backup-standby=skip`


dwsteele:
> I think the least surprising behavior for users may be to skip the backup whenever the command detects that it is running against a standby, regardless of whether a primary can also be found.

For the use case proposed in #1550 the primary and standby would never be reachable at the same time, so agreed, I think if both are found then skip means skip the backup.

I can see how you might want to interpret this as an error, however, given the configuration.

Let's see if @df7cb or @mbanck have thoughts.

test/src/module/command/backupTest.c
```diff
@@ -3032,6 +3059,98 @@ testRun(void)
...
+            TEST_RESULT_VOID(hrnCmdBackup(), "backup");
+
+            TEST_RESULT_LOG(
+                "P00   WARN: unable to check pg1: [DbConnectError] unable to connect to 'dbname='postgres' port=5432': error\n"
+                "P00   WARN: unable to find primary cluster but standby cluster found, skipping backup since"
+                " backup-standby=skip");
+
+            // Confirm no new backup was created -- latest still points at the backup from the prior test
+            TEST_RESULT_STR_Z(
+                testBackupValidateP(storageRepo(), STRDEF(STORAGE_REPO_BACKUP "/latest")),
```
dwsteele: I think this is too verbose. It can be simplified by just checking that the latest link equals the expected backup label. There is no value in enumerating all the files in the backup.

test/src/module/command/backupTest.c
```diff
@@ -3032,6 +3059,98 @@ testRun(void)
...
+                HRN_PQ_SCRIPT_OPEN(1, "dbname='postgres' port=5432", PG_VERSION_96, strZ(pg1Path), false, NULL, NULL),
+                HRN_PQ_SCRIPT_OPEN(2, "dbname='postgres' port=5433", PG_VERSION_96, strZ(pg2Path), true, NULL, NULL));
+
+            TEST_RESULT_VOID(hrnCmdBackup(), "backup");
+
+            TEST_RESULT_LOG("P00   WARN: standby cluster found, skipping backup since backup-standby=skip");
+
+            // Confirm no new backup was created -- latest still points at the backup from the prior test
+            TEST_RESULT_STR_Z(
+                testBackupValidateP(storageRepo(), STRDEF(STORAGE_REPO_BACKUP "/latest")),
```
dwsteele: Same as above.

build/config.yaml
hunleyd, line 1: `backup-standby` allow-list is shared by both `backup` and `check` commands, so `skip` is now valid for `pgbackrest check` too.

src/command/backup/backup.c
hunleyd, line 1: pre-existing offline guard (line ~89) silently demotes `skip` to `n` when `--no-online` is set, i think. so an offline scheduled run on a standby proceeds and backs up the demoted PGDATA

hunleyd (changes requested):
auto-expire still runs after a "skipped" backup (cmdBackup returning early doesn't skip the post-backup expire step). we cool with that?

test/src/module/command/backupTest.c
pgstef: Right, fixed in 473f2cc :+1:

test/src/module/command/backupTest.c
pgstef: Fixed in 473f2cc

pgstef:
> auto-expire still runs after a "skipped" backup (cmdBackup returning early doesn't skip the post-backup expire step). we cool with that?

Oh, yeah, you're right. It makes sense to skip the auto-expire too (I'm usually having that option disabled in my test environments). 

Suggested fix in 7ad7e15 to automatically disable expire-auto if the backup is skipped ;-)

build/config.yaml
pgstef: As far as I can see, allowing `skip` for `check` does not actually change the command's behavior.

`check` currently uses `if (cfgOptionSeq(cfgOptBackupStandby) != CFGOPTVAL_BACKUP_STANDBY_N)` to decide whether to log a `WARN` when no standby is found among the configured hosts.

So `prefer`, `y`, and now `skip` are treated identically by `check`. In other words, `skip` becomes a valid value for the command because the allow-list is shared, but it does not introduce any new `check` behavior.

Do you think we should have some specific behavior for `check` when `backup-standby=skip` is set?

Maybe log a `WARN` when a standby is found, to make it clear that a backup would be skipped with that configuration? :thinking:

I've always mostly seen `check` as "check that archiving is working", so I hadn't really thought about looking at that command specifically for this feature :shrug:

src/command/backup/backup.c
pgstef: If I read it correctly, and based on my tests, with only a stopped standby configured and `--no-online` used, `y`, `prefer`, and `skip` all behave the same way.

The backup proceeds and logs `WARN: option backup-standby is enabled but backup is offline - backups will be performed from the primary`.

That warning already feels a bit misleading in the general case, since with an offline cluster we cannot actually determine whether the stopped cluster is a primary or a standby.

I wonder if the more reasonable approach would be to document this limitation explicitly: `backup-standby` has no effect with `--no-online`, because the cluster role cannot be detected without a database connection.

In that sense, I would probably treat the current behavior as a known limitation of offline backups rather than something specific to `skip` that needs to be fixed here.

See 15fb8ac for a proposal along those lines.

build/config.yaml
hunleyd: i look at check's job as validating a working backup config end to end. so if someone set `skip` they probably want to know it is actually being skipped. i wouldn't block this PR over it, can be done later. also not a hill i'm willing to die on :)

src/command/backup/backup.c
hunleyd: that makes sense, yeah. and 15fb8ac works for me

hunleyd (approved):

build/config.yaml
pgstef: Hm, maybe just a notice like INFO "skip is set, so any backup command run with this setup is likely to be skipped"?
I can think about the text a bit more over the weekend and push something about it on Monday if that can help the users, I'm on-board with the idea ;-)

pgstef: @hunleyd What would you think about 4092154?

> INFO: standby cluster found and backup-standby=skip is set - online backups using this configuration will be skipped

